PR #3, opened 2019-06-25
Basic contract metadata proposal

bowenwang1996:
@vgrichina feel free to change things/add more stuff.

### DIFF
--- a/text/0000-contract-metadata.md
+++ b/text/0000-contract-metadata.md
@@ -8,8 +8,10 @@
 This NEP introduces contract metadata, which summarizes the content of a given contract in json format.
 Contract metadata allows developers to easily list the methods in a contract and potentially display them to users to
 provide transparency. 
-It also provides the ability to generate bindings in different languages for the contract, thanks to the class information
-also contained in the metadata.
+It also provides the ability to validate input data in frontend thanks to the signatures of contract methods provided
+by the contract metadata.
+Contract metadata is similar to contract abi in some other blockchains such as Ethereum or Substrate but is more geared
+towards readability and developer friendliness.
 
 # Motivation
 [motivation]: #motivation
@@ -21,6 +23,8 @@ As a result, if developers want to display the methods of a contract to an user 
 especially in the case of financial apps, they have no choice but to hardcode them in the front end, which is suboptimal in many ways.
 Furthermore, if developers want to get the list of methods in some contract for some downstream task like data analysis, or interacting
 with other contracts, they have no way of doing so.
+In addition, because of the lack of signature of the contract functions, there is no way that the frontend can validate
+that input data is of the right format, which might lead to some spamming attacks.
 Contract metadata aims to solve the aforementioned problems by providing a convenient way 
 for developers to list the methods of contracts.
 
@@ -35,13 +39,75 @@ also provide, in addition to function annotations, class annotations, which main
 Finally, since developers might want to have some contract-level description that provides an overview of what the contract does,
 in metadata we also have contract-level annotation.
 
-More specifically, every contract will have a `metadata` method that returns a json that serializes the aforementioned information.
-The overall format looks like `{"methods": [<method1_info>, ..], "classes": [<class1_info>, ..], "contract": <contract annotation>}`.
-For each method, the json serialization is of the form `{"name": <name>, "parameters": [{"name": <param_name1>, "type": <param_type1>, ..}, .. ], "returnType": <return_type>}`
-where the `returnType` key will only be present for functions that have non-void return types.
-For each class, the json serialization is of the form `{"name": <name>, "fields": [{"name": <field_name>", "type": <field_type>"}, ..]}`.
+The format of contract metadata is similar to [Ethereum json contract abi](https://solidity.readthedocs.io/en/v0.5.3/abi-spec.html#json).
+For each function, the metadata is a json object with the fields:
+* `name`: the name of the function;
+* `parameters`: an array of objects, each of which contains:
+  * `name`: the name of the parameter
+  * `type`: the type of the parameter (more below)
+* `returnType`: The return type of the function. If there is no return value, this field is omitted.
+* `stateMutability`: a string that is either `view` or `change`. It specifies whether the function can mutate state.
 
-## Simple Example
+For types, we try to be as general and language agnostic as possible. Therefore we follow mostly the types that [serde](https://serde.rs/data-model.html)
+ uses. The following types are available:
+* primitive types:
+  * `bool`
+  * `i8`, `i16`, `i32`, `i64`, `i128`
+  * `u8`, `u16`, `u32`, `u64`, `u128`
+  * `string`
+
+* `byteArray`: an array of bytes. In Rust this is `[u8]` whereas in AssemblyScript this is typedarrays.
+* `Option<T>`: Either none or some value. In Rust this is `Option<T>` whereas in AssemblyScript this is `T | null` provided
+that `T` is nullable.
+* `Seq<T>`: A variably sized homogeneous sequence of values, for example `Vec<T>` or `HashSet<T>` in Rust, `Array<T>` in
+AssemblyScript.
+* tuple: A statically sized heterogeneous sequence of values, for example `(u8, string)`. This doesn't apply to AssemblyScript.
+* `Map<K, V>`: A variably sized heterogeneous key-value pairing, for example `BTreeMap<K, V>` in Rust and `Map<K, V>` in AssemblyScript.
+* object: `struct` in Rust or `class` in AssemblyScript. The metadata for an object type is a json object with the fields:
+    * `name`: Name of the class
+    * `fields`: An array of objects each of which contains:
+      * `name`: the name of the parameter
+      * `type`: the type of the parameter
+
+The object types will not be fully unrolled in their json representation. For example, if we have the following classes
+```rust
+pub struct Person {
+  pub name: String,
+  pub address: Address,
+}
+
+pub struct Address {
+  pub city: String,
+  pub street: String,
+  pub zip: u16,
+}
+```
+
+The json representation for `Person` is 
+```json
+{
+  "name": "Person",
+  "fields": 
+    [
+      {"name": "name", "type": "string"},
+      {"name": "address", "type":  "Address"}
+    ]
+}
+```
+
+The type `Address` is not unrolled to avoid repetitive information. Instead, the overall contract metadata also contains
+metadata for each exported class. 
+
+In addition to function metadata and class metadata, the contract metadata also includes information about the contract
+as a whole. To provide this information, a developer can choose to implement a method called `description` in their contract
+which returns a json string that contains `name`, the name of the contract, and `description`, annotation of what the 
+contract does as a whole.
+
+Contract metadata will also have a version associated with it so that it is clear to the developers when the format changes.
+Overall, every contract will have a `metadata` method generated at compile time that returns a json that serializes all the aforementioned information.
+The overall format looks like `{"methods": [<method1_info>, ..], "classes": [<class1_info>, ..], "contract": <contract annotation>, "version": <version>}`.
+
+## A Simple Example
 
 As an concrete example, suppose we have a contract that maintains a counter on chain:
 
@@ -65,9 +131,15 @@ export function decrementCounterBy(amount: i32 = 1): void {
 export function getCounter(): i32 {
   return storage.get<i32>("counter");
 }
+
+@view
+export function description(): string {
+  return '{"name": "Counter", "description": " A contract that maintains a counter with the ability to increase and decrease by the given amount"}';
+} 
 ```
 
-This contract has two change methods, `incrementCounter` and `decrementCounter`, as well as one view method, `getCounter`.
+This contract has two change methods, `incrementCounter` and `decrementCounter`, as well as two view methods, `getCounter`
+and `description`.
 In this case, the metadata we want looks like 
 ```json
 {
@@ -76,33 +148,39 @@ In this case, the metadata we want looks like
                  "name": "getCounter",
                  "parameters": [{"name": "amount", "type": "i32"}],
                  "returnType": "i32",
-                 "methodType": "view"
+                 "stateMutability": "view"
                },
                {
                  "name": "incrementCounterBy",
                  "parameters": [{"name": "amount", "type": "i32"}],
-                 "methodType": "change"
+                 "stateMutability": "change"
                },
                {
                  "name": "decrementCounterBy",
                  "parameters": [{"name": "amount", "type": "i32"}],
-                 "methodType": "change"
+                 "stateMutability": "change"
+               },
+               {
+                 "name": "description",
+                 "parameters": [],
+                 "stateMutability": "view"
                }
              ],
   "classes": [],
-  "contract": "A contract that maintains a counter with the ability to increase and decrease by the given amount"
+  "contract": {"name": "Counter", "description": "A contract that maintains a counter with the ability to increase and decrease by the given amount"},
+  "version": "1.0"
 }
 
 ```
 and the generated `metadata` method looks like:
 ```typescript
 export function metadata(): string {
-    return '{"methods": [{"name": "getCounter", "parameters": [{"name": "amount", "type": "i32"}], "returnType": "i32", "methodType": "view"}, {"name": "incrementCounterBy", "parameters": [{"name": "amount", "type": "i32"}], "methodType": "change"}, {"name": "decrementCounterBy", "parameters": [{"name": "amount", "type": "i32"}], "returnType": "void"}], "classes": [], "contract": "A contract that maintains a counter with the ability to increase and decrease by the given amount"'
+    return '{"methods": [{"name": "getCounter", "parameters": [{"name": "amount", "type": "i32"}], "returnType": "i32", "stateMutability": "view"}, {"name": "incrementCounterBy", "parameters": [{"name": "amount", "type": "i32"}], "stateMutability": "change"}, {"name": "decrementCounterBy", "parameters": [{"name": "amount", "type": "i32"}]}], {"name": "description", "parameters": [], "returnType": "string", "stateMutability": "view"}, "classes": [], "contract": {"name": "Counter", "description": "A contract that maintains a counter with the ability to increase and decrease by the given amount"}, "version": "1.0"}'
 }
 ```
 
-## Real-world Example
-Now let's consider a real-world example that involves more features such as class and arrays. 
+## A More Complex Example
+Now let's consider a more complex example that involves more features such as class and arrays. 
 Suppose one wants to build a todo list app on blockchain, which is modeled as
 ```typescript
 export class Todo {
@@ -148,6 +226,11 @@ export function getAllTodos(): Array<Todo> {
   }
   return loaded;
 }
+
+@view
+export function description(): string {
+  return '{"name": "Todo list", "description": "A todo list on blockchain!"}'
+}
 ```
 
 For this contract, `setTodo` is a change method while `getTodo` and `getAllTodos` are view methods.
@@ -159,19 +242,25 @@ In this case, the metadata we want looks like
                  "name": "setTodo",
                  "parameters": [{"name": "id", "type": "string"}, {"name": "todo", "type": "Todo"}],
                  "returnType": "void",
-                 "methodType": "change"
+                 "stateMutability": "change"
                },
                {
                  "name": "getTodo",
                  "parameters": [{"name": "id", "type": "string"}],
                  "returnType": "Todo",
-                 "methodType": "view"
+                 "stateMutability": "view"
                },
                {
                  "name": "getAllTodos",
                  "parameters": [],
                  "returnType": "Array<Todo>",
-                 "methodType": "view"
+                 "stateMutability": "view"
+               },
+               {
+                 "name": "description",
+                 "parameters": [],
+                 "returnType": "string",
+                 "stateMutability": "view"
                }
              ],
    "classes": [
@@ -180,7 +269,8 @@ In this case, the metadata we want looks like
                   "fields": [{"name": "id", "type": "string"}, {"name":  "title", "type": "string"}, {"name": "completed", "type": "bool"}]
                 }
               ],
-   "contract": "a contract that implements a todo list on blockchain"
+   "contract": {"name":  "Todo list", "description": "A todo list on blockchain!"},
+   "version": "1.0"
 }
 
 ```
@@ -188,10 +278,9 @@ In this case, the metadata we want looks like
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-To implement this NEP, we just need to modify the binding generation to generate a method called `metadata` that returns
-json serialization of contract metadata described in the previous section. This involves walking through the exported methods
-in `main.ts`, get the metadata of each method, and serialize them in json. Metadata of a given method, including decorators,
-are easily extractable from the assemblyscript IR and serialized into json format.
+To implement this NEP, we need to modify the binding generation procedure to generate a method called `metadata` that returns
+json serialization of contract metadata described in the previous section. This involves an AST walk to collect the relevant
+ information about functions and classes, as well as mapping types to the types used in metadata.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -208,11 +297,10 @@ It is unclear to me what the alternative is.
 [unresolved-questions]: #unresolved-questions
 
 * What other information, besides those mentioned in [guide-level-explanation], should we include in the metadata?
-* Most of this NEP is concerned with contract metadata for assemblyscript, for the rust API, it is not yet unclear what
-needs to be done given that it is not yet stabilized.
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
 
 Under the framework proposed in this NEP, it is also not difficult to add annotations to methods in natural language.
-We can also add contract-level annotation as part of the contract metadata json.
+Another interesting possibility is that for data validation, the contract can provide some predicate (in javascript for example)
+to the frontend to validate input data.
--- a/text/0000-contract-metadata.md
+++ b/text/0000-contract-metadata.md
@@ -27,9 +27,8 @@ For developers, there will be two main changes:
 - Instead of annotating view and change methods in the `initContract` function in `main.js`,
 they will instead annotate the methods of a contract by decorators.
 More specifically, every method is by default a change method, unless annotated by `@view_method`.
-- Every contract will have a `metadata` method that returns a json that serializes the contract methods. For each method,
-the json serialization is of the form `{"parameters": {<param1>: <type1>, .. }, "returnType": <return_type>}`.
-The overall serialization is of the form `{"viewMethods": {<method_name>: <method_metadata>, .. }, "changeMethods": {<method_name>: <method_metadata>, .. }}`. 
+- Every contract will have a `metadata` method that returns a json that serializes the contract methods in an array.
+For each method, the json serialization is of the form `{"name": <name>, "parameters": [{"name": <param_name1>, "type": <param_type1>, ..}, .. ], "returnType": <return_type>}`.
 
 As an concrete example, suppose we have a contract that maintains a counter on chain:
 
@@ -57,31 +56,31 @@ export function getCounter(): i32 {
 This contract has two change methods, `incrementCounter` and `decrementCounter`, as well as one view method, `getCounter`.
 In this case, the metadata we want looks like 
 ```json
-{
-  "viewMethods": 
+[
   {
-      "getCounter": {
-        "parameters": {"amount": "i32"},
-        "returnType": "i32"
-      }
+    "name": "getCounter",
+    "parameters": [{"name": "amount", "type": "i32"}],
+    "returnType": "i32",
+    "methodType": "view"
   },
-  "changeMethods": 
   {
-      "incrementCounterBy": {
-        "parameters": {"amount": "i32"},
-        "returnType": "void"
-      },
-      "decrementCounterBy": {
-        "parameters": {},
-        "returnType": "void"
-      }
+    "name": "incrementCounterBy",
+    "parameters": [{"name": "amount", "type": "i32"}],
+    "returnType": "i32",
+    "methodType": "change"
+  },
+  {
+    "name": "decrementCounterBy",
+    "parameters": [{"name": "amount", "type": "i32"}],
+    "returnType": "void",
+    "methodType": "change"
   }
-}
+]
 ```
 and the generated `metadata` method looks like:
 ```typescript
 export function metadata(): string {
-    return "{\"viewMethods\": {\"getCounter\": {\"parameters\": {}, \"returnType\": \"i32\"},\"changeMethods\": {\"incrementCounterBy\": {\"parameters\": {\"amount\": \"i32\"}, \"returnType\": \"void\"}}, {\"decrementCounterBy\": {\"parameters\": {\"amount\": \"i32\"}, \"returnType\": \"void\"}}}"
+    return '[{"name": "getCounter", "parameters": [{"name": "amount", "type": "i32"}], "returnType": "i32", "methodType": "view"}, {"name": "incrementCounterBy", "parameters": [{"name": "amount", "type": "i32"}], "returnType": "i32", "methodType": "change"}, {"name": "decrementCounterBy", "parameters": [{"name": "amount", "type": "i32"}], "returnType": "void", "returnType": "void"}]'
 }
 ```
 
--- a/text/0000-contract-metadata.md
+++ b/text/0000-contract-metadata.md
@@ -28,7 +28,7 @@ For developers, there will be two main changes:
 they will instead annotate the methods of a contract by decorators.
 More specifically, every method is by default a change method, unless annotated by `@view_method`.
 - Every contract will have a `metadata` method that returns a json that serializes the contract methods. For each method,
-the json serialization is of the form `{"name": <method_name>, "parameters": {<param1>: <type1>, .. }, "returnType": <return_type>}`.
+the json serialization is of the form `{"parameters": {<param1>: <type1>, .. }, "returnType": <return_type>}`.
 The overall serialization is of the form `{"view_methods": {<method_name>: <method_metadata>, .. }, "change_methods": {<method_name>: <method_metadata>, .. }}`. 
 
 As an concrete example, suppose we have a contract that maintains a counter on chain:
@@ -61,7 +61,6 @@ In this case, the metadata we want looks like
   "view_methods": 
   {
       "getCounter": {
-        "name": "getCounter",
         "parameters": [],
         "returnType": "i32"
       }
@@ -69,12 +68,10 @@ In this case, the metadata we want looks like
   "change_methods": 
   {
       "incrementCounter": {
-        "name": "incrementCounter",
         "parameters": [],
         "returnType": "void"
       },
       "decrementCounter": {
-        "name": "decrementCounter",
         "parameters": [],
         "returnType": "void"
       }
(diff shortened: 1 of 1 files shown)


bowenwang1996:
@vgrichina updated example

DanielRX:
After reading this I had a few questions:

1. Does `metadata()` get included in the returned `JSON` from `metadata()`?
2. Would it work to store the length of the metadata (when tightly packed) and the length of the contract bytecode, and simply read from `memory(contractStartLocation + contractLen, metadataLength)` in order to get the metadata? It would mean no more metadata function, but *slightly* more work on the side of the reader (however any client could easily build a quick memory read + parse function)

bowenwang1996:
1. `metadata` itself is not included because the purpose is to get the list of existing methods that is otherwise hard to obtain. As every contract has this method, I don't quite see the point of including `metadata` itself in the returned json.
2. The idea is that we don't want developers to deal with memory directly. So if we store the location in memory and length, we still want to have some method that reads it, which could be something to consider, but it is more involved on the implementation side.
3.  The point of metadata is not to evaluate the function, but rather to show what methods (and their types) a contract has. We potentially want to also have annotations included in metadata as well. @vgrichina can provide more info on this.

DanielRX:
1) Then the definition `Every contract will have a metadata method that returns a json that serializes the contract methods` is not correct, the metadata method is a method on the contract, so falls under the contracts methods surely?
2) Do you intend for developers to directly access `metadata()` or for them to get it via a client implementation? If you have a ts library for accessing a contract, that library just simply needs to add the memory function there. The question is more catering to size, storing another method, and a full JSON object, is going to be large compared to a fixed memory location
3) The type narrowing is not evaluation of a function, it's a feature in `ts`
The type of `const x = 1;` is not `number`, it's `1`, so it gives better auto completion and intellisense
If you have a function `() => 1` the return type comes out as `1` because it is a constant value, meaning you don't need to evaluate it to use the value, it becomes referentially transparent in it's type

bowenwang1996:
1. It depends on what you mean by `contract methods`. What I meant is all the contract methods that a developer wrote. `metadata` is generated so even though it is technically a method on the contract, I didn't consider it to be a contract method. I'll revise the wording to be more precise.
2. Yes the intent is that developers can directly call `metadata` and return a json to the frontend. We have a library and could do it there as you suggested. I just need to think more thoroughly about implementation.
3. Ah I see. Please forgive my ignorance. But I guess autocompletion doesn't really matter here and the point is to make it easily understandable, so we'll probably stick with the current design.
Also the current implementation is [here](https://github.com/AssemblyScript/assemblyscript/pull/697) in case anyone is interested.

DanielRX:
As an aside, is this issue meant to be discussing the implementation of the metadata, or the general idea of _some_ metadata being added? As I noticed a small change to the JSON that can save a few bytes (and a different format that would reduce the size by ~60%)

bowenwang1996:
The implementation detail should probably be discussed in this [PR](https://github.com/nearprotocol/assemblyscript/pull/38). Please note that the commit history is somewhat messed up right now because of some backlog on our end.

vgrichina:
@DanielRX we use JSON at many places cause it's easy to work with, especially when debugging. At some point we'll add compression and may switch to some more efficient (but semantically compatible) format like MessagePack.

I wouldn't worry too much about optimizing size before compression, cause it's not really scalable – compressing all data going to be bigger win then saving on field names.

eriktrautman:
Some UI usage: References https://github.com/nearprotocol/near-wallet/issues/66#issue-451033267

DanielRX:
Given the addition of the decorators for the metadata, are they enforced? 
Can I use `@view_method` on a function that is clearly not view? If so, is it not possible to infer the modifier where one is missing (so as to add it for the compiler to output)
Something like "function `add` was missing a state modifier, it looks to be a view function, please add `@view_method` if that is correct"

bowenwang1996:
I don't think that's possible to do given that the compiler knows nothing about state. So if you annotate a state-changing function with `@view_method`, it will return an error during execution https://github.com/nearprotocol/nearcore/blob/0d163dfa44dd333a0a2a50fe7edd023e392662aa/runtime/runtime/src/state_viewer.rs#L158.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,118 @@
...
+
+For developers, there will be two main changes:
+- Instead of annotating view and change methods in the `initContract` function in `main.js`,
+they will instead annotate the methods of a contract by decorators.
+More specifically, every method is by default a change method, unless annotated by `@view_method`.
+- Every contract will have a `metadata` method that returns a json that serializes the contract methods. For each method,
+the json serialization is of the form `{"parameters": {<param1>: <type1>, .. }, "returnType": <return_type>}`.
+The overall serialization is of the form `{"view_methods": {<method_name>: <method_metadata>, .. }, "change_methods": {<method_name>: <method_metadata>, .. }}`. 
+
+As an concrete example, suppose we have a contract that maintains a counter on chain:
```
vgrichina: I think we need example where methods actually take parameters.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,118 @@
...
+export function getCounter(): i32 {
+  return storage.get<i32>("counter");
+}
+```
+
+This contract has two change methods, `incrementCounter` and `decrementCounter`, as well as one view method, `getCounter`.
+In this case, the metadata we want looks like 
+```json
+{
+  "view_methods":
```
vgrichina: inconsistent usage of snake_case and camelCase throughout JSON. Let's still to single thing, preferably camelCase to match JS expectations.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,118 @@
...
+```json
+{
+  "view_methods": 
+  {
+      "getCounter": {
+        "parameters": [],
+        "returnType": "i32"
+      }
+  },
+  "change_methods":
```
vgrichina: I feel like it might be better to put method type inside of method info vs use it as a group. Seems like it would play better with addition of new method types.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,118 @@
...
+  "view_methods": 
+  {
+      "getCounter": {
+        "parameters": [],
+        "returnType": "i32"
+      }
+  },
+  "change_methods": 
+  {
+      "incrementCounter": {
```
vgrichina: I feel like using an object vs an array here might have impact on how you process this in statically-typed languages. E.g. is it easy to work with such structure in Rust, i.e  iterate over individual methods deserializing info about each individual method?

text/0000-contract-metadata.md
bowenwang1996: But that way it is harder to just get info of all methods of a certain type.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,118 @@
...
+        "parameters": [],
+        "returnType": "void"
+      }
+  }
+}
+```
+and the generated `metadata` method looks like:
+```typescript
+export function metadata(): string {
+    return "{\"view_methods\": {\"getCounter\": {\"parameters\": {}, \"returnType\": \"i32\"},\"change_methods\": {\"incrementCounter\": {\"parameters\": {}, \"returnType\": \"void\"}}, {\"decrementCounter\": {\"parameters\": {}, \"returnType\": \"void\"}}}"
```
vgrichina: if you use single quotes for string escaping double quote won't be necessary

bowenwang1996: Sure I can change it here, but this is consistent with what the code does.

text/0000-contract-metadata.md
bowenwang1996: Rust has this https://docs.serde.rs/serde_json/value/enum.Value.html thing that handles json serialization pretty well. Are you suggesting we should use array instead?

text/0000-contract-metadata.md
bowenwang1996: Sure. But the decorator should be in snake_case right?

text/0000-contract-metadata.md
bowenwang1996: Changed json format

text/0000-contract-metadata.md
bowenwang1996: using array now

text/0000-contract-metadata.md
bowenwang1996: Fixed

text/0000-contract-metadata.md
bowenwang1996: Fixed

text/0000-contract-metadata.md
bowenwang1996: Fixed

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+@view_method
+export function getCounter(): i32 {
+  return storage.get<i32>("counter");
+}
+```
+
+This contract has two change methods, `incrementCounter` and `decrementCounter`, as well as one view method, `getCounter`.
+In this case, the metadata we want looks like 
+```json
+[
```
vgrichina: This looks good but kinda still seems not diverse enough. E.g. it is not clear how do we represent nested objects, arrays, strings, uint128, etc.
Maybe we need few examples to cover it doesn't have to be single contract.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+[summary]: #summary
+
+This NEP introduces contract metadata, which summarizes the content of a given contract in json format.
+Contract metadata allows developers to easily list the methods in a contract and potentially display them to users to
+provide transparency.
+
+# Motivation
+[motivation]: #motivation
+
+Currently there is no direct way for a developer to programmatically list the methods of a contract that is deployed on chain
```
nearmax: Technically, this is not correct. If names were mangled host wouldn't be able to call the method from Wasm. The argument names might be mangled though.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+because the contract code stored on chain is the compiled wasm code, where the methods name and parameters are already mangled.
+As a result, if developers want to display the methods of a contract to an user of the app to provide transparency,
+especially in the case of financial apps, they have no choice but to hardcode them in the front end, which is suboptimal in many ways.
+Furthermore, if developers want to get the list of methods in some contract for some downstream task like data analysis,
+they have no way of doing so. Contract metadata aims to solve the aforementioned problems by providing a convenient way 
+for developers to list the methods of contracts.
+
+# Guide-level explanation
```
nearmax: This NEP should be rephrased in language-agnostic way. We now have two contract languages: AssemblyScript and Rust, since Rust will be the primary language for financial instruments it makes it the first-class citizen on par with AssemblyScript.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
+- Proposal Name: `contract_metadata`
```
nearmax: Consider renaming to contract ABI -- this is the common term used among blockchains.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+they have no way of doing so. Contract metadata aims to solve the aforementioned problems by providing a convenient way 
+for developers to list the methods of contracts.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+For developers, there will be two main changes:
+- Instead of annotating view and change methods in the `initContract` function in `main.js`,
+they will instead annotate the methods of a contract by decorators.
+More specifically, every method is by default a change method, unless annotated by `@view_method`.
```
nearmax: I suggest that all our methods are callable as `near call` and `near view` independently on whether they change the state or not. We should remove the current constraint the prohibits calling `near view` on methods that mutate the state. There are not safety concerns to have this constraint. @evgenykuzyakov @ilblackdragon

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+[
+  {
+    "name": "getCounter",
+    "parameters": [{"name": "amount", "type": "i32"}],
+    "returnType": "i32",
+    "methodType": "view"
+  },
+  {
+    "name": "incrementCounterBy",
+    "parameters": [{"name": "amount", "type": "i32"}],
```
nearmax:
```suggestion
    "parameters": {"name": "amount", "type": "i32"},
```

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+}
+```
+
+This contract has two change methods, `incrementCounter` and `decrementCounter`, as well as one view method, `getCounter`.
+In this case, the metadata we want looks like 
+```json
+[
+  {
+    "name": "getCounter",
+    "parameters": [{"name": "amount", "type": "i32"}],
```
nearmax:
```suggestion
    "parameters": {"name": "amount", "type": "i32"},
```

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+  },
+  {
+    "name": "incrementCounterBy",
+    "parameters": [{"name": "amount", "type": "i32"}],
+    "returnType": "i32",
+    "methodType": "change"
+  },
+  {
+    "name": "decrementCounterBy",
+    "parameters": [{"name": "amount", "type": "i32"}],
```
nearmax:
```suggestion
    "parameters": {"name": "amount", "type": "i32"},
```

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+    "returnType": "void",
```
nearmax: I think `null` is more commonly used to denote absence in json, than `void`. I suggest we research the json conventions.

text/0000-contract-metadata.md
vgrichina: it's not just ABI though, we also need UX-specific info (like contract name, icon, etc).

text/0000-contract-metadata.md
vgrichina: it's not about safety, it's about developer experience. If I call the function and it didn't purist state change silently I won't know WTF is going on.

text/0000-contract-metadata.md
vgrichina: why?

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,117 @@
...
+
+@view_method
+export function getCounter(): i32 {
+  return storage.get<i32>("counter");
+}
+```
+
+This contract has two change methods, `incrementCounter` and `decrementCounter`, as well as one view method, `getCounter`.
+In this case, the metadata we want looks like 
+```json
```
vgrichina: let's return object at the root instead of returning array, so that we can have contract-level metadata, not just method-level

text/0000-contract-metadata.md
nearmax: Why do we need to put a map into an array of one element?

text/0000-contract-metadata.md
nearmax: If you are calling `near view` then you know that it has effect no permanent effect on the state, because you read  `near view --help` before calling it :)

text/0000-contract-metadata.md
DanielRX: If you have a single element for one param, but an array for 2+, the parser must have that extra complexity. If you have 0 - 2 ^64 items, an array will work for all cases and therefore would provide a simple parsing: "The array represents the list of params, an empty array <=> empty params"

text/0000-contract-metadata.md
DanielRX: Are the names stored in a table or are they in a header file, or something else? If the names and types and offsets of all functions are programmatically available already, the metadata wouldn't need any of that?

text/0000-contract-metadata.md
DanielRX: Wouldn't this make the type of `returnType` into `string | null` instead of `string` requiring an explicit typecheck (vs a switch case for `"void"`)?

text/0000-contract-metadata.md
nearmax: I leave this research to @bowenwang1996 . I just know for a fact that names are not mangled, because host calls functions in Wasm by their names.

text/0000-contract-metadata.md
nearmax: Standard JSON serialization/deserialization libraries know how to handle null.

DanielRX: I'm talking about the consumers

I can do `switch(returnType)` with cases `"void"`, `"u32"` and `default` whereas with `null` I must do

`if(returnType === null) {}`
`switch(returnType)` with cases `"u32"` and `default`

text/0000-contract-metadata.md
DanielRX: On this point, does a `view` contract function have the restriction that it can't call a `non-view` function?

text/0000-contract-metadata.md
bowenwang1996: I don't think the names are stored anywhere. Only the function names are available. Also because wasm only has i32, i64, f32, and f64, all the type information is erased.

nearmax: We should figure out it for a fact, for instance, find it in the Wasm spec, or ask it from someone developing Wasm or Wasm-related infra (you can ask Max).

>Also because wasm only has i32, i64, f32, and f64, all the type information is erased.

I don't see how this:
>Also because wasm only has i32, i64, f32, and f64

implies this:
>all the type information is erased.

I suggest we do not do own our deduction, and either find it in the Wasm spec or talk to the Wasm team.

text/0000-contract-metadata.md
bowenwang1996: I don't see where in the description there is language-specific information. In examples, yes, but it is not possible to have real examples without being language-specific. Same with reference-level explanations. If you want to add rust examples, feel free to do so.

text/0000-contract-metadata.md
bowenwang1996: Yes. Right now if you call a state-changing function within a view function, the execution will fail.

text/0000-contract-metadata.md
nearmax: If you are proposing a change to the system, make sure it also includes all the components of the system. Rust is one of such components.

text/0000-contract-metadata.md
bowenwang1996: Do we really want to denote absence here? It seems to me that `void` is just a type.

text/0000-contract-metadata.md
bowenwang1996: Yep. Will change this.

text/0000-contract-metadata.md
bowenwang1996: Sure but I am not familiar with how rust contracts work, so I think it would be better if someone who's familiar with how rust bindings are implemented write the corresponding part.

nearmax: The exact format of how we encode argument types and return types into JSON will affect Rust. Because if we come up with some non-standard method not supported by standard Rust JSON serializer/deserializer then it will be a huge problem and a waste of time tweaking Rust JSON serializer/deserializer to understand it.

nearmax: > Sure but I am not familiar with how rust contracts work, so I think it would be better if someone who's familiar with how rust bindings are implemented write the corresponding part.

I suggest you familiarize yourself. Rust is going to be a huge part of our system, all important contracts are going to be written on it. I cannot be the only person knowing how it works.

text/0000-contract-metadata.md
bowenwang1996: http://webassembly.github.io/spec/core/text/types.html

text/0000-contract-metadata.md
nearmax: We kinda mix type notation from different languages. `void` is JS or C? `f32` is Rust? Let's agree on one thing, I suggest we take the type system from Wasm.

text/0000-contract-metadata.md
nearmax: Can you elaborate on this link?

nearmax: Also, did you look into how Wasm preserves the names of the functions? I am pretty sure they are not mangled, otherwise we would not be able to call functions by name from Wasmer.

text/0000-contract-metadata.md
bowenwang1996: It is actually all assemblyscript notation. Wasm does not have a separate type for `void`.

nearmax: Then I suggest we get rid of `void`. The metadata should have a level of generality of Wasm.

bowenwang1996: So what do we do when there is no return type?

nearmax: Also, we can pass non-primitive types as arguments to the contracts, because we can encode then in JSON. What would be a type description for them in metadata?

text/0000-contract-metadata.md
bowenwang1996: The link says the parameter of a function takes valtype, which is one of `i32`, `i64`, `f32`, `f64`.

text/0000-contract-metadata.md
nearmax: > So what do we do when there is no return type?

Either do not specify "returnType" key in the map, or specify "returnType": null

bowenwang1996: Yes so the plan is to have all the class info listed separately so that binding generation is possible with metadata alone. I'll update the json format to reflect this.

text/0000-contract-metadata.md
nearmax: Cool, I see now that argument names are not preserved. Thank you. Could you do now the research on how function names are handled in Wasm?

text/0000-contract-metadata.md
bowenwang1996: I think it's probably better to not specify `returnType` key as it conforms better to wasm and can also save some space.

nearmax: Thanks, serialization of class type is a problem if we want to make it language-agnostic.

nearmax: > I think it's probably better to not specify returnType key as it conforms better to wasm and can also save some space.

Sounds good.

text/0000-contract-metadata.md
DanielRX: So calling a state-changing function with `near view` is fine, but calling a `view` function that calls a state changing function (with `near view`) is not?

bowenwang1996: No. `near view` is the proposed change, not the status quo.

nearmax:
I think we should compare how other blockchains are doing metadata for their contracts. We need to make sure this design is solid because we won't be able to iterate on it neither through governance nor forks; at least not without losing contracts that are already compiled to Wasm.

Please check https://github.com/paritytech/ink

text/0000-contract-metadata.md
DanielRX: I'm saying that right now, it is an error to call a `view` function in a `non-view` function, so should it not be the same for `near view` to throw an error when calling a `non-view` function?

DanielRX:
> I think we should compare how other blockchains are doing metadata for their contracts. We need to make sure this design is solid because we won't be able to iterate on it neither through governance nor forks; at least not without losing contracts that are already compiled to Wasm.
> 
> Please check https://github.com/paritytech/ink and other blockchains that you can find.

Would it make sense to add some form a `version` to the schema, which would in effect allow iteration (as older contracts can "work" via an compatibility tool for parsing). This would also allow for some level of flexibility for backwards compatible changes to the metadata.

An alternative I would suggest is make the metadata "output" as close to binary and low level as possible, and support the `JSON` or `YAML` or any other human readable format, as part of the near client instead of the chain data directly (this means you can have simple parsers for the basic serialisation formats and enables all developers to work in the format they are already using for configs, static data and general use).

Binary data (simply the name, types and some other basic data like view and public/private) would be fairly robust since you aren't adding anything extra (and anything extra added after such data, could easily be combined with extension).
If you choose `JSON` as a direct output by the contract, you risk being very locked into it.  The output from the contract (in my mind) should be a compact, machine readable output, and the clients should support parsing such data into human readable format. 

One final comment I have is "Why must the data be returned inside of the contract?" Is there no "data" protocol you could use to store this, which would allow the same UX, but without bundling it onto the contract itself. Something like an IPFS hash of the file (with all files pinned) would provide the data in a computer readable format, while having an option to update the format in the future. 
If the files were hosted on a "data" section, you could generate a system for updating a contracts "ABI" or "Metadata" from the data section. This would mean you wouldn't need to worry about already compiled contracts.

text/0000-contract-metadata.md
DanielRX: It was my understanding from the latest call (with @vgrichina) that typescript would be the focus, if rust is the main language you plan to use, would it serve us better to write our contracts in rust (and if so, is there a benefit to rust vs typescript from this POV?)

text/0000-contract-metadata.md
bowenwang1996: I think you misunderstood. Calling a view function in a non-view function is fine, the converse is not.

bowenwang1996:
> I think we should compare how other blockchains are doing metadata for their contracts. We need to make sure this design is solid because we won't be able to iterate on it neither through governance nor forks; at least not without losing contracts that are already compiled to Wasm.
> 
> Please check https://github.com/paritytech/ink and other blockchains that you can find.

Does parity have anything related to metadata? I didn't find any but maybe I missed something.

text/0000-contract-metadata.md
DanielRX: That was a typo yes, 

Calling a `non-view` function with a `view` one, should error.
Therefore, calling a `non-view` function with `near view` should be an error as well?

bowenwang1996: In the proposed change by max, that will not error but the state will not be changed instead.

DanielRX: The two should be consistent though? Either they both error or neither error?

nearmax:
> > I think we should compare how other blockchains are doing metadata for their contracts. We need to make sure this design is solid because we won't be able to iterate on it neither through governance nor forks; at least not without losing contracts that are already compiled to Wasm.
> > Please check https://github.com/paritytech/ink and other blockchains that you can find.
> 
> Does parity have anything related to metadata? I didn't find any but maybe I missed something.

As I mentioned above, the conventional name for contract metadata is ABI. Search for ABI in the link. But also, please don't wait for me to do the research and send you the links, there are prominent blockchains that have ABI, e.g. Ethereum https://solidity.readthedocs.io/en/develop/abi-spec.html#

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,218 @@
...
+[summary]: #summary
+
+This NEP introduces contract metadata, which summarizes the content of a given contract in json format.
+Contract metadata allows developers to easily list the methods in a contract and potentially display them to users to
+provide transparency. 
+It also provides the ability to generate bindings in different languages for the contract, thanks to the class information
+also contained in the metadata.
+
+# Motivation
+[motivation]: #motivation
```
janedegtiareva: I think we need to also include app title in the metadata. 
If we look at playstore app submission process, it requires devs to fill in some display info for the app so that it can be listed in various UX contexts. For us, this might mean displaying the name of the app in the wallet for authorization, and/or NEAR app listing. This data may need to be localized if we want to support multiple languages in the future. 

For more info on play store app submission process see https://themanifest.com/app-development/how-publish-app-google-play-step-step-guide -- see product details section.

vgrichina:
> Would it make sense to add some form a `version` to the schema, which would in effect allow iteration (as older contracts can "work" via an compatibility tool for parsing). This would also allow for some level of flexibility for backwards compatible changes to the metadata.

@DanielRX @bowenwang1996 yes, we totally should have version of our build tools in metadata.

bowenwang1996:
Agree. Will update this when I have time

text/0000-contract-metadata.md
DanielRX: This sounds very much to me as following on from <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md> With the idea of being able to add domain specific values (like app title) in a way that humans can verify it. Would it make sense to include all of this on contract, or out of band / on a separate part of the protocol?

DanielRX:
### ABI vs Metadata
> As I mentioned above, the conventional name for contract metadata is ABI. Search for ABI in the link. But also, please don't wait for me to do the research and send you the links, there are prominent blockchains that have ABI, e.g. Ethereum https://solidity.readthedocs.io/en/develop/abi-spec.html#

The other item that appears for `ABI` when searched for, is in `swift`: <https://github.com/apple/swift/blob/master/docs/ABIStabilityManifesto.md> but this is to do with how binary objects interface, while this proposal is more tailored towards how various properties of a contract can be accessed from the libraries/developer tools.

This to me seems like it is a feature not typically put onto the protocol level but the contract level (which is then put into the "protocol" level due to support for a specific form of ABI).

**The way I would view the naming of the spec (ignoring the fact this is bikeshedding at this point):**
The ABI defines the `interface` of the contract, the metadata encompasses this *and* the static / related properties of the contract (size, name, shard (if it were static), owner (if it were to be a standard to have all contracts have owners)) so I would say ABI as a name only covers the interface.


### Location of storage
**Is there a more logical location to store the metadata of the contract?**

**Should it be stored in the code storage of the contract, or in a separate area (which lends itself to a much more "focused" area to find the data).**

If it were to be stored in a separate area, you could almost imagine it something like
```
/* on deploy (or compiled into constructor) */
metaStorage.set(contractLocation, metaDataString);
```
or better:
```
/* on deploy (or compiled into constructor) */
// metaStorage would use `context.sender` as the key to store it under,
// preventing anyone but the contract editing it
metaStorage.set(metaDataString);
```

DanielRX:
Should the metadata string be signed, allowing for uses and consumers of the data to verify that the metadata belongs to the same contract they intend?

It would be possible to do this on deploy (since the name it's stored at is not included on compile) or it could be bundled with the compiled code if the name was compiled with it.

nearmax:
> The ABI defines the interface of the contract, the metadata encompasses this and the static / related properties of the contract (size, name, shard (if it were static), owner (if it were to be a standard to have all contracts have owners)) so I would say ABI as a name only covers the interface.

I agree, this sounds like a valid separation of abstractions. 
Metadata then consists of:
* ABI -- pure syntactic description of contract interface;
* Semantic ABI -- e.g. minimum value of gas that the function expects, under which conditions it will reject the call, etc.
* static data, like author's name, description, license (similarly to metadata of Rust crates, [example](https://github.com/nearprotocol/near-bindgen/blob/master/near-bindgen/Cargo.toml#L1) );

text/0000-contract-metadata.md
nearmax: Rust contracts-writing UX is optimized for safety while typescript contracts-writing UX is optimized for convenience. It is a trade-off that the contract developers have to make.

text/0000-contract-metadata.md
bowenwang1996: I still think there is some advantage of being explicit. The annotation serves as documentation for developers who might interact with the contract and with metadata, the list of view methods can be easily extracted and displayed, which is better, in my opinion, than looking at the source code and tries to understand for each function, whether it changes the state or not.

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,306 @@
...
+Contract metadata is similar to contract abi in some other blockchains such as Ethereum or Substrate but is more geared
+towards readability and developer friendliness.
+
+# Motivation
+[motivation]: #motivation
+
+Currently there is no convenient way for a developer to programmatically list the methods of a contract that is deployed on chain
+because the contract code stored on chain is the compiled wasm code, where the parameters and type information are already mangled (even though
+method name is not, it is still cumbersome to extract them from wasm binary).
+As a result, if developers want to display the methods of a contract to an user of the app to provide transparency,
```
nearmax: As discussed with @bowenwang1996 before I think the only real application of the contract metadata is input validation. The case when front-end is somehow implemented to not have hardcoded dependencies on method or argument names has several issues:
* As discussed, it is very unlikely that front-end developer will somehow use a contract method without using the name of the method anywhere in their code (because they at least need to identify it) so if the contract method name changes they would have to manually modify the code anyway;
* It opens a huge security concern and such usage should be in general discouraged. We need to have a section in this NEP and documentation explaining the following example: Suppose someone malicious implements a contract with method `burn(gas_in_thousands: u64)` and lets other developers to use it in their front-ends. Then malicious contract developer changes the signature of the function to be `burn(gas: u64)` without announcing to anyone. If front-end did not break because it has automatically rebound to the the argument name the users now are burning 1000 more gas without having any notification. The core problem with metadata is that **it does not describe semantics, it describes syntax**

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,306 @@
...
+In addition, because of the lack of signature of the contract functions, there is no way that the frontend can validate
+that input data is of the right format, which might lead to some spamming attacks.
+Contract metadata aims to solve the aforementioned problems by providing a convenient way 
+for developers to list the methods of contracts.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+With contract metadata, developers, instead of annotating view and change methods in the `initContract` function in `main.js`,
+will instead annotate the methods of a contract by decorators. We propose that view functions be annotated with the `@view` decorator
```
nearmax:
```suggestion
will instead annotate the methods of a contract by decorators. We propose that view functions be annotated with the `@view` decorator in AssemblyScript (in Rust there is no need for additional decorators)
```

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,306 @@
...
+   "contract": {"name":  "Todo list", "description": "A todo list on blockchain!"},
+   "version": "1.0"
+}
+
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+To implement this NEP, we need to modify the binding generation procedure to generate a method called `metadata` that returns
```
nearmax:
```suggestion
To implement this NEP, we need to modify the binding generation procedure in AssemblyScript compiler to generate a method called `metadata` that returns
```

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,306 @@
...
+json serialization of contract metadata described in the previous section. This involves an AST walk to collect the relevant
+ information about functions and classes, as well as mapping types to the types used in metadata.
+
```
nearmax:
```suggestion

In Rust we would need to augment the procedural macro to inject the metadata method.
```

text/0000-contract-metadata.md
```diff
@@ -0,0 +1,306 @@
...
+                 "stateMutability": "view"
+               }
+             ],
+   "classes": [
+                {
+                  "name": "Todo",
+                  "fields": [{"name": "id", "type": "string"}, {"name":  "title", "type": "string"}, {"name": "completed", "type": "bool"}]
+                }
+              ],
+   "contract": {"name":  "Todo list", "description": "A todo list on blockchain!"},
```
vgrichina: contract name / description need to be localizable 

wonder if maybe it makes sense to keep it in separate metadata, so that user can call it like `description(userLocale)` and get this info for given locale.

text/0000-contract-metadata.md
bowenwang1996: How do we plan to do this in Rust? This decorator doesn't enforces anything but rather serves as a way to document the function.

text/0000-contract-metadata.md
frol: From today's hackathon, the distinction between view and change methods should be either strictly enforced (and extensively documented), or their functionality should be merged into a single kind. Otherwise, it gets really confusing and error-prone.

bowenwang1996: Okay it seems like the distinction is confusing. I think we should eliminate the difference and allow dry run of change methods. One thing that is not clear to me is how we should support cross contract view calls. It seems that if you pass 0 gas, the instantiation of wasm module for the call on the second contract will fail.

bowenwang1996:
Update: due to technical difficulties, we will not have contract metadata support for rust in the near future.

text/0000-contract-metadata.md
fckt: Technically it's distinguished by the borrowing type of ABI method in implementation (`fn method(&mut self)` or `method(&self)` for view). See
https://github.com/nearprotocol/near-bindgen/blob/master/examples/status-message/src/lib.rs

nearmax:
We should investigate whether WAI https://medium.com/wasmer/introducing-webassembly-interfaces-bb3c05bc671 can be used here as a drop-in universal solution. Pros: If it works we will be compatible with other Wasm infrastructure that supports WAI, e.g. Wasm package manager.

fckt:
Consider to use [custom section](https://webassembly.github.io/spec/core/appendix/custom.html) as a good place to store metadata. `wasm-bindgen` uses it to store metadata as well as in Wasm Interface Types proposal.

text/0000-contract-metadata.md
fckt: > As discussed with @bowenwang1996 before I think the only real application of the contract metadata is input validation.

As well as it allows to generate a wrapper which would automatically encode/decode ABI (which could be considered as a validation too).

nearmax:
We have implemented metadata for Rust contract in this PR: https://github.com/nearprotocol/near-sdk-rs/pull/99 The format is quite different from the one described in this PR.

The explanation of the metadata format consists of mostly explanation of Borsh Schema format introduced here: https://github.com/nearprotocol/borsh/pull/56 and https://github.com/nearprotocol/borsh/pull/57 Since we don't have that documentation yet, I suggest we do not push this PR and instead later simply add documentation for Borsh Schema and metadata as a part of documentation work. CC @amgando @mikedotexe Therefore I am closing this PR

PR #4, opened 2019-06-28
Nonfungible Token Standard

potatodepaulo:
EDIT: Updated 18-May-2020 with simpler specification

# Nonfungible Token

## Summary

[summary]: #summary

A standard interface for non-fungible tokens allowing for ownership and transfer.

## Motivation

[motivation]: #motivation

Non-fungible tokens (NFTs) have been described in many ways: digital goods, collectible items, unique online assets etc. The current core use case for NFTs is speculating on value and trading unique items. The use case of trading NFTs should be natively supported. This is the most basic set of functions needed to create an interoperable NFT that works in an asynchronous environment.

Prior art:

- [ERC-20 standard](https://eips.ethereum.org/EIPS/eip-20)
- [ERC-721 standard](https://eips.ethereum.org/EIPS/eip-721)
- [ERC-1155 standard](https://eips.ethereum.org/EIPS/eip-1155)

## Guide-level explanation

This token should allow the following:

- Initialize contract once. The given total supply will be owned by the given account ID.
- Get the total supply of created tokens per contract.
- Transfer a token to a new owner.
- Grant access to all token to a third party.
  - A third party account ID will be able to transfer the tokens that they have access to
- Get current balance for a given account ID.
- Transfer tokens from one user to another.

There are a few concepts in the scenarios above:

- **Total supply**. The total number of tokens in circulation.
- **Token owner**. An account ID that owns one or more tokens.
- **Transfer**. Action that moves some amount from one account to another account.
- **Escrow**. A different account from the balance owner who has permission to one or more tokens.
- **Access**. The specific token ID that another account has access to.

#### **Assumptions**

- the Corgi nft contract is `corgi`
- Alice's account is `alice`
- Jeraldo's account is `jerry`
- The NFT contract has been initialized with a nonzero token supply
- There exists a token with the ID of `3`
#### **High-level**

Alice needs to issue one transaction to the Corgi NFT contract to transfer one corgi token to Jeraldo.

#### **Technical calls**

1. `alice` calls `corgi::transfer({"new_owner_id":"jerry", "token_id":3})`

### **Token swap through a third party escrow**

Alice wants to transfer one Corgi NFT through a third party escrow to Jeraldo in exchange for one Sausage NFT.

#### **Assumptions**

- the Corgi nft contract is `corgi`
- the Sausage nft contract is `sausage`
- Alice's account is `alice`
- Jeraldo's account is `jerry`
- The Escrow contract is `escrow`
- The NFT contract has been initialized with a nonzero token supply
- There exists a Corgi token with the ID of `3` and a Sausage token with the ID of `5`
- The Escrow contract manages how the transfer is facilitated and guarantees requirements are met for transfer

#### **High-level**

Both Alice and Jerry will issue asynchronous transactions to their respective contracts, `corgi` and `sausage` to grant access to the escrow to trade tokens on their behalf. `escrow` will call the `sasuage` token contract asynchrounously to transfer the Sausage token to `escrow`. After, `escrow` will also call the `corgi` contract to asynchornously transfer the Corgi token to `escrow`. Then, `escrow` will conduct a transfer to both parties.

- If both of the `transfer_from` calls succeed, then Alice will now own one Sausage token and Jerry will own one Corgi token.
- If one or both of the final `transfer_from` calls fail, then nothing will happen and `escrow` should attempt reissuing the failed transaction.

#### **Technical calls**

1. `alice` makes an async call to `corgi::grant_access({"escrow_account_id":"escrow"})`
2. `jerry`  makes an async call to ``sausage::grant_access({"escrow_account_id":"escrow"})`
3. `escrow` calls `sausage::transfer_from({"owner_id":"jerry", "new_owner_id:"escrow", "token_id": 5})`
    - attaches callback `escrow::on_transfer({"owner_id":"jerry", "token_contract":"sausage", "token_id": 5})`
4. `escrow` calls `corgi::transfer_from({"owner_id":"alice", "new_owner_id:"escrow", "token_id": 3})`
    - attaches callback `escrow::on_transfer({"owner_id":"alice", "token_contract":"corgi", "token_id": 3})`
5. In one Promise:
    1. `escrow` calls `corgi::transfer_from({"owner_id":"escrow", "new_owner_id:"jerry", "token_id": 3})`
        - attaches callback `escrow::on_transfer({"owner_id":"alice", "token_contract:"corgi", "token_id": 3})`
    2. `escrow` calls `sausage::transfer_from({"owner_id":"jerry", "new_owner_id:"escrow", "token_id": 5})`
        - attaches callback `escrow::on_transfer({"owner_id":"jerry", "token_contract":"corgi", "token_id": 3})`
## Template for smart contract in AssemblyScript
At time of writing, this standard is established with several constraints found in AssemblyScript. The first is that interfaces are not an implemented feature of AssemblyScript, and the second is that classes are not exported in the conversion from AssemblyScript to WASM. This means that the entire contract could be implemented as a class, which might be better for code organization, but it would be deceiving in function.
```TypeScript
  type TokenId = u64;
  /******************/
  /* CHANGE METHODS */
  /******************/
  
  // Grant the access to the given `accountId` for the given `tokenId`.
  // Requirements:
  // * The caller of the function (`predecessor_id`) should have access to the token.
  export function grant_access(escrow_account_id: string): void;
  // Revoke the access to the given `accountId` for the given `tokenId`.
  // Requirements:
  // * The caller of the function (`predecessor_id`) should have access to the token.
  export function revoke_access(escrow_account_id: string): void;
  // Transfer the given `tokenId` from the given `accountId`.  Account `newAccountId` becomes the new owner.
  // Requirements:
  // * The caller of the function (`predecessor_id`) should have access to the token.
  export function transfer_from(, owner_id: string, new_owner_id: string, tokenId: TokenId): void;
  // Transfer the given `tokenId` to the given `accountId`.  Account `accountId` becomes the new owner.
  // Requirements:
  // * The caller of the function (`predecessor_id`) should have access to the token.
  export function transfer(new_owner_id: string, token_id: TokenId): void;
  /****************/
  /* VIEW METHODS */
  /****************/
// Returns `true` or `false` based on caller of the function (`predecessor_id) having access to the token
  export function check_access(token_id: TokenId): boolean;

  // Get an individual owner by given `tokenId`.
  export function get_token_owner(token_id: TokenId): string;

```
## Drawbacks
[drawbacks]: #drawbacks
The major design choice to not use a system of approvals for escrow in favor of performance means that it is up to implementors of markets to decide how they manage escrow themselves. This is a dilemma because it increases freedom, while increasing risk of making a mistake on the market side. Ultimately, it will be up to markets and their users to find the best solution to escrow, and we don't believe that approvals is the way to do it. This allows for that solution to be discovered with trail and error. The standard for the market will change, but not the token itself.
This token standard has been whittled down to the simplest fundamental use cases. It relies on extensions and design decisions to be useable.
There are some things that have been in contention in the design of this standard. Namely, the tokenId system relies on unique indices to function. This might cause a problem with use cases that need the `lock` and `unlock` functionality.
In addition, the `grant_access` and `revoke_access` functions act similarly to approvals, but must operate asynchronously and in batch transactions where appropriate.
## Rationale and alternatives
A multi-token standard was considered, as well a standard that allowed for the transfer of any type of token along with the assets associated with this contract. This was foregone for the sake of decoupling the market contracts from the token contracts. The emphasis of this standard is now on simplicity and flexibility. It allows for any type of token to interface with any type of market that accepts this standard. The explicit goal is to maximize developer freedom with a rigid enough foundation to make a standard useful.
## Unresolved questions
[unresolved-questions]: #unresolved-questions
Primarily edge cases for various applications should be surfaced. For example, the use case of creating an in-game store is different than creating a token for tracking real-world objects digitally. This token attempts to create a standard for both.
Neither a market standard nor an escrow system is addressed here. These should exists in the future, but are purposefully left separate. An item should not care about the place it is sold or agreed on.
The ability to `lock` and `unlock` tokens is a likely requirement for many use cases, but there are many challenges around this. The initial solution to solely rely on callbacks was abandoned in favor of an access system that allows escrow contracts to lock and transfer tokens.
Finally, in the original draft, metadata was included in the model for tokens. It was clear through some basic implementations that this is not ideal since users may want to store metadata elsewhere. This could be entirely offchain, or in a separate contract. This creates an unsolved problem of synchronizing metadata with contracts, and needs more design work. 
## Future possibilities
The next step in the development of this standard is extending it further in a new standard that addresses spcifically how a generic and safe escrow would function, and how metadata should be handled based on the specific use cases of tokens implemented. In addition, an importable module should be developed, allowing developers to integrate a token system with little overhead. Alternative uses of this token are of high interest. Known uses for nonfungible tokens include collectible items online, and item systems in games as discussed throughout. There are many uses cases yet to be invented. These might include tokens for supply chain or even tokens for shared custody of physical items. The possibilities are ultimately going to be driven by community use.


0004-market-integration-token-standard.md
```diff
@@ -0,0 +1,123 @@
...
+
+```TypeScript
+class MarketStandardToken {
+  getTokenTypes():TokenType[]
+  getToken(tokenId:string):Token
+  getTokensByOwner(ownerId:string):Token[]
+
+  // getCount is equivalent of balanceOf
+  getCount(tokenType: TokenType, ownerId: string)
+  getTokenValue(tokenId:string):u64
```
evgenykuzyakov: This should be u128

0004-market-integration-token-standard.md
```diff
@@ -0,0 +1,123 @@
...
+  getTokenData(tokenId:string):bytes
+
+  // It's up to the developer of a token to decide how tokens are generated.
+  // Minting should return the Token with a unique ID attached.
+  mintToken(token:Token):Token
+
+  // Note that transferring can be done on 1 or more tokens
+  transfer(recipientId:string, ownerId:string, tokenIds:string[]):void
```
evgenykuzyakov: When we discussed this for ERC-20 example, we realized we need a callback to the receiver's contract. E.g.
On ERC-20 you would have a function:

// Token's owner would be the immediate sender_id (that initiated this call).

    transfer(receiver_id: string, amount: u128, method_name: string, args: bytes): void

The contract should initiate an async call to `receiver` with `method_name`, and pass the `sender_id`, `amount` sent and `args`. And revert this transfer if the callback fails with error. This ensures composability of the contracts in the async settings. While the transfer is in progress, the amount is locked and can't be withdrawn.

evgenykuzyakov: Also we may want to separate `new_owner_id` and the `callback_receiver_id` in the transfer method.
Maybe we need to have 2 functions, simple transfer and transfer with callback

0004-market-integration-token-standard.md
evgenykuzyakov: This function also need comments

potatodepaulo: It looks like we're going to do away with value associated directly with tokens. The consensus is that this can be implemented better in any good market contract.

0004-market-integration-token-standard.md
```diff
@@ -137,25 +157,28 @@ At time of writing, this standard is established with several constraints found
...
 
 ```TypeScript
 export class Token {
-  id: string;
-  tokenType: TokenType;
-  data: bytes;
+  id:u64;
+  tokenTypeId:u32;
+  // serialized json
+  data:string;
```
DanielRX: Is this data intended to be `readonly` ? 
If not, how would a consumer parse the data in a deterministic way to arrive at the same JSON as the original token.
It seems to be that having the data change would "promote" people using `data` as a way to store things like "level" or "value" of an NFT, leading to immutability issues.

0004-market-integration-token-standard.md
```diff
@@ -65,38 +80,43 @@ At time of writing, this standard is established with several constraints found
...
+    unimplemented();
+  }
+
+  export function getCountByOwner(ownerId: string):u64 {
     unimplemented();
   }
 
-  export function transfer(recipientId:string, tokenIds:string[], onTransfer:PayloadCallback = null):void {
+  export function transfer(recipientId:string, tokenIds:u64[], onTransfer:PayloadCallback = null):void {
     unimplemented();
```
DanielRX: Will this not work better as an overloaded function (allowing for optimised transfers when `tokenIds.length === 1`)?
In terms of `onTransfer`, would a default of `() => {}` or `no-op` be clearer (since `null` would imply that you would need to do null checking, instead of unconditionally applying the cb.

0004-market-integration-token-standard.md
```diff
@@ -65,38 +80,43 @@ At time of writing, this standard is established with several constraints found
...
   }
 
-  export function getCountByOwner(tokenType: TokenType, ownerId: string) {
+  export function getAllTokenTypes():TokenType[] {
     unimplemented();
   }
-  
-  export function getTokenData(tokenId:string):bytes {
+
+  export function getTokensByOwner(ownerId:string, startTokenId:u64 = 0, limit:u32 = 10):Token[] {
```
DanielRX: Is `startTokenId` an index, or an id (an owner may have `ids: [3, 7, 9])`?

0004-market-integration-token-standard.md
```diff
@@ -65,38 +80,43 @@ At time of writing, this standard is established with several constraints found
...
     unimplemented();
   }
 
-  export function getTokenOwner(tokenId:string):string {
+  export function getToken(tokenId:u64):Token {
     unimplemented();
   }
 
-  export function getTokenTypes():TokenType[] {
+  export function getTokenType(tokenTypeId:u32) {
```
DanielRX: No return type here

0004-market-integration-token-standard.md
DanielRX: This also is one `user`, many tokens.
Is there scope for many `user` + many tokens (much like a `zip` + `map`)

Is the onTransfer called once? After the first transfer? After the last?

0004-market-integration-token-standard.md
```diff
@@ -107,24 +127,24 @@ At time of writing, this standard is established with several constraints found
...
     unimplemented();
   }
 
-  export function approveEscrow(escrowId:string, tokenIds:string[]):void {
+  export function approveEscrow(escrowId:string, tokenIds:u64[]):void {
     unimplemented();
   }
 
-  export function cancelEscrow(escrowId:string, tokenIds:string[]):void {
+  export function cancelEscrow(escrowId:string, tokenIds:u64[]):void {
```
DanielRX: To save the git diffs in the future, would it be worth having a `tokenId` type (merely an alias to `u64` at present) so it is simpler to edit them all?

potatodepaulo: Good idea, integrating this for all of these.

0004-market-integration-token-standard.md
potatodepaulo: Added explanations in latest push

0004-market-integration-token-standard.md
potatodepaulo: The idea of overloading is a good idea, but is not yet supported in AssemblyScript. This is likely going to come soon.

potatodepaulo: Also, onTransfer will be called each transfer. Transfers happen as batches. This is not truly a callback, since it may be called before the actual transfer occurs.

0004-market-integration-token-standard.md
```diff
@@ -0,0 +1,196 @@
...
+
+# Summary
+[summary]: #summary
+
+A standard interface for non-fungible tokens allowing for ownership and transfer, specifically targeting third-party marketplace integration.
+
+# Motivation
+[motivation]: #motivation
+
+Non-fungible tokens (NFTs) have been described in many ways: digital goods, collectible items, unique online assets etc. The current core use case for NFTs is speculating on value and trading unique items. The use case of trading NFTs should natively supported for this emerging market.
```
bowenwang1996: "should natively supported" -> "should be natively supported"

0004-market-integration-token-standard.md
```diff
@@ -0,0 +1,196 @@
...
+
+```
+
+## Models
+
+```TypeScript
+export class Token {
+  id:u64;
+  tokenTypeId:TokenTypeId;
+  data:string;
```
bowenwang1996: I think it would be helpful if there is more explanation on how the data field should be used,

potatodepaulo: I'll add an example

0004-market-integration-token-standard.md
```diff
@@ -0,0 +1,197 @@
...
+
+  export function getCountByOwner(ownerId: string):u64 {
+    unimplemented();
+  }
+
+  export function transfer(ownerId:string, recipientId:string, tokenIds:TokenId[], onTransfer:PayloadCallback = null):void {
+    unimplemented();
+  }
+
+  class PayloadCallback {
```
vgrichina: why do we need callback? Like if you call this from other smart contract you already have built-in ability to receive callbacks.

See https://github.com/nearprotocol/NEARStudio/pull/102/files#diff-041db770fcda1cafea3812c8c8561f8dR47

It's somewhat ugly now, but we'll be able to codegen once metadata lands.

evgenykuzyakov:
Here is a comment on approvals from discord chat (to keep a record):

In async setting like NEAR other contracts can't rely on approvals being valid. Cause there are always can be a transaction in between the initial approval and removing the approval. In NEP#4 we use transfer with callback to ensure the transfer can either succeed or reverted in secure way.
One reason to have approvals is to share access to a single token by multiple contracts at the same time, an example for this can be listing an NFT on multiple exchanges. Once the sale is in the process the async transaction from Ex will fly to the NEP#4 contract to call transferFrom with callback. If the approval is still valid, the sale will be completed, if not, the sale will be rejected.
Multiple approvals create O(N) operation when moving a token, to remove all approvals at unlock. It's doable, but we need to build an abstract class for this to make it simple to use by developers.

0004-market-integration-token-standard.md
vgrichina: yes, it's expected that you want to store stuff like crypto-corgi DNA when you mint a token and then not change it

people can mess it up, but you really can mess up anything if not following convention while implementing token

0004-market-integration-token-standard.md
evgenykuzyakov: That's not a callback. The name is bad. It's a functionCall that the token contract will do to confirm the transfer. If the function call fails, the transfer will be reverted. While the function call happens, the token is locked.

evgenykuzyakov:
After more offline discussions and talking to a few folks @potatodepaulo and I decided to update the spec to look similar to Eth spec with a few key differences. We propose the following API:

```typescript
// Grant the access to the given `account_id` for the given `tokenId` .
// Requirements:
// * The caller of the function (`predecessor_id`) should have access to the token.
// * The token should not be locked.
function grantAccess(tokenId: TokenId, account_id: string): void;

// Revokes the access to the given `account_id` for the given `tokenId` .
// Requirements:
// * The caller of the function (`predecessor_id`) should have access to the token.
// * The token should not be locked.
function revokeAccess(tokenId: TokenId, account_id: string): void;

// Locks the given `tokenId` to the caller of the function (`predecessor_id`).
// The `predecessor_id` becomes the owner of the lock.
// Requirements:
// * The caller of the function (`predecessor_id`) should have access to the token.
// * The token should not be locked.
function lock(tokenId: TokenId): void;

// Unlocks the given `tokenId`. Removes lock owner.
// Requirements:
// * The caller of the function (`predecessor_id`) should be the owner of the lock.
// * The token should be locked.
function unlock(tokenId: TokenId): void;

// Transfers the given `tokenId` to the given `account_id`.  Account `account_id` becomes the new owner.
// The token unlocks (if it was locked) and all access is revoked except for
// the new owner.
// Requirements:
// * The caller of the function (`predecessor_id`) should have access to the token.
// * If the token is locked, the locked owner should be `predecessor_id`.
function transfer(tokenId: TokenId, account_id: string): void;
```

NOTES:
* In async setting the `predecessor_id` is the account ID of the immediate previous caller. If a user initiates a new transaction, then the user's account ID is the predecessor. If a `contract_a` creates an async call towards the token contract, then the `contract_a` is the predecessor.
* Each token contains the following properties:
  - owner_id - account ID of the owner
  - lock_owner_id - optional field. An account ID of the owner of the lock. Or the field is not set, if the token is unlocked.
  - access - a set of account IDs who currently hold access to the token.
* Token owner always has access to the token. But the owner can't override locks.

Examples on how it work:
* When DEX tries to exchange tokens between 2 NFT contracts, it calls 2 async functions in parallel to lock exchanged tokens.
The DEX contract receives 2 callbacks and makes a decision:
  - If both locks succeeded, DEX calls transfer to the new corresponding owners. The trade considered successful.
  - If one of the locks succeeds, but another fails, the DEX issues `unlock` for the successful lock. And the trade considered failed.
  - The benefit of such design is the token can be simultaneously listed on multiple exchanges.
When a trade fails by the other side, the token doesn't lose access and doesn't get delisted from other exchanges.
* Simple transfer. The owner can just call `transfer` to any other user.
* Temporary lock. The owner can call `lock` to temporary put a hold on the token which prevents other accounts with access from acting.

k06a:
I think the best way for swap NFT is to use escrow smart contract. When escrow will own NFT no one will be able to make permitted changes to it. But if you still wish to use your NFT while it is in the order we can do it in 2 steps. Buyer can call a method on escrow contract and escrow will `transferFrom` NFT from the seller. Then if escrowed NFT looks good buyer can swap it according to initial sellers order.

evgenykuzyakov:
@k06a The following doesn't work in async design unless you own both tokens.
```
order.asset.transferFrom(msg.sender, this, order.assetAmount);
order.nft.transferFrom(order.seller, this, order.nft_id);
```

k06a:
@evgenykuzyakov what do you mean? Approve should be enough to do `transferFrom`. I expected that async environment not adding any breaking restrictions not to break existing concepts.

evgenykuzyakov:
approve can be invalidated before `transferFrom` arrives. This can have one of `transferFrom` succeeds and another fails. I'm writing a DEX implementation example to demonstrate the concept. I'm not convinced we have to separate the escrow and NFT contracts, because exchanges are either have to be able to handle both direct listing by NFT contracts and through escrow contracts, or only escrow contracts. For the 2nd, people are forced to use escrow for selling which doesn't simplify the wallet designs. For the 1st case, DEX design becomes more complicated, but still require handle coins on escrow and in the wallet.

k06a:
@evgenykuzyakov we should just revert in case of any `transferFrom` failed. Will this cancel already transferred one?

evgenykuzyakov:
>>> we should just revert in case of any transferFrom failed. Will this cancel already transferred one?

No, since it's async. We can only revert the local state in case of failure. Any async operation that was successfully issued will not be canceled. So when you call 2 `transferFrom`, they are not executed immediately, but just scheduled to execute on corresponding contracts.

Here is the DEX example:

```
// Arbitrary number, but has to be enough for the worst case.
const GAS_FOR_A_CALL = 1000000;
const GAS_TO_PROCESS_CALLBACKS = 3 * GAS_FOR_A_CALL;
const RESULT_STATUS_SUCCESS = 1;

export function buy(orderId: OrderId) {
  let order = orders.get(tokenId);
  // NFT for sale
  let tokenId = order.tokenId;
  let tokenContract = order.tokenContract;
  // Buying price (assuming some fungible tokens)
  let price = order.price;
  let assetContract = order.priceContract;

  // Should lock the order and save it.
  order.lock();

  // Who's trying to buy (the immediate predecessor account id who called this method, it's can be
  // different from the signer of the transaction).
  let orderBuyer = context.predecessor;

  // Locking NFT token
  let sellSideLockArgs: TokenContractLockArgs = {
    tokenId,
  };
  let promiseSellSideLock = ContractPromise.create(tokenContract, "lock", sellSideLockArgs.encode().serialize(), GAS_FOR_A_CALL);

  // Locking amount token (we don't have a standard for this, so improvising)
  let buySideLockArgs: AssetContractLockArgs = {
    owner: orderBuyer,
    amount: price,
  };
  let promiseBuySideLockArgs = ContractPromise.create(assetContract, "lock", buySideLockArgs.encode().serialize(), GAS_FOR_A_CALL);

  // Joining promises to attach a callback for both of them.
  let promiseJoin = ContractPromise.all([promiseSellSideLock, promiseBuySideLockArgs]);

  // Attaching a callback to process the results.
  let processCallbackArgs: ProcessCallbackArgs = {
    orderId,
    orderBuyer,
  }
  let promiseProcessCallbacks = promiseJoin.then(context.contractName, "processCallbacks", processCallbackArgs.encode().serialize(), GAS_TO_PROCESS_CALLBACKS);

  promiseProcessCallbacks.returnAsResult();
}

export function processCallbacks(orderId: OrderId, orderBuyer: string) {
  // Checking that this function was called by us.
  assert(context.predecessor == context.contractName);

  let order = orders.get(tokenId);

  // NFT for sale
  let tokenId = order.tokenId;
  let tokenContract = order.tokenContract;
  let tokenOwner = order.tokenOwner;
  // Buying price (assuming some fungible tokens)
  let price = order.price;
  let assetContract = order.priceContract;

  // Collecting results
  let results = ContractPromise.getResults();
  if (results[0].status == RESULT_STATUS_SUCCESS && results[1].status == RESULT_STATUS_SUCCESS) {
    // Success, exchanging assets to the new owners and closing the order.

    // Transferring NFT token
    let sellSideTransferArgs: TokenContractTransferArgs = {
      tokenId,
      account_id: orderBuyer,
    };
    // We don't care about result of the execution, because it should succeed after lock.
    ContractPromise.create(tokenContract, "transfer", sellSideLockArgs.encode().serialize(), GAS_FOR_A_CALL);

    // Transferring the ERC20 like asset
    let buySideTransferArgs: AssetContractTransferArgs = {
      owner: orderBuyer,
      amount: price,
      new_owner: tokenOwner,
    };
    ContractPromise.create(assetContract, "transfer", buySideTransferArgs.encode().serialize(), GAS_FOR_A_CALL);

    // Closing the order. The sale was done.
    order.close();
  } else {
    // One or both of the locks failed. Unlocking successful locks and the order
    if (results[0].status == RESULT_STATUS_SUCCESS) {
      // Token lock succeeded, unlocking
      let sellSideUnlockArgs: TokenContractUnlockArgs = {
        tokenId,
      };
      ContractPromise.create(tokenContract, "unlock", sellSideUnlockArgs.encode().serialize(), GAS_FOR_A_CALL);
    }

    if (results[1].status == RESULT_STATUS_SUCCESS) {
      // Asset lock succeeded, unlocking
      let buySideUnlockArgs: AssetContractUnlockArgs = {
        owner: orderBuyer,
        amount: price,
      };
      ContractPromise.create(assetContract, "unlock", buySideUnlockArgs.encode().serialize(), GAS_FOR_A_CALL);
    }

    // Unlocking the order
    order.unlock();
  }
}
```

k06a:
@evgenykuzyakov but I think we can easily send fetched token back in case of second failure:
```
const promise1 = order.asset.transferFrom(msg.sender, this, order.assetAmount);
const promise2 = order.nft.transferFrom(order.seller, this, order.nft_id);
await Promise.all([promise1, promise2]);

if (!promise1 && !promise2) {
    return;
}
if (promise1 && !promise2) {
    order.asset.transfer(msg.sender, order.assetAmount);
    return;
}
if (promise2 && !promise1) {
    order.nft.transfer(order.seller, order.nft_id);
    return;
}


```

k06a:
Or better provide something like this `shouldSuccessAll`:
```
function shouldSuccessAll(calls: callback[], cancels: callback[]) {
    const results = await calls.map(call => call.execute());
    if (!results.all()) {
        await results.map((v,i) => {
            if (!v) {
                cancels[i].execute();
            }
        });
        return false;
    }

    return true;
}

await ([
    () => { order.asset.transferFrom(msg.sender, this, order.assetAmount); },
    () => { order.nft.transferFrom(order.seller, this, order.nft_id); },
], [
    () => { order.asset.transfer(msg.sender, order.assetAmount); },
    () => { order.nft.transfer(order.seller, order.nft_id); },
])
```

evgenykuzyakov:
And that's why you need locks to not lose approvals in case you revert the transfer. Otherwise, if the token1 fails to transfer, the token2 reverted and all approvals are removed, which delists token2 from any future transfers until the owner approves it again.

k06a:
Ok, got it. Let me think of it.

k06a:
> @k06a The following doesn't work in async design unless you own both tokens.
> 
> ```
> order.asset.transferFrom(msg.sender, this, order.assetAmount);
> order.nft.transferFrom(order.seller, this, order.nft_id);
> ```

It seems maker allowance will not be wasted in case we first try to grab funds from taker. Which have single tx with multiple actions: approve and exchange calls.
```
require(order.asset.transferFrom(msg.sender, this, order.assetAmount));
require(order.nft.transferFrom(order.seller, this, order.nft_id));
```

evgenykuzyakov:
@k06a It works for the trade that is initiated by the buyer. But without locks we are limited some use-cases. E.g. exchange of multiple NFTs at the same (e.g. 3 NFTs for asset). Offline trade, e.g. when both parties have approvals for DEX, but someone else triggers the trade (maybe the condition becomes satisfied over time).
Without lock pattern we also put responsibility for the correct order implementation. It would require more steps (trade, take from buyer, confirm, take from seller, confirm, exchange).

So I would prefer to keep the locks. So long as we provide an abstract class for locks, transfers, etc.

potatodepaulo:
Talk in office documented here:

Flow Diagram NEP4

The current challenge that's not completely solved is reducing the number of transactions each party would have to send or sign.

In this case, the `lock` acts basically as "transfer in-progress". If lock doesn't work, the transfer is killed, but the granted access is not cleared. Access is only cleared manually by the user, OR on a completed transfer.

DanielRX:
@potatodepaulo is "grantAccess" for more than one token? If not, couldn't grant access and lock happen in one go? i.e have the escrow forward a signature to the NFTor have the NFT trigger the escrow (on grant access)

potatodepaulo:
> @potatodepaulo is "grantAccess" for more than one token? If not, couldn't grant access and lock happen in one go? i.e have the escrow forward a signature to the NFT or have the NFT trigger the escrow (on grant access)

They could be one step. The reason to keep them separate in this case is for when you want to list the token in multiple places (DEX, marketplace etc.). This way, you grant access for a future lock so you can sell while you sleep (E.g. no need to sign anything after granting access unless something goes wrong).

> EDIT: Would it not work to have "lockFor" which takes a user and locks the token such that only the user can unlock it (meaning you would effectively lock the token and give access in one go)?

A problem with "only the user passed can unlock" is in the case of rollback for some kind of failure in the transaction.

DanielRX:
> A problem with "only the user passed can unlock" is in the case of rollback for some kind of failure in the transaction.

But is that not how your graph works? If the user can unlock the token after the Escrow locked it, that would make the whole lock idea pointless, and the escrow then _could_ have one token fail after locking?

potatodepaulo:
> But is that not how your graph works? If the user can unlock the token after the Escrow locked it, that would make the whole lock idea pointless, and the escrow then _could_ have one token fail after locking?

Good point. We're implementing starting this week. We'll see how `lock` works if we it is functionally the same as `lockFor`.

Thanks :)

potatodepaulo:
Just a quick update on this. From several conversations about this and #21 , there are problems with locks as we originally planned them. The current solution to this is https://github.com/nearprotocol/NEPs/pull/26. When this lands, we can merge this NEP.

robert-zaremba:
Hi,
Would love to contribute to this standard. We can do it much better than what's ERC20, and make it more friendly for real world especially regulatory environments (PS: I was working with few projects facing regulations).
I've just had a chat with Jasper (from Flux) to improve the collaboration process. Github pull request is not very friendly. The latest "changes" are hard to parse (especially if other file formats are include). Idea would be to use hackmd and collaborate there before doing doing a PR.

robert-zaremba:
So... if I would like to propose some alternative standard for fungible tokens, what's the best way? Create a spec and do it as a new Issue? I could then paste a link to HackMD - so we can test it out?

bowenwang1996:
@robert-zaremba we want to keep all discussions within github so that people don't have to go to multiple different places to check out proposals and discussions. Feel free to suggest changes in this PR. But if you want to suggest something entirely different, feel free to open another PR.

potatodepaulo:
> Hi,
> 
> Would love to contribute to this standard. We can do it much better than what's ERC20, and make it more friendly for real world especially regulatory environments (PS: I was working with few projects facing regulations).
> 
> I've just had a chat with Jasper (from Flux) to improve the collaboration process. Github pull request is not very friendly. The latest "changes" are hard to parse (especially if other file formats are include). Idea would be to use hackmd and collaborate there before doing doing a PR.

Hey @robert-zaremba, make sure to check out https://github.com/nearprotocol/NEPs/pull/21. That’s the fungible token standard we’ve worked on. That’s the ERC-20-like token.

janedegtiareva:
// Create a new type of token within the same contract with given `data` as metadata/display data and `totalSupply`.
  // Requirements:
  // * token types should be stored in collection ordered by index with index serving as TokenTypeId.
  export function mintTokenType(data: string, totalSupply: u64): TokenTypeId;
This won't work if we allow deleting token types. Are we okay with that?

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### **Technical calls**
+
+1. `alice` makes an async call to `corgi::grantAccess({"escrow_account_id":"escrow", "token_id":3})`
+2. `jerry`  makes an async call to ``sausage::grantAccess({"escrow_account_id":"escrow", "token_id":5})`
+3. In one Promise
+    1. `escrow` calls `sausage::transferFrom({"owner_id":"jerry", "new_owner_id:"alice", "token_id": 5})`
+        1. attaches callback `escrow::onTransfer({"owner_id":"jerry", "token_contract":"sausage", "token_id": 5})`
+    2. `escrow` calls `corgi::transferFrom({"owner_id":"alice", "new_owner_id:"jerry", "token_id": 3})`
+        1. attaches callback `escrow::onTransfer({"owner_id":"alice", "token_contract:"corgi", "token_id": 3})`
```
evgenykuzyakov, line 90: This is wrong example, because if one promise failed, the other succeeds and escrow can't do anything with the successful transfer. This would mean the owner has both assets. You need to adjust this example to follow similar logic to https://nomicon.io/Standards/Tokens/FungibleToken.html#multi-token-swap-on-dex

Basically escrow first takes both assets into ownership. This way escrow contract can be sure that it holds both assets, then it can swap them respectively.

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,196 @@
...
+  type TokenId = u64;
+
+  /******************/
+  /* CHANGE METHODS */
+  /******************/
+  
+  // Grant the access to the given `accountId` for the given `tokenId`.
+  // Requirements:
+  // * The caller of the function (`predecessor_id`) should have access to the token.
+  export function grantAccess(tokenId: TokenId, accountId: string): void;
```
evgenykuzyakov, line 131: Why give access per `tokenId` instead of per `accountId`? You may not want to approve every asset to a given contract. Likely if you trust the contract for one tokenId, then you can trust it with all of your tokens.

Also we need to follow NEP#21 Rust code-style for method names and variables, to be consistent across standards, so method should be like this:
```
 export function grant_access(escrow_account_id: string): void;
```

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,196 @@
...
+  export function checkAccess(tokenId: TokenID): boolean;
+
+  // Get an individual owner by given `tokenId`.
+  export function getTokenOwner(tokenId: TokenId): string;
+
+  // Allows for the retrieval of tokens from given `ownerId` starting from given `startTokenIdx`.
+  // up to the provided `limit.`
+  // Requirements:
+  // * Starting index should not be larger than total number of tokens.
+  export function getTokensByOwner(ownerId: string, startTokenIdx: u64 = 0, limit: u32 = 10): Token[];
```
evgenykuzyakov, line 164: We can't get this in the default implementation, because it needs complicated data structure to iterate on every token. Plus the initial owner need to own all tokens, which is not trivial as well.

So this can go into an extension.

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,196 @@
...
+
+  // Revoke the access to the given `accountId` for the given `tokenId`.
+  // Requirements:
+  // * The caller of the function (`predecessor_id`) should have access to the token.
+  export function revokeAccess(tokenId: TokenId, accountId: string): void;
+
+  // Transfer the given `tokenId` from the given `accountId`.  Account `newAccountId` becomes the new owner.
+  // Requirements:
+  // * The caller of the function (`predecessor_id`) should have access to the token.
+  export function transferFrom(tokenId: TokenId, accountId: string, newAccountID: string): void;
```
evgenykuzyakov, line 141: If you follow NEP#21 standard, then the order of the arguments may also be changed. But for JSON it's not important, just for consistency

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,196 @@
...
+
+
+  // Transfer the given `tokenId` to the given `accountId`.  Account `accountId` becomes the new owner.
+  // The token unlocks (if it was locked) and all access is revoked except for the new owner.
```
evgenykuzyakov, line 145: If you don't grant access per token, then you don't need to change `access` for the transferred token, which makes implementation simpler.

specs/Standards/Tokens/NonFungibleToken.md
potatodepaulo: We had discussed this a couple weeks ago and I thought the consensus was to to use language specific standards. We can make it so the front end always has the same style calls either way.

janedegtiareva:
The latest version seems to not have a way to create tokens. Is that supposed to be not part of the standard?

janedegtiareva:
We should remove transfer_from.

Copy of slack discussion for posterity/visibility
Evguenia Degtiareva  2 hours ago
Q about transfer_from: there is a param owner_id. We should never trust this since we have this data stored internally. Can we remove this param? The consequence is that transfer_from becomes the same as transfer, and we can get rid of it completely and only have transfer @Peter (edited) 

Peter DePaulo  1 hour ago
So the point is that for transfer, you check predecessor id. For transfer_from, you have access to someone else’s token.

Peter DePaulo  1 hour ago
For more complex stuff, it’s valuable which is why it’s in the base standard  (edited) 

Peter DePaulo  1 hour ago
It’s specifically when owner_id is not equal to predecessor_id

Evguenia Degtiareva  1 hour ago
it's not about whether owner_id is equal to predecessor_id

Evguenia Degtiareva  1 hour ago
we have owner_id stored on chain. so if have an owner_id passed in, and it doesn't match what we have on chain, we have to ignore what's passed in for security purposes. We never really want to trust the caller on what the token owner is, hence there is no need to have that as a parameter (
Peter DePaulo  1 hour ago
Yeah okay, I can’t come up with a counter argument other than it will be useful for future use cases. Make sure to add a commit to the pr removing transfer_from then. @Evgeny Kuzyakov comments?

Evgeny Kuzyakov  1 hour ago
I think it’s good idea to remove it. Because we have token_id that defines the current owner

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,173 @@
...
+
+// Revoke the access to the given `accountId` for the given `tokenId`.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function revoke_access(escrow_account_id: string): void;
+
+// Transfer the given `tokenId`. Account `newAccountId` becomes the new owner.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function transfer(new_owner_id: string, tokenId: TokenId): void;
```
evgenykuzyakov, line 126:
```suggestion
export function transfer(new_owner_id: string, token_id: TokenId): void;
```

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,173 @@
...
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function transfer(new_owner_id: string, token_id: TokenId): void;
+
+
+/****************/
+/* VIEW METHODS */
+/****************/
+
+// Returns `true` or `false` based on caller of the function (`predecessor_id`) having access to a user's tokens
+export function check_access(account_id: string): boolean;
```
evgenykuzyakov, line 139: View methods don't have `predecessor_id`. Need to explicitly provide `owner_id` and `escrow_account_id `

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,173 @@
...
+// Transfer the given `tokenId`. Account `newAccountId` becomes the new owner.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function transfer(new_owner_id: string, tokenId: TokenId): void;
+
+// Transfer the given `tokenId` to the given `accountId`.  Account `accountId` becomes the new owner.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function transfer(new_owner_id: string, token_id: TokenId): void;
+
```
evgenykuzyakov, line 132: duplicate

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,173 @@
...
+At time of writing, this standard is established with several constraints found in AssemblyScript. The first is that interfaces are not an implemented feature of AssemblyScript, and the second is that classes are not exported in the conversion from AssemblyScript to WASM. This means that the entire contract could be implemented as a class, which might be better for code organization, but it would be deceiving in function.
+
+```TypeScript
+type TokenId = u128;
+
+/******************/
+/* CHANGE METHODS */
+/******************/
+
+// Grant the access to the given `accountId` for all tokens that account has.
```
evgenykuzyakov, line 113: comments needs to be updated

evgenykuzyakov:
For some reason nomicon doesn't render it. https://nomicon-pr-4.onrender.com/Standards/Tokens/NonfungibleToken.html

Might need to be added to https://nomicon-pr-4.onrender.com/Standards/README.html

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,167 @@
...
+
+// Revoke the access to the given `account_id` for the given `token_id`.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function revoke_access(escrow_account_id: string): void;
+
+// Transfer the given `token_id`. Account `new_account_id` becomes the new owner.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function transfer(new_owner_id: string, token_id: u128): void;
```
evgenykuzyakov, line 125: Does assembly script interface uses string for `u128` serialization and deserialization?

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,167 @@
...
+
+```TypeScript
+
+/******************/
+/* CHANGE METHODS */
+/******************/
+
+// Grant access to the given `escrow_account_id` for all tokens that account has.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the tokens.
```
evgenykuzyakov, line 114: That's a little confusing in terms or requirements and part for `for all tokens that account has.` Maybe better to restructure it, to remove requirements and explain that the `predecessor_id` gives access to `escrow_account_id` for all tokens

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,167 @@
...
+export function grant_access(escrow_account_id: string): void;
+
+// Revoke the access to the given `account_id` for the given `token_id`.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
```
evgenykuzyakov, line 119: Please update this as well

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,167 @@
...
+export function transfer(new_owner_id: string, token_id: u128): void;
+
+/****************/
+/* VIEW METHODS */
+/****************/
+
+// Returns `true` or `false` based on `escrow_account_id` having access to tokens owned by `owner_id`
+export function check_access(escrow_account_id: string, owner_id: string): boolean;
+
+// Get an individual owner by given `token_id`.
```
evgenykuzyakov, line 134:
```suggestion
// Get the owner's account ID of the given `token_id`.
```

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,167 @@
...
+Both Alice and Jerry will issue asynchronous transactions to their respective contracts, `corgi` and `sausage` to grant access to the escrow to trade tokens on their behalf. `escrow` will call the `sasuage` token contract asynchrounously to transfer the Sausage token to `escrow`. After, `escrow` will also call the `corgi` contract to asynchornously transfer the Corgi token to `escrow`. Then, `escrow` will conduct a transfer to both parties.
+
+- If both of the `transfer` calls succeed, then Alice will now own one Sausage token and Jerry will own one Corgi token.
+- If one or both of the final `transfer` calls fail, then nothing will happen and `escrow` should attempt reissuing the failed transaction.
+
+#### **Technical calls**
+
+1. `alice` makes an async call to `corgi::grant_access({"escrow_account_id":"escrow"})`
+2. `jerry`  makes an async call to `sausage::grant_access({"escrow_account_id":"escrow"})`
+3. `escrow` calls `sausage::transfer({"new_owner_id:"escrow", "token_id": 5})`
```
evgenykuzyakov, line 87: This might be a security issue. When escrow wants to transfer the token 5 from `alice` by the given `token_id`, the token may not be owned by `alice` anymore. Since `alice` can transfer it to `jerry`. If jerry also gave the access to `escrow` (pretty common for marketplaces), then `escrow` will take the token out of `jerry` and assign ownership to `alice.

The easiest way to address it is to explicitly provide `owner_id` in the `transfer_from` function. So might need to put it back.

@janedegtiareva @potatodepaulo

specs/Standards/Tokens/NonFungibleToken.md
chadoh: "_for_ the given `token_id`" sounds better to my ears

specs/Standards/Tokens/NonFungibleToken.md
janedegtiareva: Is escrow malicious in this case, or are we trying to prevent some sort of race condition/working with stale state?

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,167 @@
...
+    - Recommondation: attach callback `escrow::on_transfer({"owner_id":"jerry", "token_contract":"sausage", "token_id": 5})`
```
chadoh, line 88:
```suggestion
    - Recommendation: attach callback `escrow::on_transfer({"owner_id":"jerry", "token_contract":"sausage", "token_id": 5})`
```

specs/Standards/Tokens/NonFungibleToken.md
chadoh: > `escrow` will take the token out of `jerry` and assign ownership to `alice`

Assuming a trustworthy escrow, wouldn't this situation imply that the tokens still get where they were intended?

* alice gives access to escrow
* jerry gives access to escrow
* alice & jerry swap directly, without escrow
* escrow performs "swap", but ends up taking target tokens away from jerry & alice just to give them back again

seems like a waste of gas, but like it wouldn't result in a mis-swap _in this particular case_

However, it seems like it might create an attack vector, though I haven't thought of a particular attack scenario yet. Can you think of one?

specs/Standards/Tokens/NonFungibleToken.md
chadoh: We [implemented this](https://github.com/near-examples/NFT/blob/426fb986fadf10f00abca6e6f28269497bcb690d/contracts/assemblyscript/nep4-basic/main.ts#L8) with `u64` in demo to simplify the code; should the demo change or the spec?

specs/Standards/Tokens/NonFungibleToken.md
evgenykuzyakov: `escrow` is not malicious. So it's somewhat race condition.

`escrow` can't rely on the result of `get_token_owner` to determine the current owner, because by the time it gets the result, the owner may change. This means `alice` can call trade with `jerry` once and then trade the same token with `bob` again. The 2nd time, the token is owned by `bob`, but `escrow` can't know about it.

If `escrow` first calls `get_token_owner` to determine the current owner, then by the time the result got back to `escrow` the owner may already change. So `alice` can trade on `escrow2` with `jerry` in parallel, if `get_token_owner` executed before the `escrow2` trade, then `escrow` thinks that the `alice` still owns the token.

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -131,7 +131,7 @@ export function transfer(new_owner_id: string, token_id: u128): void;
 // Returns `true` or `false` based on `escrow_account_id` having access to tokens owned by `owner_id`
 export function check_access(escrow_account_id: string, owner_id: string): boolean;
 
-// Get an individual owner by given `token_id`.
+// Get the owner's account ID of the given `token_id`.
```
chadoh, line 134:
```suggestion
// Get the owner's account ID for the given `token_id`.
```

robert-zaremba:
> @robert-zaremba we want to keep all discussions within github so that people don't have to go to multiple different places to check out proposals and discussions. Feel free to suggest changes in this PR. But if you want to suggest something entirely different, feel free to open another PR.

Sorry for late reply. In Ethereum EIPs are discussed using _issues_. I think the major motivation for that is to centralize the whole spec discussion in a single issue. I think github doesn't offer a best solution for that.

BTW: render.com integration is NUTS!

robert-zaremba:
> @robert-zaremba we want to keep all discussions within github so that people don't have to go to multiple different places to check out proposals and discussions. Feel free to suggest changes in this PR. But if you want to suggest something entirely different, feel free to open another PR.

Sorry for late reply. In Ethereum EIPs are discussed using issues. I think the major motivation for that is to centralize the whole spec discussion in a single issue. I think github doesn't offer a best solution for that.

BTW: render.com integration is NUTS!

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,173 @@
...
+- Transfer tokens from one user to another.
+- Get current number of tokens for a given account ID. (Not in template).
+- Initialize contract once. The given total supply will be owned by the given account ID. (Not in template).
+
+There are a few concepts in the scenarios above:
+
+- **Total supply**. The total number of tokens in circulation.
+- **Token owner**. An account ID that owns one or more tokens.
+- **Transfer**. Action that moves some amount from one account to another account.
+- **Escrow**. A different account from the balance owner who has permission to one or more tokens.
```
robert-zaremba, line 40: Let's call it an **Operator**. It's cleary indicating the use of that role: operating someones else token, (rather than holding it).

robert-zaremba:
> We should remove transfer_from.
> 
> Copy of slack discussion for posterity/visibility
> Evguenia Degtiareva 2 hours ago
> Q about transfer_from: there is a param owner_id. We should never trust this since we have this data stored internally. Can we remove this param? The consequence is that transfer_from becomes the same as transfer, and we can get rid of it completely and only have transfer @peter (edited)
> 
> Peter DePaulo 1 hour ago
> So the point is that for transfer, you check predecessor id. For transfer_from, you have access to someone else’s token.
> 
> Peter DePaulo 1 hour ago
> For more complex stuff, it’s valuable which is why it’s in the base standard (edited)
> 
> Peter DePaulo 1 hour ago
> It’s specifically when owner_id is not equal to predecessor_id
> 
> Evguenia Degtiareva 1 hour ago
> it's not about whether owner_id is equal to predecessor_id
> 
> Evguenia Degtiareva 1 hour ago
> we have owner_id stored on chain. so if have an owner_id passed in, and it doesn't match what we have on chain, we have to ignore what's passed in for security purposes. We never really want to trust the caller on what the token owner is, hence there is no need to have that as a parameter (
> Peter DePaulo 1 hour ago
> Yeah okay, I can’t come up with a counter argument other than it will be useful for future use cases. Make sure to add a commit to the pr removing transfer_from then. @evgeny Kuzyakov comments?
> 
> Evgeny Kuzyakov 1 hour ago
> I think it’s good idea to remove it. Because we have token_id that defines the current owner

@potatodepaulo , @evgenykuzyakov  - having an `operator` and a `transfer_from` is very useful feature. We could remove `transfer` but we can't remove `transfer_from` - otherwise there is no way to transfer tokens on someone else behalves. Few use cases:

1. I want to add an access to an exchange to get my token whenever there is a trade to match.
2. I want to temporally give an access to my employee / partner without doing it through a wallet.

specs/Standards/Tokens/NonFungibleToken.md
robert-zaremba: I think the right question is what value is better there. `u128` is a safe bet, but this will force slightly bigger token size (1% ?). `u64` is very big, hard to reason if it's sufficient or not for all possible use-cases. `max(u64) ~ 1.8e19`. There is about `2e23` atoms in known Universe, so only 1k times more then what `u64` can hold.

specs/Standards/Tokens/NonFungibleToken.md
robert-zaremba: There is an easy way to avoid it. Since there is only one `owner` of each token, then we don't need to call `get_token_owner` and `transfer_from` in two different calls. In fact `transfer_from` can call `get_token_owner` inside, and check if the callee is authorized for executing transfer on behalves of the owner. Since all this operations are happening inside the same smart contract (when calling `transfer_from`) then we don't have race condition.

specs/Standards/Tokens/NonFungibleToken.md
robert-zaremba: Same as in my previous comment, I would recommend to rename `escrow_acount_id` to `operator_id`. Also, is `predecessor_id` the name of a context parameter which indicates the `callee`?

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,173 @@
...
+// Transfer the given `token_id`. Account `new_account_id` becomes the new owner.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function transfer_from(owner_id: string, new_owner_id: string, token_id: u128): void;
+
+// Transfer the given `tokenId` to the given `accountId`.  Account `accountId` becomes the new owner.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should be the owner of the token. Callers who have
+// escrow access should use transfer_from.
+export function transfer(new_owner_id: string, token_id: TokenId): void;
```
robert-zaremba, line 131: How about renaming the parameters to:
```
transfer(dest: string, token_id: TokenId)
```

specs/Standards/Tokens/NonFungibleToken.md
```diff
@@ -0,0 +1,173 @@
...
+
+// Revoke the access to the given `account_id` for the given `token_id`.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function revoke_access(escrow_account_id: string): void;
+
+// Transfer the given `token_id`. Account `new_account_id` becomes the new owner.
+// Requirements:
+// * The caller of the function (`predecessor_id`) should have access to the token.
+export function transfer_from(owner_id: string, new_owner_id: string, token_id: u128): void;
```
robert-zaremba, line 125: As noted in the PR discussion, how let's remove `owner_id` - in case of NFT it can be simple determined inside the `transfer_from` function.

robert-zaremba:
If I would like to propose more changes in the code - is there any way how I can edit it and make a commit without doing a separate PR?
The change I would like to propose:
+ clarifying the discussion above
+ better explanation and motivation of the interface
+ updated role names and parameter names.

robert-zaremba:
There is one more thing I would strongly recommend: adding a `ref` (bytes / string) parameter to both `transfer` and `transfer_from` functions. 

This would tremendously help with all compliance integrations!

evgenykuzyakov:
@robert-zaremba Can you clarify on `ref`?

> As noted in the PR discussion, how let's remove `owner_id` - in case of NFT it can be simple determined inside the `transfer_from` function.

We can't remove `owner_id` from transfer_from, because of the async execution. If the 2 owners authorized access to the same `marketplace`, the marketplace will be able to transfer tokens from both owners. So any of these 2 owners can call a method on the marketplace that will call the `transfer_from` assuming the ownership.
The token contract can't verify the who's initiated the transfer because we use immediate predecessor instead of the use signer of the transaction as a proof of ownership.

specs/Standards/Tokens/NonFungibleToken.md
evgenykuzyakov: We can't remove `owner_id` from transfer_from, because of the async execution. If the 2 owners authorized access to the same `marketplace`, the marketplace will be able to transfer tokens from both owners. So any of these 2 owners can call a method on the marketplace that will call the `transfer_from` assuming the ownership.
The token contract can't verify the who's initiated the transfer because we use immediate predecessor (in this case marketplace) instead of the use signer of the transaction as a proof of ownership.

evgenykuzyakov:
> If I would like to propose more changes in the code - is there any way how I can edit it and make a commit without doing a separate PR?

You can use suggestion feature in the github reviews. https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/reviewing-proposed-changes-in-a-pull-request

robert-zaremba:
> Can you clarify on ref?

`ref` is a reference string. The goal is to be able to link an on-chain transaction with off-chain data. Data can be in a `ref` string, an URI, or a simple ID. This helps (a lot!) carrying out extra data when needed and bind it in a trustless way.

robert-zaremba:
> @robert-zaremba Can you clarify on `ref`?
> 
> > As noted in the PR discussion, how let's remove `owner_id` - in case of NFT it can be simple determined inside the `transfer_from` function.
> 
> We can't remove `owner_id` from transfer_from, because of the async execution. If the 2 owners authorized access to the same `marketplace`, the marketplace will be able to transfer tokens from both owners. So any of these 2 owners can call a method on the marketplace that will call the `transfer_from` assuming the ownership.
> The token contract can't verify the who's initiated the transfer because we use immediate predecessor instead of the use signer of the transaction as a proof of ownership.

Ah, right. We might transfer a token assuming wrong ownership. But, since this is NFT - each token has only one owner, and to transfer it we always have to have allowance of the _current owner_. So, security wise, we will not be able to transfer tokens we don't have authorization for if the owner query is happening inside the `transfer_from` (so both the transfer and ownership check is happening in the same context) - since calling functions within the same contract won't create an async call.

evgenykuzyakov:
> Ah, right. We might transfer a token assuming wrong ownership. But, since this is NFT - each token has only one owner, and to transfer it we always have to have allowance of the current owner. So, security wise, we will not be able to transfer tokens we don't have authorization for if the owner query is happening inside the transfer_from (so both the transfer and ownership check is happening in the same context) - since calling functions within the same contract won't create an async call.

Yes it's true, but for this `transfer_from` has to know who's the expected owner. Since it's called not by the owner, but from the contract (let's call it `marketplace`) then the setup will be the following:
- `current_account_id` - `token` - the account ID of the contract 
- `predecessor_account_id` - `marketplace` - the account ID of the contract that called token contract
- `signer_account_id` - `owner` - the account ID of the user that initiated the transaction.

Now from the token perspective the `token` can be owned either by `owner` or by `other_owner`. Depends on the access, both of them might authorized the `marketplace` contract to act their behalf because they both use `marketplace`.

The token contract as for NEAR security model always relies on the `predecessor_account_id` instead of `signer_account_id`, because the signer might call one contract which can indirectly call `token` and steal tokens. To avoid it, the for permissions we always check the immediate predecessor instead of the original signer.

This means the `token` contract in `transfer_from` can't check who initiated the transaction for transfer, because it shouldn't rely on the `signer_account_id`. It guarantees that when you authorize some application on NEAR through restricted access keys, it can't steal your tokens, because the `predecessor_account_id` will be different.

robert-zaremba:
Finally some serious work is starting on the enterprise side with defining token standards:
https://cointelegraph.com/news/microsoft-nasdaq-and-others-to-set-global-tokenization-standards

specs/Standards/Tokens/NonFungibleToken.md
potatodepaulo: Trying to keep consistent with ERC20. I think that `Escrow` is more accurate since they actually take ownership of the tokens in the current model.

robert-zaremba: Escrow is a very specific process. Her we have more general functionality. According to [Investopedia](https://www.investopedia.com/terms/e/escrow.asp), escrow:
> Escrow is a legal concept describing a financial instrument whereby an asset or escrow money is held by a third party on behalf of two other parties that are in the process of completing a transaction. 
> An escrow agent is an entity that has fiduciary responsibilities in the transfer of property from one party to another. 

The differences here are:
1. It's an interface, and it's up to the implementation if the asset is changing an owner or not. In most of the implementations, we don't change an ownership - instead an allowance record is created. BTW: also you description doesn't say that we want to change the owner: _A different account from the balance owner who has permission to one or more tokens._
2. IMHO it's better to keep it general for different processes.

zahhar:
Hi all,

Are there any news or updates on potential standard composition? e.g.

1. mandatory features to be implemented by smartcontract?
2. how strictly method & their attributes must be named when implementing a feature?
3. what is the minimum set of metadata (public-assessible attributes of the NFT, if any) that must be stored in the blockchain?

My view on metadata:

- Token_ID
- Owner_Account_ID
- Last_Owner_Account_ID
- Created_At
- Last_Changed_At
- Valid_From
- Valid_To
- User_Data (string of serialized into json or similar format custom data that contract designer decided to store on-chain for whatever reason)

Am I missing something?

chadoh:
Hey @zahhar, see discussion on improved NFT standard at https://github.com/near/NEPs/discussions/171

chadoh:
@potatodepaulo @mikedotexe I think it makes sense to close this for now. It was never fully finalized, and we now have a replacement NEP that likewise has not yet been finalized. We may as well signal to people that they should refer to the new discussion instead of this one.

New discussion at #171 

Feel free to re-open this one if you think we should keep it around for a bit longer.

PR #5, opened 2019-07-11
Access Keys

evgenykuzyakov:


0000-access-keys.md
```diff
@@ -0,0 +1,286 @@
...
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Here are proposed changes for the AccessKey and Account structs.  
+
+```rust
+/// `account_id,public_key` is a key in the state
+struct AccessKey {
+  /// The nonce for this.
+  pub nonce: Nonce(u64),
```
bowenwang1996: The type in parentheses should be in comment, like 
```rust
pub nonce: Nonce //u64
```
The current format is confusing

0000-access-keys.md
```diff
@@ -0,0 +1,286 @@
...
+The user also selects the allowance to some reasonable amount, enough for the application to issue transactions. The application might hint the user about this allowance.
+Now the app can issue function call transactions on behalf of the user’s account towards the app’s contract without requiring the user to sign each transaction.
+
+An access key example for chess app from user `vasya.near` (using old permissions model):
+```rust
+/// vasya.near,c5d312f3
+AccessKey {
+    nonce: 0,
+    
+    allowance: 1_000_000_000,
```
bowenwang1996: Why is this allowance not 0?

0000-access-keys.md
```diff
@@ -0,0 +1,286 @@
...
+    method_name: Some("proxy"),
+
+    full_access: false,
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+- Access keys are going to be stored with the `account_id` prefix and the `public_key` suffix.
```
bowenwang1996: It would be better to explain the exact composition of this key.

0000-access-keys.md
```diff
@@ -0,0 +1,286 @@
...
+Allowance is the amount of tokens the AccessKey can spend from the account balance.
+When some amount is spent, it's subtracted from both the allowance of the access key and from the account balance.
+
+NOTE: In the previous iteration of access keys, we used balance instead of the allowance. But it required to sum up all access keys balances to get the the total account balance.
+It also prevented sharing of the account balance between access keys.
+
+#### Permissions
+
+Almost all desired use-cases of access keys can be achieved by using the old permissions model.
+It restricts allows access keys to only issue function call with no attached tokens.
```
bowenwang1996: remove "allows"

0000-access-keys.md
```diff
@@ -0,0 +1,286 @@
...
+The function calls are restricted to the selected `receiver_id` and potentially restricted to a single `method_name`.
+Anything non-trivial can be done by the contract that receives this call, e.g. through `proxy` function.
+
+If we want to remove public keys from the account, we also need to add a new permission that allows unlimited access to the account.
```
bowenwang1996: Unless a contract is locked, why would there not be a full access key?

0000-access-keys.md
```diff
@@ -0,0 +1,286 @@
...
+
+```rust
+/// `account_id,public_key` is a key in the state
+struct AccessKey {
+  /// The nonce for this.
+  pub nonce: Nonce(u64),
+
+  /// The amount that can be spent for transaction fees by this access key from the account balance.
+  /// When used, both account balance and the allowance is decreased.
+  pub allowance: Balance(u128),
```
bowenwang1996: It feels better if have `Option<Balance>` here for access keys without limit.

0000-access-keys.md
evgenykuzyakov: By default access keys allow only function calls

0000-access-keys.md
evgenykuzyakov: Good idea

0000-access-keys.md
evgenykuzyakov: replaced permissions model, to make it clear

0000-access-keys.md
evgenykuzyakov: added a comment

evgenykuzyakov: Allowance has to be non-zero for the transaction to even be considered by the runtime.

0000-access-keys.md
evgenykuzyakov: Updated permissions model

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+
+  /// Gives full access to the account.
+  /// NOTE: It's used to replace account-level public keys.
+  FullAccess,
+}
+
+pub struct FunctionCallPermission {
+  /// The amount that can be spent for transaction fees by this access key from the account balance.
+  /// When used, both account balance and the allowance is decreased.
+  pub allowance: Balance,  // u128
```
nearmax: Can allowance be reset by the account when access key runs out of it?

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+Here are proposed changes for the AccessKey and Account structs.  
+
+```rust
+/// `account_id,public_key` is a key in the state
+struct AccessKey {
+  /// The nonce for this.
+  pub nonce: Nonce,  // u64 
+  
+  /// Defines permissions for the AccessKey 
+  pub permission: AccessKeyPermission,
```
nearmax: Should it be a `Vec` for access keys that can call more than one method, or more than one smart contract?

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+  /// The amount that can be spent for transaction fees by this access key from the account balance.
+  /// When used, both account balance and the allowance is decreased.
+  pub allowance: Balance,  // u128
+
+  /// The AccountID of the receiver of the transaction. The access key will restrict transactions to
+  /// only this receiver.
+  pub receiver_id: AccountId,  // String
+  
+  /// If not None, the access key would be restricted to call only the given method name.
+  pub method_name: Option<String>,
```
nearmax: What if it's `None`? Will it be allowed to call any method on the contract? I thought it's what `FullAccess` is for.

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+}
+
+/// NOTE: This change removes account-level nonce and public keys.
+/// Key is `account_id`
+struct Account {
```
nearmax: We need to spec out here how transactions from different access keys are going to be ordered with respect to each other. For example:
3 access keys (A,B,C) issue 3 transactions each:
A1, A2, A3; B1,B2,B3; C1, C2, C3;
All these transactions operate on the same state so they need to have an order. First transaction to execute is one of {A1,B1,C1} that has lowest hash, let's say it is B1. Second transaction to execute is one of {A1,B2,C1} with lowest hash, etc.

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+  pub balance: Balance(u128),
+  pub code_hash: Hash,
+  /// Storage usage accounts for all access keys      
+  pub storage_usage: StorageUsage(u64),
```
nearmax: I don't think we should be storing storage usage explicitly. It would need to be recomputed every time an allowance of some access key changes, because there is actually a small mechanism compression in protos. This needs to be investigated carefully. Instead we can compute storage on every function call, which is not that hard to do since all access keys are easy to access and we just need to measure their size. If some node implementation wants to precompute storage usage then they can do it in their own way, but I don't think this should be a part of the protocol spec.

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+    }),
+}
+```
+
+#### AccessKey issued by a dApp
+
+This is an example where the dApp wants to pay for the user, or it doesn't want to go through the user's sign-in flow.
+For whatever reason the dApp decided to issue an access key directly for their account.
+
+For this to work there should be one account with funds (that dApp controls on the backend) which creates access keys for the users.
```
nearmax: This does not describe where dApp gets public key to generate access key from.

nearmax: Also, the case when user wants to map their existing account to public key I think should be described separately.

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+#### AccessKey through a proxy
+
+This examples demonstrates how to have more granular control on top of built-in access key restrictions.
+
+Let's say a user wants to:
+- limit the number of calls the access key can make per minute
+- support multiple contracts with the same access key
+- select which methods name can be called and which can't
+- transfer funds from the account up to a certain limit 
+- stake from the account, but prevent withdrawing funds
```
nearmax: Add example with social network to explain how we can use proxy to attribute account id with a public key.

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+### Technical changes
+
+#### `nonce` on the AccessKey level instead of account level
+
+Since access keys can be used by the different people or parties at the same time, we need to be able to 
+have separate nonce for each key instead of a single nonce at the account level.
+With a single nonce on the account level, there is a high probability that 2 apps would use the same nonce for 2 different transactions and one of this transactions would be rejected.
+
+Previously we were ordering transactions by nonce and rejecting transactions with a duplicated or lower nonce.
+With the access key nonce, we still need to order transactions by nonce, but now we need to group them by `account_id,public_key` key instead of just account_id.
```
nearmax: Better group them by hash like suggested above to prevent one access key having constant priority over others.

0000-access-keys.md
```diff
@@ -0,0 +1,313 @@
...
+
+In the previous design, the `receiver_id` was called `contract_id` and was an option field. But it didn't remove the requirement for the receiver when it was `None`. Instead it was assuming the access key is pointed to the owner's account. 
+We can potentially use `None` to mean unlimited key, and require user to explicitly specify their own account_id if they want to use proxy function.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+#### Permissions
+
+Not clear whether a single pair of `receiver_id`/`method_name` is enough to cover all use-cases at the moment.
```
nearmax: You would have to redeploy the contract code with merged in `proxy` method.

0000-access-keys.md
nearmax: Idea: if allowance is equal to max u128 value then we do not decrement.

evgenykuzyakov: Best way I think is to recreate the `AccessKey` with `SwapKey` transaction

0000-access-keys.md
evgenykuzyakov: I don't think we should support it on the runtime side and instead rely on the contract code.
Otherwise we introduce additional complexity on the runtime for verification of the access keys transaction unless we index all permissions in some way.

0000-access-keys.md
evgenykuzyakov: None would mean key is not restricted to a single method name, but can call any method name on the given `receiver_id` contract.

0000-access-keys.md
evgenykuzyakov: Good point. It should be similar to how we order transactions with conflicting nonces. E.g. which transaction is taken first for a given `account_id` key. Now with the access key we should do the same but for the `account_id,public_key` keys.

nearmax: I am not sure I understand this:

>E.g. which transaction is taken first for a given account_id key. Now with the access key we should do the same but for the account_id,public_key keys.

0000-access-keys.md
evgenykuzyakov: Yeah, it might be an issue. The best option would be is to change storage_usage only when the access key is created or removed. But changes to the allowance value shouldn't change the storage_usage. For this to work, we might need to update the storage computation formula for the access key. Especially, because we don't use proto size for the storage_usage for the account.

0000-access-keys.md
evgenykuzyakov: We decided to avoid this proxy example and rely on `authorize` public key to link an identity with the given public key

0000-access-keys.md
evgenykuzyakov: The better option would be what @bowenwang1996 suggested with using Option<Balance> type (where None means unlimited allowance), but I can't think of an example when we actually need to have u128::max value instead of a full permission key.

nearmax: Using Option is more type-safe. I think it is a good idea.

0000-access-keys.md
evgenykuzyakov: Is there a reason to ever accept a transaction with `new_nonce != previous_nonce + 1` ?

nearmax: When transaction is lost or delayed due to how transactions are propagated through gossip. Should be a very common thing.

evgenykuzyakov: But this is exactly what we can avoid by having a strict requirement. Asking this question in discord for broader discussion

nearmax: Copying from discord channel:

Transactions are propagated asynchronously through p2p network. If someone issues two transactions with nonces X+1 and X+2 (where X is the current nonce of the account) but X+1 is getting delayed we want to allow X+2 to be accepted and processed and X+1 to be discarded later. (your suggested rule would disallow this)
Consider a more severe example: Let's say someone issues 10 TPS (they do diligent work incrementing nonce correctly) and we are producing 1 block per second. If every 1 second at least one transaction is getting delayed by p2p network by 1 second then we are going to have a 50% failure rate of all transactions issued by this person on average.

0000-access-keys.md
evgenykuzyakov: Updated comments

0000-access-keys.md
evgenykuzyakov: Added it to unresolved question. The actual TX ordering is not a blocking question, since it should just derive it from the current TX ordering with key upgrade.

ilblackdragon:
We want access keys to be programmable.
Currently all access keys are using the same contract deployed on the account. 
This is very limiting. As soon as you want to deploy another access key with a `proxy` capability, you will need to update the existing contract. Which the access key you have might not have permission to do.
Instead we should make each access key have it's own code that checks permissions. E.g. when transaction is sent, we first execute AccessKey code and check that this transaction is valid given it's permissions and only then pass this transaction to be executed.

Note, that if you only have access key to the contract, you might be able to only add access keys with lesser or equal scope than the current key. To perform that, we actually need to chain execution of the access key permissions. E.g. if there is AccessKey1, and I deployed AccessKey2 using it -> transactions signed with AccessKey2 should also pass AccessKey1 permissions.

PR #8, opened 2019-07-24
Batched Transactions

evgenykuzyakov:
Ready for review. I'll start implementing it in a meantime.


text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,327 @@
...
+For example, we want to create a new account with some account balance and one or many access keys, deploy a contract code on it and run an initialization method to restrict access keys permissions for a `proxy` function.
+
+To be able to do this now, we need to have a `CreateAccount` transaction with all the parameters of a new account.
+Then we need to handle it in one operation in a runtime code, which might have duplicated code for executing some WASM code with the rollback conditions.
+
+Alternative to this is to execute multiple simple transactions in a batch within the same block.
+It has to be done in a raw without any commits to the state until the entire batch is completed.
+We propose to support this type of transaction batching to simplify the runtime.
+
+This change
```
bowenwang1996: Something is missing here.

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,327 @@
...
+
+This NEP proposes changing it by first converting all signed transactions into receipts and then either execute them immediately before received receipts, or put them into the list of the new receipts to be routed.
+To achieve this, NEP introduces a new message `Action` that represents one of atomic actions, e.g. a function call.
+`TransactionBody` is now called just `Transaction`. It contains the list of actions that needs to be performed in a single batch and the information shared across these actions.
+
+`Transaction` contains the following fields
+- `originator_id` is an account ID of the transaction originator.
+- `public_key` is to identify the access key used to signs the transaction.
+- `nonce` is used to deduplicate and order transactions.
+- `receiver_id` is where the transaction has to be routed.
```
bowenwang1996: I thought this is the final destination

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,327 @@
...
+- `action` is the list of actions to perform.
+
+An `Action` can be of the following:
+- `CreateAccount` creates a new account. It has to be the first action, and the receiver has to be a new account. The action will fail if the account already exists.
```
bowenwang1996: Why does it *have* to be the first action?

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,327 @@
...
+- `DeployContract` deploys given binary wasm code on the account. Either the `receiver_id` equals to the `originator_id`, or the batch of actions has started with `CreateAccount`, which granted that permission.
+- `FunctionCall` executes a function call on the last deployed contract. E.g. if the previous action was `DeployContract`, then the code to execute will be the new deployed contract. `FunctionCall` has `method_name` and `args` to identify method with arguments to call. It also has `fee` and the `amount`. `fee` is the transaction fee that is prepaid for this call to be spent on gas. `amount` is the attached amount of NEAR tokens that the contract can spend, e.g. 10 tokens to pay for a crypto-corgi.
+- `Transfer` transfers the given amount of tokens from the originator to the receiver.
+- `Stake` stakes the new total amount which is given in the `amount` field with the given `public_key`. The difference in stake is taken from the account's balance at the moment when this action is executed, so it's not prepaid. There is no particular reason to stake on behalf of a newly created account, so we may disallow it.
```
bowenwang1996: The difference is only taken from the account's balance if the transaction tries to increase the stake.

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,327 @@
...
+- `DeleteKey` deletes an old `AccessKey` identified by the given `public_key` from the account. Fails if the access key with the given public key doesn't exist. All next batched actions will continue to execute, even if the public key that authorized that transaction was removed.
+- `AddKey` adds a new given `AccessKey` identified by a new given `public_key` to the account. Fails if an access key with the given public key already exists. We removed `SwapKeyTransaction`, because it can be replaced with 2 batched actions - delete an old key and add a new key.
+
+The new `Receipt` contains the shared information and either one of the receipt actions or a list of actions:
+- `sender_id` the account ID of the immediate previous sender of this receipt. It can be different from the `originator_id` in some cases.
+- `receiver_id` the account ID of the current account, on which we need to perform action(s).
+- `receipt_id` is a hash of this receipt (previously was called `nonce`). It's generated from either the signed transaction or the parent receipt.
```
bowenwang1996: It's not necessarily the hash of this receipt. It's better to call it something like "unique identifier".

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,327 @@
...
+        input_data_id: [],
+        
+        action: [...]
+    }
+}
+```
+
+This receipt will be sent to `vitalik.vasya.near`'s shard to be executed.
+In case the `vitalik.vasya.near` account already exists, the execution will fail and some amount will be refunded back to `vasya.near` account with a new `ActionReceipt` with a single transfer action.
+BTW, if the refund fails (e.g. if `vasya.near` deletes his account in between), then the next `refund_account_id` will be `system`, which means the tokens are going to be burned and no receipts are generated.
```
bowenwang1996: How do you delete an account?

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,327 @@
...
+- `action` is the list of actions to execute. The execution doesn't need to validate permissions of the actions, but need to fail in some cases. E.g. when the receiver's account doesn't exist and the action acts on the account, or when the action is a function call and the code is not present.
+
+`DataReceipt` contains the following fields:
+- `data_id` is the data ID to be used as an input.
+- `success` is true if the `ActionReceipt` that generated this `DataReceipt` finished the execution without any failures.
+- `data` is the binary data that is returned from the last action of the `ActionReceipt`. Right now, it's empty for all actions except for function calls. For function calls the data is the result of the code execution. But in the future we might introduce non-contract state reads.
+
+Data should be stored at the same shard as the receiver's account, even if the receiver's account doesn't exist.
```
bowenwang1996: What's a scenario where this would happen?

text/0000-transaction-refactoring.md
evgenykuzyakov: It's is. Clarified

text/0000-transaction-refactoring.md
evgenykuzyakov: Until we have the account deletion, it doesn't make sense for trying to create an account after some other action.

text/0000-transaction-refactoring.md
evgenykuzyakov: Either way it's not prepaid.

text/0000-transaction-refactoring.md
evgenykuzyakov: right.

text/0000-transaction-refactoring.md
evgenykuzyakov: When you run out of rent. We don't support it yet, but it will be the case

text/0000-transaction-refactoring.md
evgenykuzyakov: For callbacks on other shards, will add an example

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,508 @@
...
+
+`Transaction` contains the following fields
+- `originator_id` is an account ID of the transaction originator.
+- `public_key` is to identify the access key used to signs the transaction.
+- `nonce` is used to deduplicate and order transactions.
+- `receiver_id` is the account ID of the destination of this transaction. It's where the generated receipt will be routed for execution.
+- `action` is the list of actions to perform.
+
+An `Action` can be of the following:
+- `CreateAccount` creates a new account. It has to be the first action, and the receiver has to be a new account. Until we have the account deletion, it doesn't make sense for trying to create an account after some other action. The action will fail if the account already exists.
```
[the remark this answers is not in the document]
evgenykuzyakov: Removed that requirement

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,508 @@
...
+This NEP proposes changing it by first converting all signed transactions into receipts and then either execute them immediately before received receipts, or put them into the list of the new receipts to be routed.
+To achieve this, NEP introduces a new message `Action` that represents one of atomic actions, e.g. a function call.
+`TransactionBody` is now called just `Transaction`. It contains the list of actions that needs to be performed in a single batch and the information shared across these actions.
+
+`Transaction` contains the following fields
+- `originator_id` is an account ID of the transaction originator.
+- `public_key` is to identify the access key used to signs the transaction.
+- `nonce` is used to deduplicate and order transactions.
+- `receiver_id` is the account ID of the destination of this transaction. It's where the generated receipt will be routed for execution.
+- `action` is the list of actions to perform.
```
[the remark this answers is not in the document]
evgenykuzyakov: We can only batch actions to the same account

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,508 @@
...
+
+An `Action` can be of the following:
+- `CreateAccount` creates a new account. It has to be the first action, and the receiver has to be a new account. Until we have the account deletion, it doesn't make sense for trying to create an account after some other action. The action will fail if the account already exists.
+`CreateAccount` also grants permission for all subsequent batched action for the newly created account. For example, deploying code on the new account.
+- `DeployContract` deploys given binary wasm code on the account. Either the `receiver_id` equals to the `originator_id`, or the batch of actions has started with `CreateAccount`, which granted that permission.
+- `FunctionCall` executes a function call on the last deployed contract. E.g. if the previous action was `DeployContract`, then the code to execute will be the new deployed contract. `FunctionCall` has `method_name` and `args` to identify method with arguments to call. It also has `fee` and the `amount`. `fee` is the transaction fee that is prepaid for this call to be spent on gas. `amount` is the attached amount of NEAR tokens that the contract can spend, e.g. 10 tokens to pay for a crypto-corgi.
```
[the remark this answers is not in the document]
evgenykuzyakov: discussed

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,508 @@
...
+- `Transfer` transfers the given amount of tokens from the originator to the receiver.
```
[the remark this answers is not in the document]
evgenykuzyakov: It opens the question to arguments serialization. Instead of this we better use separate protos for actions

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,508 @@
...
+It also increased the complexity of the runtime implementation.
+
+This NEP proposes changing it by first converting all signed transactions into receipts and then either execute them immediately before received receipts, or put them into the list of the new receipts to be routed.
+To achieve this, NEP introduces a new message `Action` that represents one of atomic actions, e.g. a function call.
+`TransactionBody` is now called just `Transaction`. It contains the list of actions that needs to be performed in a single batch and the information shared across these actions.
+
+`Transaction` contains the following fields
+- `originator_id` is an account ID of the transaction originator.
+- `public_key` is to identify the access key used to signs the transaction.
+- `nonce` is used to deduplicate and order transactions.
```
nearmax: Add note that `nonce` is per access key.

text/0000-transaction-refactoring.md
```diff
@@ -0,0 +1,508 @@
...
+- `receiver_id` is the account ID of the destination of this transaction. It's where the generated receipt will be routed for execution.
+- `action` is the list of actions to perform.
+
+An `Action` can be of the following:
+- `CreateAccount` creates a new account. It has to be the first action, and the receiver has to be a new account. Until we have the account deletion, it doesn't make sense for trying to create an account after some other action. The action will fail if the account already exists.
+`CreateAccount` also grants permission for all subsequent batched action for the newly created account. For example, deploying code on the new account.
```
nearmax: We should clarify that for action `A_i` in list of actions `A_0,...,A_i,..., A_n` it has permission to operate only on the following account.

* As we execute actions we maintain the stack of account ids.
* Before we start executing them we pre-populate the stack with one account id -- the one that signed the transaction.
* Every `CreateAccount` puts new account id on the top of the stack;
* Every `DeleteAccount` removes account id from the stack;
* An action has rights to operate on the account id on the top of the stack.

evgenykuzyakov:
Implementation branch in progress: https://github.com/nearprotocol/nearcore/tree/batched-tx-1

DanielRX:
Something I noticed with this, does this not violate some basic logical cases?

If I can run `tx1, tx2, tx3 and tx4` as a batch, it should be possible to run them all on their own, in sequence (assuming no `tx`s interrupt them)? This will not hold for batched transactions, since if `tx1` is a deleteKey op, `tx2` could then fail?

If I add a key in `tx1` and perform `tx2` which needs the key to work, would this be valid? If so, the `addKey` must be changing the state, but the `deleteKey` is not.

k06a:
Wish wallet to use this kind of transaction after bunch of transactions were "staked/collected" instead of "signed and sent". So wallet should allow grouping even for dapps that not initially support it.

evgenykuzyakov:
@DanielRX 

So we actually batching actions instead of transactions. The batched actions form a transaction which is signed. Once the transaction is authorized (e.g. with a full access key), all actions are executed and they don't require public key to be present. So the first action can be delete the access key and the 2nd action is something else, e.g. a function call.

This is done to support operations like deploy+migrate_call, or deploy+init.

@k06a 
All Batched Actions are from one signer to one receiver. It's not always makes sense to batch actions, e.g. 2 function calls might need to go in a parallel or fail independently

k06a:
@evgenykuzyakov but we need at least cover approve+transferFrom ERC20 flow. Approve should be called on token smart contract and second tx should be to receiver, which will internally make transferFrom call.

evgenykuzyakov:
@k06a As we discussed in other channel, it should be done with something like `approveThen`, which approves and locks the token for `contract_x`, then calls given method on `contract_x` with a callback on itself to unlock a token if the `contract_x` doesn't act.

k06a:
@evgenykuzyakov, hope both calls will be performed from account, not from token smart contract. Else this would behave like ERC827.

evgenykuzyakov:
Ready to be merged

k06a:
@evgenykuzyakov can you point to the locks implementation example?

evgenykuzyakov:
@k06a I think locks are for the different NEP.

PR #9, opened 2019-07-25
Revisited Bindings API

nearmax:
Implementation steps:
- [x] Modify bindings APIs in nearcore
- [x] Modify near-bindgen
- [x] Modify bindings APIs in near-runtime-ts
- [x] Fix any integration tests that got broken
- [x] Fix templates in studio
- [x] Create issue for @mikhailOK to make sure the Trie behavior follows the spec. Any non-matching behavior is considered to be an implementation bug;


0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+
+###### Panics
+* If the scratch buffer size exceeds the limit panics with `MemoryAccessViolation`;
+
+###### Current bugs
+* Not implemented.
+
+---
+```rust
+refund_id(register_id: u64)
```
evgenykuzyakov: `refund_account_id` makes more sense

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+###### Panics
+* If `balance_ptr + 16` points outside the memory of the guest with `MemoryAccessViolation`;
+
+###### Current bugs
+* Use a different name;
+* Unclear how they behave with the callbacks.
+
+---
+```rust
+deposit(min_amount_ptr: u64, max_amount_ptr: u64, result_ptr: u64)
```
evgenykuzyakov: We'll have to revisit deposit and withdrawals if we split `fee` and `amount` balances

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+* If `promise_idx` does not correspond to an existing promise panics with `InvalidPromiseIndex`.
+* If `account_id_len + account_id_ptr` or `method_name_len + method_name_ptr` or `arguments_len + arguments_ptr`
+or `amount_ptr + 8` points outside the memory of the guest or host, with `MemoryAccessViolation`.
+
+###### Returns
+* Index of the new promise that uniquely identifies it within the current execution of the method.
+
+---
+```rust
+promise_and(promise_id1: u64, promise_id2: u64) -> u64
```
evgenykuzyakov: That can be changed to take a C style array instead of 2 indices

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+Note, we are only going to have incomplete callbacks once we have `promise_or` combinator.
+###### Normal execution
+* If there is only one callback `promise_results_count()` returns `1`;
+* If there are multiple callbacks (e.g. created through `promise_and`) `promise_results_count()` returns their number.
+* If the function was called not through the callback `promise_results_count()` returns `0`.
+
+
+---
+```rust
+promise_result(promise_idx: u64, register_id: u64)
```
evgenykuzyakov: this is where we need to distinguish `promise_idx` and `result_index`. In this case this is a result index and it's only available within callbacks. So the actual promise_idx would be different, unless you suggest to inject it there. Might be an interesting concept, but right now positional results are enough.

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+* Index of the new promise that uniquely identifies it within the current execution of the method.
+
+---
+```rust
+promise_results_count() -> u64
+```
+If the current function is invoked by a callback we can access the execution results of the promises that
+caused the callback. This function returns the number of complete and incomplete callbacks.
+
+Note, we are only going to have incomplete callbacks once we have `promise_or` combinator.
```
evgenykuzyakov: For promise_or we may always expect to have only one result provided (the first data receipt). But to identify which result it is, we may simulate positional args.

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+###### Panics
+* If `value_len + value_ptr` points outside the memory or scratch buffer uses more memory than the limit with `MemoryAccessViolation`.
+
+###### Current bugs
+* Current name `hash` is not specific to what hash is being used.
+* We have `hash32` that largely duplicates the mechanics of `hash` because it returns the first 4 bytes only.
+
+---
+```rust
+check_ethash(block_number_ptr: u64,
```
evgenykuzyakov: We should completely remove it until ethash bridge works

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+If `min_amount <= attached_balance` moves `min(max_amount, max(min_amount, account_balance))`. The moved amount is
+saved into `u128` variable pointed by `result_ptr`. `min_amount_ptr` and `max_amount_ptr` point to `u128` variables.
+
+###### Panics
+* If `min_amount_ptr + 16` or `max_amount_ptr + 16` or `result_ptr + 16` point outside the memory of the guest with `MemoryAccessViolation`;
+
+## Math
+
+```rust
+random_buf(buf_len: u64, buf_ptr: u64)
```
evgenykuzyakov: It makes sense to switch to writing into a new register instead to be consistent.
`random_buf(len: u64, register_id: u64)`
The `register_id` is where to write

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+* `data_read` currently has mixed functionality -- it can be used both for reading data from trie and to read data from
+the context. In former it expects pointers to be passed as arguments, in later it expects enum to be passed. It achieves
+juxtaposition by casting pointer type in enum when needed;
+* **Economics API.** The functions that provide access to balance and such might need to be added or removed since we
+now consider splitting attached balance into two.
+
+# Specification
+
+## Scratch Buffer
+Scratch buffer allows the host function to return the data into a buffer located inside the host oppose to the buffer
```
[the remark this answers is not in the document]
bowenwang1996: I see. Was going to ask exactly the same thing

0006-bingings.md
bowenwang1996: I don't think so. This is needed for implementing the bridge

nearmax: Discussed with @ilblackdragon . We keep it for the team that will be working on it.

0006-bingings.md
nearmax: The identification can be done with `promise_result`.

0006-bingings.md
nearmax: Got it.

0006-bingings.md
nearmax: Done

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+
+###### Panics
+* If the scratch buffer size exceeds the limit panics with `MemoryAccessViolation`;
+
+###### Current bugs
+* Currently we conflate `originator_id` and `sender_id` in our code base.
+
+---
+```rust
+initiator_key(register_id: u64)
```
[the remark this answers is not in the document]
nearmax: Done

0006-bingings.md
nearmax: Done

0006-bingings.md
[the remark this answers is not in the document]
nearmax: Done.

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+* `data_read` does not help with renaming. If later we decide to rename `initiator_id` to `originator_id` then one could
+argue that contracts that rely on `data_read` would not break, while contracts relying on `initiator_id()` would. However
+the name change often means the change of the semantics, which means the contracts using this function are no longer safe to
+execute anyway.
+
+However there is one reason to not have `data_read` -- it makes `API` more human-like which is a general direction Wasm APIs, like WASI are moving towards to.
+
+---
+```rust
+initiator_id(register_id: u64)
```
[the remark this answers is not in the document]
nearmax: Discussed the naming offline.

0006-bingings.md
nearmax: Refund account is now irrelevant with new balance model.

0006-bingings.md
nearmax: Done.

0006-bingings.md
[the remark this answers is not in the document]
nearmax: Clarified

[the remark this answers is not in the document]
nearmax: Fixed

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+some access key and submitted into a memory pool (either through the wallet using RPC or by a node itself). This function returns the id of that account.
+
+###### Normal operation
+* Saves the bytes of the initiator account id into the register.
+
+###### Panics
+* If the scratch buffer size exceeds the limit panics with `MemoryAccessViolation`;
+
+###### Current bugs
+* Currently we conflate `originator_id` and `sender_id` in our code base.
```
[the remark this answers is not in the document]
nearmax: Discussed naming offline.

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+* If iterator is not empty, copies the next key into `key_register_id` and value into `value_register_id` and returns the length of the copied value;
+* If iterator is empty returns `u64::MAX`.
+
+This allows us to iterate over the keys that have zero bytes stored in values.
+
+###### Panics
+* If `key_register_id == value_register_id` panics with `MemoryAccessViolation`;
+* If the scratch buffer size exceeds the limit panics with `MemoryAccessViolation`;
+* If `iterator_id` does not correspond to an existing iterator panics with  `InvalidIteratorId`
+* If between the creation of the iterator and calling `storage_iter_next` the range over each it iterates was modified panics with `IteratorWasInvalidated`.
```
mikhailOK:
```suggestion
* If between the creation of the iterator and calling `storage_iter_next` the range over which it iterates was modified panics with `IteratorWasInvalidated`.
```

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+* If `prefix_len + prefix_ptr` exceeds the memory container it panics with `MemoryAccessViolation`;
+
+---
+```rust
+storage_iter_range(start_len: u64, start_ptr: u64, end_len: u64, end_ptr: u64) -> u64
+```
+Similarly to `storage_iter_prefix`
+creates an iterator object inside the host.
+###### Normal operation
+Unless lexicographically `start < end`, it creates empty an iterator.
```
mikhailOK:
```suggestion
Unless lexicographically `start < end`, it creates an empty iterator.
```

0006-bingings.md
```diff
@@ -0,0 +1,554 @@
...
+Iterates over all key-values such that keys are between `start` and `end`, where `start` is inclusive and `end` is exclusive.
+
+Note, this definition allows to either `start` or `end` keys to not actually exist on the given trie.
```
mikhailOK:
```suggestion
Note, this definition allows for `start` or `end` keys to not actually exist on the given trie.
```

0006-bingings.md
```diff
@@ -0,0 +1,554 @@
...
+
+###### Panics:
+* If `start_len + start_ptr` or `end_len + end_ptr` exceeds the memory container or points to an unused register it panics with `MemoryAccessViolation`;
+
+---
+```rust
+storage_iter_next(iterator_id: u64, key_register_id: u64, value_register_id: u64) -> u64
```
mikhailOK: why does this return the length of the value but storage_read for example doesn't?

0006-bingings.md
```diff
@@ -0,0 +1,591 @@
...
+`storage_write(u64::MAX, 0, u64::MAX, 1, 2)` -- insert key-value into storage, where key is read from register 0,
+value is read from register 1, and result is saved to register 2.
+
+#### Specification
+```rust
+read_register(register_id: u64, ptr: u64)
+```
+Writes the entire content from the scratch buffer `register_id` into the memory of the guest starting with `ptr`.
+###### Panics
+* If the content extends outside the memory allocated to the guest. In Wasmer, it returns `MemoryAccessViolation` error message;
```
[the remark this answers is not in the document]
nearmax: Added a comment.

This is pretty much true for any API that deals with raw pointers. We cannot avoid raw pointers with Wasm because we need to work with blobs of data of arbitrary size. The advantage of registers is that we now have one function that can corrupt the guest memory. Oppose to having multiple specialized functions like `read_account_id(ptr: u64)`  that can do that.

0006-bingings.md
nearmax: `storage_iter_next` does not have to return the length but should be able to signal the following situations upon calling `storage_iter_next`:
* Iterator has advanced to key/value;
* Iterator has stopped.

We can use `u64`-encoded bool to differentiate between them (1 for not stopped iterator, 0 for stopped), but then we also optionally want to know about the following case:
* Iterator has advanced to key/value, but value is zero bytes, so in some cases there is no need to read `value` from the register.

This means we need three values (0 for stopped iterator, 1 for not stopped but non-zero value size, 2 for not stopped but zero value size).

We could have this encoding, but we also could just return the length of the value with `u64::MAX` signifying the stopped iterator.

mikhailOK: detecting zero byte value doesn't seem like a useful case, and anyway the result of this function is stored in `register_len(value_register_id)` so returning is just for convenience.
but then the question is, why not have a return value in storage_read?
and also, why not `register_len(key_register_id)`

nearmax: Because it is more common when someone write `map` of `filter` over iterator to skip values that are empty. That's why it is the only place where it is useful. But I agree that `register_len` can be used anyway and looks more consistent. So I will change it.

nearmax: Changed.

ilblackdragon:
Two points:

1. Randomness currently is exposed as `random_buf` and `random_u64`.
We either need to explicitly define which random function we use, e.g. actual formula. 
Or just return random seed coming from the blockchain, and let std library of the language implement the actual randomness.

2. We need to surface timestamp from the `block`.

ilblackdragon:
Also we need to have crypto primitives in bindings (my assumption they will be tedious to implement and expensive to run otherwise). Specifically:
 - `erecover` that [Ethereum uses](https://github.com/ethereum/go-ethereum/blob/bd6879ac518431174a490ba42f7e6e822dcb3ee1/core/vm/contracts.go#L52)
 - ed25519 verify
 - also we should support BLS (seems ppl standardize on BLS12-381), but we can add it later.

nearmax:
Sure, we can add these crypto primitives later.

0006-bingings.md
```diff
@@ -0,0 +1,584 @@
...
+---
+```rust
+prepaid_gas() -> u64
+used_gas() -> u64
+```
+
+## Math
+
+```rust
+random_seed(register_id: u64)
```
bowenwang1996: Why does this need to take register_id?

nearmax: That's how we pass blobs between host and guest.

bowenwang1996: Do we need this for every function? It is somewhat confusing from outside. Maybe we can agree for certain functions which registers we write to.

nearmax: Blockchain Wasm APIs are in general ugly, we make it pretty by wrapping it with `near.ts`. Motivation on why we need registers is provided in the registers sections of this doc. It applies to every function that operates with blobs.

Also, this API is not subject to modifications anymore. We had several last call announcements for changes two weeks ago.

text/0006-bingings.md
```diff
@@ -0,0 +1,584 @@
...
+log_utf8(len: u64, ptr: u64)
+```
+Logs the UTF-8 encoded string. See https://stackoverflow.com/a/5923961 that explains
+that null termination is not defined through encoding.
+
+###### Normal behavior
+If `len == u64::MAX` then treats the string as null-terminated with character `'\0'`;
+
+###### Panics
+* If string extends outside the memory of the guest with `MemoryAccessViolation`;
```
fckt: Late for a party..

Right now, if contract passes invalid utf8 to log it will mean that whole transaction will became invalid. But isn't it just a matter of a client interpretation? I think the log interpretation (as well as we args) must orthogonal to the virtual machine.

PR #10, opened 2019-08-05
Wallet API

vgrichina:


mesquka:
Just checking the status of this as I have a usecase that could benefit. The only interaction with the user's wallet in my app is calling a contract function with an attached deposit, going through the process of adding a key to the users wallet to only use once and then remove has a lot of steps that could be removed with a general purpose transaction signing function.

mikedotexe:
Thanks for taking a stab at this Vlad, and having the foresight in the summer of 2019. I think we'll possibly make a new NEP, using the updated process described here:
https://github.com/near/NEPs/blob/95e2db9b315c0a801659034c6b76a5895ea78d55/README.md#process

Also, I've been looking into deep linking for MetaMask, as I'm quite impressed with the product, especially the mobile app. This has led me to the parsing here:
https://github.com/MetaMask/metamask-mobile/blob/598331385e1335d12d6d8a5712ab08ef1abc3005/app/core/DeeplinkManager.js#L27-L30

which uses https://www.npmjs.com/package/eth-url-parser

I believe we can take lessons from the following Ethereum EIPs:
- https://github.com/nagydani/EIPs/blob/master/EIPS/eip-681.md
- https://eips.ethereum.org/EIPS/eip-831

Our URI / deep linking will be quite different, since there are more Actions than just send or sign.
I'm also a bit hesitant to use paths like `login` (although we can keep this for backwards compatibility) and instead move to a system that's technically correct using [Actions](https://nomicon.io/RuntimeSpec/Actions.html), perhaps camel-cased.

So:
`{base-uri}/addKey?type=full&curve=ed25519&publicKey=abc…319&…`

where **addKey** is the camel-cased version of the Action.

Now this _doesn't_ mean that every path/slug must be an atomic Action. There will also be common needs like:
- Offline signing (like [this cookbook recipe](https://github.com/near/near-api-js/blob/75e3d77455f8d2bf5a780fd4a53ade65e3346508/examples/cookbook/utils/verify-signature.js))
- Batch Actions in a transaction, where we'd likely send a base64-encoded message that fits some defined structure, like we already have in NEAR Wallet.
- Other common needs…

I'm afraid I'm not terribly familiar with the NEAR Wallet codebase, but notice that there seem to be paths set up here, in case it's a useful resource for this discussion.
https://github.com/near/near-wallet/blob/0db73eef9fa21e45cf3485de93bb5f8a9cf55f20/packages/frontend/src/utils/wallet.js#L44-L49

frol:
@vgrichina Is this still relevant?

ori-near:

PR #11, opened 2019-08-22
Structured error messages

vgrichina:


nearmax:
I don't think there is a need for our own format. Rust Error messages are types, that are sometimes complex and nested sometimes with string data or blobs, e.g.:
https://github.com/nearprotocol/nearcore/blob/3978e027c2ab5e8606cd1c920e1bddce5c6837de/runtime/near-vm-runner/src/errors.rs#L55

or

https://github.com/nearprotocol/nearcore/blob/3978e027c2ab5e8606cd1c920e1bddce5c6837de/runtime/near-vm-logic/src/errors.rs#L4

Currently we provide a so-called "display" format of these errors, which is a string. If we want something structural we can simply serialize them as JSON. nearlib and other tools already know how to work with JSON-serialized Rust objects.

DanielRX:
Maybe not your own format but the NEP is tailored towards "There should be _some_ way for a computer to get the error code of an error when returned from the contract/RPC/nearlib/near-runtime-ts"
 Is it really the case that you have to use RegExps to work out which error you are dealing with? I did this for our tests but hoped I was missing something

bowenwang1996:
The problem I saw with this is that sometimes we convert between error types when propagating error messages and it is not clear to me how to compose error codes. For example, inside runtime we can have some error saying "account blah.blah.near cannot be created by near" but then at the client level, we might have some enum for errors and this error might get into the `Error::Runtime` variant for example, which might add some additional error code on top.

nearmax:
Here is an example of JSON-serialized nested error message:
```json
{"External":{"InvalidMethodNameDemo":"-12122BadN@me"}}
```
We can return both structured error message and its human readable equivalent like this:
```json
{"structured": {"External":{"InvalidMethodNameDemo":"-12122BadN@me"}},
"display": "External error: VM Logic returned an invalid method name -12122BadN@me" }
```
Both are constructed using:
```rust
let structured = serde_json::to_string(&err).unwrap();
let display  = err.to_string();
```

For the reference, this is how error message looks like in Rust:
```rust
HostError::External(ExternalError::InvalidMethodNameDemo(
            "-12122BadN@me".to_string()
        ))
```

vgrichina:
@nearmax I am not sure this nested representation is going to be very convenient. Writing code dealing with nested objects is definitely harder than comparing string keys. Looking up such errors also going to be harder than grepping / googling for given string key.

Like e.g. you want to load string from localized resources – how do you approach that? At some point you'll have to serialize to string key anyway.

I assume we can also automate serialization to something like `External.InvalidMethodNameDemo: External error: VM Logic returned an invalid method name -12122BadN@me`?

nearmax:
@vgrichina Inventing our own language should be the last thing we do. There are multiple structured representations of data that should work for us. JSON is the simplest one that comes to mind since all our tools already work with return data in JSON format.

Maintaining such error language would be difficult, and besides that I am not sure all error messages are going to be expressible with such language, for instance what do we do with the payload if error message has an associated payload, like string, blob, or number? Consider error message saying a string is not in UTF8 and carries the bytes in the wrong format.

>  Writing code dealing with nested objects is definitely harder than comparing string keys.

We are already doing it for return data, arguments, etc in our applayer tools since JSON is the most convenient format for JS. Our own format would require writing our parser (even if it is just a "." splitter, but then again what do we do with payloads or errors that have two suberrors?). JSON does not require writing our own parser.

> Like e.g. you want to load string from localized resources – how do you approach that? At some point you'll have to serialize to string key anyway.

I am not following what do you mean. Can you elaborate?

> Looking up such errors also going to be harder than grepping / googling for given string key.

Googling error message in JSON format is just as easy as in any other format. Google figured it out long time ago. C++/Rust compilation errors are coming in a random format and people are able to google them anyway, by just copypasting it.

In short:

|                                                                 | our own language | JSON |
|-----------------------------------------------------------------|------------------|------|
| can work with any error structure (payload, mutliple suberrors) | no               | yes  |
| does not require writing our own serializer/deserializer        | no               | yes  |
| does not require maintenance of the serializer/deserializer     | no               | yes  |
| already supported by every language possible                    | no               | yes  |
| development community is familiar with it                       | no               | yes  |
| Googlable errors?                                               | yes              | yes  |

DanielRX:
As I said above, surely the keep element of this issue is:

All errors, however they are serialized, stored, parsed, passed, should contain a unique `ErrorCode` that can be used to pinpoint the error
Nested errors shouldn't be a special case, you just treat it like a stack trace:
The error contains an Error Code, Error Message and "Sub Error" section that takes the format returned by the suberror and just puts it there

For the user/dev/machine, it parses the top error code, finds X_1:
If it knows X_1 is the error to handle, it stops there
If it doesn't expect X_1, it goes to `SubErrors` and checks there, it then finds X_2, repeating above with X_2 replacing X_1.

TL:DR
Errors should come with a code, and a place to look that code up, everything else is secondary to that

9999-structured-error-messages.md
```diff
@@ -1,77 +1,53 @@
-- Proposal Name: (fill me in with a unique ident, `my_awesome_feature`)
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
 - NEP PR: [nearprotocol/neps#0000](https://github.com/nearprotocol/neps/pull/0000)
-- Issue(s): link to relevant issues in relevant repos (not required).
 
 # Summary
 [summary]: #summary
 
-One paragraph explanation of the proposal.
+Make error messages structured and include machine-readable error codes.
```
DanielRX: Removing the structured part of this change might make the whole thing much simpler

9999-structured-error-messages.md
```diff
@@ -1,77 +1,53 @@
...
 
 # Rationale and alternatives
 [rationale-and-alternatives]: #rationale-and-alternatives
 
-- Why is this design the best in the space of possible designs?
-- What other designs have been considered and what is the rationale for not choosing them?
-- What is the impact of not doing this?
+Alternative is to pass error code as a separate field. This was abandoned as it's more complex to implement: all places where errors get converted from one type to another and/or serialized would need special attention.
+
+Dot-separated strings are used as codes as they are both machine and human readable. They also fit well into existing UI localization systems.
```
DanielRX: I propose a simple `N-*` prefix (like `TS1009` for a typescript error)

vgrichina:
> Our own format would require writing our parser (even if it is just a "." splitter, but then again what do we do with payloads or errors that have two suberrors?).

I don't want to parse error code splitting it on `.`.  I just want some error code string I can either plug as is (or with some prefix like `error.`) into localization system (e.g. https://ryandrewjohnson.github.io/react-localize-redux-docs/#add-translations-to-components) or use to make decisions in UI code.

I agree with @DanielRX, this is most important part:

> Errors should come with a code, and a place to look that code up, everything else is secondary to that

So @nearmax if you have alternative proposal that you are willing to implement – feel free to come up with alternate NEP.  I've mostly designed from standpoint of being able to plumb this myself through whole stack within 1-2 days without knowing much Rust, so I lean towards naivest solution.

nearmax:
> So @nearmax if you have alternative proposal that you are willing to implement – feel free to come up with alternate NEP. I've mostly designed from standpoint of being able to plumb this myself through whole stack within 1-2 days without knowing much Rust, so I lean towards naivest solution.

@vgrichina That's not how NEPs work (or RFCs that NEPs were designed after). When you propose something expect to iterate on it based on the arguments in the discussion. If someone gives an argument that you don't like reply back with a contrargument or agree with it, but do not tell the person to go somewhere else and make their own NEPs.

The runtime code of nearcore has the highest standards of quality because it is responsible for "safety" of our blockchain. I am completely against plumbing some code in 1-2 days into the runtime of nearcore especially by someone not familiar with Rust just because we don't want to write the very same code in nearlib+nearshell using JS. We need to completely agree on the code changes to the runtime in this NEPs **before** anyone starts plumbing it into runtime.

@DanielRX Let me give you more context to understand what we are discussing here. @vgrichina  wants to plumb some code into the nearcore runtime that formats the error messages in some desirable to him way. I suggest that this formatting should be rather done in the applayer, specifically: nearlib and nearshell. As a contract developer you will never know whether we format error messages in nealib+nearshell (and spit out something like `error.message.code: human readable error message`) or we do this formatting in the nearcore itself. Because contract developers and users always communicate with blockchain, deploy contracts, run function through nearlib+nearshell. What @vgrichina wants for unknown to me reason is to have this formatting code in nearcore , I suggest to have this code in nearlib+nearshell.

Implementation-wise it is the same code whether we do this formatting in nearlib+nearshell or nearcore. For nearcore we would have to take structure error messages expressed as Rust structures and write mapping from them into "error.message.code: human readable error message". For JS in nearlib we would have to take structured error messages expressed as JSON received from nearcore (JSON is a native format to JS so there is not complexity overhead) and write the very same mapping but in JS. We should have this code in JS instead of Rust because:
* runtime of nearcore is kept very clean and tested, so we need strong reason to add features to it;
* this is pure UX thing. Nearcore provides us with all needed structured information about error messages. UX-components like nearlib and nearshell choose whichever format for error messages they want.

In general, I find this discussion disorganized and this NEPs to be low-effort.
First of all, before choosing the error format we need to decide the strict technical criteria for error message format, and only after we agree on criteria decide the format that satisfies the criteria the best. @vgrichina what are the criteria for the message format in your opinion?
* That the error is googlable?
* what else?

The body of the NEPs itself is not well-thought through. Take for instance motivation section:
>We are doing this to allow client-side code (e.g. nearlib, studio, wallet) to make decisions based on errors returned by nearcode. Currently it requires hardcoding regexps that depend on error mesages (which often are edited for better readability, etc).

It does not explain many things:
* Why this code necessarily needs to live in nearcore and not nearlib/nearshell?
* Why we cannot pass messages in JSON format? -- JSON does not require parsing or regexps.
* Are you aware that some error messages are coming from the dependencies, like Wasmer/Cranelift/LLVM and we cannot tie decision making to them because they change with every version of Wasmer/Cranelift/LLVM.

The Rationale and alternatives section is also lacking:
> Alternative is to pass error code as a separate field. This was abandoned as it's more complex to implement: all places where errors get converted from one type to another and/or serialized would need special attention. Dot-separated strings are used as codes as they are both machine and human readable. They also fit well into existing UI localization systems.

That's not a single alternative. Passing errors in JSON-serialized structured form is another big one. I am pretty sure there are other alternatives, did this NEPs do a research?

@vgrichina Also, please do not skip the questions and arguments that me and @bowenwang1996 left in this thread. Please address them.

P.S. @DanielRX It is really hard for me to address the following comment of yours:
>As I said above, surely the keep element of this issue is: All errors, however they are serialized, stored, parsed, passed, should contain a unique ErrorCode that can be used to pinpoint the error Nested errors shouldn't be a special case, you just treat it like a stack trace: The error contains an Error Code, Error Message and "Sub Error" section that takes the format returned by the suberror and just puts it there For the user/dev/machine, it parses the top error code, finds X_1: If it knows X_1 is the error to handle, it stops there If it doesn't expect X_1, it goes to SubErrors and checks there, it then finds X_2, repeating above with X_2 replacing X_1. TL:DR Errors should come with a code, and a place to look that code up, everything else is secondary to that

This whole idea with stack trace etc is hard to grasp. Could you give some specific code examples in JS or Rust?

DanielRX:
@nearmax My point on the whole NEP is it should be reduced to simply "including Error Codes in all Error objects, and carrying them with the errors across layers":

Near core (for example) should return an object with an error code, and nearlib, near-shell, near-runtime-ts can take that error (and add their own if need be) and give that to the user.

By stack trace:

```ts
Error1 = {ErrorCode, ErrorMessage, SubErrors: [SubError1, SubError2]}
SubError1 = {ErrorCode, ErrorMessage, SubErrors: [...]}
````

So I can do
```ts
if(Error1.ErrorCode == Expected) { handleError(); }
else { checkSubErrors(Error1); } // Recursive, or nested, or any other form of traversal
```
and I will catch `WrappedErrors` with that, and if you were to print the Object, it would show much like a stack trace, where the error propagated through.

```ts
ReferenceError: notDefined is not defined
    at c (/stacktrace.js:9:3)
    at b (/stacktrace.js:6:3)
    at a (/stacktrace.js:3:3)
    at Object.<anonymous> (/stacktrace.js:11:1)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:456:32)
    at tryModuleLoad (module.js:415:12)
    at Function.Module._load (module.js:407:3)
    at Function.Module.runMain (module.js:575:10)
```
and for near it would let you see the path the error took, which helps for debugging

However: 
I agree that formatting should not be done in nearcore at all @nearmax (passing JSON to the nearlb/nearshell works just as good), I merely think that the sub issue of the Error Codes is needed, as having
```TS
const NEARRegex = /Transaction [a-zA-Z0-9]{1,100} failed. ABORT: "(.*)" filename: "(.*)" line: (.*) col: (.*)/;
```
in our code is not exactly robust.

As a developer, I don't care how the errors are serialised, I care about getting the ErrorCode so I can actually catch errors and handle them without having to use a RegExp to check if it's a runtime error or an assert that is designed to fail.

nearmax:
The result of the discussion is https://github.com/nearprotocol/nearcore/issues/1839 . Closing this NEP since the issue has a more up to date info.

PR #12, opened 2019-08-27
Network specification

ilblackdragon:


text/0000-network-spec.md
```diff
@@ -0,0 +1,160 @@
...
+  - Known peers, peers that this node has connected before in previous sessions and stored on disk to persist between runs. This list can be empty if this is first time to start the node.
+
+Both of these lists are merged and node starts by connecting to `max_peers` random peers from this list.
+If there is not enough peers in the lists, node periodically tries to connect to newly discovered peers with a exponential back off timeout (in case network is smaller than K in the first place) until it switches to *regular* mode. *Regular* mode means that node periodically checks if there is not enough peers (due to disconnects) and conncects to new peers.
+
+Additionally, to prevent various eclipse attacks where attacker forces node to have all incoming connection to them, we require node to have at least `min_outgoing_peers` outgoing connections.
+
+In pseudo code, this would look like:
+```
+  peers = bootstrap & known_peers - banned_peers
```
mfornet: Shouldn't this be `|` (union) instead of `&` (intersection)?
`peers = (bootstrap | known_peers) - banned_peers`

text/0000-network-spec.md
```diff
@@ -0,0 +1,160 @@
...
+Additionally, active participants, such as block producers and other nodes that want to link their account information to the peer information have a string `account_id` and appropriate `public_key` which is associated with this account. Accounts maintained by blockchain itself, and can not be forged without breaking the blockchain itself.
+
+### Storing network information
+
+A client stores information about other nodes in three data structures. `known_peers` a long-term database, which is stored on disk and persists between reboots. And two short-term databases `active_peers` and `routing` that start empty.
+
+`known_peers`. The `known_peers` is stored on disk and contains information about each node that client has ever connected or recevied gossip about. There is no limit to the size of the `known_peers`.
+Each entry of `known_peers` is tuple of `<node ID, IP address, TCP port, last_seen, status>`. Where `last_seen` is a timestamp when we received last message from this peer. If this was received from gossip, `last_seen` is 0. Status contains if this peer is currently unknown, connected, disconnected or banned.
```
DanielRX: Is this IPv4 or IPv6?

IPv6 would invalidate any attempt at IP bans, since you can have millions of addresses for a single computer with little effort

IPv4 is a smaller space and is more prone to directly identifying the peer, which is bad.

If known peers are tuples of `node ID` and `IP address` does that mean that a client that moves from one IP to another would come out as a brand new client, even though it has the same `node ID` as a previous peer?

text/0000-network-spec.md
```diff
@@ -68,42 +86,116 @@ When one peer creates outgoing connection to another peer, they send a `Handshak
...
-- 
+- *Create connection between two nodes*. We should only broadcast based on [localization]. If a new connection is relevant to us we broadcast to other nodes as an important hop. Note: We can force a maximum depth, so we don't send a connection if its exceed some depth. If we connect to other node this is relevant to us. (This way we keep track of all peer at distance 2 from us).
+
+- *Remove connection between two nodes*. We should not broadcast a removal if we didn't broadcast the creation. If node A receives a Remove connection [A, B] signed by B we should disconnect from B (probably malicious action by B). Also it is expected that nodes disconnect from each other when rebalancing network on epoch changes.
+
+Broadcasting messages at the network level is the main source of network overhead unrelated to blockchain protocol level. We should keep this as low as possible.
+
+### Syncing graphs
+
+When we connect to a new peer we should exchange relevant portions of the network to each other progressively. Portions that involve only high reputable nodes and validators.
```
ilblackdragon: would be good to have more detail here.
Because I estimate one edge to be 144 bytes (ip address + port + public key = 38 bytes, sig = 64 bytes, + timestamp 4 bytes + edge weight 4 bytes?).
And let's target 10k nodes with 40 connections in the network for back-of-envelope. That's ~27 MB of graph data.

text/0000-network-spec.md
```diff
@@ -68,42 +86,116 @@ When one peer creates outgoing connection to another peer, they send a `Handshak
...
+
+Edges are signed from both parties with timestamp. They should expire automatically after some period of time So they are re-broadcasted again. This period of time should be reasonably large so incur in small network overhead penalty.
+
+### Routing
+
+Routing mechanism specification.
+
+- *Routing to*: A peer can only send a message to another peer (target) if it knows a path to it in its current view of the network. It will send a message through shortest path using Round Robin (among all direct peers that have shortest path to target it will pick one).
```
ilblackdragon: how we going to do retry in cases of lost message / malicious actor in the middle / broken connection?

text/0000-network-spec.md
```diff
@@ -68,42 +86,116 @@ When one peer creates outgoing connection to another peer, they send a `Handshak
...
+
+- *Routing back*: The network design establish that path to reputable nodes are know, so for messages that need to be routed back (to probably some not reputable node) every node in the path should be aware of a message that needs a response and will keep track of this submission hash and predecessor (only predecessor since full route might be unknown). This will allow new nodes to fetch information from other nodes in the network that doesn’t know him (only those who are directly connected should know him).
```
ilblackdragon: I'm curious how this is done on the regular networking routers.

text/0000-network-spec.md
```diff
@@ -68,42 +86,116 @@ When one peer creates outgoing connection to another peer, they send a `Handshak
...
+
+We maintain a graph with all connections from our point of view. We should build a DAG/Tree (this should be similar to have several routes for some peer) in such a way that every path from us to other nodes is:
+
+- reasonably short.
+- use high reputation nodes.
+
+One idea to localize nodes in the network:
+
+- From our POV we have infinite reputation.
+- Build a directed graph where weight of edge [u -> v] is reputation of node [u].
```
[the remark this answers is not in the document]
mfornet: 1. We will only use real edges on the network. The answer is yes [me -> other node] has infinite weight, but we will only use active connection.

2. Can we trust in what reputation `u` tells about `v`? In my mind we keep reputation as a local value, since peers might forge high reputation values of other peers. What you are telling makes sense, since that path up to `u` will be important/short if intermediates hops also have high reputation. We should take care of malicious actor directly (or indirectly) telling about other nodes (the same actor in disguise) having high reputation.

text/0000-network-spec.md
mfornet: I will investigate into it.

text/0000-network-spec.md
mfornet: Should we handle this at the network level. Aren't we assuming at the protocol level that messages can get lost / dropped. What we can try is to identify at which position the issue take place and try to:
- ban or 
- remove edge and update network view.

text/0000-network-spec.md
mfornet: What is your exact concern here? 
I see this is a lot of data, but in theory this data will not travel the network too often.

text/0000-network-spec.md
mfornet: We are trying to address this issue in the new network design using a reputation system.

text/0000-network-spec.md
mfornet: AFAIU from [this answer](https://networkengineering.stackexchange.com/a/10216) the approach used by routers is almost the same as it is proposed in this document. The main difference is that instead of changing/mapping internal port to external port we are using nonces to describe each routed connection and create mapping.

I'll write implementation details.

text/0000-network-spec.md
```diff
@@ -20,40 +20,96 @@ This document should serve as reference for all the clients to implement network
...
+If view of the network differ by a large margin nodes will send their full view of the graph. Useful when new nodes join for the first time.
+Otherwise they will use bloom filters to find differences.
+
+**AddEdge/RemoveEdge** messages should contains a timestamp/nonce to avoid situations when removal gets to a node before addition (though in practice this should be extremely rare). When adding a new edge, both nodes after having signature from the other peer start broadcasting this new edge. Use multiple source broadcast mechanism.
+
+## Connecting Service
+
+This service will handle active connections.
+
+- It will determine if an incoming connection should be established. When some node is trying to establish a new connection if peers slots is full we will ping peer with latest last message. If ping fails, we drop our connection and establish connection with new peer.
```
ilblackdragon: hm, that means you are keeping connection open with newly connected peer during ping (what is the timeout on ping?). Eg. attack would be to spam with bunch of connections and which may get node to "too many files open".

Generally TCP/IP I think handles disconnects within reasonable time, so this might not really be needed.

text/0000-network-spec.md
```diff
@@ -20,40 +20,96 @@ This document should serve as reference for all the clients to implement network
...
+- It will try to establish new connections in several situations:
+    - Bootstrapping (When joining the network for the first time). Nodes will try to learn new nodes running
```
ilblackdragon: running...?

text/0000-network-spec.md
```diff
@@ -20,40 +20,96 @@ This document should serve as reference for all the clients to implement network
...
+
+## Sync Service
+
+This service will handle keeping the network view for every peer up to date using Broadcast mechanism. This will manage MST. When new edge is created this information is propagated through the network.
+
+Also there is a hard sync mechanism for two nodes, that is used when new connection between two nodes is established.  Periodically nodes will run this hard sync to become up to date. The period to hard sync is always large enough so that changes are allowed to get propagated through the spanning tree.
+If view of the network differ by a large margin nodes will send their full view of the graph. Useful when new nodes join for the first time.
+Otherwise they will use bloom filters to find differences.
+
+**AddEdge/RemoveEdge** messages should contains a timestamp/nonce to avoid situations when removal gets to a node before addition (though in practice this should be extremely rare). When adding a new edge, both nodes after having signature from the other peer start broadcasting this new edge. Use multiple source broadcast mechanism.
```
ilblackdragon: let's actually have data structure here - to explicitly show that this requires signature of both parties

text/0000-network-spec.md
```diff
@@ -20,40 +20,96 @@ This document should serve as reference for all the clients to implement network
...
+## Connecting Service
+
+This service will handle active connections.
+
+- It will determine if an incoming connection should be established. When some node is trying to establish a new connection if peers slots is full we will ping peer with latest last message. If ping fails, we drop our connection and establish connection with new peer.
+- It will try to establish new connections in several situations:
+    - Bootstrapping (When joining the network for the first time). Nodes will try to learn new nodes running
+    - Network balancing. We will have high tolerance to support connections to nodes that need to be close to use (same shard or block producers). Though we will have this number capped in case
+    - Low number of outbound connections.
+- Remove some of the active connections if the number of active peer goes over some threshold. We allow number of active peers to be unbound in case we need to establish new connections because lack of outbound connections, or because we becoming part of new shard.
```
ilblackdragon: I think totally unbounded number of connections is dangerous as can knock node into "too many files open" mode. Which is hard to recover from as our storage layer also uses files and starts to fall apart.

text/0000-network-spec.md
```diff
@@ -80,6 +136,27 @@ In pseudo code, this would look like:
...
+### API
+
+- **Add epoch_hash/height info:** Add information of all validators that belong to the same shard as we do for every epoch.
+- **Start active epoch_hash/height:** Active epoch are usually current and next epoch. When new epoch gets active this service will start to establish connection between peers in that epoch.
+- **Finish active epoch_hash/height:** This epoch_hash already finished so, connections made because of are marked as not important anymore, so can be dropped if the routing table grows
+
+### Policy
+
+- At all moments we should try to have a minimum number of outbound connections.
+- We will accept (or try) connections to other validators in our same shard.
```
ilblackdragon: shards -> each validator can be allocated to multiple shards.
Also only block + chunk producers really need to be fully interconnected.

"hidden" validators just need a sparse connecitivity.
Also non-validators are "tracking" subset of shards -> shards they want to have state for.
Non-validators don't need to connect to every single block/chunk producer either. But should be connected to at least to one node that tracks shard they are tracking and with reduced distance to chunk producers.

ilblackdragon:
Just want to leave it here for later:

# Network level privacy 
 - A compact and provable secure mix format - https://cypherpunks.ca/~iang/pubs/Sphinx_Oakland09.pdf
 - Low latency communication mixer that provides sender/receiver anonymity - https://arxiv.org/abs/1703.00536 (300 messages per second, ~1.5 s latency)
 - Coconut: Threshold Issuance Selective Disclosure Credentials with Applications to Distributed Ledgers - https://arxiv.org/abs/1802.07344 (identity layer for privacy networking rooted in blockchain)

text/0000-network-spec.md
mfornet: That is what I meant for block+chunk producers.

What about this tracking subset of shards? How do you imagine this subset will be selected, I mean do we care really about tracking some shards or is it ok to have sparse connectivity in general?

text/0000-network-spec.md
mfornet: This is how I imagine it. Each connection have some of this properties (maybe more than once at the same time).
- Node is block/chunk producer with us.
- Is outbound connection.
- Is important regarding topology efficiency (edge in the MST).
- Is trusted (this is ambiguous it can be hardcoded/old/reliable connection).

We can fix a maximum number per property (notice again that several connection can share some properties). When a connection drops some properties and become not meaningful to us we can stop it gracefully. I think it would be very good if we don't cap connections with first property (though it might be an issue as you mentioned).

mfornet: This is how I imagine it. Each connection have some of this properties (maybe more than once at the same time).
- Node is block/chunk producer with us.
- Is outbound connection.
- Is important regarding topology efficiency (edge in the MST).
- Is trusted (this is ambiguous it can be hardcoded/old/reliable connection).

We can fix a maximum number per property (notice again that several connection can share some properties). When a connection drops some properties and become not meaningful to us we can stop it gracefully. I think it would be very good if we don't cap connections with first property (though it might be an issue as you mentioned).

text/0000-network-spec.md
mfornet: Actually I got this idea from Ethereum. Maybe for new incoming connections we should use UDP instead, to avoid wasting resources on inactive connections. And yes, we can get rid of this ping message, just need to think if it can be easily attacked somehow.

ilblackdragon:
@mfornet @Kouprin any chance we can refurbish this into spec format instead of current NEP and merge?

mfornet:
> @mfornet @Kouprin any chance we can refurbish this into spec format instead of current NEP and merge?

This is part of the stability push effort, however we haven't started writing specs yet.

PR #13, opened 2019-09-04
System methods in runtime API

evgenykuzyakov:


0000-system-methods.md
```diff
@@ -0,0 +1,221 @@
...
+- Start Date: 2019-09-03
+- NEP PR: [nearprotocol/neps#0000](https://github.com/nearprotocol/neps/pull/0000)
+- Issue(s): link to relevant issues in relevant repos (not required).
+
+# Summary
+[summary]: #summary
+
+Adds new ability for contracts to perform some system functions:
+- create new accounts (with possible code deploy and initialization)
+- deploy new code (upgrades)
```
nearmax:
```suggestion
- deploy new code (or redeploying code for upgrades)
```

0000-system-methods.md
```diff
@@ -0,0 +1,221 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Contracts should have the ability to create new accounts, transfer money without calling code and
+stake. It will enable full functionality of contract-based accounts.
+
+# Reference
+
+We introduce additional promise APIs to support batched actions.
```
nearmax: Somewhere in this NEP we should emphasize that the existing promise API is the special case of batched promise API and we should give a specific way to reproduce the existing promise API through this promise API.

0000-system-methods.md
```diff
@@ -0,0 +1,221 @@
...
+
+Firstly, we enable ability to create empty promises without any action. They act similarly to
+traditional promises, but don't contain function call action.
+
+Secondly, we add API to append individual actions to promises. For example we can create
+a promise with a function_call first using `promise_create` and then attach a transfer action on top
+of this promise. So the transfer will only deposit tokens if the function call succeeds. Another example
+is how we create accounts now using batched actions. To create a new account, we create a transaction with
+the following actions: `create_account`, `transfer`, `add_key`. It creates a new account, deposit some funds on it and the adds a new key.
```
nearmax: This is not the full list of actions. Could we list all of them?

evgenykuzyakov: This is just examples. I think they are covered well in the batched transaction NEP

0000-system-methods.md
evgenykuzyakov: done

evgenykuzyakov:
> We need a separate error message for when the sequence of actions is currently not supported. AFAIR there is one.

During the creation of a receipt we don't validate the sequence of actions. Only later once the receipt is sent and routed we validate actions during execution.

PR #14, opened 2019-09-13
WIP: Split contract execution

nearmax:


text/0014-split-contract-execution.md
```diff
@@ -0,0 +1,69 @@
...
+that calls itself;
+2. When receipt of the promise calls this method second time it performs one more iteration, this time using the second
+element and issues the promise to itself again. Step (2) is then performed until the collection is exhausted;
+3. When collection is exhausted the execution stops and the result is propagated using `promise_then` such that from
+RPC standpoint it looks like the contract call returned that aggregated result.
+
+
+## Distributed execution
+
+For general use-cases we can implement a framework that allows doing MapReduce. We can decide implementation details
```
bowenwang1996: Will this be part of this NEP?

text/0014-split-contract-execution.md
```diff
@@ -0,0 +1,69 @@
...
+final signature it stops calling itself.
+
+Similarly, when contract execution requires too much space we can partition it across multiple contracts (in expectation
+that they will end up on different shards). For instance ethash computation requires storing a very large graph (~1GiB)
+which we can split across multiple contract instances that perform their part of verification in parallel before calling
+the "aggregator" to call the result, i.e. MapReduce.
+
+## Non-distributed execution
+
+For non-distributed execution we can provide nice convenience functions to the contract developer using near-bindgen.
```
bowenwang1996: Is there anything specific to rust that enables this?

text/0014-split-contract-execution.md
```diff
@@ -0,0 +1,69 @@
...
+
+For specialized cases, like ethash computation we can implement distributed computation without the framework.
+
+## Alternatives
+
+I was not able to find the alternative approaches by other blockchains. In most cases they solve expensive computation
+by either making a fast host-function or a precompile, which does not help if computation is by itself very slow
+and heavy.
+
+### Preserving the entire Wasm state
```
DanielRX: Would it be fair to describe this similar to a "breakpoint" where another agent (the reducer, the sub contract) must then "step forward" the contract, and it will continue as normal?

text/0014-split-contract-execution.md
```diff
@@ -0,0 +1,69 @@
...
+## Distributed execution
+
+For general use-cases we can implement a framework that allows doing MapReduce. We can decide implementation details
+once we have a specific use case.
+
+For specialized cases, like ethash computation we can implement distributed computation without the framework.
+
+## Alternatives
+
+I was not able to find the alternative approaches by other blockchains. In most cases they solve expensive computation
```
DanielRX: The alternative in my eyes (while not a direct replacement) would be something like strong support for ZK-SNARKs and the like, which allow heavy computation to be done off chain and verified in a clear way on chain

Given the specific example, is there no reduction that can be made for the problem such that an external host could produce a proof the on chain contract can consume?

text/0014-split-contract-execution.md
```diff
@@ -0,0 +1,69 @@
...
+Similarly, when contract execution requires too much space we can partition it across multiple contracts (in expectation
+that they will end up on different shards). For instance ethash computation requires storing a very large graph (~1GiB)
+which we can split across multiple contract instances that perform their part of verification in parallel before calling
+the "aggregator" to call the result, i.e. MapReduce.
+
+## Non-distributed execution
+
+For non-distributed execution we can provide nice convenience functions to the contract developer using near-bindgen.
+
+Here is an example of BLS aggregator:
```
DanielRX: I assume this is for few, large calculations (or batches of), vs lots of small ones? As this much provide some overhead for each async call?

text/0014-split-contract-execution.md
nearmax: ZK-SNARK are not suitable for general purpose computing, since they only work if computation can be expressed as a polynomial function.

ilblackdragon:
@nearmax I'm thinking we should start policy for NEPs as well. Because we have super old PRs here still sitting.

nearmax:
> @nearmax I'm thinking we should start policy for NEPs as well. Because we have super old PRs here still sitting.

I think we should allow NEPs to sit as long as they wish, similarly to Rust RFCs https://github.com/rust-lang/rfcs/pulls (some of them are ~4y.o.). NEPs allow expressing large and ambitious ideas that should be allowed to sit there for long time without ETA on implementation.

ilblackdragon:
Should this be closed @nearmax ?

nearmax:
While I think this is a very promising direction that can unlock transaction size (in terms of execution gas), this proposal got stale and we should close it until someone else decides to champion it.

PR #17, opened 2019-09-23
Execution Outcome

evgenykuzyakov:


text/0017-execution-results.md
```diff
@@ -0,0 +1,91 @@
...
+    pub logs: Vec<LogEntry>,
+    /// Receipt IDs generated by this transaction or receipt.
+    pub receipt_ids: Vec<CryptoHash>,
+    /// The amount of the gas burnt by the given transaction or receipt.
+    pub gas_burnt: Gas,
+}
+
+pub enum ExecutionResult {
+    /// The execution of the given transaction or receipt is pending.
+    Pending,
```
ailisp: I feel ExecutionStatus/FinalExecutionStatus is still a better name then *Result. Since pending here and Not Started/Started in FinalExecutionResult is not the result but more like a status that will make progress overtime

evgenykuzyakov: Yeah, we can use `Status`. I just don't want to have both `result` and `status`

text/0017-execution-results.md
```diff
@@ -0,0 +1,91 @@
...
+    Failure,
+    /// The final action succeeded and returned some value or an empty vec.
+    SuccessValue(Vec<u8>),
+    /// The final action of the receipt returned a promise or the signed transaction was converted
+    /// to a receipt. Contains the receipt_id of the generated receipt.
+    SuccessReceiptId(CryptoHash),
+}
+
+// TODO: Need a better name
+pub struct ExecutionOutcomeWithId {
```
frol: I am not that familiar with the terminology, but my take on this would be `Receipt` or even two separate (or type alias) structures: `Receipt` and `Transaction`.

PR #18, opened 2019-09-26
Improve view/change methods

bowenwang1996:
A simple approach to improving and stabilizing the difference between view and change methods in contracts.

### DIFF
--- a/text/0000-view-change-method.md
+++ b/text/0000-view-change-method.md
@@ -15,22 +15,26 @@ to differentiate between the two in frontend or through near-shell.
 [motivation]: #motivation
 
 From the feedback we received it seems that developers are confused by the results they get from view calls, which are
-mainly caused by the context variables such as `signer_account_id`, `current_account_id`, `attached_deposit` do not make
-sense in a view call. To avoid such confusion and create better developer experience, it is better if those context variables
+mainly caused by the fact that some binding methods such as `signer_account_id`, `current_account_id`, `attached_deposit`
+do not make sense in a view call. 
+To avoid such confusion and create better developer experience, it is better if those context variables
 are prohibited in view calls.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
 
-Among context variables, some do make sense in a view call, such as `block_index`, while the majority does not. Here we
-explicitly list the variables are not allowed in a view call and, in case they are invoked, the contract will panic with
+Among binding methods that we expose from nearcore, some do make sense in a view call, such as `block_index`,
+while the majority does not. 
+Here we explicitly list the methods are not allowed in a view call and, in case they are invoked, the contract will panic with
 `Context variable <context_variable> is not allowed in view calls`.
 
-The only allowed context variables are:
-    - `block_index`
-    - `storage_usage`
-
-All others are prohibited.
+The following methods are prohibited:
+    - `signer_account_id`
+    - `signer_account_pk`
+    - `predecessor_account_id`
+    - `attached_deposit`
+    - `prepaid_gas`
+    - `used_gas`
 
 From the developer perspective, if they want to call view functions from command line on some contract, they would just
 call `near view <contractName> <methodName> [args]`. If they are building an app and want to call a view function from the
@@ -40,9 +44,11 @@ frontend, they should follow the same pattern as we have right now, specifying `
 # Reference-level explanation
 [reference-level-explanation]: #reference-level-explanation
 
-To implement this NEP, we need to change how context is handled in runtime. More specifically, we can add a field `is_view: bool`
-in `VMContext` to indicate whether we are processing a view call. If `is_view` is true, then all the access to the prohibited
-variables will error with the error message described above.
+To implement this NEP, we need to change how binding methods are handled in runtime. More specifically, we can rename
+`free_of_charge` to `is_view` and use that to indicate whether we are processing a view call. In addition we can add
+ a variant `ProhibitedInView(String)` to `HostError` so that if `is_view` is true,
+then all the access to the prohibited
+methods will error with `HostError::ProhibitedInView(<view method name>)`.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -61,7 +67,7 @@ requires much more work and is not currently feasible for Rust contracts.
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-- Should we allow any context variables to be accessed at all? 
+- Should we allow promise API to be called in view methods?
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
--- a/text/0000-view-change-method.md
+++ b/text/0000-view-change-method.md
@@ -26,7 +26,7 @@ are prohibited in view calls.
 Among binding methods that we expose from nearcore, some do make sense in a view call, such as `block_index`,
 while the majority does not. 
 Here we explicitly list the methods are not allowed in a view call and, in case they are invoked, the contract will panic with
-`Context variable <context_variable> is not allowed in view calls`.
+`<method_name> is not allowed in view calls`.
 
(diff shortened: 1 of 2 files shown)


evgenykuzyakov:
I suggest the following approach. Instead of introducing a binary `is_view`, allow view RPC to inject  some context fields.
For example for nearlib view methods it makes sense to inject `singer_id`, `predecessor_id` and `singer_public_key` when there user logged in with the AccessKey. But if the user is not logged in then RPC should not have `signer_id`, etc.

Now in the Runtime, we can change some fields in the VMContext with optional values. E.g.  `signer_id: Option<AccountId>`. Then when `account_signer_id` is called, Runtime logic will panic with HostError that `signer_id` is not provided for `None`.
Same can be done for `predecessor_id` and `singer_public_key`.

It's also possible that we don't need to change a lot on the RPC level when introducing optional fields (especially with JSON). But need to recompile them with borsh serialization

evgenykuzyakov:
An example of the use-case when you want to inject context is `metanear`. You call something like:
`lookAround()` and if you're logged in, it would give a map around you.

Also it make sense to inject different context in test. E.g. you want to check how your function react based on authorized or non-authorized access.

text/0000-view-change-method.md
```diff
@@ -0,0 +1,69 @@
...
+
+From the feedback we received it seems that developers are confused by the results they get from view calls, which are
+mainly caused by the context variables such as `signer_account_id`, `current_account_id`, `attached_deposit` do not make
+sense in a view call. To avoid such confusion and create better developer experience, it is better if those context variables
+are prohibited in view calls.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Among context variables, some do make sense in a view call, such as `block_index`, while the majority does not. Here we
```
nearmax: It is confusing to refer to "context variables" because what we are banning is some functions Bindings APIs. Instead of stating which functions are allowed it is better to state which functions are prohibited. For instance, this document implies that `input()` function is prohibited (because it is a part of the context), but actually we don't want to prohibit it. `current_account_id()` also should not be prohibited (which is again implied to be prohibited by this doc), but `signer_account_id()` should, etc. Please follow https://github.com/nearprotocol/NEPs/blob/master/text/0006-bingings.md and list explicitly what functions are prohibited.

text/0000-view-change-method.md
```diff
@@ -0,0 +1,69 @@
...
+explicitly list the variables are not allowed in a view call and, in case they are invoked, the contract will panic with
+`Context variable <context_variable> is not allowed in view calls`.
```
nearmax: I suggest we have an error type `HostError::ProhibitedInView(String)` where `String` refers to the function name and  `ProhibitedInView` implements `Display` trait that returns `Method <method_name> is not allowed in view calls`.

text/0000-view-change-method.md
nearmax: Also, some functions are not part of `Context API` but are rather part of `Economics API`, like `attached_deposit()`, so it makes sense to not refer to "Context variables" but instead talk about "Bindings methods" because that's what contract developers are going to be interfacing with.

text/0000-view-change-method.md
```diff
@@ -0,0 +1,69 @@
...
+
+This design is very simple and requires very little change to the existing infrastructure. An alternative solution is
+to distinguish between view methods and change methods on the contract level. One way to do it is through decorators, as
+described [here](https://github.com/nearprotocol/NEPs/pull/3). However, enforcing such distinction on the contract level
+requires much more work and is not currently feasible for Rust contracts. 
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Should we allow any context variables to be accessed at all?
```
nearmax: We totally should :) `input`, `block_index`, `storage_usage`, `account_balance`, `random_seed` , etc are all useful.

The bigger question is, should we prohibit "Promise API" since calling it does not make sense in view method? @evgenykuzyakov  However, I think it is not as confusing as `signer_account_id` so we can keep it for now.

nearmax:
@evgenykuzyakov 

Regarding injecting the info, how is it going to look on RPC level? Do we require RPC caller to pass all context, e.g. `signer_account_id`, `signer_account_pk`, `predecessor_account_id`, `attached_deposit`, `prepaid_gas`, `used_gas`, etc even if the function does not use it? Because there is no way to tell the data that function will try to access during its call, without looking into Wasm. Or do we allow injecting anything, and then if some data was not injected but has been tried to be accessed we panic? Then it sounds like an extension of this proposal. Should we first implement this NEP and later maybe extend it with your idea?

> I suggest the following approach. Instead of introducing a binary is_view, allow view RPC to inject some context fields.

We already have this field in context it is called `free_of_charge`.

> For example for nearlib view methods it makes sense to inject signer_id, predecessor_id and signer_public_key when there user logged in with the AccessKey. But if the user is not logged in then RPC should not have signer_id, etc.

This behavior seems very specific no nearlib use case. Any ideas where else it can be useful?

> Also it make sense to inject different context in test. E.g. you want to check how your function react based on authorized or non-authorized access.

It is not clear what test are you talking about. Can you give more details? If we are talking about having integration or unit tests for AssemblyScript contracts, then you can already call them with injected context, that's what https://github.com/nearprotocol/nearcore/tree/master/runtime/near-vm-runner-standalone is for and @janedegtiareva is already using it in near-runtime-ts tests.

text/0000-view-change-method.md
bowenwang1996: Good point. I suggest we prohibit it because otherwise people might call it in view by mistake and getting strange errors.

bowenwang1996:
@nearmax @evgenykuzyakov Regarding injecting context, it seems to me that only injecting certain context like `current_account_id` is really useful. If that's the case, we can either require some specific context to be set in RPC or make it optional and if not set, the call to this method will error at runtime.

evgenykuzyakov:
`current_account_id` is the contract. You already know it

bowenwang1996:
I meant `signer_account_id`

nearmax:
As I said before, we can always add context injection through RPC later if there is a demand. Tests can already use https://github.com/nearprotocol/nearcore/tree/master/runtime/near-vm-runner-standalone  that allows context injection.

nearmax:
@bowenwang1996 Let me know when I can look at it again.

text/0000-view-change-method.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Among binding methods that we expose from nearcore, some do make sense in a view call, such as `block_index`,
+while the majority does not. 
+Here we explicitly list the methods are not allowed in a view call and, in case they are invoked, the contract will panic with
+`<method_name> is not allowed in view calls`.
+
+The following methods are prohibited:
```
nearmax: Formatting is off.

bowenwang1996:
I am leaning towards explicitly banning promise API in view calls. @evgenykuzyakov what do you think?

bowenwang1996:
Okay we decide to not allow promise API to be called in view calls.

DanielRX:
> Okay we decide to not allow promise API to be called in view calls.

Does this also mean no cross contract calls since they are async or is it only the explicit use of async?

text/0000-view-change-method.md
```diff
@@ -35,6 +35,23 @@ The following methods are prohibited:
   * `attached_deposit`
   * `prepaid_gas`
   * `used_gas`
+  * `promise_create`
```
DanielRX: Would it be better to use `promise_*` to exclude all future promise methods without having to add to the list?

bowenwang1996: I think it's better to be explicit so that if some other promise methods is added we can remember to do the proper update.

PR #20, opened 2019-10-10
Block limiters

nearmax:


text/0020-block-limiters.md
```diff
@@ -0,0 +1,56 @@
...
+[motivation]: #motivation
+
+We have several problems at the moment:
+
+* Currently our block production can be easily abused. For instance one can submit 20 transactions each of which
+deploys a 1MiB contract. Together these transactions would take > 1 sec to process, however block producer will happily
+include them into the block and when the next block is produced (potentially by a different block producer) node will choke on
+processing it;
+
+* Currently we have a limit on how many transactions+receipts we include in the block, but if we have number of receipts
```
bowenwang1996: Actually this is not true. We only have a limit on the number of transactions and include all the receipts.

text/0020-block-limiters.md
```diff
@@ -0,0 +1,56 @@
...
+
+1. `block_size_limit` -- as block producer includes receipts and transactions into the block
+they count the total size of serialized receipts and transactions. Block producer always starts by including receipts
+from the previous block and cross-shard receipts, as they do this they do not check the total size against `block_size_limit` (because we never drop the receipts).
+Then block producer starts including transactions into the block, if total size exceeds `block_size_limit` they stop. 
+
+    Note, it is possible to have no transactions in the block and it is possible for receipts in the block to exceed the threshold.
+    In other words, it is possible for shard to get spammed by other shards and get slowed down;
+    
+2. `block_gas_limit` -- as block producer executes transactions and receipts from the previous block they keep track
```
bowenwang1996: So are we going to remove gas limit on transactions?

bowenwang1996: Actually I think the current `gas_limit` is not intended to be used as gas limit per transaction.

nearmax: We still keep the gas limit per transaction to avoid having infinite loop contracts.

bowenwang1996: What I meant is whether we are going to have a separate parameter for gas limit per transaction (we can use the block gas limit for each transaction as well).

nearmax: We have two gas limits:
* For the entire block -- since we cannot predict transactions before we execute them when we put them into the block;
* For a single function call -- since we want to have a sane cut-off for smart contracts that attempt doing infinite loop.

I don't think we need to have any other gas limits.

nearmax:
Closing it until we start working on cross-shard receipt congestion. CC @evgenykuzyakov

PR #23, opened 2019-11-14
Move operator

ilblackdragon:


evgenykuzyakov:
As discussed offline, it's currently not easy to do, so might not happen

ilblackdragon:
Moving design forward with https://github.com/nearprotocol/NEPs/pull/24

PR #28, opened 2020-01-17
Dapp Nodes

k06a:


ilblackdragon:
We should also consider alternative all view calls become transactions, and get executed within consensus (charge for reading data / computation). This removes need in data witnesses (merkalization of the data)

frol:
It seems that [near-workspaces](https://github.com/near/workspaces) allows achieving the behavior described in this NEP. Feel free to open a new NEP with more details fleshed out if near-workspaces is not sufficient.

PR #31, opened 2020-02-14
NEAR CLI enhancements (#31)

damons:
Initial draft of NEP for NEAR Shell enhancements.  Comments required and appreciated.

### DIFF
--- a/text/0000-near-shell.md
+++ b/text/0000-near-shell.md
@@ -20,11 +20,11 @@ These enhancements are intended to simplify and enhance developer and user produ
 
 ## Settings, config, and key management
 
-Project-level **settings**, connection **configuration**, and **key management** are the three types of stored data necessary for developers to reliably build and deploy projects.
+Shell-experience **settings**, connection **configuration**, and **key management** are the three types of stored data necessary for developers to reliably build and deploy projects.
  
 ```
 ─ awesome-near-project  ⟵ NEAR dApp
- ├── .near-config       ⟵ Stores project-level settings and connection configuration
+ ├── .near-config       ⟵ Stores shell-experience settings and connection configuration
  │  ├── connections     ⟵ Contains files used to connect and deploy a contract (except keys)
  │  │  ├── default.js   ⟵ Default configuration is for development
  │  │  └── localnet.js  ⟵ Custom connection added by user for localnet
@@ -37,7 +37,7 @@ Project-level **settings**, connection **configuration**, and **key management**
 ### Definitions:
 
 #### Settings
-[project-level-settings]: #project-level-settings
+[shell-experience-settings]: #shell-experience-settings
 
 ---
 
@@ -45,7 +45,7 @@ These are key-values reflecting how `near-shell` behaves when called within a NE
     
 Example: a developer using OS X uses `near login` with access to a browser, whereas a validator runs `near login` on a CentOS box with no UI or browser. A user prompt may ask "Is this login from a computer with a browser?" The answer is then stored in key-value format in the project-level directory so it can be referenced later, skipping the prompt in the future.
 
-Besides answers to user prompts, project-level settings also store information about the last version of `near-shell` used in this project. For more information, please see [Upgradability](#upgradability).
+Besides answers to user prompts, shell-experience settings also store information about the last version of `near-shell` used in this project. For more information, please see [Upgradability](#upgradability).
 
 As shown in the directory structure above, this file is located in the project directory at:
 
@@ -53,7 +53,7 @@ As shown in the directory structure above, this file is located in the project d
 
 In summary, "settings" are key-value pairs that are set by `near-shell` and relate to the behavior of how it operates in a given project directory. They do not relate to the development or deployment of smart contracts, except for providing default values like `accountId`.
 
-**Note**: project-level settings can exist in the user's home directory as well. In this case, projects without their own `.near-config/settings.js` will use the home directory settings. `near-shell` will be quite verbose when running a command, letting the user know which settings and configuration files are being used when the user runs commands.
+**Note**: shell-experience settings can exist in the user's home directory as well. In this case, projects without their own `.near-config/settings.js` will use the home directory settings. `near-shell` will be quite verbose when running a command, letting the user know which settings and configuration files are being used when the user runs commands.
 
 To make this more clear, there are two nested arrays that differentiate settings that should only be applied from a project-level file from those in the home directory. This is how a `settings.js` file may look:
 
@@ -69,7 +69,7 @@ To make this more clear, there are two nested arrays that differentiate settings
 
 As defined later, there is a command `near settings set-default` which copies the current project's settings into the home directory. Keys inside the `project-only` object will not be moved.
 
-If a new project is created and has no project-level settings yet and the home directory has a settings file, this will be used as the template. Otherwise, a project-level settings file will be created and populated based on the user prompts given during regular usage of `near-shell`. 
+If a new project is created and has no project-level settings file yet, and the home directory has a settings file, the home directory's file will be used as the template. Otherwise, a project-level settings file will be created and populated based on the user prompts given during regular usage of `near-shell`. 
     
 #### Configuration
 
@@ -161,7 +161,7 @@ It's important to invite the international community into developing with NEAR. 
 
 Language preference can be set in two ways:
 - Using environment variables (i.e., parsed from `process.env.LANG`, and the default assignment)
-- Set explicitly in project-level settings using [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
+- Set explicitly in shell-experience settings using [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
 
 Example content from `<project-path>/.near-config/settings.js`:
 ```json
@@ -181,7 +181,7 @@ This proposal declares that `near-shell` will have a mechanism to make such upgr
 
 **Example scenario**: a user begins developing a dApp on NEAR using `near-shell` version `0.19.0`. The user takes a sabbatical for a few months and returns with a new computer, cloning the old project. The new computer installed `near-shell` version `0.23.1`. The first time this user runs a command in this old project, migrations are run ensuring the project stays current.
 
-Of the three types of storage mentioned before (project-level settings, connection configuration, and key management) the saved version information belongs to the project-level settings. It's possible that a user will have multiple dApps that are used less frequently than others and become out of date. Therefore, each project must have their own record of which version of `near-shell` was most recently used. This setting will never be placed in the home directory.
+Of the three types of storage mentioned before (shell-experience settings, connection configuration, and key management) the saved version information belongs to the project-level settings. It's possible that a user will have multiple dApps that are used less frequently than others and become out of date. Therefore, each project must have their own record of which version of `near-shell` was most recently used. This setting will never be placed in the home directory.
 
 The location of these settings is:
 
@@ -278,7 +278,7 @@ This category is used to create, select, and configure accounts on NEAR networks
 * `near account delete` : Delete an account or sub-account.
 * `near account create-key` : Delete an account or sub-account.
 * `near account revoke-key` : Delete an account or sub-account.
-* `near account select` : Selects a default account from the list of accounts with keys. Stored in project-level settings with key `defaultAccount`.
+* `near account select` : Selects a default account from the list of accounts with keys. Stored in shell-experience settings with key `defaultAccount`.
 * `near account login`  : Log in the current active or specified account.
 * `near account stake`  : Stake Ⓝ to a given staking contract
 * `near account secure` : Finds the key file for an account name, converts it to use OS-level key management via a command line wizard/instructions.
@@ -289,7 +289,7 @@ This category is used to create, select, and configure accounts on NEAR networks
 * `near config`         : List the location of active configuration file if it's loaded, or default config settings
 * `near config add`     : Set a key in the active configuration, or if using default, create a config file in the home directory with defaults and the specified key and value
 * `near config remove`  : Removes a configuration file. (Example: `near config remove localnet` removes `.near-config/connections/localnet.js`)
-* `near config select`  : Selects a default connection configuration file from the list of those available. Stored in the project-level settings with key `defaultConnection`. Default is `default`.
+* `near config select`  : Selects a default connection configuration file from the list of those available. Stored in the shell-experience settings with key `defaultConnection`. Default is `default`.
 * `near config wizard`  : Runs through a command-line wizard for an environment, allowing user to modify the values.
 
 ### `near settings <command>`
@@ -362,7 +362,7 @@ Reading from a file may be added to other commands in the future.
 
 Usage analytics will be used anonymously and purely to measure key performance indicators. [Mixpanel](https://mixpanel.com/) will be the chosen integration at this time. No private data will ever be sent, only metrics on the command usage and possible crash reporting.
 
-There will be the ability to opt-out as well. While the majority of settings are project-level, this is the only setting that will be saved directly to the user's home directory settings instead of the project-level settings file. (That is, stored in: `~/.near-config/settings.js`)
+There will be the ability to opt-out as well. While the majority of shell-experience settings are project-level, this is the only setting that will be saved directly to the user's home directory settings instead of the project-level settings file. (That is, stored in: `~/.near-config/settings.js`)
 
 ### User Stories
 
@@ -418,7 +418,7 @@ Adding the [upgradability mechanism](#upgradability) is key to `near-shell` bein
 
 Out of scope is smart contract execution and debugging. However, future versions of `near-shell` might include specific commands that enable the debugging and optimization of smart contract deployments. Gas estimation is the first step that will likely bring execution much closer in feasibility.
 
-In this proposal, `near account select` and `near config select` add to the project-level settings. These could arguably exist in the connection configuration.
+In this proposal, `near account select` and `near config select` add to the shell-experience settings. These could arguably exist in the connection configuration.
 
 #### Additional unresolved thoughts/questions
 
@@ -443,3 +443,4 @@ This section highlights actionable items needed to get from the current state of
 * Keys stored in `neardev` need to have an additional key added: `"type": "unencrypted"`.
 * The `neardev` folder needs to be renamed to `.near-credentials`.
 * Add `--env` flag so that config will load proper file.
+* Begin implementing yargs' [locale ability](https://yargs.js.org/docs/#api-locale) for translation.
--- a/text/0000-near-shell.md
+++ b/text/0000-near-shell.md
@@ -6,35 +6,17 @@
 # Summary
 [summary]: #summary
 
-NEAR Shell is one of the primary tools NEAR developers must use to develop and interact with NEAR networks and smart contracts.  Today, NEAR Shell contains the base commands for account creation and deletion, login, viewing account state and keys, sending tokens, creating staking transactions, and building, deploying, and calling smart contracts. There are more opportunities to integrate core NEAR features to expose them directly to users.  This proposal is to enhance NEAR Shell to represent all major functionality of NEAR.  For example, validator node management, network selection, native OS key management and integration, smart contract development (including testing, and interaction) could be added.  
+`near-shell` is one of the primary tools NEAR developers must use to develop and interact with NEAR networks and smart contracts.  Today, `near-shell` contains the base commands for account creation and deletion, login, viewing account state and keys, sending tokens, creating staking transactions, and building, deploying, and calling smart contracts. There are more opportunities to integrate core NEAR features to expose them directly to users.  This proposal is to enhance `near-shell` to represent all major functionality of NEAR.  For example, validator node management, network selection, native OS key management and integration, smart contract development (including testing, and interaction) could be added.  
 
-NEAR Shell is an excellent tool to attract and engage new NEAR developers, both core and contract level, and if it contains all the required features to select and configure and deploy NEAR networks, nodes, crypto assets (i.e., keys), accounts, and wallets, it can be the ideal single-entry point for those potential new hackers.  
+`near-shell` is an excellent tool to attract and engage new NEAR developers, both core and contract level, and if it contains all the required features to select and configure and deploy NEAR networks, nodes, crypto assets (i.e., keys), accounts, and wallets, it can be the ideal single-entry point for those potential new hackers.  
 
 # Motivation
 [motivation]: #motivation
 
-These enhancements are intended to simplify and enhance developer and user productivity by providing a single command-line tool which exercises all major features of NEAR.  As each major piece of NEAR functionality is added to Near Shell, we should see utilization of each of those features increase.  Additionally, community mindshare and understanding of NEAR should increase as NEAR Shell will contain detailed help documentation for each major feature with clear examples of use, just like Linux man pages and/or other major CLI tools which have integrated help for each command.
-
+These enhancements are intended to simplify and enhance developer and user productivity by providing a single command-line tool which exercises all major features of NEAR. As each major piece of NEAR functionality is added to `near-shell`, we should see utilization of each of those features increase.  Additionally, community mindshare and understanding of NEAR should increase as `near-shell` will contain detailed help documentation for each major feature with clear examples of use, just like Linux man pages and/or other major CLI tools which have integrated help for each command.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
-
-## Configuration and key management
-
-Configuration and key management are two types of stored settings necessary for developers to build and deploy projects. The options for key storage and the way configuration is loaded will benefit with customization.
-
-Definitions:
-- Configuration - a key-value storage (typically JSON) containing information.
-    - `networkId` - similar to an environment to work on (ex: 'staging')
-    - `nodeUrl` - URL to RPC
-    - `contractName` - NEAR account name for smart contract
-    - `walletUrl` - URL to NEAR Wallet
-    - `helperUrl` - URL to project helping with token dissemination
-    - `masterAccount` - NEAR account used for continuous integration
-    - `keyPath` - for validators, the path to a file containing an account's private key
-- Keys - the storage of an account id and a corresponding private key.
-    - `account_id`
-    - `private_key`
 
 ### Projects (currently)
 
@@ -42,7 +24,7 @@ Currently, NEAR projects (for example, one generated using `create-near-app`) ho
 
 `src/config.js`
 
-**Note**: the active configuration can have keys overridden by flags.
+**Note**: the active configuration can have values overridden by flags.
 
 After the user completes the instructions from running `near login`, the project has a new directory where private key(s) are stored:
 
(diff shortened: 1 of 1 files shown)


amgando:
I like this interface

adding this to the mix
https://github.com/lirantal/nodejs-cli-apps-best-practices

text/0000-near-shell.md
```diff
@@ -0,0 +1,191 @@
...
+
+* `near tokens send`	: Send tokens to another NEAR account.
+* `near tokens status`	: Display tokens in wallet.
+* `near tokens stake`
+* `near tokens help`
+
+### `near key <command>`
+
+### `near contract <command>` 
+* `near contract list`
```
mikedotexe: This would be incredibly useful. Would this information be stored from the user's history? Is there some way we can use a library/RPC to get this info?

mikedotexe:
Yeah, this is great. I think besides adding the validator aspect to `near-shell`, we can also take the approach we intend to add all functionality on the command line that can currently only be done with other, external interactions. For instance, if a person wants to give function-level access to their smart contract, you must log in via Wallet. Or if a person wants to deauthorize (aka revoke) an access key you must use Wallet, as far as I am aware.
These are the kind of examples that could be moved to `near-shell` as well.
I like the idea of laying it out like this:

> While interacting with NEAR as a developer or validator, there is a relatively small number of atomic actions that are needed. Currently, some of these actions are tied to other NEAR offerings, such as Wallet or Explorer. The ideal future of `near-shell` allows a command-line interface for these in a highly secure and configurable way.

mikedotexe:
Still have more work to do on this NEP, committing progressively

text/0000-near-shell.md
bowenwang1996: Not sure what this is supposed to do. Will it just return a list of contract account ids, or contract metadata, or contract code itself?

damons: What is the unique identifier for a contract?  Do contracts have names?  The idea was to list contracts for a given account.  It would be good to have an ID and an contract name.

mikedotexe: The unique identifier is the account name, which is a regular, human-readable NEAR account. Thanks for bringing this up, Bowen. I am considering removing that list of commands.

bowenwang1996: @damons what do you mean by "contracts for a given account"? There can be zero or one contract deployed on a given account. If you are talking about contracts that an account interact with, that is much harder to do because user data are stored under the contract account, not the user account.

text/0000-near-shell.md
```diff
@@ -0,0 +1,313 @@
...
+* `near server tail`    : Tail the log of the NEAR validator server on current host.
+* `near server help`    : Display help on server subcommands.
+* `near server stake`   : Stake tokens on the configured network with the active or specified account
+
+### `near contract <command>` 
+* `near contract list`
+* `near contract status`
+* `near contract add`
+* `near contract remove`
+* `near contract build`
```
potatodepaulo: Just highlighting a discussion here: https://github.com/nearprotocol/near-shell/pull/275

Since the build process for AssemblyScript and Rust differs, it may make sense to either:

1) Remove this and lean on the tools for building in the language ecosystems
2) In implementation, the build process has a step where the language is determined.

@vgrichina makes good points for (1) here: https://github.com/nearprotocol/near-shell/pull/275#issuecomment-595419971

vgrichina: There is another important aspect to this. 
We don't currently have support for multi-contract projects.

And I don't think we can easily provide anything like `near build` for these, as they can easily be mix of Rust / AssemblyScript / whatever else contracts which all need to be deployed together.

We need to think of some build system supporting dependencies (i.e. something like `make` or `gulp`) to put this all together and provide ability for developer to customize individual steps as necessary.

mikedotexe: > We need to think of some build system supporting dependencies (i.e. something like make or gulp) to put this all together
`near build` currently runs gulp. if the suggestion is to remove `near build` would we be adding this back in?
https://github.com/nearprotocol/near-shell/blob/7d9e8601b0eeb05872406e8bf5eca3e14098224c/bin/near-cli.js#L118-L122

vgrichina: I think we should:
1) remove `near build`
2) have design for multi contract projects if we want to move away from gulp

ilblackdragon:
Few comments:
* We should use different folder from ~/.near, for example ~/.neardev
* ~/.neardev/config.json should contain list of network configurations
* We should deprecate and remove --nodeUrl and other overriders for networks - instead people should just add alternative networks. Right now it's confusing because there are multiple ways to do the same thing.

mikedotexe:
@ilblackdragon when you say:
>Make sure to NEVER to override keys.
I want to confirm that having environment variables containing private keys is not considered overriding, correct? If I'm spinning up a Docker instance on the cloud, it would be great to have the option to include these are env vars, fallback to key files in the project (cwd), fallback to key files in the home directory…
Are you wanting to avoid this?

mikedotexe:
ready for re-review, @ilblackdragon 
https://github.com/nearprotocol/NEPs/blob/near-shell-enhancements/text/0000-near-shell.md

text/0000-near-shell.md
mikedotexe: Removed `near build` from the spec
Haven't added a recommended method for multi-contract projects, as I think we haven't discussed the alternatives yet. Honestly, I don't have a good gut instinct for that task and would like to hear from others.

text/0000-near-shell.md
```diff
@@ -0,0 +1,449 @@
...
+### Definitions:
+
+#### 1. Settings
+[shell-experience-settings]: #shell-experience-settings
+
+---
+
+These are key-values reflecting how `near-shell` behaves when called within a NEAR project.
+    
+Example: a developer using OS X uses `near login` with access to a browser, whereas a validator runs `near login` on a CentOS box with no UI or browser. A user prompt may ask "Is this login from a computer with a browser?" The answer is then stored in key-value format in the project-level directory so it can be referenced later, skipping the prompt in the future.
```
ilblackdragon, line 48: this is global configuration, not per project

text/0000-near-shell.md
```diff
@@ -0,0 +1,449 @@
...
+
+Besides answers to user prompts, shell-experience settings also store information about the last version of `near-shell` used in this project. For more information, please see [Upgradability](#upgradability).
```
ilblackdragon, line 50: This seems like work for `package.json` file not for a `.<something>` folder. Usually `.` folders are added to gitignore as they contain local configuration.

https://stackoverflow.com/questions/10065564/add-custom-metadata-or-config-to-package-json-is-it-valid

text/0000-near-shell.md
```diff
@@ -0,0 +1,449 @@
...
+
+As shown in the directory structure above, this file is located in the project directory at:
+
+`.near-config/settings.js`
+
+In summary, "settings" are key-value pairs that are set by `near-shell` and relate to the behavior of how it operates in a given project directory. They do not relate to the development or deployment of smart contracts, except for providing default values like `accountId`.
```
ilblackdragon, line 56: default app name can also go to `package.json`

text/0000-near-shell.md
```diff
@@ -0,0 +1,449 @@
...
+
+This NEP will jump straight into the concepts necessary to understand the components needed in `near-shell`, and generally an extensible CLI tool. Unlike other NEPs, the guide-level and reference-level explanations are combined. There is nothing overly technical in the following sections, but more conceptual in nature. With that, the discussion begins by separating the three types of stored information needed for a successful, long-term `near-shell`.
+
+## Settings, config, and key management
+
+Shell-experience **settings**, connection **configuration**, and **key management** are the three types of stored data necessary for developers to reliably build and deploy projects.
+ 
+```
+─ awesome-near-project  ⟵ NEAR dApp
+ ├── .near-config       ⟵ Stores shell-experience settings and connection configuration
```
ilblackdragon, line 29: I would really prefer not to have any settings folders inside the project folder.

text/0000-near-shell.md
```diff
@@ -0,0 +1,449 @@
...
+ │  ├── connections     ⟵ Contains files used to connect and deploy a contract (except keys)
+ │  │  ├── default.js   ⟵ Default configuration is for development
+ │  │  └── localnet.js  ⟵ Custom connection added by user for localnet
+ │  └── settings.js     ⟵ Stores near-shell settings, how shell behaves when run in this project
+ └── .near-credentials  ⟵ Previously "neardev" this directory contains private key inforamtion
```
ilblackdragon, line 34: credentials for sure should not be here. this is both easiest way to delete them (delete & reclone the project) or expose them because forgot to add the folder into .gitignore.

text/0000-near-shell.md
```diff
@@ -0,0 +1,449 @@
...
+This NEP will jump straight into the concepts necessary to understand the components needed in `near-shell`, and generally an extensible CLI tool. Unlike other NEPs, the guide-level and reference-level explanations are combined. There is nothing overly technical in the following sections, but more conceptual in nature. With that, the discussion begins by separating the three types of stored information needed for a successful, long-term `near-shell`.
+
+## Settings, config, and key management
+
+Shell-experience **settings**, connection **configuration**, and **key management** are the three types of stored data necessary for developers to reliably build and deploy projects.
+ 
+```
+─ awesome-near-project  ⟵ NEAR dApp
+ ├── .near-config       ⟵ Stores shell-experience settings and connection configuration
+ │  ├── connections     ⟵ Contains files used to connect and deploy a contract (except keys)
```
ilblackdragon, line 30: connections should be global for the computer. In cases where developers really need their non-local and non-standard environment, we can be `near-provider.js` or something in project folder to extend and be similar to truffle?

text/0000-near-shell.md
```diff
@@ -0,0 +1,449 @@
...
+* `networkId`       : Similar to an environment to work on (ex: 'staging')
+* `nodeUrl`         : URL to the NEAR node
+* `contractName`    : NEAR account name for the smart contract
+* `walletUrl`       : URL to NEAR Wallet
+* `helperUrl`       : URL to the contract helper that provides tokens upon account creation
+* `masterAccount`   : When creating "child" accounts, `masterAccount` is the "parent" and source of initial Ⓝ balance
+
+As shown in the directory structure earlier, this file is located in the project directory at:
+
+`.near-config/connections/default.js`
```
ilblackdragon, line 95: let's stop use default, as it's confusing. we have testnet and mainnet

text/0000-near-shell.md
```diff
@@ -0,0 +1,449 @@
...
+* `type`        : Options include 
+    - `unencrypted`     : An unencrypted file containing an account's private key
+    - `native_osx`      : Private key stored with OS X key management 
+    - `native_linux`    : Private key stored with Linux-based system's key management 
+    - `native_windows`  : Private key stored with Windows key management
+* `account_id`  : The NEAR account name
+* `network`     : The NEAR network (example, "betanet", "testnet", etc.)
+* `private_key` : The plain-text private key, used when `type = "unencrypted"` 
+
+`near-shell` will look for keys in a specific order. This list is in the prioritized order and can be understood to mean, "if the key is not found here, then try the next location/store."
```
janedegtiareva, line 130: What do you think about not having a complex order, but instead a simple default + a way to set this in config

ilblackdragon:
@mikedotexe I also wanted to mention that there are a bunch of cases, when user wouldn't want to sign transaction on the same device.

People would use custody, CloudHSM, HSMs that our shell doesn't support, etc.

E.g., near shell should be able to create transactions without signing or sending them.
And also optionally sending signed transaction (as input prob combination of data from above and signature).

We should make this at least possible and as secure as we can.

potatodepaulo:
Just wanted to x-link this request from a while back: https://github.com/near/near-shell/issues/57

text/0000-near-shell.md
Sladuca: How about something like [lerna](https://lerna.js.org/)? We could probably abstract away all of the lerna operations with `near` commands.

vgrichina: @Sladuca Lerna solves somewhat orthogonal problem, I'm talking more about stuff how you can have one project with multiple deployments (e.g. given this is shared system you may want to deploy single contract on multiple accounts). 

Other than that I think it should be possible to use Lerna with near-shell as an option, no need to wrap around with near-shell commands. We want to provide dev tools that fit into different workflows (i.e. not tied to specific build tools ideally) + provide reasonable default workflows in examples.

text/0000-near-shell.md
vgrichina: Do we even need complex configurable order? Seems like it always can be:

project folder -> global folder -> protected OS key store 

External stuff like `Ledger` likely makes sense to specify explicitly (as it's generally not going to be based on account ID).

text/0000-near-shell.md
```diff
@@ -0,0 +1,452 @@
...
+
+Configuration are key-value pairs that relate to development and deployment of smart contracts and refer to *connection* information.
+
+Example: As a user, I want to deploy a contract to my localnet instead of testnet during development. I will provide flags and/or modify configuration so that `near-shell` can determine where to connect, for what contract, and on behalf of which NEAR account.
+
+Contains:
+
+* `networkId`       : Similar to an environment to work on (ex: 'staging')
+* `nodeUrl`         : URL to the NEAR node
+* `contractName`    : NEAR account name for the smart contract
```
chadoh, line 88: I doubt this should be here. It's commonly included in `src/config.js` today, but that doesn't seem right to me. 

Let's consider your example command below for calling a contract:

    near contract view near-game topPlayers --env localnet

What would having `contractName` in my config file give me? It doesn't make sense to call a function on a contract without specifying which contract. I doubt any sort of "default contract" makes sense.

frol:
I want to dump my vision on near-cli and tooling design in general. (moving it from a private discussion to this public NEP discussion)

**TL;DR: A well-designed tool should drive its own usage instead of requiring the user to learn about everything before they can even start using it.**

I want to make the number of available options on each step to be as minimal as possible (as opposed to 18 top-level subcommands in near-cli JS), and I don't want to hide flows behind the scenes (`near login` is 100% confusing to users at first, even though the idea behind it is quite straightforward).

For example, I consider having the following command to do a transfer:

```
$ near-cli \
    construct-transaction \
    online \
        --network testnet \
    send-from frol4.testnet \
    send-to qq.frol4.testnet \
    transfer 3.14NEAR \
    sign-with-secret-key ed25519:... \
    submit
```

it is quite a lengthy command, but a user is going to be guided through it as follows:

1. Let's see what we can do:

```
$ near-cli
```

```
...
SUBCOMMANDS

construct-transaction
    Initiate transaction creation with follow up actions to optionally sign and submit the transactions to the network

utils
   Low-level helpers for common operations (key pair generation, transaction verification, signing, etc)
```

2. Ok, let's create a transaction:

```
$ near-cli construct-transaction
```

I initially considered printing a help message here listing all the required options, but we can have it for `--help`, and instead opt into an interactive mode here:

```
To construct a transaction you will need to provide information about sender (signer) and receiver accounts, and actions that needs to be performed.

Do you want to derive some information required for transaction construction automatically querying it online?
> Yes, I keep it simple
  No, I want to work in no-network (air-gapped) environment

Select NEAR protocol RPC server:
> testnet (https://rpc.testnet.near.org)
  mainnet (https://rpc.mainnet.near.org)
  betanet (https://rpc.betanet.near.org)
  Custom (will prompt you to provide RPC URL)

What is the account ID of the sender? ... frol4.testnet

What is the account ID of the receiver? ... qq.frol4.testnet

Select an action that you want to add to the transaction:
> Transfer NEAR Tokens
  Call a Function
  Stake NEAR Tokens
  Create an Account
  Delete an Account
  Add an Access Key
  Detete an Access Key
  [Skip adding a new action]

How many NEAR Tokens do you want to transfer? ... 3.14 NEAR

Select an action that you want to add to the transaction:
...
> [Skip adding a new action]

Would you like to sign the transaction? ...
> Yes, I want to sign the transaction with my private key
  No, I want to construct the transaction and sign it somewhere else

Select your key chain: // <-- if "Yes", otherwise, skip this step
...

Some extra information needs to be filled in before we proceed signing...

The account ID has 2 access keys, select one that you will use to sign the transaction:  // <-- Only in online mode with "No" reply to signing since we don't know the access key; in offline mode, we should prompt a user to enter their public access key manually
> ed25519:11111...
  ed25519:22222...

Constructing the Transaction...
Transaction {
...
}
Base64-encoded Transaction: ...(base64 value)...
Transaction Hash (sha256): 83ffe880d749ea081c12cb7aa96b481bea3ee0c30e13650d363c2a28edfc0971

Constructing the Signed Transaction...
SignedTransaction {
...
}

Confirm submitting the transaction:
> Everything is correct, send it
  Oops! Stop it!
```

<details><summary>Offline flow example</summary>
<p>

```
To construct a transaction you will need to provide information about sender (signer) and receiver accounts, and actions that needs to be performed.

Do you want to derive some information required for transaction construction automatically querying it online?
 Yes, I keep it simple
>  No, I want to work in no-network (air-gapped) environment

What is the account ID of the sender? ... frol4.testnet

What is the account ID of the receiver? ... qq.frol4.testnet

Select an action that you want to add to the transaction:
> Transfer NEAR Tokens
  Call a Function
  Stake NEAR Tokens
  Create an Account
  Delete an Account
  Add an Access Key
  Detete an Access Key
  [Skip adding a new action]

How many NEAR Tokens do you want to transfer? ... 3.14 NEAR

Select an action that you want to add to the transaction:
...
> [Skip adding a new action]

Would you like to sign the transaction? ...
  Yes, I want to sign the transaction with my private key
> No, I want to construct the transaction and sign it somewhere else

Some extra information needs to be filled in before we proceed signing...

Enter transaction nonce (query the access key information with `near-cli utils view-access-key frol4.testnet ed25519:...` incremented by 1):

Enter recent block hash (see above or `near-cli inspect-data blocks latest`):

Constructing the Transaction...
Transaction {
...
}
Base64-encoded Transaction: ...(base64 value)...
Transaction Hash (sha256): 83ffe880d749ea081c12cb7aa96b481bea3ee0c30e13650d363c2a28edfc0971

If you want to sign the transaction using `near-cli` on another device, run: `near-cli utils sign-transaction ...(base64 value)...`.

Once you sign the Transaction Hash, we can proceed.
Enter the signature (base58-encoded string):
Signature is valid.

Constructing the Signed Transaction...
SignedTransaction {
...
}

Here is the Signed Transaction ready for submittion through RPC: ...(base64 value)...
```

</p>
</details>

<details><summary>Merged flows into a single scenario (just to understand the full picture)</summary>
<p>

```
To construct a transaction you will need to provide information about sender (signer) and receiver accounts, and actions that needs to be performed.

Do you want to derive some information required for transaction construction automatically querying it online?
> Yes, I keep it simple
  No, I want to work in no-network (air-gapped) environment

Select NEAR protocol RPC server: // <-- Only in online mode
> testnet (https://rpc.testnet.near.org)
  mainnet (https://rpc.mainnet.near.org)
  betanet (https://rpc.betanet.near.org)
  Custom (will prompt you to provide RPC URL)

What is the account ID of the sender? ... frol4.testnet

What is the account ID of the receiver? ... qq.frol4.testnet

Select an action that you want to add to the transaction:
> Transfer NEAR Tokens
  Call a Function
  Stake NEAR Tokens
  Create an Account
  Delete an Account
  Add an Access Key
  Detete an Access Key
  [Skip adding a new action]

How many NEAR Tokens do you want to transfer? ... 3.14 NEAR

Select an action that you want to add to the transaction:
...
> [Skip adding a new action]

Would you like to sign the transaction? ...
  Yes, I want to sign the transaction with my private key
  No, I want to construct the transaction and sign it somewhere else

Select your key chain: // <-- if "Yes", otherwise, skip this step
...

Some extra information needs to be filled in before we proceed signing...

The account ID has 2 access keys, select one that you will use to sign the transaction:  // <-- Only in online mode with "No" reply to signing since we don't know the access key; in offline mode, we should prompt user to enter their public access key manually
> ed25519:11111...
  ed25519:22222...

Enter transaction nonce (query the access key information with `near-cli utils view-access-key frol4.testnet ed25519:...` incremented by 1): // <-- Only for offline mode

Enter recent block hash (see above or `near-cli inspect-data blocks latest`): // <-- Only in offline mode

Constructing the Transaction...
Transaction {
...
}
Base64-encoded Transaction: ...(base64 value)...
Transaction Hash (sha256): 83ffe880d749ea081c12cb7aa96b481bea3ee0c30e13650d363c2a28edfc0971

NOTE: if you want to sign the transaction using `near-cli` on another device, run: `near-cli utils sign-transaction ...(base64 value)...`. // <-- Only in offline mode

Once you sign the Transaction Hash, we can proceed. // <-- Only in offline mode
Enter the signature (base58-encoded string): // <-- Only in offline mode
Signature is valid. // <-- Only in offline mode

Constructing the Signed Transaction...
SignedTransaction {
...
}

Here is the Signed Transaction ready for submittion through RPC: ...(base64 value)... // <-- Only in offline mode

Confirm submitting the transaction: // <-- Only in online mode
> Everything is correct, send it
  Oops! Stop it!
```

</p>
</details>

The flow is lengthy with various branches for online/offline and early-exit scenarios, but it guides the users through quite confidently.

Here is a demo of the CLI promts from [enquirer crate](https://github.com/termapps/enquirer):



Based on some feedback I want to clarify that the non-interactive mode will co-exist and it will just have the values prepopulated into the interactive flow, and those will not be prompted for.

There are basically four scenarios I see based on the **online** or **offline** mode of fetching nonce and recent block hash, and **automatic** or **manual** signing process:

1. [simplest, automatic signing in online mode] Online mode with keys availble on the machine or a device attached to it [e.g. Ledger]:

```
$ near-cli \
    construct-transaction \
    online \
        --network testnet \
    send-from frol4.testnet \
    send-to qq.frol4.testnet \
    transfer 3.14NEAR \
    sign-with-sender-secret-key ed25519:... \
    submit
```
2. [secured with air-gapped device for signing] Online mode with manual signing:

```
$ near-cli \
    construct-transaction \
    online \
        --network testnet \
    send-from frol4.testnet \
    send-to qq.frol4.testnet \
    transfer 3.14NEAR \
    sign-manually \
    submit
```
3. [secured air-gapped device constructs the transaction completely] Offline mode, signing with provided keys (either via args, or in a local key store):

```
$ near-cli \
    construct-transaction \
    offline \
        --recent-block-hash Etd1FevM62PPi57YXF4T6r56cPPPCcmF3onJUhcuF3M3 \
        --nonce 123
    send-from frol4.testnet \
    send-to qq.frol4.testnet \
    transfer 3.14NEAR \
    sign-with-sender-secret-key ed25519:...
    display
```
4. [paranoid air-gapped device constructs the transaction, but signing is performed somewhere else] Offline mode with manual signing:

```
$ near-cli \
    construct-transaction \
    offline \
        --recent-block-hash Etd1FevM62PPi57YXF4T6r56cPPPCcmF3onJUhcuF3M3 \
        --nonce 123
    send-from frol4.testnet \
    send-to qq.frol4.testnet \
    transfer 3.14NEAR \
    sign-manually
    display
```

Extra examples:

Ledger support:

```
$ near-cli \
    construct-transaction \
    online \
        --network testnet \
    send-from frol4.testnet \
    send-to qq.frol4.testnet \
    transfer 3.14NEAR \
    sign-with-ledger <path> \
    submit
```

Having all of those as nested subcommands we can have help message on every single level with only the relevant options to choose from. For example:

```
$ near-cli construct-transaction online --help
`online` subcommand is a part of transaction construction which defines that the extra information required for transaction construction will get fetched automatically and directly from RPC node.

ARGUMENTS:

--network  [testnet | mainnet | betanet] predefined set of networks with hardcoded rpc-server links; use `--rpc-server` if you want to use a custom NEAR JSON RPC server
--rpc-server HTTP URL to NEAR node serving JSON RPC providing access to access keys and transaction submission

SUBCOMMANDS:

send-from  specify the account ID of the transaction sender
```

The next stop would be:

```
$ near-cli construct-transaction online send-from --help
...

send-to    specify the account ID of the transaction receiver
```

Notice how you progressed from `online` help to `send-from` help guided with the only available option to choose (`send-from`)

/cc @khorolets @vgrichina @nearmax @ilblackdragon

frol:
FYI, I want to share that there has been great progress on my proposed design of the new near-cli. The project lacks the README and a proper show-case at the moment and lacks a few features here and there, but overall, it is on track, and it can be ready for release in April!

image

you can also skip a few steps providing more arguments:

image

The project is implemented from scratch here: https://github.com/FroVolod/near-cli

jimmy3dita:
Do we still retain the possibility to use in-line command parameters instead of this interactive interface? This is an example of the many creations we have out there: https://gist.github.com/gaia/cff45baf3fa710a42c3fc4cdaafe8edc

frol:
@jimmy3dita Look at the second screenshot, it is there :smile:

chadoh:
This looks awesome. Thanks!

I do wonder if some of the subcommands given to `construct-transaction` should be flags instead. `online` especially seems like sort of parenthetical information, not the main text of the command. Seeing the word `skip` in the non-interactive version also strikes me as strange.

Maybe something like:

```bash
near-cli \
  construct-transaction \
    from frol4.testnet \
    to illia.testnet \
    --online \ # could have alias `--network` and a `--no-network` flag to accompany it
    --rpc testnet \ # RPC endpoint seems like a separate concern from `--online`
    --action transfer-near-tokens 1NEAR \ # allow adding multiple `--action`s
    sign-private-key \ # do we need this? maybe just use flags below
      --signer-public-key ed25519:3oN25HE9tf9YUGXQyXCFB2rEtpomijG8ghVV2JvrA4zJ \
      --signer-secret-key ed25519:...

**Comment by frol** [comment]
@chadoh I [initially started](https://github.com/frol/near-cli) with an idea that looks somewhat similar to yours:

image

I ended up with tons of unstructured arguments, that are hard to follow for a newbie, and it was not even feature-complete (there is no online/offline mode, network selection, and signing options). In my new proposal, I decided to try a guided step-by-step process with exhaustive parameters matching, for example, when you select `online` mode, you are presented with a finite number of options to select: `betanet`, `testnet`, `mainnet`, or `custom`, where `custom` has a required option `--url`, and other options don't have any mention of the `--url` option:

image

So I defined the constraints that I wanted to satisfy:

1. The command line arguments should not overwhelm users when they add `--help` to the end. Ideally, help should guide users through the flow (one step at a time)
2. The questionary and the CLI arguments order should be in sync, otherwise, it is confusing
3. In questionary mode we want to be able to validate user input and immediately ask to type valid values, e.g. to validate the sender and the receiver account ids we need to confirm online/offline status first and know the RPC endpoints to communicate with, that is why in my CLI design online/offline status is so early on the list

The first point is my main pain-point with the current near-cli (in JS):

image

I find this approach much more friendly:

image

(surely help messages needs to be added, but you have a clear idea which options you need to specify at a given stage)

**Comment by frol** [comment]
`construct-transaction` is mostly done, though there are still some papercuts. Some utilities are already implemented as well: generate a keypair, sing a transaction with a secret key in plain text (useful for teaching low-level concepts), combine transaction with a signature (also useful for teaching the low-level concepts).

The next step would be to design a bunch of shortcuts, and I plan to group them into five categories: transfer, execute, add, delete, and view. All of them should have feature-complete CLI and interactive modes.

I want to invite everyone to review the proposed flows.

## Transfer NEAR

```sh
near-cli transfer NEAR \
    network testnet \
    sender 'frol.testnet' \
    receiver 'illia.testnet' \
    amount '1 NEAR' \
    sign-with-keychain
```

Note: I use `transfer NEAR` since I think we may want to support other tokens in the future.

## Execute a change method (function)

Note: the reasoning for the argument order is similar to the `construct-transaction` design as it should have the same flexibility in online/air-gapped transaction construction, and various ways of signing it.

```sh
near-cli execute change-method \
    network testnet \
    caller 'frol.testnet' \
    contract 'meta.pool.testnet' \
    call 'distribute_staking' '{}' \
        --attached-deposit '0 NEAR' \
        --prepaid-gas '1 Tgas' \
    sign-with-keychain
```

Note: we may consider the following options to the call arguments syntax:

* `call 'distribute_staking' '{"account_id": "testnet"}'` (as in the example)
* `call 'distribute_staking({"account_id": "testnet"})'`
* `call 'distribute_staking(account_id="testnet")'`

## Execute a view method (read-only function)

Note: to do a view method call, we don't need to create a transaction, we don't need a signer, we don't need keys etc. It is just a [JSON RPC call to `query` endpoint with `call_function` type](https://docs.near.org/docs/develop/front-end/rpc#call-a-contract-function)

```sh
near-cli execute view-method \
    network testnet \
    contract 'meta.pool.testnet' \
    call 'distribute_staking' '{}'
```

## Add Access Key (and "login")

```sh
near-cli add access-key \
    network testnet \
    public-key 'ed25519:...' \
    account 'frol.testnet' \
    grant-full-access \
    sign-with-keychain
```

```sh
near-cli add access-key \
    network testnet \
    generate-keypair \
    account 'frol.testnet' \
    grant-function-call-access \
        --receiver-id 'meta.pool.testnet' \
        --allowance '10 NEAR' \
        --method-names 'set_a, set_b' \
    sign-with-keychain
```

Here is how `near-cli login` could look like:

```sh
near-cli add access-key generate-keypair grant-by-near-wallet testnet
```

One could argue that `login` is much shorter, but I argue that we would spend less time explaining to people over and over again how login actually works.

## Delete Access Key

```sh
near-cli delete access-key \
    network testnet \
    public-key 'ed25519:...' \
    account 'frol.testnet' \
    sign-with-keychain
```

## Stake

```sh
near-cli add stake-proposal \
    network testnet \
    validator 'staking.poolv1.testnet' \
    transactions-signing-public-key 'ed25519:...' \
    amount '10_000_000 NEAR' \
    sign-with-keychain
```

## Create New Account

TBD: we probably need to consider the use-cases since it is going to be confusing to have `near-cli add account`. I think having `near-cli add sub-account` and `near-cli add implicit-account` could be more useful.

## Delete Account itself

```sh
near-cli delete account \
    network testnet \
    account 'frol.testnet' \
    beneficiary 'illia.testnet' \
    sign-with-keychain
```

## Deploy Contract Code

```sh
near-cli add contract-code \
    network testnet \
    account 'frol.testnet' \
    contract-file './frol.testnet.wasm' \
    sign-with-keychain
```

When you need to initialize the contract:

```sh
near-cli add contract-code \
    network testnet \
    account 'frol.testnet' \
    contract-file './frol.testnet.wasm' \
    initialize 'new' '{}' \
        --attached-deposit '0 NEAR' \
        --prepaid-gas '1 TGas' \
    sign-with-keychain
```

## View Account Details

Displays the non-staked balance, staked balance, storage usage, whether a contract is deployed to it, and lists access keys:

```sh
near-cli view account 'frol.testnet'
```

## Download Contract Code

```sh
near-cli view contract-code 'meta.pool.testnet' download
```

NOTE: by default, it will save the code to the`./account-id.wasm` file.

## View Contract State

```sh
nenar-cli view contract-state 'meta.pool.testnet'
```

## View Transaction Status

```sh
near-cli view transaction by 'frol.testnet' 'G5as3gqdivCPhYbk2et3YRU9Gc14LAWFvzgTgjXYWqmA'
```

Note: due to sharding we need to specify the signer account id when lookup the transaction status

frol:
https://near.cli.rs is now in a pretty usable state (it does everything that was laid out above, and has NEAR Wallet login and Ledger support)

There is also the next iteration of a discussion on where to move forward with NEAR CLI: https://gov.near.org/t/new-near-cli-design-proposal/3616

Issue #29, opened 2020-02-09
NEAR Spec and standards


ilblackdragon:
Want to start discussion for where and how we are going to maintain NEAR Spec and standards.

The NEPs process is mostly around proposal to modify / add things, but we also need the final version that lists the overall specification for the protocol and also all the accepted / used standards.

My proposal, we actually rework this repository instead of been a list of proposals into maintaining the released version of Spec + all accepted standards.

E.g. if would be a combination of https://github.com/ethereum/eth2.0-specs and https://github.com/ethereum/EIPs. 

Where as NEP gets accepted it modifies the staging version of the spec and gets released with the new version. Alternatively if NEP represents new standard and doesn't require changes to protocol - it goes to the standards list.

Why standards mixed with spec?
Mostly because standards end up representing a big part of the protocol itself.
If you think of ERC-20 specification - it actually is a core part of the Ethereum protocol, even though doesn't live on blockchain level itself. Same thing can go for any number of other protocols.

I also think we can repurprose https://nomicon.io/ to serve final version of spec. That website already maintains mostly the spec details with few implementation details that we currently have in Rust client. We can split Rust specifics into the docs inside Rust repo (which currently is not pointing to nomicon.io anyway).

nearmax:
I suggest we make a separate repo for specs. NEPs is a great place to discuss all sorts of major ideas, like large rewrites, performance improvements. I would even use NEPs for discussions around the processes!

+1 For moving specs and standards into https://nomicon.io/ but let's not limit it to specs and standards only. Nomicon is a great place to guide through the details of the internals, which makes it easier for external people to contribute. If we foresee several versions of the implementation then let's give our implementation a code name, e.g. `linear` (other names for our client: `bilinear`, `collinear`, `nonlinear`, `overnear`) and have a section that talks specifically about its implementation.

fckt:
Would be nice to have a clear discriminator. 
I'm for having specification and proposals which describe/alter consensus in a separate place from other documentation.
If it's a ERC-20-like standard or a different serialization/deserialization format for contracts, or an addition of some RPC method: since it's a not altering consensus - better to keep it in a different category.

ilblackdragon:
Here is my reasoning:
 - The more places we have, harder it is to keep track of things. I for example can not keep up already with all the different sources of truth for different parts of the ecosystem.
[...]

PR #32, opened 2020-02-24
[Spec] Adding initial specification & changing NEPs process

ilblackdragon:
# Summary

Addresses #29, by moving NEAR Protocol specification and standards into one repository. Both of them will be served at nomicon.io and linked in the docs.nearprotocol.com

# Motivation

Currently, there are no single place where the up to date specification of the Protocol can be found.
Additionally, as standards evolve and replace each other, it is great to have the up to date version repository of them.

See more details in the #29 for discussion.


nearmax:
I see, it is about current pull requests.

PR #39, opened 2020-03-10
Add StateRecord page

evgenykuzyakov:
- Proposal Name: `add-state-record-spec`
- Start Date: 2020-03-09
- Issue(s): https://github.com/nearprotocol/nearcore/issues/2215

# Summary
[summary]: #summary

Add StateRecord spec page.

Switch from `StateRecord::Data { key: String, value: String }` to `StateRecord::Data { account_id: AccountId, key: String, value: String }`

Also add `DelayedReceipt(Box<ReceiptView>)` for delayed receipts.

# Motivation
[motivation]: #motivation

We plan to hash account keys in the linked issue. For this we need to update Data records to keep `account_id` separately, so it doesn't have to be part of the base64 `key`.

# Reference-level explanation

New `StateRecord::Data` contains key in base64 format, but it doesn't include `account_id,` prefix.


bowenwang1996:
Why does it change so many files? I cannot find where I should be looking...

evgenykuzyakov:
> Why does it change so many files? I cannot find where I should be looking...

It updates nomicon.io

evgenykuzyakov:
@bowenwang1996 Updated `.gitattributes` to ignore `docs`

specs/GenesisConfig/StateRecord.md
```diff
@@ -0,0 +1,131 @@
...
+### account
+
+_type: [AccountView]_
+
+The account structure in JSON-friendly format. See [Account](../DataStructures/Account.md) for details.
+
+
+## Data
+
+_type: Unnamed struct_
```
bowenwang1996: What is `Unnamed struct` here? Why not give the struct information?

evgenykuzyakov: It's Rust concept meaning a struct without a name. All fields of this structure are listed below.

PR #41, opened 2020-03-16
[Spec] Switch to state staking in economics

### LINKED ISSUE #40: State Rent vs State Staking
Moved from https://commonwealth.im/near/proposal/discussion/380-state-rent-vs-state-staking

State is the most important resource of the blockchain network.
This is what all nodes in the network need to maintain and what they all come with consensus.

There are two major ways to charge for storage without undercharging (what happens now on ETH):
 - Charge state rent
 - Require some amount of $NEAR token on the account to occupy a byte.

There are also more complex alternatives, like what EOS is using, where RAM (their name for state) needs to be bought explicitly at the current market price. Price then fluctuates depending on the available RAM. These kinds of mechanics lead to lots of weird market manipulations and disadvantaged developers.

# State Rent

State rent mechanics is when we charge each block some amount of $NEAR from the account for the occupied size. For example if account is `1kb`, and price is `7e-15 $NEAR for block / byte`, we will be deducting `7.168E-12 $NEAR` per block or `~0.00022 $NEAR` per year.

Pros:
 - Developers who need to deploy their contracts and applications, that might be taking ~100kb - 1mb of space will be paying in as-you-go mode.

Cons:
 - Account balance becomes "virtual", because we are not updating it all the time as we charge. Currently for example in NEAR - we never show the real balance still.
 - Accounts must be deleted when rent runs out. This creates both weird user experience and potential vectors where attacker will try to drain user's account in some way, delete their account and re-create account with the same name to get all of their resources. Simplest way is to get user to sign a tx that will transfer _all_ of their assets to some lending application, leaving account without enough to pay for rent and then by taking over the account - receive all their funds back.
 - We have some weird code around validators, to make sure they have enough money to pay for rent.
 - Even though we can burn these fees, at the end it doesn't require to have that much balance on accounts - removing this as a Token Sink.
 - Price is fixed in $NEAR

# State Staking

Alternative model in the simplest form is to require some amount of $NEAR on account's balance to store `1kb` of data.

This means that total supply of $NEAR will map to some amount of storage. For example `1B` of tokens maps to `1TB`. This is roughly `1 $NEAR` for `1kb` of storage.

Specifically, the mechanics will work next time:

```
def process_transaction(account, tx):
   ... do actual work ...
   if sizeOf(account) > account.amount * config.price_of_bytes:
       fail_and_revert(tx)

```
(e.g. [replacing check_rent function](https://github.com/nearprotocol/nearcore/blob/6d07a2d408a58ae1feeefecb815ad7351cdc6086/runtime/runtime/src/actions.rs#L60) with different condition and removing `apply_rent` completely with extra fields from the account)

Pros:
 - Improves user experience around account ownership: doesn't require to deduct balance (if you have 10 $NEAR, you will still have 10 $NEAR in a year), doesn't have account deletion problems due to low balance. Tx that reduce your balance below required to store data - will just fail with meaningful error.
 - Creates very solid Token Sink, both because of useful storage and "excess" of $NEAR.
 - Creates lending and borrowing market for $NEAR for storage purposes.
 - Price can adjust as we add more storage into the system by adding extra shards.

Cons:
 - Making user-centric contracts where we deploy 500kb contracts would require a large enough balance (e.g. 500 $NEAR) and to store any extra data that might require more .
 - Developers now required to have a larger amount of $NEAR to deploy their contracts. There might be more considerations around pricing over time that they must consider.

The last item is actually very interesting, because it creates new opportunity for something similar to Polkadot's "Initial Parachain Offering". 

Let's say developer expects to need 10Mb of storage for their application but they don't really have expected amount of $NEAR.

On one hand, they can borrow it for their contract (this can be implemented in trustless way, to not require collateral on such loan).

On the other hand, developer can sell their third-party token for $NEAR or for loaning $NEAR for some period of time. Benefits of such approach:

 - $NEAR holders have an alternative way to lend or provide their tokens in exchange for new application token which gives a way to participate in the new ecosystem on top of $NEAR and capture value there as it grows.

 - These $NEAR token holders immediately interested in success of such application and become it's promoters. This also removes some of the tensions between third-party tokens and native token-based applications. Like Bancor vs Uniswap, where Uniswap which doesn't have it's own token is preferred by community because it didn't create "another token". There are few models for such third-party tokens to even further reduce this tension.

Balance between validators staking and using $NEAR for storage will be maintained via the fact that as more storage is required - more $NEAR won't be staked but will be allocated for storage. This means that each staked $NEAR will capture bigger portion of inflation.

Additionally, we will need to have a clear program to onboard developers initially. For example grants to developers who built interesting applications on TestNet to provide them with grant funds to launch on MainNet.

*Thoughts?*

ilblackdragon:
Switch spec for economics to do state staking. See details of the proposal and discussion #40.
Fixes #40

### DIFF
--- a/specs/Economics/README.md
+++ b/specs/Economics/README.md
@@ -31,6 +31,7 @@
 | `TREASURY_ACCOUNT_ID` | `treasury` |
 | `TREASURY_PCT` | `0.01` |
 | `TOTAL_SEATS` | `100` |
+| `ACCOUNT_LENGTH_BASELINE_COST` | `2.07 * 10**20` yoctoNEAR |
 
 ## General Variables
 
@@ -69,15 +70,37 @@ Where `gasUsed[t] = sum([sum([gas(tx) for tx in chunk]) for chunk in block[t]])`
 Amount of `NEAR` on the account represents right for this account to take portion of the blockchain's overall global state. Transactions fail if account doesn't have enough balance to cover the storage required for given account.
 
 ```python
+def check_storage_cost(account):
+    accountLengthAmount = 0 if len(account_id) > 10 else ACCOUNT_LENGTH_BASELINE_COST / 3 ** (len(account_id) - 2)
+    # Compute requiredAmount given size of the account and additional cost for shorter names.
+    requiredAmount = sizeOf(account) * storageAmountPerByte + accountLengthAmount
+    return account.amount + account.locked < requiredAmount
+
 # After account touched / changed, we check it still has enough balance to cover it's storage.
 def on_account_change(block_height, account):
-    # Compute requiredAmount given size of the account.
-    requiredAmount = sizeOf(account) * storageAmountPerByte
-    if account.amount + account.locked < requiredAmount:
-        assert "Transaction fails due to not enough balance to cover state stake"
+    # ...
+    # Validate post-condition and revert if it fails.
+    if not check_storage_cost(account):
+        assert LackBalanceForState(signer_id: ctx.signer_id, amount: requiredAmount)
 ```
 
-Where `sizeOf(account)` includes size of `account` structure and size of all the data stored under the account.
+Where `sizeOf(account)` includes size of `account_id`, `account` structure and size of all the data stored under the account.
+
+Account can end up with not enough balance for next two reasons:
+ - Created new account and didn't give it enough funds.
+ - Account got slashed.
+
+This account still can receive transfers, hence can be saved.
+But to prevent grinding with accounts that don't have balances, we allow for anyone to delete accounts that don't have enough balance.
+
+```python
+def action_delete_account(account_id, account, ...):
+    if ctx.signer == account_id or not check_storage_rent(account):
+        # proceed with deleting
+    else:
+        assert DeleteAccountHasEnoughBalance()
+```
+
 
 ## Validators
 
--- a/specs/Economics/README.md
+++ b/specs/Economics/README.md
@@ -17,11 +17,13 @@
 | - | - |
 | `INITIAL_SUPPLY` | `10**33` yoctoNEAR |
 | `NEAR` | `10**24` yoctoNEAR |
+| `INITIAL_GAS_PRICE` | `10**5` yoctoNEAR |
 | `REWARD_PCT_PER_YEAR` | `0.05` |
 | `BLOCK_TIME` | `1` second |
 | `EPOCH_LENGTH` | `43,200` blocks |
 | `EPOCHS_A_YEAR` | `730` epochs |
 | `POKE_THRESHOLD` | `500` blocks |
+| `INITIAL_MAX_STORAGE` | `10 * 2**40` bytes == `10` TB |
 | `STORAGE_AMOUNT_PER_BYTE` | `7E-5` NEAR per byte | 
 | `TREASURY_PCT` | `0.1` |
 | `CONTRACT_PCT` | `0.3` |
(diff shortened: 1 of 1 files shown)


specs/Economics/README.md
```diff
@@ -65,40 +61,23 @@ Gas is priced dynamically in `NEAR` tokens. At each block `t`, we update `gasPri
...
-def can_delete_account(block_height, account):
-    maxFee = sizeOf(account) * storagePrice * pokeThreshold
-    if maxFee < account.amount:
-        assert "Account still has enough funds"
-    # Return to caller the rest of the amonut after charging fees.
-    owedFee = sizeOf(account) * storagePrice * (block_height - account.storagePaidAt)
-    return account.amount - owedFee
+    # Compute requiredAmount given size of the account.
+    requiredAmount = sizeOf(account) * storageAmountPerByte
+    if account.amount + account.lock < requiredAmount:
```
bowenwang1996:
```suggestion
    if account.amount + account.locked < requiredAmount:
```

specs/Economics/README.md
```diff
@@ -22,7 +22,7 @@
 | `EPOCH_LENGTH` | `43,200` blocks |
 | `EPOCHS_A_YEAR` | `730` epochs |
 | `POKE_THRESHOLD` | `500` blocks |
-| `STORAGE_PRICE` | `7E-15` NEAR per byte per block | 
+| `STORAGE_AMOUNT_PER_BYTE` | `7E-5` NEAR per byte |
```
bowenwang1996: How is `7e-5` determined?

specs/Economics/README.md
```diff
@@ -17,25 +17,28 @@
 | - | - |
 | `INITIAL_SUPPLY` | `10**33` yoctoNEAR |
 | `NEAR` | `10**24` yoctoNEAR |
+| `INITIAL_GAS_PRICE` | `10**5` yoctoNEAR |
```
bowenwang1996: how is this determined

specs/Economics/README.md
```diff
@@ -31,6 +31,7 @@
 | `TREASURY_ACCOUNT_ID` | `treasury` |
 | `TREASURY_PCT` | `0.01` |
 | `TOTAL_SEATS` | `100` |
+| `ACCOUNT_LENGTH_BASELINE_COST` | `2.07 * 10**20` yoctoNEAR |
```
bowenwang1996: how is this number determined?

specs/Economics/README.md
```diff
@@ -69,15 +70,37 @@ Where `gasUsed[t] = sum([sum([gas(tx) for tx in chunk]) for chunk in block[t]])`
 Amount of `NEAR` on the account represents right for this account to take portion of the blockchain's overall global state. Transactions fail if account doesn't have enough balance to cover the storage required for given account.
 
 ```python
+def check_storage_cost(account):
+    accountLengthAmount = 0 if len(account_id) > 10 else ACCOUNT_LENGTH_BASELINE_COST / 3 ** (len(account_id) - 2)
```
bowenwang1996: there seems to be a unit mismatch. `ACCOUNT_LENGTH_BASELINE_COST` is in yotcoNEAR while `3 ** (len(account_id) - 2)` computes the cost in NEAR

evgenykuzyakov: Account length should not be part of storage cost, but instead part of the rent.

If we would switch to state staking, then account names have to be rethought as well. Larger stake is one option, but naming has to change.

ilblackdragon: What is wrong with naming? `ACCOUNT_LENGTH_BASELINE_AMOUNT` ?

ilblackdragon: @bowenwang1996  I don't understand your comment. `ACCOUNT_LENGTH_BASELINE_COST` in yoctoNEAR, the resulting `accountLengthAmount` is in yoctoNEAR, and it compares to account balance below in yoctoNEAR. `3 ** (len(account_id) - 2)` is just a rescaling.

specs/Economics/README.md
ilblackdragon: By solving gas_limit * gas_price = 0.1 N

specs/Economics/README.md
bowenwang1996: Nvm I got confused.

ilblackdragon:
Following discussion in #40 we decided to remove extra balance requirements for short TLAs. Instead #45 and following work with target a fair auction with release schedule to create a fair market for TLAs.

PR #45, opened 2020-03-23
[Spec] Top Level Accounts by Registrar

ilblackdragon:
Specify that short top level accounts are allowed only by registrar account. Delaying details of the registrar spec post MainNet.

### DIFF
--- a/specs/DataStructures/Account.md
+++ b/specs/DataStructures/Account.md
@@ -25,12 +25,24 @@ Regex for a full account ID, without checking for length:
 ```
 ### Top Level Accounts
 
-Top level accounts are very valuable as they provide root of trust and discoverability for companies, applications and users.
-To prevent take over by squatters, the top level account names that are shorter than 32 characters they are going to be auctioned off.
+| Name | Value |
+| REGISTRAR_ACCOUNT_ID | `registrar` |
+| MIN_ALLOWED_TOP_LEVEL_ACCOUNT_LENGTH | 32 |
 
-Specifically, only `registrar` account can create new top level accounts that are shorter than 32 characters. `registrar` implements standard Account Naming interface to allow create new accounts.
+Top level account names (TLAs) are very valuable as they provide root of trust and discoverability for companies, applications and users.
+To allow for fair access to them, the top level account names that are shorter than `MIN_ALLOWED_TOP_LEVEL_ACCOUNT_LENGTH` characters going to be auctioned off.
 
-*Note: we are not going to deploy `registrar` auction at launch, instead allow to deploy it by Foundation after initial launch. The details of the auction will be added here in the next spec release post MainNet.*
+Specifically, only `REGISTRAR_ACCOUNT_ID` account can create new top level accounts that are shorter than `MIN_ALLOWED_TOP_LEVEL_ACCOUNT_LENGTH` characters. `REGISTRAR_ACCOUNT_ID` implements standard Account Naming (link TODO) interface to allow create new accounts.
+
+```python
+def action_create_account(predecessor_id, account_id):
+    """Called on CreateAccount action in receipt."""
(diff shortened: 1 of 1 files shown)


evgenykuzyakov:
Do we want to hardcode it in the `nearcore`. It'll break some UI for now, since there is no error indicating you can't create a top-level account ID.

The scope of work is the following:
- Hardcode top-level account ID creation to be only allowed by `registrar`. Can be part of genesis config. Or just hardcoded
- Add an error to indicate that the top-level account can only be created by `registrar` account.
- Update all app-layer to know about the new requirement and the new error.

ilblackdragon:
The app layer work is already in progress ATM - https://github.com/nearprotocol/near-wallet/issues/407

I think config having it in genesis config is fine.
Btw, what's our plan for error kinds spec?

specs/DataStructures/Account.md
```diff
@@ -24,7 +24,15 @@ Regex for a full account ID, without checking for length:
 ^(([a-z\d]+[\-_])*[a-z\d]+\.)*([a-z\d]+[\-_])*[a-z\d]+$
 ```
 
-There is a rent for the account ID length in case it's less than 11 characters long. See [Economics] for the details.
+There is extra requirements for the amount that account has for the account ID length in case it's less than 11 characters long. See [Economics] for the details.
+
+### Top Level Accounts
+
+Top level accounts are very valuable as they provide root of trust and discoverability for companies, applications and users. To prevent take over by squatters, the top level account names that are shorter than 10 characters they are going to be auctioned off.
```
evgenykuzyakov: Let's go with 10 characters or shorted.

specs/DataStructures/Account.md
```diff
@@ -24,7 +24,15 @@ Regex for a full account ID, without checking for length:
...
+
+Specifically, only `registrar` account can create new top level accounts. `registrar` implements standard Account Naming interface to allow create new accounts.
```
evgenykuzyakov: Should any top-level account ID be allowed to create if the length is larger than 10 characters?

ilblackdragon:
Few considerations:
 * for localnet we don’t care about registrar per se, so in theory can omit that
 * what we DO care, is that account ids between networks are different to avoid confusion
Because compared to Ethereum, where at least whoever has address on testnet also can access address on mainnet - in our case this address can be taken by completely someone else
 * this is why we wanted to make all accounts on testnet/devnet/betanet to be prefixed by .testnet/.devnet/.betanet
 * should we maintain the same for .localnet as well?
 * and then, given these considerations we should not have 11 characters limit at all:
      - it’s totally arbitrary and just launch the auction sooner than later with release starting with longer names
       - and with this limit people will register accidentally on localnet / testnet accounts and mess up.

bowenwang1996:
@ilblackdragon as I said here https://github.com/nearprotocol/nearcore/pull/2355#issuecomment-607034684 it seems undesirable to create this kind of restriction on testnet. Also for testnet will there also be auction? If not it seems that certain account names (TLA with less than 11 characters) will be permanently banned.

ilblackdragon:
@bowenwang1996 Why is it undesirable?
The idea is that all networks except MainNet will actually have accounts namespaced under `.{network}`. See comment https://github.com/nearprotocol/NEPs/pull/45#issuecomment-607051828 for more details.

bowenwang1996:
@ilblackdragon if that is the plan I think it's fine.

specs/DataStructures/Account.md
ilblackdragon: I suggest we increase this to 64, to only allow hex(publicKey) in general.
Otherwise, people will create longer account names on the testnet/devnet and will confuse account names with mainnet.

ilblackdragon: If we are doing 2nd price auction for TLA registration going forward, ppl can get longer names for free if no one is contesting.

PR #57, opened 2020-04-02
[Spec] Consensus

SkidanovAlex:


abacabadabacaba:
Why do we include `prev_hash` in skip messages?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,147 @@
...
+    epoch_id: EpochId,
+    last_final_block_hash: BlockHash,
+    approvals: Vec<Approval>
+    ...
+}
+```
+
+Each block belongs to a particular epoch, and has a particular height. The highest height block known to a particular node is its head.
+
+Each epoch has a set of block producers who are assigned to produce blocks in the epoch, and has an assigned block producer for each height in the epoch. The block producers and the assignments in a particular epoch are known from the first block of the *preceding* epoch.
```
ilblackdragon, line 19: link to epoch manager for more details of the assignment? or for high level API needed?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,147 @@
...
+    prev_height: BlockIndex,
+    target_height: BlockHeight,
+    signature: Signature,
+    account_id: AccountId
+}
+```
+
+Where `prev_hash` and `prev_height` are the hash and the height of the block being approved, `target_height` is the specific height at which the approval can be used (an approval with a particular `target_height` can be only included in the `approvals` of a block that has `height = target_height`), `account_id` is the account of the block producer who created the approval, and `signature` is their signature on the tuple `(prev_hash, target_height)` if `target_height == prev_height+1` and the tuple `(prev_height, target_height)` otherwise.
+
+Any valid block must contain approvals from block producers whose cumulative stake exceeds 2/3 of the total stake in the epoch. For a block `B` and its previous block `B'` each approval in `B` must have `prev_height` to be equal to the height of `B'`. If (and only if) `B.height == B'.height + 1`, each approval's `prev_hash` must also be exactly the hash of `B'`. See [this section](#approval-condition) below for details on why the condition must be exactly such.
```
ilblackdragon, line 41: we are ending up with 2/3 liveness requirement at this point?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,147 @@
...
+```python
+def on_block(self, block):
+    header = block.header
+
+    if header.height <= self.head_height:
+        return
+
+    last_final_block = store.get_block(header.last_final_block_hash)
+
+    self.head_height = header.height
```
ilblackdragon, line 61: it would be good to describe state that chain needs to maintain:
 - blocks // last 3 epochs of blocks
 - head {heigh, hash}
 - largest_final_height
 - approvals -- is this all approvals in last 3 epochs?
 - what else?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,147 @@
...
+If a block `B` at some height `h` builds on top of a block `B_prev` that has height `h-1`, and `B_prev` in turn builds on top of a block `B_prev_prev` that has height `h-2` (i.e. if the last three blocks in the chain ending at `B` have sequential heights), and all three blocks are in the same epoch, the block `B_prev_prev` is final, and cannot be reverted unless block producers with more than `2/3` cumulative stake deviate from the protocol. 
+
+TODO: talk about epoch boundaries here
+
+Block producers in the particular epoch exchange many kinds of messages. The two kinds that are relevant to the consensus are **Blocks** and **Approvals**. The approval contains five fields:
+
+```rust
+struct Approval {
+    prev_hash: BlockHash,
+    prev_height: BlockIndex,
```
bowenwang1996, line 32:
```suggestion
    prev_height: BlockHeight,
```

bowenwang1996, line 32: Why do we need `prev_height`? It can be derived.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,147 @@
...
+Any valid block must contain approvals from block producers whose cumulative stake exceeds 2/3 of the total stake in the epoch. For a block `B` and its previous block `B'` each approval in `B` must have `prev_height` to be equal to the height of `B'`. If (and only if) `B.height == B'.height + 1`, each approval's `prev_hash` must also be exactly the hash of `B'`. See [this section](#approval-condition) below for details on why the condition must be exactly such.
+
+On receival of the approval message the participant just stores it in the collection of approval messages.
+
+```python
+def on_approval(self, approval):
+    self.approvals.append(approval)
+```
+
+Whenever a participant receives a block, the operations relevant to the consensus include updating the `head` and initiating a timer to start sending the approvals on the block to the block producerse at the consecutive `target_height`s. The timer delays depend on the height of the last final block, so that information is also persisted.
```
bowenwang1996, line 50:
```suggestion
Whenever a participant receives a block, the operations relevant to the consensus include updating the `head` and initiating a timer to start sending the approvals on the block to the block producers at the consecutive `target_height`s. The timer delays depend on the height of the last final block, so that information is also persisted.
```

bowenwang1996, line 50: Is this paragraph necessary? It seems that it is explaining the code below.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,147 @@
...
+```python
+def get_delay(n):
+    min(MAX_DELAY, MIN_DELAY + DELAY_STEP * (n-2))
+
+def process_timer(self):
+    now = time.time()
+
+    skip_delay = get_delay(self.timer_height - self.largest_final_height)
+
+    if self.endorsement_pending and \
```
bowenwang1996, line 82: Why do you need to break the line here?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,147 @@
...
+                         approval.prev_hash == self.head_hash)]
+```
+
+A block producer assigned for a particular height produces a block at that height whenever they have `get_approvals` return approvals from block producers whose stake collectively exceeds 2/3 of the total stake.
+
+## Safety
+
+We call approvals for which `prev_height + 1 = target_height` endorsements, and all other approvals skips.
+
+Note that with the implementation above a honest block producer can never produce two endorsements with the same `prev_height` (call this condition *conflicting endorsements*), neigher can they produce a skip message `s` and an endorsement `e` such that `s.prev_height > e.prev_height and s.target_height <= e.target_height` (call this condition *conflicting skip and endorsement*).
```
bowenwang1996, line 122:
```suggestion
Note that with the implementation above a honest block producer can never produce two endorsements with the same `prev_height` (call this condition *conflicting endorsements*), neither can they produce a skip message `s` and an endorsement `e` such that `s.prev_height > e.prev_height and s.target_height <= e.target_height` (call this condition *conflicting skip and endorsement*).
```

SkidanovAlex:
@abacabadabacaba yes, `parent_hash` can be technically removed from the skip approvals.

specs/ChainSpec/Consensus.md
SkidanovAlex: Even though not part of the spec, each implementation will need to have some sort of cache for the approvals on blocks that we haven't seen yet. Since the signature on the skip is on (prev_height, target_height), you won't be able to verify the signature for an approval that arrived before the block. I don't think a non-obusable approval cache is possible in such a case.

specs/ChainSpec/Consensus.md
SkidanovAlex: Each implementation can choose what and why to maintain, for as long as they can perform the processes described in the spec. I don't think the implementation of orphan pools / caches / garbage collection belong to the spec.

specs/ChainSpec/Consensus.md
SkidanovAlex: I don't think we have epoch manager spec yet?
As we have other sections written, we can add the cross links.

specs/ChainSpec/Consensus.md
SkidanovAlex: Yes

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,146 @@
...
+    self.timer_started = time.time()
+
+    self.endorsement_pending = True
+```
+
+The timer needs to be checked periodically, and contain the following logic:
+
+```python
+def get_delay(n):
+    min(MAX_DELAY, MIN_DELAY + DELAY_STEP * (n-2))
```
abacabadabacaba, line 75: This doesn't satisfy the conditions on _T_(_x_) in the Doomslug paper, so liveness proof doesn't apply.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,146 @@
...
+    prev_height: BlockHeight,
+    target_height: BlockHeight,
+    signature: Signature,
+    account_id: AccountId
+}
+```
+
+Where `prev_hash` and `prev_height` are the hash and the height of the block being approved, `target_height` is the specific height at which the approval can be used (an approval with a particular `target_height` can be only included in the `approvals` of a block that has `height = target_height`), `account_id` is the account of the block producer who created the approval, and `signature` is their signature on the tuple `(prev_hash, target_height)` if `target_height == prev_height+1` and the tuple `(prev_height, target_height)` otherwise.
+
+Any valid block must contain approvals from block producers whose cumulative stake exceeds 2/3 of the total stake in the epoch. For a block `B` and its previous block `B'` each approval in `B` must have `prev_height` to be equal to the height of `B'`. If (and only if) `B.height == B'.height + 1`, each approval's `prev_hash` must also be exactly the hash of `B'`. See [this section](#approval-condition) below for details on why the condition must be exactly such.
```
abacabadabacaba, line 41: Why is this condition different from the one described in Doomslug paper? In the paper, if `B.height == B'.height + 1`, a witness can contain both endorsements and skips, and if `B.height > B'.height + 1`, it can contain skips of different lengths.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,146 @@
...
+```python
+def get_delay(n):
+    min(MAX_DELAY, MIN_DELAY + DELAY_STEP * (n-2))
+
+def process_timer(self):
+    now = time.time()
+
+    skip_delay = get_delay(self.timer_height - self.largest_final_height)
+
+    if self.endorsement_pending and now > self.timer_started + ENDORSEMENT_DELAY:
```
abacabadabacaba, line 82: Why do we need `ENDORSEMENT_DELAY`?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,146 @@
...
+
+        self.endorsement_pending = False
+
+    if now > self.timer_started + skip_delay:
+        assert not self.endorsement_pending
+
+        self.largest_target_height = max(self.largest_target_height, self.timer_height + 1)
+        self.send_approval(self.timer_height + 1)
+
+        self.timer_started = now
```
abacabadabacaba, line 96: Do we have `ENDORSEMENT_DELAY <= MIN_DELAY`? Otherwise, resetting the timer here would mean that endorsements won't be sent on time.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,146 @@
...
+                         approval.prev_hash == self.head_hash)]
+```
+
+A block producer assigned for a particular height produces a block at that height whenever they have `get_approvals` return approvals from block producers whose stake collectively exceeds 2/3 of the total stake.
+
+## Safety
+
+We call approvals for which `prev_height + 1 = target_height` endorsements, and all other approvals skips.
+
+Note that with the implementation above a honest block producer can never produce two endorsements with the same `prev_height` (call this condition *conflicting endorsements*), neither can they produce a skip message `s` and an endorsement `e` such that `s.prev_height > e.prev_height and s.target_height <= e.target_height` (call this condition *conflicting skip and endorsement*).
```
abacabadabacaba, line 121: Did you mean `s.prev_height <= e.prev_height and s.target_height >= e.target_height`?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,146 @@
...
+
+1. `Bp.height < B0.height`. Then each approval in `Bx` is a skip. For each block producer that included a skip in `Bx` and an endorsement in `B1` these two messages conflict, thus at least 1/3 of stake-weighted block producers deviated from the protocol.
+2. `Bp.height == B0.height and Bx.height == B1.height`. Then each approval in `B1` is an endorsement. For each block producer that included an endorsement in `Bx` and an endorsement in `B1` these two messages conflict, thus at least 1/3 of stake-weighted block producers deviated from the protocol.
+3. `Bp.height == B0.height and Bx.height > B1.height`. Then each approval in `Bx` is a skip. For each block producer that included a skip in `Bx` and an endorsement in `B2` these two messages conflict, thus at least 1/3 of stake-weighted block producers deviated from the protocol.
+
+Thus, for any block with height higher than `B0.height` that doesn't have `B0` in its ancestry to be produced at least 1/3 of stake-weighted block producers must deviate from the protocol.
```
abacabadabacaba, line 131: I think that we should use some different term that "deviate from the protocol". Not every deviation from the protocol is slashable, and it is important that these particular deviations are.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,146 @@
+# Consensus
+
+The fields in the Block header relevant to the consensus process are:
+
+```rust
+struct BlockHeader {
+    ...
+    prev_hash: BlockHash,
+    height: BlockHeight,
+    epoch_id: EpochId,
```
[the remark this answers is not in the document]
bowenwang1996: It is possible to have two epochs with the same height, even though there might be slashing involved.

specs/ChainSpec/Consensus.md
bowenwang1996: To control the block production speed. We do not want blocks to be produced too fast.

specs/ChainSpec/Consensus.md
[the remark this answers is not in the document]
abacabadabacaba: Is it possible to have two epochs with the same height on the same chain?

specs/ChainSpec/Consensus.md
abacabadabacaba: Then I think that it is better to measure the delay not from the time the block was received but from the time the approval (or skip) for the previous height was sent. In this case, the delay is just the desired time between blocks. If we measure the delay from the time the block was received, the time between blocks is the delay plus the time it takes for the approvals from 2/3 of the block producers to travel to the producer responsible for the next block plus the time it takes the block to travel to the 2/3 of the block producers.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,146 @@
...
+    last_final_block_hash: BlockHash,
+    approvals: Vec<Approval>
```
abacabadabacaba, line 12: We don't need full approvals here, only `(account_id, signature)` tuples. Remaining information is already present in the block header.

SkidanovAlex: Replacing with `Vec<Option<Signature>>`

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,152 @@
...
+    inner: ApprovalInner,
+    target_height: BlockHeight,
+    signature: Signature,
+    account_id: AccountId
+}
+```
+
+Where the parameter of the `Endorsement` is the hash of the approved block, the parameter of the `Skip` is the height of the approved block, `target_height` is the specific height at which the approval can be used (an approval with a particular `target_height` can be only included in the `approvals` of a block that has `height = target_height`), `account_id` is the account of the block producer who created the approval, and `signature` is their signature on the tuple `(inner, target_height)`.
+
+Any valid block must contain approvals from block producers whose cumulative stake exceeds 2/3 of the total stake in the epoch. For a block `B` and its previous block `B'` each approval in `B` must be an `Endorsement` with the hash of `B'` if and only if `B.height == B'.height + 1`, otherwise it must be a `Skip` with the height of `B'`. See [this section](#approval-condition) below for details on why the condition must be exactly such.
```
mfornet, line 45: The final sentence looks unfinished.

`... must be exactly such.`

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,152 @@
...
+
+def process_timer(self):
+    now = time.time()
+
+    skip_delay = get_delay(self.timer_height - self.largest_final_height)
+
+    if self.endorsement_pending and now > self.timer_started + ENDORSEMENT_DELAY:
+
+        if self.head_height >= self.largest_target_height:
+            self.larget_target_height = self.head_height + 1
```
mfornet, line 89:
```suggestion
            self.largest_target_height = self.head_height + 1
```

specs/ChainSpec/Consensus.md
```diff
@@ -120,22 +118,35 @@ def get_approvals(self, target_height):
...
 
-We call approvals for which `prev_height + 1 = target_height` endorsements, and all other approvals skips.
+- Any block `B` with height between `h` and `h + epoch_length - 3` (both inclusive) is in the epoch `e_cur` and must have approvals from more than 2/3 of `BP(e_cur)` (stake-weighted).
+- Any block `B` with height `h + epoch_length - 2` or higher for which `LFH(prev(B)) < h + epoch_length - 3` is in the epoch `e_cur` and must have the approvals from both more than 2/3 of `BP(e_cur)` and more than 2/3 of `BP(e_next)` (both stake-weighted).
+- The first block `B` with height `h + epoch_length - 2` or higher and `LFH(prev(B)) >= h + epoch_length - 3` is in the epoch `e_next` and only needs the approvals from more than 2/3 of `BP(e_next)` (stake-weighted).
```
bowenwang1996, line 126: This means that an epoch only has `epoch_length - 2` blocks under normal circumstances, which seems suboptimal.

SkidanovAlex: Incorrect, this makes most epochs have exactly `epoch_length` blocks.

Let's say some block at height `h'` triggers the last condition. The the block at height `h'-1` has a final block with height `h + epoch_length - 3`. Since the final block is 2 heights before the block that observes it, it means that `h'-1-2 = h + epoch_length - 3`, or `h' = h + epoch_length`.

bowenwang1996: 👍

specs/ChainSpec/Consensus.md
```diff
@@ -99,6 +101,8 @@ def process_timer(self):
 
 Where `send_approval` creates an approval with `prev_hash` and `prev_height` equal to `self.head_hash` and `self.head_height` correspondingly, `target_height` to the passed argument, and sends to the block producer at `target_height`.
```
abacabadabacaba, line 102: The fields `prev_height` and `prev_hash` are gone now, so every mention of them should be reworded.

specs/ChainSpec/Consensus.md
```diff
@@ -9,36 +9,38 @@ struct BlockHeader {
...
 
-Any valid block must contain approvals from block producers whose cumulative stake exceeds 2/3 of the total stake in the epoch. For a block `B` and its previous block `B'` each approval in `B` must have `prev_height` to be equal to the height of `B'`. If (and only if) `B.height == B'.height + 1`, each approval's `prev_hash` must also be exactly the hash of `B'`. See [this section](#approval-condition) below for details on why the condition must be exactly such.
+Any valid block must contain approvals from block producers whose cumulative stake exceeds 2/3 of the total stake in the epoch. For a block `B` and its previous block `B'` each approval in `B` must be an `Endorsement` with the hash of `B'` if and only if `B.height == B'.height + 1`, otherwise it must be a `Skip` with the height of `B'`. See [this section](#approval-condition) below for details on why the condition must be exactly such.
```
abacabadabacaba, line 43: There are no more approvals in blocks, just signatures, so this paragraph should be updated.

specs/ChainSpec/Consensus.md
```diff
@@ -114,21 +118,34 @@ def get_approvals(self, target_height):
...
 
-We call approvals for which `prev_height + 1 = target_height` endorsements, and all other approvals skips.
+- Any block `B` with height between `h` and `h + epoch_length - 3` (both inclusive) is in the epoch `e_cur` and must have approvals from more than 2/3 of `BP(e_cur)` (stake-weighted).
+- Any block `B` with height `h + epoch_length - 2` or higher for which `LFH(prev(B)) < h + epoch_length - 3` is in the epoch `e_cur` and must have the approvals from both more than 2/3 of `BP(e_cur)` and more than 2/3 of `BP(e_next)` (both stake-weighted).
+- The first block `B` with height `h + epoch_length - 2` or higher and `LFH(prev(B)) >= h + epoch_length - 3` is in the epoch `e_next` and only needs the approvals from more than 2/3 of `BP(e_next)` (stake-weighted).
```
abacabadabacaba, line 126: I assume that the intent is that under ideal conditions, the first block in `e_next` has height `h + epoch_length`, and it can never be less. Am I right? This implies that `LFH(B)` is defined such that for every `B`, `LFH(B).height <= B.height - 2`, which is absolutely not clear from the definition of `LFH`.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,220 @@
...
+}
+```
+
+Each block belongs to a particular epoch, and has a particular height. The largest height block accepted by a particular node is its head.
+
+Each epoch has a set of block producers who are assigned to produce blocks in the epoch, and has an assigned block producer for each height in the epoch (to whom we refer as *block proposer at `h`*). The block producers and the assignments in a particular epoch are known from the first block of the *preceding* epoch.
+
+Consecutive blocks do not necessarily have sequential heights. A block at height `h` can have as its previous block a block with height `h-1` or lower.
+
+If a block `B` at some height `h` builds on top of a block `B_prev` that has height `h-1`, and `B_prev` in turn builds on top of a block `B_prev_prev` that has height `h-2`, and all three blocks are in the same epoch, the block `B_prev_prev` is final, and cannot be reverted unless block producers with more than `2/3` cumulative stake deviate from the protocol.
```
abacabadabacaba, line 23: I think that with the current epoch switching rule the requirement that all three blocks are in the same epoch is not necessary. Also, replace "more than 2/3" with "more than 1/3".

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Approvals Requirements
+
+A block `B` at height `h` that has some other block `B'` as its previous block must logically contain approvals of a form described in the next paragraph from block producers whose cumulative stake exceeds 2/3 of the total stake in the current epoch, and in specific conditions described in section [epoch switches](#epoch-switches) also the approvals of the same form from block producers whose cumulative stake exceeds 2/3 of the total stake in the next epoch.
+
+If a block to be produced will have height `h` and previous block `B'`, the approvals locically included in it must be an `Endorsement` with the hash of `B'` if and only if `h == B'.height + 1`, otherwise it must be a `Skip` with the height of `B'`. See [this section](#approval-condition) below for details on why the endorsements must contain the hash of the previous block, and skips must contain the height.
```
abacabadabacaba, line 47: typo: locically

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,220 @@
...
+    next_epoch = get_next_block_next_epoch(prev_block)
+
+    account_ids = get_epoch_block_producers_ordered(cur_epoch)
+    if next_block_needs_approvals_from_next_epoch(prev_block):
+        for account_id in get_epoch_block_producers_ordered(next_epoch):
+            if account_id not in account_ids:
+                account_ids.append(account_id)
+
+    return account_ids
+```
```
abacabadabacaba, line 63: So, we take a sorted set of approvals from the current epoch, and then append the sorted set of approvals from the next epoch excluding those that are also present in the current epoch. Why not instead sort the union of both sets?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,220 @@
...
+
+        self.largest_target_height = max(self.largest_target_height, self.timer_height + 1)
+        self.send_approval(self.timer_height + 1)
+
+        self.timer_started = now
+        self.timer_height += 1
+
+def send_approval(self, target_height):
+    if target_height == self.head_height + 1:
+        inner = new Endorsement(self.head_hash)
```
abacabadabacaba, line 127: Given that this is Python syntax, perhaps it is better to omit `new`.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,220 @@
...
+We first define a convenience function to fetch approvals that can be included in a block at particular height:
+
+```python
+def get_approvals(self, target_height):
+    return [approval for approval
+                     in self.approvals
+                     if approval.target_height == target_height and
+                        approval.prev_height == self.head_height and
+                        (approval.prev_height + 1 != target_height or
+                         approval.prev_hash == self.head_hash)]
```
abacabadabacaba, line 149: Rewrite to use `inner`.

specs/ChainSpec/Consensus.md
SkidanovAlex: Because the set that comes from the epoch manager is not "sorted" per se, it is in the ordrer of producing blocks. Not sorting here allows me not to leak any sorting key from the epoch manager into the consensus.

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,219 @@
...
+    epoch_id: EpochId,
+    last_final_block_hash: BlockHash,
+    approvals: Vec<Option<Signature>>
+    ...
+}
+```
+
+Each block belongs to a particular epoch, and has a particular height. The largest height block accepted by a particular node is its head.
+
+Each epoch has a set of block producers who are assigned to produce blocks in the epoch, and has an assigned block producer for each height in the epoch (to whom we refer as *block proposer at `h`*). The block producers and the assignments in a particular epoch are known from the first block of the *preceding* epoch.
```
ailisp, line 19: why block producer assignments is calculated that early?

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,219 @@
...
+    signature: Signature,
+    account_id: AccountId
+}
+```
+
+Where the parameter of the `Endorsement` is the hash of the approved block, the parameter of the `Skip` is the height of the approved block, `target_height` is the specific height at which the approval can be used (an approval with a particular `target_height` can be only included in the `approvals` of a block that has `height = target_height`), `account_id` is the account of the block producer who created the approval, and `signature` is their signature on the tuple `(inner, target_height)`.
+
+## Approvals Requirements
+
+A block `B` at height `h` that has some other block `B'` as its previous block must logically contain approvals of a form described in the next paragraph from block producers whose cumulative stake exceeds 2/3 of the total stake in the current epoch, and in specific conditions described in section [epoch switches](#epoch-switches) also the approvals of the same form from block producers whose cumulative stake exceeds 2/3 of the total stake in the next epoch.
```
ailisp, line 45: why also need approvals from bp that has >2/3 stake in next epoch

ailisp: I got it

specs/ChainSpec/Consensus.md
ailisp: got it

specs/ChainSpec/Consensus.md
```diff
@@ -0,0 +1,219 @@
...
+```python
+def on_approval(self, approval):
+    self.approvals.append(approval)
+```
+
+Whenever a participant receives a block, the operations relevant to the consensus include updating the `head` and initiating a timer to start sending the approvals on the block to the block producers at the consecutive `target_height`s. The timer delays depend on the height of the last final block, so that information is also persisted.
+
+```python
+def on_block(self, block):
+    header = block.header
```
ailisp, line 80: does on_approval and on_block all comes after "check approval/block valid & not repetitive"?

bowenwang1996: yes

specs/ChainSpec/Consensus.md
```diff
@@ -1,5 +1,23 @@
 # Consensus
 
+## Definitions and notation
+
+For the purpose of maintaining consensus, transactions are grouped into *blocks*. There is a single preconfigured block \(G\) called *genesis block*. Every block except \(G\) has a link pointing to the *previous block* \(prev(B)\), where \(B\) is the block, and \(G\) is reachable from every block by following those links (that is, there are no cycles).
```
bowenwang1996: I don't think markdown has latex support. Am I wrong?

abacabadabacaba: We have MathJax support enabled in [book.toml](https://github.com/nearprotocol/NEPs/blob/master/book.toml). You can look at the generated page here: https://nomicon-pr-57.onrender.com/ChainSpec/Consensus.html.

bowenwang1996: EPIC

PR #60, opened 2020-04-23
[Spec] Light Client

SkidanovAlex:


specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+4. `approvals_next` and `approvals_after_next` contain signatures that check out against the block producers for the epoch of the block (see next section);
+5. The signatures present in both `approvals_next` and `approvals_after_next` correspond to more than 2/3 of the total stake.
+6. If `next_bps` is not none, `sha256(borsh(next_bps))` corresponds to the `next_bp_hash` in `inner_lite`.
+
+## Signature verification
+
+By construction by the time the `LightClientBlockView` is being validated, the block producers set for its epoch is known.
+
+If the `LightClientBlockView` corresponds to some block at height `h`, then the signatures in the `LightClientBlockView::approvals_next` are signatures on endorsements in the block at height `h+1`, and the signatures in the `LightClientBlockView::approvals_after_next` and signatures on endorsements in the block at height `h+2`. To simplify the protocol we require that the next block and the block after next are both in the same epoch as the block that `LightClientBlockView` corresponds to. It is guaranteed that each epoch has at least one final block for which the next two blocks that build on top of it are in the same epoch.
```
nearmax, line 70: When it says:
>then the signatures in the `LightClientBlockView::approvals_next` are signatures on endorsements in the block at height `h+1`

Does it mean?
> then the signatures in the `LightClientBlockView::approvals_next` are signatures on endorsements in the block at height `h+1` of the block at height `h`

nearmax, line 70: Are signatures signing the block hash?

nearmax, line 70: > To simplify the protocol we require that the next block and the block after next are both in the same epoch as the block that `LightClientBlockView` corresponds to.

It is not clear what happens to the blocks that are on the edge of the epoch switch.
Suppose blocks X and X+1 are in epoch E, and block X+2 is in epoch E+1. Are we not creating `LightClientBlockView` from block X and block X+1? How are we going to verify content of these blocks using light client?

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+
```
nearmax, line 71: Also, is light client supposed to handle fork resolution or can is safely assume that the `LightClientBlockView` come only from the canonical chain?

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+## RPC end-point
+
+There's a single end-point that full nodes exposed that light clients can use to fetch new `LightClientBlockView`s:
+
+```
+http post http://127.0.0.1:3030/ jsonrpc=2.0 method=next_light_client_block params:="[<last known hash>]" id="dontcare"
+```
+
+The RPC returns the `LightClientBlock` for the block as far into the future from the last known hash as possible for the light client to still accept it. Specifically, it either returns the last final block of the next epoch, or the last final known block. If there's no newer final block than the one the light client knows about, the RPC returns an empty result.
```
nearmax, line 82: Can we have an RPC that returns `LightClientBlock` given the block id (either index or hash)? Without it I don't understand how light client will be able to close the gaps if it misses one block and the canonical chain moves forward.

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+If the `LightClientBlockView` corresponds to some block at height `h`, then the signatures in the `LightClientBlockView::approvals_next` are signatures on endorsements in the block at height `h+1`, and the signatures in the `LightClientBlockView::approvals_after_next` and signatures on endorsements in the block at height `h+2`. To simplify the protocol we require that the next block and the block after next are both in the same epoch as the block that `LightClientBlockView` corresponds to. It is guaranteed that each epoch has at least one final block for which the next two blocks that build on top of it are in the same epoch.
+
+Since both the hashes of the current and next block, and the block producers who signed them are known, the signatures can be validated.
+
+## RPC end-point
+
+There's a single end-point that full nodes exposed that light clients can use to fetch new `LightClientBlockView`s:
+
+```
+http post http://127.0.0.1:3030/ jsonrpc=2.0 method=next_light_client_block params:="[<last known hash>]" id="dontcare"
```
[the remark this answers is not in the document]
frol: yeah, definitely let's use object-based input parameters to enable extensibility (positional arguments bit us already several times):

```
http post http://127.0.0.1:3030/ jsonrpc=2.0 id="dontcare" \
    method=next_light_client_block \
    'params:={"last_known_hash": <last known hash>}'
```

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+Light clients operate by periodically fetching instances of `LightClientBlockView` via particular RPC end-point desdribed [below](#rpc-end-point).
+
+```rust
+pub struct ValidatorStakeView {
+    pub account_id: AccountId,
+    pub public_key: PublicKey,
+    pub stake: Balance,
+}
+
+pub struct LightClientBlockView {
```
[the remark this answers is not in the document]
SkidanovAlex: Bowen is pretty adamant that the spec should not explain motivation that is not relevant in the context. I.e. the user of the light client doesn't need to understand why `prev_block_hash` is not in the inner.

If I understand correctly your concern.

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+    pub prev_block_hash: CryptoHash,
+    pub next_block_inner_hash: CryptoHash,
+    pub inner_lite: BlockHeaderInnerLiteView,
+    pub inner_rest_hash: CryptoHash,
+    pub next_bps: Option<Vec<ValidatorStakeView>>,
```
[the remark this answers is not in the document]
SkidanovAlex: Because `next_bps` here is optional. Next BPs can be ~half of the size of the block, and is the same throughout the epoch, so if a light client queries blocks more frequently than once per epoch, there's no need to send the full set each time.

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+    pub approvals_next: Vec<Option<Signature>>,
+    pub approvals_after_next: Vec<Option<Signature>>,
+}
+```
+
+Recall that the hash of the block is
```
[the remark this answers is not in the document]
SkidanovAlex: This "recall" refers to the spec that doesn't exist yet :)

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+    ))
+))
+
+next_block_hash = sha256(concat(
+    current_block_hash,
+    block_view.next_block_inner_hash
+))
+```
+
+The light client updates its head with the information from `LightClientBlockView` iff:
```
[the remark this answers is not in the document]
SkidanovAlex: There's no requirements beyond those in the list below. `iff` stands for `if and only if`, as in, the set of requirements is exact, any fewer is not enough, and having these is sufficient.

Correspondingly, to accept any block in a particular epoch, the light client current head must be either in the same epoch or in the previous. It is reflected in (2) below.

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. The height of the block is higher than the height of the current head;
+2. The epoch of the block is equal to the `epoch_id` or `next_epoch_id` known for the current head;
+3. If the epoch of the block is equal to the `next_epoch_id` of the head, then `next_bps` is not `None`;
```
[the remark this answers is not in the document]
SkidanovAlex: If you first receive block from E+1, if all the conditions here are satisfied, it will get accepted, otherwise it will get rejected.
If it got accepted, and you later receive a block from E, that block from E will be rejected, because it will violate (1).

specs/ChainSpec/LightClient.md
SkidanovAlex: `LightClientBlockView` can only exist for a final block. If there's a fork of final blocks (safety is violated), then the light client applies the fork choice rule (which is just the height) locally, due to (1) above.

specs/ChainSpec/LightClient.md
SkidanovAlex: `LightClient` should not have to receive blocks in the gaps. The idea is that it has the block close to the current head, and uses proofs against that block, it doesn't need any other blocks.

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,170 @@
...
+
+    approval_message = concat(
+        borsh(ApprovalInner::Endorsement(next_block_hash)),
+        little_endian(block_view.inner_lite.height + 2)
+    )
+
+    return (current_block_hash, next_block_hash, approval_message)
+```
+
+The light client updates its head with the information from `LightClientBlockView` iff:
```
nearmax, line 87:
```suggestion
The light client updates its head with the information from `LightClientBlockView` if:
```
Iff is a logical biconditional `<->`. If we use "iff" we mean that if points 1-6 then light client must update its head, which is not the case, it might choose to not update it.

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,170 @@
...
+4. `approvals_after_next` contain valid signatures on `approval_message` from the block producers of the corresponding epoch (see next section);
+5. The signatures present in `approvals_after_next` correspond to more than 2/3 of the total stake (see next section).
+6. If `next_bps` is not none, `sha256(borsh(next_bps))` corresponds to the `next_bp_hash` in `inner_lite`.
+
+```python
+def validate_and_update_head(block_view):
+    global head
+    global epoch_block_producers_map
+
+    current_block_hash, next_block_hash, approval_message = reconstruct_light_client_block_view_fields
```
nearmax, line 101:
```suggestion
    current_block_hash, next_block_hash, approval_message = reconstruct_light_client_block_view_fields(block_view)
```

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,170 @@
...
+
+    epoch_block_producers = epoch_block_producers_map[block_view.inner_lite.epoch_id]
+    for maybe_signature, block_producer in zip(block_view.approvals_after_next, epoch_block_producers):
+        total_stake += block_producer.stake
+
+        if maybe_signature is None:
+            continue
+
+        approved_stake += block_producer.stake
+        if not verify_signature(
```
nearmax, line 127: We should check whether it was a skip message here and not return False if it was.

specs/ChainSpec/LightClient.md
SkidanovAlex: Then it's a light client that is not conformant to this spec.

specs/ChainSpec/LightClient.md
SkidanovAlex: Not sure I understand.
All the approval messages in the block are identical, and for the purposes of a light client block they are all endorsements of a very particular form, exactly the form that we validate the signature against here.

specs/ChainSpec/LightClient.md
SkidanovAlex: (similar to how when describing the block production / acceptance of full nodes, the spec will specify that if a full node receives a valid block that has higher height, it **must update**, not **may update** its head)

specs/ChainSpec/LightClient.md
SkidanovAlex: > Are we not creating LightClientBlockView from block X and block X+1

Yes

> How are we going to verify content of these blocks using light client?

The beginning of the spec says that you receive block view for a subset of blocks, not for all the blocks, and that it is sufficient to validate any proofs against any block in their ancestry. In particular, you will not receive light client block views for the blocks on the epoch boundary, but it will not prevent you from validating statements against those blocks once at least 3 blocks are built in the next epoch (and thus a light client block view can be created and sent to you)

SkidanovAlex: The spec defines exactly what the signatures sign below (they do not sign the block hash, they sign approval messages)

specs/ChainSpec/LightClient.md
nearmax: I don't remember what `ApprovalInner::Skip` corresponds to but it looks like this is something a validator can sign instead of `ApprovalInner:: Endorsement` and then `approvals_after_next` might contain `None`, and signatures of `Endorsement` and `Skip`. We are only taking care of `None` and signature of `Endorsement` here.

SkidanovAlex: Incidentally, this is one of the few components of the chain that have a full detailed spec:

https://nomicon.io/ChainSpec/Consensus.html

The block producer might have signed endorsements or skips, but a paritcular **block** can only include signatures on a very particular kind of approval (endorsement if its previous block has sequential height, or skip if there are skipped heights). A final block always has two blocks ahead of it with sequential heights, and thus they always only contain signatures on endorsements.

specs/ChainSpec/LightClient.md
SkidanovAlex: (while this particular sentence is gone, a similar one remains for `approvals_after_next`)

Yes, but endorsement is always on a block with height `h`, so it's a redundant information. If it was a block with a lower height, it would be a `Skip`

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,86 @@
...
+
+next_block_hash = sha256(concat(
+    current_block_hash,
+    block_view.next_block_inner_hash
+))
+```
+
+The light client updates its head with the information from `LightClientBlockView` iff:
+
+1. The height of the block is higher than the height of the current head;
```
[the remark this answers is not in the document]
SkidanovAlex: It's (3). `next_bps` is set if the light client block view the full node is returning is in an epoch different from the last hash known to the light client (which is an argument to the RPC end point).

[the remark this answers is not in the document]
SkidanovAlex: (alternatively, a full node can choose to always include `next_bps`, it would still be conformant to this spec, but redundant for the light client)

specs/ChainSpec/LightClient.md
```diff
@@ -0,0 +1,170 @@
...
+    pub next_bp_hash: CryptoHash,
+}
+
+pub struct LightClientBlockView {
+    pub prev_block_hash: CryptoHash,
+    pub next_block_inner_hash: CryptoHash,
+    pub inner_lite: BlockHeaderInnerLiteView,
+    pub inner_rest_hash: CryptoHash,
+    pub next_bps: Option<Vec<ValidatorStakeView>>,
+    pub approvals_after_next: Vec<Option<Signature>>,
```
nearmax, line 46: It is missing `block_merkle_root` and the corresponding spec for it.

SkidanovAlex: Yeah, the field is missing.
Regarding the spec for it, we also presently lack the spec for outcome proofs and state proofs.
I'm curious if all such info should be in the light client spec, or if we should have a separate specification for proving statements against hashes in the block.

(I don't mind including it in this spec, but if we add the spec for merkle root of blocks, we should also add all the spec for state root and outcomes root here as well I guess?)

nearmax: I think we should have a separate spec for our state and merkle trees of transactions and outcomes.

bowenwang1996: Agree that we should have a separate spec for it.

SkidanovAlex: @nearmax @bowenwang1996 not sure if you are suggesting (1) or (2) below:

1) Have a separate spec for proofs of outcomes and state, but explain proofs of block inclusion in this spec
2) Have a separate spec for proofs of outcomes and state, and the proofs of block inclusion.

I think we should go with (2), in which case the only outstanding item for this PR is to add the missing field in this data structure.

bowenwang1996: (2) sounds good to me

PR #62, opened 2020-05-04
feat: voting contract

bowenwang1996:
This NEP describes the voting contract we need for various governance related activities, such as when to reset the network and when to unlock token transfer. There has been some preliminary work around the contract implementation https://github.com/near/initial-contracts/pull/6 and necessary protocol changes https://github.com/nearprotocol/nearcore/pull/2504.


bowenwang1996:
@evgenykuzyakov Two questions:
1. if there is no poll then how do you know whether the proposals fall into the same category? This is important for calculating how much you can vote as well as for viewing all proposals for the same topic.
2. If there is no expiration height the state might grow very large.

evgenykuzyakov:
> if there is no poll then how do you know whether the proposals fall into the same category? This is important for calculating how much you can vote as well as for viewing all proposals for the same topic.

If we decide that each proposal is independent, then each proposal can either be accepted or expire. For transfers, you should get to the state where transfers can expire.

For network upgrades, we can have a separate contract that tracks current poll by having a list of proposals. 

> If there is no expiration height the state might grow very large.

Might need to limit who can create proposals through whitelist. Then network upgrade contract will be able to create proposals.

bowenwang1996:
@evgenykuzyakov I see. That is an interesting approach. So pretty much you can vote on both "unlocking transfer at height 10000" and "unlocking transfer at height 1000" with all your stake. It definitely simplifies things a lot but at the same time it feels a bit weird...

bowenwang1996:
> Might need to limit who can create proposals through whitelist. Then network upgrade contract will be able to create proposals.

I think that will result in bad UX. You have to go to one contract to create a proposal and another to actually vote on it.

bowenwang1996:
@evgenykuzyakov 
> For transfers, you should get to the state where transfers can expire.

What does this mean?

evgenykuzyakov:
> For transfers, you should get to the state where transfers can expire.

Sorry. I meant you should NOT get to the state where transfers vote can expire.

bowenwang1996:
Update: discussed with @evgenykuzyakov offline. We decide that:
* we will have one voting contract per poll (which supports multiple proposals). This NEP will provide a template for voting contracts.
* staking pool will act on behalf of delegators so that there is no need to separately deal with delegator votes.

proposals/0000-template.md
```diff
@@ -0,0 +1,154 @@
...
+aims to make a first step in that direction by laying out the framework of a general-purpose voting contract that allows
+validators as a community to make decisions on chain. 
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+The structure of this contract looks like the following:
+
+```rust
+pub type ProposalId = u64;
```
evgenykuzyakov, line 28: This has to be `U64` as json type

proposals/0000-template.md
```diff
@@ -0,0 +1,154 @@
...
+The voting contract has the following methods to allow creation of polls and proposals, as well as voting for proposals.
+
+```rust
+impl Poll {
+    /// Initialize the poll on some topic.
+    pub fn new(topic: String, threshold: Fraction) -> Self;
+    /// Create a proposal for a given poll.
+    pub fn create_proposal(&mut self, description: String, metadata: String);
+    /// Vote on a given proposal with certain amont of stake.
+    pub fn vote(&mut self, proposal_id: ProposalId, stake: Balance);
```
evgenykuzyakov, line 79: Balance has to be `U128` as string type.

proposals/0000-template.md
```diff
@@ -0,0 +1,154 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+A major difficulty in managing the votes comes from the fact that validator stake can change from epoch to epoch, which
+means that we need to carefully update validator votes. For example, if a validator has 100 stake in the previous epoch and
+they voted 80 on some proposal, but in the current epoch their stake has decreased to 80, and now if they try to vote 20
+on some other proposal, it should fail. To address this, we introduce a helper function `resolve`, which is called
+every epoch before any call to the contract:
+```rust
+fn resolve(&mut self) {
+    // go through all the proposals and for each account, scale the stake on votes to current stake, i.e,
```
evgenykuzyakov, line 96: This is not going to work. Need to pass the `proposal_id`. Otherwise you'll exceed the gas limit.

proposals/0000-template.md
```diff
@@ -0,0 +1,154 @@
...
+/// # Cost
+///
+/// For not nul-terminated account id:
+/// `base + read_memory_base + read_memory_byte * num_bytes + utf8_decoding_base + utf8_decoding_byte * num_bytes + memory_write_base + memory_write_size * 16`
+///
+/// For nul-terminated account id :
+/// `base + (read_memory_base + read_memory_byte) * num_bytes + utf8_decoding_base + utf8_decoding_byte * num_bytes + memory_write_base + memory_write_size * 16`
+pub fn validator_stake(
+    &mut self,
+    account_len: u64,
```
evgenykuzyakov, line 119: account_id_len

proposals/0000-template.md
```diff
@@ -0,0 +1,154 @@
...
+at the same time, provided that the sum of stake voted does not exceed their current stake.
+
+The voting contract has the following methods to allow creation of polls and proposals, as well as voting for proposals.
+
+```rust
+impl Poll {
+    /// Initialize the poll on some topic.
+    pub fn new(topic: String, threshold: Fraction) -> Self;
+    /// Create a proposal for a given poll.
+    pub fn create_proposal(&mut self, description: String, metadata: String);
```
evgenykuzyakov, line 77: This is likely has to be payable function to avoid spamming the contract state. But it doesn't have to be.
Also it should return the `ProposalId` of the new proposal.

proposals/0000-template.md
bowenwang1996: Make sense. But how do we make sure that we resolve all proposals?

evgenykuzyakov: If one of the proposals is resolved, then the entire poll is resolved.

bowenwang1996: That doesn't make sense. Here `resolve` scales the stake voted on proposals. Resolving one proposal doesn't help.

proposals/0000-template.md
```diff
@@ -0,0 +1,151 @@
...
+}
+```
+Here `metadata` is the json serialized content of the proposal. For example, if the poll is about when the network upgrade
+should happen, `metadata` will be something like `{"height": 10000}`. Validators can vote on multiple different proposals
+at the same time, provided that the sum of stake voted does not exceed their current stake.
+
+The voting contract has the following methods to allow creation of polls and proposals, as well as voting for proposals.
+
+```rust
+impl Poll {
```
evgenykuzyakov, line 73: Need a method get the result

FelixLutsch:
Does this mean NEAR will also have on-chain governance? Who could make such proposals? If it's any NEAR account it will probably be useful to add some deposit to the contract that might be burned/given to the community pool or whatever to prevent spam

bowenwang1996:
> does this mean NEAR will also have on-chain governance? Who could make such proposals? If it's any NEAR account it will probably be useful to add some deposit to the contract that might be burned/given to the community pool or whatever to prevent spam

Yes as I said in the proposal this will be the first step towards on-chain governance. The current plan is that anyone can make proposals, but as you point out, we need some economics around that to prevent attacks. That part is not finalized yet and I'll update the proposal when I flesh out the details a bit more.

ilblackdragon:
Posting my comment from here https://github.com/near/initial-contracts/pull/6#discussion_r430652532:

So I thought we are going to do next thing:

- Proposal are sent with date of unlock all the time, there is a window to vote for proposal is time to unlock - 2 weeks - 4 epochs
 - Validators vote for proposals as they see fit until voting is allowed: either yay or nay
 - Proposal voting locks for 4 epochs and at that point we evaluate who voted for it, and what is their stake. If 66% of total stake voted -> proceed on 2 week waiting period before unlocking.

Why 4 epochs? Because we want to allow delegates to switch their validators if they don't think the validator they are delegating right now represents them.

Why 2 weeks? Because transfer unlock will require various parties to be prepared for it (wallets, possible exchange listings, users, financial participants, etc). Giving them a specific timeline heads up is important. Also if this vote happens right after Phase 1 is done, at least there are couple of weeks between those two events (this is also needed for some internal timelines).

bowenwang1996:
@ilblackdragon this voting contract design is not specific to voting for unlocking transfer. I am not sure how to best reconcile this suggestion with the general template.

ilblackdragon:
Give me an example usage of general template? E.g. I'm not sure why would general template not follow the same rules without understanding where else are you planning to use this general template?

bowenwang1996:
@ilblackdragon the spec for the next version of the protocol for example. I was concerned with this part
```
Proposal are sent with date of unlock all the time
```
I am not sure whether you are talking about proposals in general or proposals in one specific contract. Also 2 weeks seem specific to the concerns of unlocking transfers as well.

evgenykuzyakov:
> Also 2 weeks seem specific to the concerns of unlocking transfers as well.

This can be done on the receiving part. The lockup contracts can use timestamp +2weeks.

ilblackdragon:
Fine with moving 2 weeks into lockups.

proposals/0000-template.md
```diff
@@ -0,0 +1,175 @@
...
+
+A major difficulty in managing the votes comes from the fact that validator stake can change from epoch to epoch, which
+means that we need to carefully update validator votes. For example, if a validator has 100 stake in the previous epoch and
+they voted 80 on some proposal, but in the current epoch their stake has decreased to 80, and now if they try to vote 20
+on some other proposal, it should fail. To address this, we introduce a helper function `resolve_proposal`, which is called
+before `vote` to ensure consistency of voted stake within a proposal:
+```rust
+fn resolve_proposal(&mut self, proposal_id: ProposalId) {
+    // if the epoch height has changed, then
+    // for each vote in the proposal, its stake is changed to orginal_stake * current_total_account_stake / previous_total_account_stake.
```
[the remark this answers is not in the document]
bowenwang1996: > total stake didn't change

Why? with reward the total stake increases

> one validator withdrew 10k from their account

Stake doesn't get returned immediately so I don't think there is a problem.

proposals/0000-template.md
```diff
@@ -0,0 +1,175 @@
...
+    /// Serialized metadata of the proposal.
+    metadata: String,
+    /// When this proposal expires.
+    expiration_height: BlockHeight,
+    /// Current votes on this proposal.
+    votes: Map<AccountId, Balance>,    
+}
+```
+Here `metadata` is the json serialized content of the proposal. For example, if the poll is about when the network upgrade
+should happen, `metadata` will be something like `{"height": 10000}`. Validators can vote on multiple different proposals
```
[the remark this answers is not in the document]
bowenwang1996: > multiple voting will work

The rule is that the sum of your voted stake cannot exceed your current stake.

proposals/0000-template.md
```diff
@@ -0,0 +1,175 @@
...
+pub type ProposalId = U64;
+
+#[near_bindgen]
+pub struct Poll {
+    /// Human readable description of the poll.
+    description: String,
+    /// All proposals for this poll.
+    proposals: Map<ProposalId, Proposal>,
+    /// Accounts that have participated in this poll and the corresponding stake voted.
+    accounts: Map<AccountId, Balance>,
```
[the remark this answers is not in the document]
bowenwang1996: Isn't it the same thing? If we store fractions here we need the actual stake to compute whether a proposal should be finalized and to update it when a validator changes their vote.

frol:
@bowenwang1996 I am closing this NEP as it seems that [`voting` contract has been implemented a long time ago](https://github.com/near/core-contracts/tree/master/voting)

PR #64, opened 2020-05-12
Initial spec for upgradability

ilblackdragon:
This is outcome of discussions:
 - https://commonwealth.im/near/proposal/discussion/261-upgradability
 - https://github.com/nearprotocol/nearcore/issues/1994
 - https://github.com/nearprotocol/nearcore/issues/2498
 - https://near-community.slack.com/archives/C0132RCT6AC/p1589190011058000

And in short provides smooth upgrading that is defined by % of validators switching to new version and reaching consensus within an epoch on that.

### DIFF
--- a/specs/ChainSpec/Upgradability.md
+++ b/specs/ChainSpec/Upgradability.md
@@ -1,0 +1,71 @@
+# Upgradability
+
+This part of specification describes specifics of upgrading the protocol, and touches on few different parts of the system.
+
+High level upgradability consists of three parts:
+1. Updating without any changes to underlaying data structures or protocol;
+2. Updating when underlaying data structures changed (config, database or something else internal to the node and probably client specific);
+3. Updating with protocol changes that all validating nodes must adjust to.
+
+## Versioning
+
+There are 2 different important versions:
+- Version of binary defines it's internal data structures / database and configs. This version is client specific and doesn't need to be matching between nodes.
+- Version of the protocol, defining the "language" nodes are speaking.
+
+```rust
+/// Latest version of protocol that this binary can work with.
+type ProtocolVersion = u32;
+```
+
+Open questions: do we want to use something better than a number of versioning protocol?
+Ideally it should be some hash of protocol spec that this binary supports, but we don't have a canonical protocol spec versioning yet. We can use commit hash of `spec` / `NEPs` repo.
+
+## Client versioning
+
+Clients should follow [semantic versioning](https://semver.org/).
+Specifically:
+ - MAJOR version defines protocol releases.
+ - MINOR version defines changes that are client specific but require database migration, change of config or something similar. This includes client-specific features. Client should execute migrations on start, by detecting that information on disk is produced by previous version and auto-migrate it to new one.
+  - PATCH version defines when bug fixes, which should not require migrations or protocol changes.
+
+Clients can define how current version of data is stored and migrations applied.
+General recommendation is to store version in the database and on binary start, check version of database and perform required migrations.
+
+## Protocol Upgrade
+
+### Block structure
+
+```rust
+struct BlockHeaderInnerRest {
+    ...
+    version: ProtocolVersion,
+}
+```
+
+Each validator when producing block includes latest version their node supports.
+
+### Consensus
+
+| Name | Value |
+| - | - |
+| `PROTOCOL_UPGRADE_BLOCK_THRESHOLD` | `80%` |
+
+The condition to switch to next protocol version is based on % of blocks in previous epoch:
+
+```python
+def next_epoch_protocol_version(last_block):
+    """Determines next epoch's protocol version given last block."""
+    epoch_info = epoch_manager.get_epoch_info(last_block)
+    versions = collections.defaultdict(0)
+    # Iterate over all blocks in previous epoch and collect which versions it had.
+    for block in epoch_info:
+        versions[block.version] += 1
+    # Sort by frequency and take first one.
+    most_frequent = sorted(versions.items(), key=lambda x: -x[1])[0]
+    # Return most frequent if the frequency is above threshold.
+    if most_frequent[1] > PROTOCOL_UPGRADE_BLOCK_THRESHOLD * epoch_info.num_blocks:
+        return most_frequent[0]
+    # Otherwise return version that was used in previous epoch.
+    return epoch_info.version
+```
--- a/specs/ChainSpec/Upgradability.md
+++ b/specs/ChainSpec/Upgradability.md
@@ -33,24 +33,24 @@ General recommendation is to store version in the database and on binary start, 
 
 Generally, we handle data structure upgradability via enum wrapper around it. See `BlockHeader` structure for example.
 
-### Block structure
+### Versioned data structures
+
+Given we expect many data structures to change or get updated as protocol evolves, few changes are required to support that.
+
+The major one is adding backward compatible `Versioned` data structures like this one:
 
 ```rust
-/// Updatable BlockHeader structure.
-/// Allows to decode and store old versions.
-/// Supports up to 256 versions. After that will need to reuse.
-enum BlockHeader {
+enum VersionedBlockHeader {
     BlockHeaderV1(BlockHeaderV1),
-    BlockHeaderV2(BlockHeaderV2),
-}
-
-/// Add `version` into block header.
-struct BlockHeaderInnerRest {
-    ...
-    /// Latest version that current producing node binary is running on.
-    version: ProtocolVersion,
+    /// Current version, where `BlockHeader` is used internally for all operations.
+    BlockHeaderV2(BlockHeader),
 }
 ```
+
+Where `VersionedBlockHeader` will be stored on disk and sent over the wire.
+This allows to encode and decode old versions (up to 256 given https://borsh.io specficiation). If some data structures has more than 256 versions, old versions are probably can be retired and reused.
+
+Internally current version is used. Previous versions either much interfaces / traits that are defined by different components or are up-casted into the next version (saving for hash validation).
 
 ### Consensus
 
@@ -59,7 +59,18 @@ struct BlockHeaderInnerRest {
 | `PROTOCOL_UPGRADE_BLOCK_THRESHOLD` | `80%` |
 | `PROTOCOL_UPGRADE_NUM_EPOCHS` | `2` |
 
-The condition to switch to next protocol version is based on % of blocks in previous epoch:
+The way the version will be indicated by validators, will be via 
+
+```rust
+/// Add `version` into block header.
+struct BlockHeaderInnerRest {
+    ...
+    /// Latest version that current producing node binary is running on.
+    version: ProtocolVersion,
+}
+```
+
+The condition to switch to next protocol version is based on % of stake `PROTOCOL_UPGRADE_NUM_EPOCHS` epochs prior indicated about switching to the next version:
 
 ```python
 def next_epoch_protocol_version(last_block):
@@ -82,7 +93,7 @@ def next_epoch_protocol_version(last_block):
     for author in authors:
         versions[authors[author] += epoch_manager.validators[author].stake
     (version, stake) = max(versions.items(), key=lambda x: x[1])
-    if stake > PROTOCOL_UPGRADE_BLOCK_THRESHOLD * epoch_info.total_stake:
+    if stake > PROTOCOL_UPGRADE_BLOCK_THRESHOLD * epoch_info.total_block_producer_stake:
         return version
     # Otherwise return version that was used in that deciding epoch.
     return epoch_info.version
--- a/specs/ChainSpec/Upgradability.md
+++ b/specs/ChainSpec/Upgradability.md
@@ -2,7 +2,7 @@
 
 This part of specification describes specifics of upgrading the protocol, and touches on few different parts of the system.
 
-High level upgradability consists of three parts:
+Three different levels of upgradability are:
 1. Updating without any changes to underlaying data structures or protocol;
 2. Updating when underlaying data structures changed (config, database or something else internal to the node and probably client specific);
 3. Updating with protocol changes that all validating nodes must adjust to.
@@ -18,9 +18,6 @@ There are 2 different important versions:
 type ProtocolVersion = u32;
 ```
 
-Open questions: do we want to use something better than a number of versioning protocol?
-Ideally it should be some hash of protocol spec that this binary supports, but we don't have a canonical protocol spec versioning yet. We can use commit hash of `spec` / `NEPs` repo.
-
 ## Client versioning
 
 Clients should follow [semantic versioning](https://semver.org/).
@@ -34,22 +31,33 @@ General recommendation is to store version in the database and on binary start, 
 
 ## Protocol Upgrade
 
+Generally, we handle data structure upgradability via enum wrapper around it. See `BlockHeader` structure for example.
+
 ### Block structure
 
 ```rust
+/// Updatable BlockHeader structure.
+/// Allows to decode and store old versions.
+/// Supports up to 256 versions. After that will need to reuse.
+enum BlockHeader {
+    BlockHeaderV1(BlockHeaderV1),
+    BlockHeaderV2(BlockHeaderV2),
+}
+
+/// Add `version` into block header.
 struct BlockHeaderInnerRest {
     ...
+    /// Latest version that current producing node binary is running on.
     version: ProtocolVersion,
 }
 ```
-
-Each validator when producing block includes latest version their node supports.
 
 ### Consensus
 
 | Name | Value |
 | - | - |
 | `PROTOCOL_UPGRADE_BLOCK_THRESHOLD` | `80%` |
+| `PROTOCOL_UPGRADE_NUM_EPOCHS` | `2` |
 
 The condition to switch to next protocol version is based on % of blocks in previous epoch:
 
@@ -57,15 +65,25 @@ The condition to switch to next protocol version is based on % of blocks in prev
 def next_epoch_protocol_version(last_block):
     """Determines next epoch's protocol version given last block."""
     epoch_info = epoch_manager.get_epoch_info(last_block)
+    # Find epoch that decides if version should change by walking back.
+    for _ in PROTOCOL_UPGRADE_NUM_EPOCHS:
+        epoch_info = epoch_manager.prev_epoch(epoch_info)
+        # Stop if this is the first epoch.
+        if epoch_info.prev_epoch_id == GENESIS_EPOCH_ID:
+            break
     versions = collections.defaultdict(0)
-    # Iterate over all blocks in previous epoch and collect which versions it had.
+    # Iterate over all blocks in previous epoch and collect latest version for each validator.
+    authors = {}
     for block in epoch_info:
-        versions[block.version] += 1
-    # Sort by frequency and take first one.
-    most_frequent = sorted(versions.items(), key=lambda x: -x[1])[0]
-    # Return most frequent if the frequency is above threshold.
-    if most_frequent[1] > PROTOCOL_UPGRADE_BLOCK_THRESHOLD * epoch_info.num_blocks:
-        return most_frequent[0]
-    # Otherwise return version that was used in previous epoch.
+        author_id = epoch_manager.get_block_producer(block.header.height)
+        if author_id not in authors:
+            authors[author_id] = block.header.rest.version
+    # Weight versions with stake of each validator.
+    for author in authors:
+        versions[authors[author] += epoch_manager.validators[author].stake
+    (version, stake) = max(versions.items(), key=lambda x: x[1])
+    if stake > PROTOCOL_UPGRADE_BLOCK_THRESHOLD * epoch_info.total_stake:
+        return version
+    # Otherwise return version that was used in that deciding epoch.
     return epoch_info.version
 ```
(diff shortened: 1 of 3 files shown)


specs/SUMMARY.md
```diff
@@ -9,6 +9,7 @@
 - [Architecture](Architecture.md)
 - [Chain specification](ChainSpec/README.md)
```
SkidanovAlex, line 10: Shall we remove a link here and delete `ChainSpec/README.md`? So that we don't have to maintain multiple lists consistent.

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,71 @@
+# Upgradability
+
+This part of specification describes specifics of upgrading the protocol, and touches on few different parts of the system.
+
+High level upgradability consists of three parts:
```
ailisp, line 5: I feel you mean "three different levels" of upgrades, a higher level means touches more parts of system". so this sentence is better to be "Upgradability consists of three levels:"

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,71 @@
...
+There are 2 different important versions:
+- Version of binary defines it's internal data structures / database and configs. This version is client specific and doesn't need to be matching between nodes.
+- Version of the protocol, defining the "language" nodes are speaking.
+
+```rust
+/// Latest version of protocol that this binary can work with.
+type ProtocolVersion = u32;
+```
+
+Open questions: do we want to use something better than a number of versioning protocol?
```
ailisp, line 21: One another is YYYYMMDD timestamp, it's more human readable and imply the date of when this version start. it's also strictly increased.

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,71 @@
...
+Ideally it should be some hash of protocol spec that this binary supports, but we don't have a canonical protocol spec versioning yet. We can use commit hash of `spec` / `NEPs` repo.
+
+## Client versioning
```
ailisp, line 24: What's the relationship between "versioning" and "client versioning", does "client versioning" corresponding to "binary version"?

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,71 @@
...
+    # Iterate over all blocks in previous epoch and collect which versions it had.
+    for block in epoch_info:
+        versions[block.version] += 1
+    # Sort by frequency and take first one.
+    most_frequent = sorted(versions.items(), key=lambda x: -x[1])[0]
+    # Return most frequent if the frequency is above threshold.
+    if most_frequent[1] > PROTOCOL_UPGRADE_BLOCK_THRESHOLD * epoch_info.num_blocks:
+        return most_frequent[0]
+    # Otherwise return version that was used in previous epoch.
+    return epoch_info.version
```
ailisp, line 70: I'm worry about this may make upgrade version never happen in some case:
Say all node use version 4 on epoch 1 (epoch[4].epoch_info.version=4). At middle of epoch 1 all of them upgrade to use version 5. so there are 50% epoch 1 block has version 4 and 50% version 5. It's not more than 80% so they will set next_epoch_protocol_version to that was used in previous epoch, that is version 4. Then in epoch 5 will they use "current epoch protocol version (4)" or "their binary implied protocol version (5)" to produce blocks? I think it must be "their binary implied protocol version" to switch next_epoch_protocol_version to 5 in epoch 3, otherwise next_epoch_protocol_version would stay 4 forever.
But then why do we need next_epoch_protocol_version, if all node produced block based on "their binary implied protocol version"?

specs/ChainSpec/Upgradability.md
ailisp: Evgeny Kapun mentioned a protocol version changed need to be tested on testnet first, and YYYYMMDD doesn't reasoning good in that case. Also it doesn't give much benefit over number so it's not good.

ilblackdragon:
@abacabadabacaba suggested from Protocol Research call:
1. Store latest version that validators are running in EpochManager
2. Validators indicate their version by resending staking transaction with indicating version
3. Agreement is achieved based on epoch, where validators achieve some super-majority on a version in proposal - and kicks in the epoch where these validators are active (e.g. if proposals are sent in epoch T, it will become active in T + 2).
4. Version should be a number, because people should agree on what next version should be first before 

I generally, like the (1) & (3).
I don't like (2), because it's both requires API change for Staking Action, and it requires additional step for validators to re-sign new transaction. @SkidanovAlex suggested that Staking / Validator Key should allow to sign staking actions (last time I proposed this, @evgenykuzyakov was against it and overall is complicated, because we then need to filter for only transactions that only have single action and more).

Overall I don't fully agree with (4): even though we expect agreement in what next version will be, if there are any kind of conflicting change that people want to introduce - this mechanism will be the final way to decide. Obviously that with hash of spec reported by binary doesn't mean binary follows that spec or that there are no issues or that there are no issues in the spec. It's more to be able to handle forks in decision making.

Let's review a simple example, there are group of validators who want to increase the minimum gas price, forked the code and pushed a new version. At the same time core dev team pushes new update of protocol with some bells & whistles.
If we use integers, both of these of these will have the same version. It's possible that core team will figure out in time and bump versions, but there is a chance of weird outcome (even if it's TestNet stalling). And this is not validators being malicious - they followed procedure from their side by publishing source code, announcing it, running testnets, etc.

specs/ChainSpec/Upgradability.md
ilblackdragon: Yes, client==binary

specs/ChainSpec/Upgradability.md
ilblackdragon: In epoch 5 they will produce blocks with reported version 5 while following protocol 4 and at the end of it will have 100% on version 5.

ilblackdragon:
Linking for context:
 - Cosmos 0 time upgradability:
   -   https://github.com/cosmos/cosmos-sdk/tree/master/x/upgrade/spec
   - https://github.com/cosmos/cosmos-sdk/blob/0b7449129af3268c1d0e4d3fef4741beb2b4aa30/x/upgrade/doc.go
- Algorand: https://algorand.foundation/protocol-development

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,71 @@
...
+- Version of binary defines it's internal data structures / database and configs. This version is client specific and doesn't need to be matching between nodes.
+- Version of the protocol, defining the "language" nodes are speaking.
+
+```rust
+/// Latest version of protocol that this binary can work with.
+type ProtocolVersion = u32;
+```
+
+Open questions: do we want to use something better than a number of versioning protocol?
+Ideally it should be some hash of protocol spec that this binary supports, but we don't have a canonical protocol spec versioning yet. We can use commit hash of `spec` / `NEPs` repo.
```
bowenwang1996, line 22: The nice property with numeric values is that you can easily compare them to see which one is newer. With hash it becomes much more difficult.

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Client versioning
+
+Clients should follow [semantic versioning](https://semver.org/).
+Specifically:
+ - MAJOR version defines protocol releases.
```
bowenwang1996, line 28: according to https://semver.org/, if we make a change that requires migration (whether it is a protocol change or a client-only change), we should update major version.

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,71 @@
...
+
+Each validator when producing block includes latest version their node supports.
+
+### Consensus
+
+| Name | Value |
+| - | - |
+| `PROTOCOL_UPGRADE_BLOCK_THRESHOLD` | `80%` |
+
+The condition to switch to next protocol version is based on % of blocks in previous epoch:
```
birchmd, line 54: As discussed in the meeting the other day, we may want to change this from blocks being labelled with a version number to having the upcoming block producers declare a version they wish to support in the next epoch and set the epoch's version based on this vote (we could probably keep the 80% threshold).

I think even in this model it is a good idea to include the version number in the block header because this will allow handling messages from previous epochs properly.

One question I have about any model of upgrades is what happens to the block producers that are not yet ready? This could be up to 20% (by stake) of the block producers. Will they simply not produce blocks until they upgrade? If they do make blocks according to the last version do we still accept them or will they be considered invalid?

ilblackdragon: The problem with upcoming block producers declaring is:
 - They would need to explicitly send staking transaction (which might require some HSM / offline / multisig custody signing which is hard)
 - Validators will be unstaking a bit of stake, because when they send re-staking transaction they need to specify the amount with the future reward to match what we do by default in rollovers. 
 - Also problem is staking pool, as we need to change version inside them to allow them to send staking transactions with new version.

So alternative is to create new Transaction Action to `VersionVote(new-version)`, that can be signed with staking key (which is not supported by any of our current infra right now) and that node automatically sends and this gets attached to proposals -> epoch manager.

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,71 @@
+# Upgradability
+
+This part of specification describes specifics of upgrading the protocol, and touches on few different parts of the system.
+
+High level upgradability consists of three parts:
+1. Updating without any changes to underlaying data structures or protocol;
+2. Updating when underlaying data structures changed (config, database or something else internal to the node and probably client specific);
```
fckt, line 7: The NEP will benefit in the clearness if the it will focus only on the protocol-level (consensus) upgradability and not mix it with implementations upgradability.

specs/ChainSpec/Upgradability.md
fckt: The version order requires consensus by itself. In the decentralized setting version ordering can be only determined by order of their acceptance in the blockchain.

specs/ChainSpec/Upgradability.md
fckt: > So alternative is to create new Transaction Action to VersionVote(new-version), that can be signed with staking key (which is not supported by any of our current infra right now) and that node automatically sends and this gets attached to proposals -> epoch manager.

Also we can just add `protocol_version` into the StakeAction
https://github.com/nearprotocol/nearcore/blob/master/core/primitives/src/transaction.rs#L124-L132

ilblackdragon: @fckt that's what I argued against above the alternative option.

specs/ChainSpec/Upgradability.md
bowenwang1996: It does only cover protocol level upgradability. I don't think simply laying out the possible levels of upgradability is the same as covering them.

abacabadabacaba:
My opinion:

1. I'm still strongly in favor of consensus version being a plain number rather than a hash. Moreover, the only supported upgrade path should be from version i to version i+1. Consensus upgrade is very bug-prone, and having only one possible scenario to test should reduce the number of possible corner cases.
2. If an upgrade that is only supported by some implementations gets activated on the mainnet, this will be a significant trouble for validators using other implementations. Thus, activating an upgrade that is not supported by implementations used by almost all validators should be avoided if possible.
3. We want to have a way for validators to specify which consensus version they'd like to switch to. This shouldn't depend on them actually producing blocks (for example, a user who is going to become a validator in a future epoch should be able to do this). Having to resubmit a staking transaction is not ideal because it needs to be signed with an account key. I think that this shouldn't be a big problem in practice, because upgrades are not that frequent. In the future, we may implement a way to sign this kind of updates with a staking key instead of an account key.
4. While we are at it, we may add a field for validators to specify which software they are running. This will have no effect on consensus, but we want to know what fraction of validators is running each implementation.
5. Regarding switch condition: switching as soon as 80% of validators upgrade will do a very bad service to the remaining 20%, as they will need to upgrade in a hurry. I propose to switch only after at least 80% (or even 90%) of validators support the upgrade in _each_ of the last 10 (or even 20) epochs.
6. Also, even if a validator is running a new version of the software, they should have a configuration option to not propose an upgrade to a new consensus version.
7. Regarding data structures: current version number should be the first field of the block header. This is because interpretation of any subsequent fields may change depending on the version. It should appear in the highest level part of the block header, that is, the part hash of which is the block hash. Also, version number should be the first field in approval messages. Version number of approvals must match version number of the block they appear in.
8. Transactions should also be versioned. Version number of transactions doesn't need to match version number of blocks. When we introduce new transaction version, we should keep accepting the previous version for some time.

> Let's review a simple example, there are group of validators who want to increase the minimum gas price, forked the code and pushed a new version. At the same time core dev team pushes new update of protocol with some bells & whistles.

It would be very irresponsible for the validators to do this without any testing and coordination with the rest of the participants. Imagine what happens if a group of companies running some of the Internet core infrastructure decide to change something in TCP/IP protocol. They fork the software their infrastructure is running and roll out a new version before telling anyone else. Would it work? I don't think so. In our case, upgrading the consensus would require agreement of a supermajority of the validators, and this includes an agreement to not do any conflicting upgrade in parallel, which means that such situation with two upgrades happening at the same time shouldn't be possible.

specs/ChainSpec/Upgradability.md
fckt: Cool! Just to make sure :)

fckt:
> Moreover, the only supported upgrade path should be from version i to version i+1.

Interesting. It changes my initial intuition from the "voting for some specific update" to the "voting for (some) update".. Which actually makes sense to me (especially b/c there should be only one canonical chain and since validators are not voting for some specific binary(es)).

bowenwang1996:
> While we are at it, we may add a field for validators to specify which software they are running. This will have no effect on consensus, but we want to know what fraction of validators is running each implementation.

We can add it to telemetry. Forcing it as part of the protocol doesn't make a lot of sense.

bowenwang1996:
> Regarding switch condition: switching as soon as 80% of validators upgrade will do a very bad service to the remaining 20%, as they will need to upgrade in a hurry. I propose to switch only after at least 80% (or even 90%) of validators support the upgrade in each of the last 10 (or even 20) epochs.

I don't think it's really a problem. In practice I don't see upgrades happening in two consecutive epochs, but it also makes sense to be more cautious.

ilblackdragon:
> 1. I'm still strongly in favor of consensus version being a plain number rather than a hash. Moreover, the only supported upgrade path should be from version i to version i+1. Consensus upgrade is very bug-prone, and having only one possible scenario to test should reduce the number of possible corner cases.

Yeah, that's fine.

> 2. If an upgrade that is only supported by some implementations gets activated on the mainnet, this will be a significant trouble for validators using other implementations. Thus, activating an upgrade that is not supported by implementations used by almost all validators should be avoided if possible.

Well, that was the point of using this as a final coordination mechanism. Social consensus is great and should be used as much as possible, but you also want to make sure to have final stops.

TCP/IP is a bad example, because it's been 10+ years as they have been trying to update. We want to update every month the slowest.

> 4. While we are at it, we may add a field for validators to specify which software they are running. This will have no effect on consensus, but we want to know what fraction of validators is running each implementation.

This is already in telemetry - https://explorer.betanet.near.org/nodes/validators

> Regarding switch condition: switching as soon as 80% of validators upgrade will do a very bad service to the remaining 20%, as they will need to upgrade in a hurry

The condition is 80% of validators indicated a switch, epoch after next will activate it. So the rest of validators have epoch+some to switch. We can add a bit more timing, but I also don't want to give infinite amount of time here, because in case of security updates etc - we want this to be timely.

bowenwang1996:
@ilblackdragon 

> security updates

We can discuss this separately but I feel like for critical vulnerabilities I am not entirely sure that using this kind of voting is the best idea.

ilblackdragon:
Quick recap from research meeting:
 - If we can, we should try to avoid adding new actions. At this point doing upgradability voting via smart contracts though is way too complicated.
 - We need a way to determine version for block/chunk when deserialized. It might be possible to determine this based on current height. But it might be hard to do given the deserialization calls.
 - Because recommended way to upgrade nodes is by starting a new node with new validator key and then re-issuing staking transaction with it. Hence we expect validators to already re-issue staking transaction.
 - It would be still great if staking transaction can be signed by signed key - but it will be complicated to implement at the moment, because verifier will need to have access to EpochManager.

specs/ChainSpec/Upgradability.md
ilblackdragon: Migration is backward compatible change when APIs don't change. So it's a MINOR version update per semver.

specs/ChainSpec/Upgradability.md
ailisp: If it's the commit hash, we have git history to compare, it's okay for scripts but more difficult human. What about
- a web page for listing spec hash <> spec
- or use a combination of <number>_<hash>

specs/ChainSpec/Upgradability.md
ilblackdragon: Updated logic also to account for the latest version that given validator produced.

ilblackdragon:
Want to follow up and close the thread on upgradability spec.

Specifically, I went back and forth, and even though @abacabadabacaba idea of using proposals to indicate next version makes sense - it’s more complicated to implement, more complicated for validators and doesn’t really change that much the behavior.

It’s worth noting with more chunk producers (more shards) and hidden validators there indeed can be problem that if only block producers upgraded there is still large % of stake didn’t update yet.
But I think adding enough buffer should mitigate this issue and still at the end upgradability is about more on informing everyone on time about update and making smooth transition that doesn’t require state dump.

Given that, I kept the version for indicating about upcoming transition in block header, but now it aggregates total stake voting for the version and it checks K epochs before the current one to give everyone time.

See details here: https://nomicon-pr-64.onrender.com/ChainSpec/Upgradability.html#consensus

Additionally, I added example of how we can handle quite a few updates to data structures like this: https://nomicon-pr-64.onrender.com/ChainSpec/Upgradability.html#versioned-data-structures We will need to work more on this as we actually get real examples of protocol changes.

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,100 @@
...
+Clients can define how current version of data is stored and migrations applied.
+General recommendation is to store version in the database and on binary start, check version of database and perform required migrations.
+
+## Protocol Upgrade
+
+Generally, we handle data structure upgradability via enum wrapper around it. See `BlockHeader` structure for example.
+
+### Versioned data structures
+
+Given we expect many data structures to change or get updated as protocol evolves, few changes are required to support that.
```
bowenwang1996, line 38:
```suggestion
Given we expect many data structures to change or get updated as protocol evolves, a few changes are required to support that.
```

bowenwang1996:
@ilblackdragon given the data structure you proposed
```rust
enum VersionedBlockHeader {
    BlockHeaderV1(BlockHeaderV1),
    /// Current version, where `BlockHeader` is used internally for all operations.
    BlockHeaderV2(BlockHeader),
}
```
when a node with the new version enters the network, how do the other nodes (with old version) process the header produced by the new version? It seems that there has to be some logic to produce the block in the old version with the new version number.

ilblackdragon:
Node running new binary will keep producing old block headers until the version transitions to the new one.
I'm trying to write this out https://github.com/nearprotocol/nearcore/pull/2701 here to nail down better how it should look like. Probably some traits will be used.

specs/ChainSpec/Upgradability.md
```diff
@@ -0,0 +1,100 @@
...
+| `PROTOCOL_UPGRADE_NUM_EPOCHS` | `2` |
+
+The way the version will be indicated by validators, will be via 
+
+```rust
+/// Add `version` into block header.
+struct BlockHeaderInnerRest {
+    ...
+    /// Latest version that current producing node binary is running on.
+    version: ProtocolVersion,
```
birchmd, line 69: Just double checking for my understanding. This field represents the highest version number the node producing the block _could_ support, but not the version of the block header itself, correct? The actual version of the block header would be determined based on the enum variant it deserializes to.

If that is the case I wonder if a different name should be used. Say `proposed_version` or something like that.

ilblackdragon: You are correct, I used `latest_protocol_version` in implementation here - https://github.com/nearprotocol/nearcore/pull/2701

ilblackdragon:
So there is a small alternation that makes it close to what @abacabadabacaba proposed, while allowing block producer to automatically do this. Instead of adding a field to the `BlockHeader`, can extend `ValidatorStake` to contain `latest_protocol_version`.

While by default Proposals from `Stake` transactions won't fill this field (until we change `Stake` action), we can extend validity of proposal if the chunk producer adds one for themself (one is important note to make sure they can't create large number of proposals).

Note my self: make sure our proposals code validation checks invalid for proposals.

bowenwang1996:
@ilblackdragon when do chunk producer add it? Also this is not future-proof. What do we do when we have hidden validators?

ilblackdragon:
@bowenwang1996 Chunk producers add it when it's their turn to produce chunk and it's the first time they produce a chunk since their binary version changed.

Hidden validators don't participate in either voting scheme, so the latest proposal is strictly more inclusive than original one (as it includes chunk producers on top of only block producers in the original proposal).

bowenwang1996:
> so the latest proposal is strictly more inclusive than original one

Currently they are the same. If you are talking about future possibilities, you are assuming that every block producer is a chunk producer.

PR #69, opened 2020-05-20
wip: Network specification

mfornet:
Follow up from #12


specs/ChainSpec/Messages.md
```diff
@@ -0,0 +1,98 @@
...
+    /// Original sender of this message
+    author: PeerId,
+    /// Signature from the author of the message. If this signature is invalid we should ban
+    /// last sender of this message. If the message is invalid we should ben author of the message.
+    signature: Signature,
+    /// Time to live for this message. After passing through some hop this number should be
+    /// decreased by 1. If this number is 0, drop this message.
+    ttl: u8,
+    /// Message
+    body: RoutedMessageBody,
```
SkidanovAlex, line 96: I would add the definition of `RountedMessageBody` also to this file.

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
+# Network
+
+Network layer constitutes the lower level of the NEAR protocol and is ultimately responsible of transporting messages between peers. To provide an efficient routing it maintains a routing table between all peers actively connected to the network, and send message between them using best paths. There is a mechanism in place that allows new peers joining the network to discover other peers, and rebalance network connections in such a way that latency is minimized. Cryptographic signatures are used to check identities from peers participating in the protocol since it is non-permissioned system.
```
SkidanovAlex, line 3: "and send message" -> "and sends messages"

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+This document should serve as reference for all the clients to implement networking layer.
+
+## Messages
+
+All data structure used for messages between peers are enumerated in [Message](Messages.md).
+
+## Discovering the network
+
+When a node start for the first time it tries to connect to a list of bootstrap nodes specified via a config file.
```
SkidanovAlex, line 13: node start -> node starts

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+It is expected that a node on a regular basis request a list of peers from its neighboring nodes to learn about other nodes in the network. This will allow every node to discover each other, and have relevant information to try to establish a new connection with it. When a node receive a message of type [`PeersRequest`](Messages.md#PeersMessage) it is expected to answer with a message of type [`PeersResponse`](Message.md#PeersMessage) with information from healthy peers known to this node. Both of this message are unrestricted, i.e. can be sent at any moment without incurring in a fault, but beware about the policy to [ban for abusive behavior](#Abusive-behavior).
```
SkidanovAlex, line 15: request a list -> requests a list
when a node receive -> when a node receives

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Handshakes
+
+To establish new connections between pair of nodes, they will follow the following protocol. Node A open a connection with node B and sends a [Handshake](Messages.md#Handshake) to it. If handshake is valid (see reasons to [decline the handshake](#Decline-handshake)) then node B will proceed to send [Handshake](Messages.md#Handshake) to node A. After each node accept a handshake it will mark the other node as an active connection, until one of them stop the connection.
```
SkidanovAlex, line 19: Either "to establish new connections between pairs of nodes" or "to establish a new connection between a pair of nodes"

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+#### Decline handshake
+
+When a node receive a handshake from other node it will decline this connection if one of the following situations happens:
+
+1. Other node has different genesis.
+2. Edge nonce is too low
+
+#### Edge
+
+Edges are used to let other know in the network that there is currently an active connection between a pair of nodes. See the signature of [this data structure](Messages.md#Edge).
```
SkidanovAlex, line 32: "to let other know in the network" -> "to let others in the network know"
"the signature of this datastructure" -> "the definition of this datastructure" (to avoid confusion with cryptographic signatures)

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+If the nonce of the edge is odd, it denotes an `Added` edge, otherwise it denotes a `Removed` edge. Each node should keep track of the nonce used for edges between every pair of nodes. Peer C believe that peer A and B are currently connected if and only if the edge with higher nonce known to C about them have odd nonce.
```
SkidanovAlex, line 34: "Peer C believe that peer A and B are currently connected if and only if the edge with higher nonce known to C about them have odd nonce." -> "Peer C believes that the peers A and B are currently connected if and only if the edge with the highest nonce known to C for them has an odd nonce."

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+When two nodes successfully connect to each other, they broadcast the new edge to let other peers know about this connection. When a node is disconnected from other node, it should bump the nonce by 1, sign the new edge and broadcast it to let other nodes know that the connection was removed.
+
+A removed connection will be valid, if it contains valid information from the added edge it is invalidating. This will prevent peers bump nonce by more than one when deleting an edge.
```
SkidanovAlex, line 38: This will prevent -> This prevents

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+When node A propose an edge to B with nonce X, it will only accept it and sign it if:
```
SkidanovAlex, line 40: propose -> proposes
also, if -> iff?

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+- X = 1 and B doesn't know about any previous edge between A and B
+- X is odd and X > Y where Y is the nonce of the edge with higher nonce between A and B known to B.
```
SkidanovAlex, line 43: higher -> the highest

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Routing Table
+
+Every node will maintain a routing table with all existing connections and relevant information to route messages. The explicit graph with all active connection is stored at all times.
```
SkidanovAlex, line 47: will maintain -> maintains

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+    /// Current view of the network. Nodes are Peers and edges are active connections.
+    raw_graph: Graph,
+}
+
+pub struct Graph {
+    adjacency: HashMap<PeerId, HashSet<PeerId>>,
+}
+```
+
+`RoutingTable` should be update accordingly the node receives updates from the network. There are four types of updates `RoutingTable` cares about:
```
SkidanovAlex, line 68: accordingly the -> accordingly when the

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+- [Adding Edge](#Add-edge)
+- [Removing Edge](#Remove-edge)
+- [Route back message](#Route-back-messages)
+- [Announce account](#Announce-account)
+
+### Routing
+
+When a node needs to send a message to another peer, it checks in the routing table if it is connected to that peer, possibly not directed but through several hops. Then it select one of the shortest path to the target peer and send a [`RoutedMessage`](Messages.md#RoutedMessage) to the first peer in the path.
```
SkidanovAlex, line 77: possibly not directed -> possibly not directly

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+When it receives a `RoutedMessage`, it check if it is the target, in that case consume the body of the message, otherwise it finds a route to the target following described approach and send the message again. Is is important that before routing a message each peer check signature from original author of the message, passing a message with invalid signature can result in ban for sender. It is not required however checking the content of the message itself.
```
SkidanovAlex, line 79: and send the message again -> and sends the message again
in ban for sender -> in ban for the sender

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+Each `RoutedMessage` is equipped with a time-to-live integer. If this message is not for us, we decrement by one this field, if value is 0, we drop this message, otherwise we forward it.
```
SkidanovAlex, line 81: "If this message is not for the node processing it, it decrement the field by one before routing it; if the value is 0, the node drops the message instead of routing it."

Generally, avoid using "we" and "us".

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### Routing back
+
+It is possible that node A is known to B but not the other way around. In case node A sends a request that requires a response to B, the response is routed back through the same path used to send the message from A to B. This are special types of messages.
```
SkidanovAlex, line 85: > This are special types of messages

Those are not really messages, right, those are some sort of routines?

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+### Synchronization
+
+Routing tables is kept up to date between the network with two mechanisms:
+
+- First sync: When two nodes connect with each other they exchange all information in their routing tables. More details in [Handshake](#Handshake).
+- Broadcasting: When a node learns new information that is added to the routing table it broadcast that information to all active node currently connected to it. More details in [Edge](#Edge) and [Announce Account](#Announce-account).
+
+## Security
+
+Messages exchanged between routed (both direct or routed) are cryptographically signed with sender private key. Nodes ID contains public key of each node that allow other peer to verify this messages. To keep secure communication most of this message requires some nonce/timestamp that forbid a malicious actor reuse a signed message out of context.
```
SkidanovAlex, line 99: between routed -> between peers?

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,119 @@
...
+
+In the case of routing messages each intermediate hop should verify that message hash and signatures are valid before routing to next hop.
+
+### Abusive behavior
+
+When a node A sends more than `MAX_PEER_MSG_PER_MIN` messages per minute to node B, it will be banned and unable to keep sending messages to it. This a protection mechanism against abusive node to avoid being spammed by some peers.
```
SkidanovAlex, line 105: This should be phrased as a recommendation, since it is not a requirement for the node to be protocol compliant, and other implementations might choose to use different protection mechanisms.

SkidanovAlex:
Lots of bikeshedding inline.

I think the level of detalization of the routing is too low. In particular, it doesn't quite explain what is the key in the routing table, how to send a message back via routing.

SkidanovAlex:
Also missing, but important:

- Definitions of `PeerId` and `PeerInfo`
- What port do we connect to? Does it matter, or the node can use any?

SkidanovAlex:
Also the spec appears to miss a critical piece of information that the messages are not just Borsh, they are 4 bytes length followed by the borshified message.

specs/ChainSpec/NetworkSpec.md
mfornet: Reworded

specs/ChainSpec/NetworkSpec.md
mfornet: Yes, but how it is implemented might affect other nodes in the network. Even if it is not mandatory, it is desired to know rules that will make my node being banned, so that I avoid violating them.

mfornet:
> * What port do we connect to? Does it matter, or the node can use any?

They can use any. They will learn about the ip/port used by other nodes through the PeerDiscovery mechanism.

SkidanovAlex:
> They can use any. They will learn about the ip/port used by other nodes through the PeerDiscovery mechanism.

What I mean is that it should be clearer in the spec :)

mfornet:
> Also the spec appears to miss a critical piece of information that the messages are not just Borsh, they are 4 bytes length followed by the borshified message.

Isn't that part of borsh standard?

- Decoding messages: https://github.com/nearprotocol/nearcore/blob/master/chain/network/src/peer.rs#L585

- Encoding message: https://github.com/nearprotocol/nearcore/blob/master/chain/network/src/peer.rs#L220

We let borsh library to encode/decode messages for us, without handling any 4 bytes length prefix.

SkidanovAlex:
> Isn't that part of borsh standard?

The way I see it, `send_message` pipes the output of `peer_message_to_bytes` into `framed`, which in turn uses `Codec`, which in turn prepends the four bytes. None of that appears to be part of borsh spec.

Though I actually don't see where we use the decoder from the `Codec`

mfornet:
> > Isn't that part of borsh standard?
> 
> The way I see it, `send_message` pipes the output of `peer_message_to_bytes` into `framed`, which in turn uses `Codec`, which in turn prepends the four bytes. None of that appears to be part of borsh spec.
> 
> Though I actually don't see where we use the decoder from the `Codec`

I see. We are using the decoder here: https://github.com/nearprotocol/nearcore/blob/master/chain/network/src/peer_manager.rs#L358

mfornet:
@SkidanovAlex just addressed all your comments, and added some pseudo-code. I plan to add a diagram for the handshake process, since it is kind of technical/complex and under-specified.

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,211 @@
...
+
+Edges are used to let other nodes in the network know that there is currently an active connection between a pair of nodes. See the definition of [this data structure](Messages.md#Edge).
+
+If the nonce of the edge is odd, it denotes an `Added` edge, otherwise it denotes a `Removed` edge. Each node should keep track of the nonce used for edges between every pair of nodes. Peer C believes that the peers A and B are currently connected if and only if the edge with the highest nonce known to C for them has an odd nonce.
+
+When two nodes successfully connect to each other, they broadcast the new edge to let other peers know about this connection. When a node is disconnected from other node, it should bump the nonce by 1, sign the new edge and broadcast it to let other nodes know that the connection was removed.
+
+A removed connection will be valid, if it contains valid information from the added edge it is invalidating. This prevents peers bump nonce by more than one when deleting an edge.
+
+When node A proposes an edge to B with nonce X, it will only accept it and sign it iff:
```
bowenwang1996, line 40:
```suggestion
When node A proposes an edge to B with nonce X, B will only accept it and sign it iff:
```

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,211 @@
...
+
+Messages exchanged between peers (both direct or routed) are cryptographically signed with sender private key. Nodes ID contains public key of each node that allow other peer to verify this messages. To keep secure communication most of this message requires some nonce/timestamp that forbid a malicious actor reuse a signed message out of context.
+
+In the case of routing messages each intermediate hop should verify that message hash and signatures are valid before routing to next hop.
+
+### Abusive behavior
+
+When a node A sends more than `MAX_PEER_MSG_PER_MIN` messages per minute to node B, it will be banned and unable to keep sending messages to it. This a protection mechanism against abusive node to avoid being spammed by some peers.
+
+## Possible attacks
```
bowenwang1996, line 199: Let's not discuss attacks here

specs/ChainSpec/NetworkSpec.md
```diff
@@ -0,0 +1,211 @@
...
+
+Create several ghost accounts with ghost connections among them and flood the network with them. Nodes will be calculating shortest path frequently and having many reachable nodes in the routing table increase the cost of such operation.
+
+## Implementation Details
+
+There are some issues that should be handled by the network layer but details about how to implement them are not enforced by the protocol, however we propose here how to address them.
+
+### Balancing network
+
+[Github comment](https://github.com/nearprotocol/nearcore/issues/2395#issuecomment-610077017)
```
bowenwang1996, line 211: We should not just post a link here if it is supposed to be part of the spec

bowenwang1996 (approved):

Issue #68, opened 2020-05-19
DID specification


ilblackdragon:
NEAR's accounts have few of the features of [DID specification](https://w3c.github.io/did-core) but not all.
We need a NEP that will provide a contract will full specification, a sample implementation and we need to file to add `did:near` to [registry](https://w3c-ccg.github.io/did-method-registry/).

Examples:
 - ERC725 DID: https://github.com/WebOfTrustInfo/rwot6-santabarbara/blob/master/topics-and-advance-readings/DID-Method-erc725.md
 - Blockstack DID: https://github.com/blockstack/stacks-blockchain/blob/stacks-1.0/docs/blockstack-did-spec.md

OntologyTech:
We have developed this specification to define a new decentralized identifier (DID) method for hosting DIDs on the NEAR blockchain, also referred to as NEAR DID, and facilitate developers to work with related contracts.

A reference implementation of NEAR DID scheme is currently under development.

Please refer to https://github.com/nearprotocol/NEPs/pull/100

gagdiez:
It seams that a [`did:near` is available in the registry](https://github.com/w3c/did-extensions/blob/315df2c944a37fcfd5971d896f2ffee84c6d91f2/methods/near.json#L4).

I understand that this issue was about making the `accounts` themselves did-compatible, but given that an entry to the registry exists, and that no NEP has been started to work on this in the past 5 years I will close this issue.

Please open again if still relevant

PR #100, opened 2020-07-23
add NEAR DID method(#68)

OntologyTech:
We have developed this specification to define a new decentralized identifier (DID) method for hosting DIDs on the NEAR blockchain, also referred to as NEAR DID, and facilitate developers to work with related contracts.

A reference implementation of NEAR DID scheme is currently under development.


text/0000-near-did-method.md
```diff
@@ -0,0 +1,170 @@
...
+The public key of the Near account's access key that was used to sign the transaction will be inserted into the `publicKey` property in the valid form and will be referred to in the `authentication` property.
+
+This method will return true if the new NEAR DID is registered successfully.
+
+## Read (Resolve)
+[read]: #read
+
+NEAR DID's associated DID document can be looked up by invoking the `getDocument` method of the registry.
+
+To ensure the smart contract invocation result is trustworthy, the client could query a certain number of nodes and then compare the return values.
```
bowenwang1996, line 76: I don't think it's necessary. If the client doesn't trust some public rpc server they can use their own.

OntologyTech:
> Two questions:
> 
> * What is the specification of the DID document? How large can it be?

Link to the NEAR DID document specification: https://github.com/ontology-tech/DID-spec-near/blob/master/NEAR/NEAR-DID-spec.md
The reference implementation is available here: https://github.com/ontology-tech/DID-NEAR-rust
Generally speaking, the size of a NEAR DID document depends on the number of the bound verification methods. A greater number of verification methods would imply a larger document size.

> * How does composability work? Can you give some examples of how a DID is used?
 
Could you describe in what context you are referring to 'composability'? 
For instance, on Ontology, the DID framework ONT ID serves to support the account system that can be used to identify individuals and entities and record and validate their actions on-chain. There is value in establishing this distinction from traditional wallet addresses, since this allows for greater logical clarity when it comes to more complex operations as traditional wallet accounts are normally linked to assets and asset related operations. 
A more specific implementation would be something like issuing and verifying credentials for entities which could denote a whole variety of things based on the subject and scenario.

W3C lays out certain potential use cases for distributed identifiers. Follow this link for more details: https://www.w3.org/TR/did-use-cases/

text/0000-near-did-method.md
OntologyTech: > I don't think it's necessary. If the client doesn't trust some public rpc server they can use their own.

Sure, this definitely isn't a necessity, just a possible solution. Clients may or may not run their own server. It is up to them to use the server of their choice.

bowenwang1996:
Maybe I missed something, but it seems that the question of authorization is not well explained. It seems that the contract will be deployed on every account that participates in the system, so what prevents the account owner from modifying the data arbitrarily (the document and service endpoint for example)? Or is there some incentive that would prevent them from doing so?

OntologyTech:
> Maybe I missed something, but it seems that the question of authorization is not well explained. It seems that the contract will be deployed on every account that participates in the system, so what prevents the account owner from modifying the data arbitrarily (the document and service endpoint for example)? Or is there some incentive that would prevent them from doing so?

Good question. So the thing is, a single contract is deployed on the chain, and not one for each user that is part of the system. A new document is generated in the contract for each new DID. DID subjects can authorize other DID subjects to update their DID documents by assigning the `controller` property. Whether an owner can perform such arbitrary or malicious actions relates to  governance model specifics.

bowenwang1996:
> So the thing is, a single contract is deployed on the chain

I see. But it doesn't look like the case from your contract implementation. I suppose the contract should store the DIDs?

OntologyTech:
That is correct. The DID documents for each ID instance are stored within the contract. 
You can refer to the implementation here for better clarity: https://github.com/ontology-tech/DID-NEAR-rust

frol:
@OntologyTech Do you still have a use-case for this proposal? Would you like to work on it and shape it together?

ori-near:

PR #108, opened 2020-08-26
feat: add more spec to transaction processing and errors

bowenwang1996:
Add more spec to transaction processing and the errors that they generate.


specs/RuntimeSpec/Actions.md
```diff
@@ -32,35 +43,48 @@ _Requirements:_
 - sets Account `storage_usage` to `account_cost` (genesis config)
 - sets Account `storage_paid_at` to the current block height
```
evgenykuzyakov, line 44: This is no more

specs/RuntimeSpec/Actions.md
```diff
@@ -32,35 +43,48 @@ _Requirements:_
...
     pub code: Vec<u8>, // a valid WebAssembly code
 }
 ```
 
-## FunctionCallAction
-
 _Requirements:_
 
-- _`tx.public_key` to be `AccessKeyPermission::FullAccess` or `AccessKeyPermission::FunctionCall`_
+- _`tx.signer_id` to be equal to `receiver_id`_
```
evgenykuzyakov, line 77: In runtime we used a concept of an actor_id. This is the temporary account ID that is either the `predecessor_id` (not `signer_id`) or the current account ID. By default it's the predecessor, but if the first action is `CreateAccount` and the account doesn't exist, it changes the actor to current account ID `receiver_id`

specs/RuntimeSpec/Actions.md
```diff
@@ -32,35 +43,48 @@ _Requirements:_
...
+- If the action tries to create an account that is neither a top-level account or a subaccount of `predecessor_id`,
+the following error will be returned
+```rust
+/// A newly created account must be under a namespace of the creator account
+CreateAccountNotAllowed { account_id: AccountId, predecessor_id: AccountId },
+```
 
-- _`tx.signer_id` to be equal to `receiver_id`_
-- _`tx.public_key` to be `AccessKeyPermission::FullAccess` for the `singer_id`_
+NOTE: for the all subsequent actions in the transaction the `signer_id` becomes `receiver_id` until [DeleteAccountAction](#DeleteAccountAction). It allows to execute actions on behalf of the just created account.
```
evgenykuzyakov, line 65: Found this note. good. See comment below for `actor_id`

Instead of `signer_id` you should use `predecessor_id`. Actions can be created through a promise API without extra transaction. Since this section talks about Actions in general context and not only actions within a transaction, then we should use `predecessor_id` whenever possible and forget about `signer_id`

specs/RuntimeSpec/Actions.md
```diff
@@ -32,35 +43,48 @@ _Requirements:_
...
+- _`tx.signer_id` to be equal to `receiver_id`_
+- _`tx.public_key` to be `AccessKeyPermission::FullAccess` for the `signer_id`_
 
-Calls a method of a particular contract. See [details](./FunctionCall.md).
+**Outcome**:
+- sets the contract code for account
+
+**Errors**
+
+- If state or storage is corrupted, it may return `StorageError`. Otherwise this action should not lead to errors.
```
evgenykuzyakov, line 85: I don't think spec should talk about storage errors. Until we get to challenges, you shouldn't be able to arrive into inconsistent state

specs/RuntimeSpec/Actions.md
```diff
@@ -91,13 +114,19 @@ pub struct TransferAction {
...
 
-- _`tx.signer_id` to be equal to `receiver_id`_
 - _`tx.public_key` to be `AccessKeyPermission::FullAccess` for the `singer_id`_
 
+**Outcome**:
+- transfers amount specified in `deposit` from `tx.signer` to a `tx.receiver_id` account
+
+**Errors**:
+- If the deposit amount plus the existing amount on the receiver account exceeds `u128::MAX`,
+a `StorageInconsistentState("Account balance integer overflow")` error will be returned.
```
evgenykuzyakov, line 126: I think you can take a look at `verifiers.rs` and tests to see what errors are thrown during transaction validation.

There are some balance overflow, but it's not inconsistent state error.

There are also execution time errors, such as `AccountExists` that are thrown in `actions.rs`

bowenwang1996: https://github.com/nearprotocol/nearcore/blob/1c1c019613e4ccb47ea6dc5a591f913af8022998/runtime/runtime/src/actions.rs#L293

specs/RuntimeSpec/Actions.md
bowenwang1996: I think it is part of the spec because it is possible that we return `StorageError` if state is somehow corrupted.

specs/RuntimeSpec/Actions.md
evgenykuzyakov: That's mostly only possible for challenges. We wrapped most errors in order to make sure when the challenge on malicious state transition arrives (where the previous root is also malicious), we don't crash the node.

specs/RuntimeSpec/Actions.md
```diff
@@ -106,18 +128,51 @@ pub struct StakeAction {
...
 ```
+
 **Outcome**:
+- A validator proposal that contains the staking public key and the staking amount is generated and will be included
+in the next block.
+
+### Errors
+
+**Validation Error**:
+- If the `public_key` is not an ed25519 key, the following error will be returned:
```
evgenykuzyakov, line 139:
```suggestion
- If the `public_key` is not an ristretto compatible ed25519 key, the following error will be returned:
```

specs/RuntimeSpec/Actions.md
```diff
@@ -126,30 +181,96 @@ pub struct AddKeyAction {
 }
 ```
 
-## DeleteKeyAction
+**Outcome**:
+- Associates an [AccessKey](AccessKey) with a `public_key` provided.
```
evgenykuzyakov, line 185: It also adds this AccessKey to the receiver's account. 
```suggestion
- Adds a new [AccessKey](AccessKey) to the receiver's account and associates it with a `public_key` provided.
```

specs/RuntimeSpec/Actions.md
```diff
@@ -126,30 +181,96 @@ pub struct AddKeyAction {
...
+
+- If the length of some method name exceed `max_length_method_name`, which is a genesis parameter (current value is 256),
+the following error will be returned 
+```rust
+/// The length of some method name exceeded the limit in a Add Key action.
+AddKeyMethodNameLengthExceeded { length: u64, limit: u64 },
+```
 
-_Requirements:_
+- If the sum of length of method names exceeds `max_number_bytes_method_names`, which is a genesis parameter (current value is 2000),
```
evgenykuzyakov, line 205:
```suggestion
- If the sum of length of method names (with 1 extra character for every method name) exceeds `max_number_bytes_method_names`, which is a genesis parameter (current value is 2000),
```

specs/RuntimeSpec/Actions.md
```diff
@@ -126,30 +181,96 @@ pub struct AddKeyAction {
...
+the following error will be returned
+```rust
+/// The total number of bytes of the method names exceeded the limit in a Add Key action.
+AddKeyMethodNamesNumberOfBytesExceeded { total_number_of_bytes: u64, limit: u64 }
+```
+
+**Execution Error**:
+- If an account tries to add an access key that already exists, the following error will be returned
```
evgenykuzyakov, line 213:
```suggestion
- If an account tries to add an access key with a given public key, but an existing access key with this public key already exists, the following error will be returned
```

specs/RuntimeSpec/Actions.md
```diff
@@ -126,30 +181,96 @@ pub struct AddKeyAction {
...
+InvalidAccountId { account_id: AccountId },
+```
+
+- If this action is not the last action in the action list of a receipt, the following error will be returned
+```rust
+/// The delete action must be a final action in transaction
+DeleteActionMustBeFinal
+```
+
+- If the account still has locked balance, the following error will be returned
```
evgenykuzyakov, line 269:
```suggestion
- If the account still has locked balance due to staking, the following error will be returned
```

PR #111, opened 2020-09-08
Define client versioning through interfaces not protocol

nearmax:
Our current client versioning is meant to be following semver, but semver is defined around the notion of backwards compatibility of interfaces, not protocol. Our versioning scheme breaks a few things today and will potentially break many things in the future. This proposal redefines client versioning through interfaces.

## Examples of issues
The main cause for all problems is that we reserve major version update for protocol changes only. The following are the examples when it is going to break already today.

### Bridge, Wallet, Explorer (and anyone who uses client RPC)

When we know that client has bumped its minor version we expect it to be backwards compatible. However, in the current terminology it just means it is not a protocol change. But it can be Rust structure change or RPC change, which would break the bridge unexpectedly. Bridge team, as well as third-party developer should not be expected to change their code when minor version bump is happening.

### Indexer (and anyone who uses client as a library)

Indexer is currently using client as a library, which means it links it as Rust dependency, where indexer is its own executable. If @khorolets can compile indexer today, it does not mean he will be able to compile it tomorrow even if he does not change the code of the indexer (unless indexer has a Cargo.lock file but we shouldn't expect it from all developers). This happens because cargo interprets minor version bump as a backwards compatible change to the interfaces. Here is specific scenario:
* @khorolets writes indexer that depends on near client version `1.2.0`. It compiles;
* Someone makes a non-protocol breaking change to near client that breaks any of the interfaces (RPC, actors, internal Rust structures), and publishes `1.3.0`;
* @khorolets tries to recompile indexer which causes `cargo` to pull `1.3.0` client which breaks compilation.

This scenario is similar to the issue that @evgenykuzyakov has encountered here: https://github.com/near/near-sdk-rs/issues/219 The main point is that semver is interpreted very strictly by all tools in terms of interface backward compatibility, not protocol backward compatibility.

Note, Cargo.lock file applies only to [executables and is useless for libraries](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries), so we cannot fix it on the side of client library by adding the lock.

---

See proposed changes in the file.


specs/ChainSpec/Upgradability.md
```diff
@@ -20,11 +20,18 @@ type ProtocolVersion = u32;
...
+   - Anything that requires database migration, even if client self-migrates its database;
+   - Error format of the RPC, including compilation errors. Which implies that major updates to Wasm runtime can cause version bump, even if they do not change the protocol;
+   - The way publicly accessible actors handle messages;
+   - When client changes its behavior after X's block -- delayed non-backward compatible API and behavior change, which includes most of the protocol changes.
+- MINOR version defines backward compatible API and behavior changes, this mostly applies to extensions Non-intuitive examples of when MINOR version upgrade is needed:
+   - New RPC endpoints;
+   - Additional optional storage columns, e.g. on-disk cache.
```
bowenwang1996, line 33: But it still requires a migration to be adopted? Also I don't fully understand why database migration implies a major version change.

frol: > why database migration implies a major version change.

If you cannot interchangeably use the binaries, those should have different major version. If you perform a migration and unable to run the previous version of binary, they are not compatible 🤷

frol (approved):

khorolets (approved):

specs/ChainSpec/Upgradability.md
bowenwang1996: I see. Makes sense

ilblackdragon: What is the point of patch then if minor version should be fully compatible for running?
From what I read it says minor is any backward compatible change but doesn't mean that you can still run old binary - because if some feature was added it's possible you can't reverse anymore due to some changes even in installable software.

nearmax: Yes, good point. AFAIU the difference between patch and minor version could be purely advisory -- patch means the users should upgrade their node (we fixed a bug or improved performance), while minor version means user don't have to upgrade their node, but they can. Minor versions can introduce deprecation warnings to old interfaces, which might cause some user not wanting to upgrade.

bowenwang1996:
This is unrelated to the protocol and should be discussed on https://github.com/near/nearcore

PR #117, opened 2020-09-29
Fees spec

evgenykuzyakov:
Adding spec for runtime fees https://nomicon-pr-117.onrender.com/RuntimeSpec/Fees/Fees.html

### DIFF
--- a/specs/RuntimeSpec/Fees/Fees.md
+++ b/specs/RuntimeSpec/Fees/Fees.md
@@ -3,7 +3,7 @@
 Runtime fees are measured in Gas. The price of gas is determined from the previous block header `block_header.inner_rest.gas_price`.
 Gas price will be discussed separately.
 
-When a transaction is converted into a Receipt. The signer account is charged for the full cost of the transaction.
+When a transaction is converted into a receipt, the signer account is charged for the full cost of the transaction.
 
 The total transaction fee is the sum of the following:
 - A fee for creation of the receipt
--- a/specs/RuntimeSpec/Fees/Fees.md
+++ b/specs/RuntimeSpec/Fees/Fees.md
@@ -1,9 +1,9 @@
 # Runtime Fees
 
-Runtime fees are measured in Gas. The price of gas is determined from the previous block header `block_header.inner_rest.gas_price`.
-Gas price will be discussed separately.
+Runtime fees are measured in Gas. Gas price will be discussed separately.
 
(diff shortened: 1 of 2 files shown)


specs/RuntimeSpec/Fees/Fees.md
```diff
@@ -0,0 +1,138 @@
+# Runtime Fees
+
+Runtime fees are measured in Gas. The price of gas is determined from the previous block header `block_header.inner_rest.gas_price`.
+Gas price will be discussed separately.
```
bowenwang1996, line 4: should we provide a link here

specs/RuntimeSpec/Fees/Fees.md
```diff
@@ -0,0 +1,138 @@
...
+
+When a transaction is converted into a Receipt. The signer account is charged for the full cost of the transaction.
```
bowenwang1996, line 6:
```suggestion
When a transaction is converted into a receipt, the signer account is charged for the full cost of the transaction.
```

bowenwang1996, line 6: should we mention that what is charged at this moment is potentially more than the transaction fee due to pessimistic gas price and the overcharged part is later refunded?

specs/RuntimeSpec/Fees/Fees.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+It has `signer_id != receiver_id` so it will use `send_not_sir` for send fees.
+
+It contains 4 actions with 2 actions that requires to compute number of bytes.
+We assume `code` in `DeployContractAction` contains `128000` bytes. And `FunctionCallAction` has
+`method_name` with length of `3` and `args` length of `26`, so total of `29`.
+
+First let's compute the the amount that will be burned immediately for sending a receipt.
+```rust
```
bowenwang1996, line 113: I think we usually use python syntax for pseudocode

specs/RuntimeSpec/Fees/Fees.md
evgenykuzyakov: Once it exists :)

specs/RuntimeSpec/Fees/Fees.md
evgenykuzyakov: Yes, but it's part of gas price discussion

bowenwang1996 (approved):

PR #118, opened 2020-10-02
Minor fixes to markdown math

adlerjohn:
Fix the consensus spec page:
- Use `\operatorname` for function names, as is tradition.
- Use `\mid` instead of `|`.
- Fix broken link to Doomslug paper.


bowenwang1996 (approved):
Thanks

PR #129, opened 2020-10-31
Contract Self-Identification NEP

luciotato:
# Contract Self-Identification ([NEP-129](https://github.com/nearprotocol/NEPs/pull/129))

### Initial Draft

see: https://github.com/luciotato/NEPs/blob/contract-self-identification/specs/Standards/Contracts/SelfIdentification.md

### DIFF
--- a/specs/Standards/Contracts/SelfIdentification.md
+++ b/specs/Standards/Contracts/SelfIdentification.md
@@ -1,15 +1,17 @@
-# Contract Self-Identification ([NEP-xx](https://github.com/nearprotocol/NEPs/pull/xx))
+# Contract Self-Identification ([NEP-129](https://github.com/nearprotocol/NEPs/pull/129))
 
-Version `0.0.1`
+Version `0.1.0`
 
 ## Summary
 [summary]: #summary
 
-A standard interface for contract allowing self-identification, version information, supported standard interfaces and audit verification.
-
-This NEP also includes author identification and gas%-benefit-sharing mechanism, in order to create incentives for smart contract authors and lure developers into the NEAR ecosystem.
+A standard interface for contract allowing self-identification, version information, supported standard interfaces, author and auditor.
 
 ## Changelog
+
+### `0.1.0`
+
+- Heavy simplification. Only contract identification.
 
 ### `0.0.1`
 
@@ -32,11 +34,10 @@ A contract implementing this standard would allow users to:
 - Determine name, version & source code of the contract, e.g. `name:"fungible-token", version:"1.0.0"`
 - Get a list of standards this contract satisfies, e.g. `standards:["NEP-21","NEP-211"]`
 - Get the url of the contract code, e.g. `source:"http://github.com/author/contract"`
-- Get audit signatures for the current contract hash
-- Get exact information on the total funds managed
-- Benefit deployers and contract authors with gas-benefits
+- Get the author near account 
+- Get the auditor near account (possibly for contract hash validation via aditor's provided mechanisms)
 
-To satisfy this NEP the contract must implement 4 (four) methods: 
+To satisfy this NEP the contract must implement 1 (one) method: 
 
 1. `get_contract_info():string as ContractInfo` returning a JSON structure as described below (typescript):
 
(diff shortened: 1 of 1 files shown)


bowenwang1996 (commented):
I feel like the part about auditing can potentially be its own NEP as there are a lot of potential problems there. Without a proper reputation system it seems that having the auditor's signature isn't really useful.

evgenykuzyakov:
Why does it have to be on-chain? 
My current thinking is to be able to just link a metadata JSON file with a hash of it. It can either be stored on IPFS or github link with a commit hash, so it doesn't change. Then the metadata file can have the hash of the contract to link it back.

When you initialize a contract you pass a link to the metadata into the initialization argument and let's say the contract shouldn't change this link.

The metadata can contain all the required info. Audits can also go off-chain and link to a particular metadata file.

luciotato:
I'm open to all suggestions. 
@evgenykuzyakov Do you think on-chain would be too expensive?
Please expand on "why not on chain", you've a ton of knowledge and it will be very useful if you can expand on that.

I'm putting all on-chain so you don't depend on nothing more than the chain, I'm with the idea that the more we use the chain for, the better. From DevX POV is better because you don't have to deal with other protocols/storage/concepts/documentation. If we use Github or IPFS, Devs need to read about/understand those services and also your code needs to include API's/libs to access those services. If everything is on-chain, you can get the data with the same near-cli/api you're already using.
If we add info/registry about auditors (as a DAO/community managed registry) then you can also check auditor reputation on-chain, you can even automatize that without leaving NEAR (e.g. "Get me a NEP-21 contract holding more than 1 million NEAR audited by someone with >98% reputation")

luciotato:
@bowenwang1996 Yes, the user must previously select auditors she trusts. But it's chicken-and-egg, this NEP could make the existence of a reputation system for auditors a necessity.

evgenykuzyakov:
My main reason for why not on-chain is that the most information doesn't have to be on-chain.

For example the contracts don't care about audits information, so it doesn't have to be on-chain so long as we can link this information to the contract. It can be linked if each audit has the account-id and the contract code hash for what's reviewed.

If we can securely link metadata file and the deployed contract, then we can put all required info into the metadata file.

If one contract wants to verify on-chain that the other contract satisfies the standard, by relying on an audit. This can be done if there is an auditor contract that whitelists contracts instead. Then you can rely on this whitelist to give you a verification that the given contract satisfies requirements. It's similarly how the Staking pool contracts are whitelisted for lockup contracts to delegate.

Eventually, all metadata can live on-chain and the link can just point to a contract on chain. But for now the audits have overhead and there are no need to validate them on-chain.

luciotato:
Heavy simplification: Based on reasonable feedback I've simplified the NEP limiting it to just contract self-identification and reducing on-chain required space to a minimum.

vgrichina:
@luciotato should we also add URL with canonical web app for given contract? This can be useful for wallet a lot. /cc @kcole16

luciotato:
Good idea @vgrichina. Done.

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,208 @@
...
+#[serde(crate = "near_sdk::serde")]
+#[allow(non_snake_case)]
+pub struct NEP129Response {
+    pub dataVersion:u16,
+    pub name:String,
+    pub version:String,
+    pub developersAccountId:String,
+    pub source:String,
+    pub standards:Vec<String>,
+    pub webAppUrl:String,
```
oysterpack, line 110: `webAppUrl` should be optional (`Option<String>`) because not all contracts will have a web app frontend. Optional fields should be always be explicit in the code.

luciotato: Ok. It applies also for auditor_account_id

luciotato: done. also for auditor_account_id

ilblackdragon: I would actually suggest to maintain audit map somewhere outside of the contract itself.
There can be multiple audits for the same contract, audits should contain some extra info (like advisories) and audits can frequently come in after the contract is deployed. Also it's ideally that audit itself shouldn't modify the commit hash so can rely on it, which this would potentially lead to.

We actually have an audit registry built (but not launched on MainNet yet) - explained more here - https://gov.near.org/t/audit-registry-add-accountability-and-visibility-of-which-applications-were-reviewed-and-by-whom/234

robert-zaremba:
I started a discussion with an alternative approach: interface registry: https://github.com/near/NEPs/discussions/154

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,207 @@
+# Contract Self-Identification ([NEP-129](https://github.com/nearprotocol/NEPs/pull/129))
```
mfornet, line 1: Add proper header per [NEP-1](https://github.com/near/NEPs/blob/master/neps/nep-0001.md). See [template](https://github.com/near/NEPs/blob/master/nep-0000-template.md).

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,207 @@
...
+[motivation]: #motivation
+
+It's almost impossible to determine what contract is deployed to an account without out-of-band data.
+
+A contract satisfying this standard can be identified and verified without resorting to pre-known data or data acquisition mechanisms other than blockchain standard access.
+
+Prior art:
+- TBD
+
+## Guide-level explanation
```
mfornet, line 31:
```suggestion
## Specification
```

specs/Standards/Contracts/SelfIdentification.md
mfornet: Add Rationale and Alternative version

mfornet: Add Copyright?

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,207 @@
...
+
+* use `auditorAccountId` to check the contracts code hash and audit info (depending on the auditor's mechanism to do that)
+
+If all the above is correct, her code adds the contract as an audited NEP-21 Token and the user can interact with it via `contractInfo.webAppUrl`
+
+
+## Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+### Reference Implementations:
```
mfornet, line 81:
```suggestion
## Reference Implementations
```

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,207 @@
...
+   version: string = "0.0.1"; //contract-code semver
+   source: string = ""; //contract source code URL, e.g. http://github.com/author/contract
+   standards: string[] = []; // standards this contract satisfies
+   developersAccountId: string=""; //near account of the author of the code
+   webAppUrl?: string; //web URL where the DApp can be accessed e.g. http://Narwallets.com/div-pool/
+   auditorAccountId?: string=""; //near account of the auditor of the code
+}
+```
+
+### Details on the fields:
```
mfornet, line 60:
```suggestion
### Details on the fields
```

Remove two points for headers

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,207 @@
...
+        developersAccountId:DEVELOPERS_ACCOUNT_ID.into(),
+        source:"https://github.com/Narwallets/diversifying-staking-pool".into(), 
+        standards:vec!("NEP-129".into()),  
+        webAppUrl:self.web_app_url.clone(),
+        auditorAccountId:self.auditor_account_id.clone()
+    }
+}
+```
+
+####  ts/pseudocode:
```
mfornet, line 129: Suggestion, consider moving, ts implementation to assets, and link it here (mostly if it is pseudocode).

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,207 @@
...
+        dataVersion:1,
+        name: CONTRACT_NAME.into(),
+        version:CONTRACT_VERSION.into(),
+        developersAccountId:DEVELOPERS_ACCOUNT_ID.into(),
+        source:"https://github.com/Narwallets/diversifying-staking-pool".into(), 
+        standards:vec!("NEP-129".into()),  
+        webAppUrl:self.web_app_url.clone(),
+        auditorAccountId:self.auditor_account_id.clone()
+    }
+}
```
mfornet, line 126: I found the current code distracting. I think either pseudo code is better (potentially moving this implementation to assets) and making sure it compiles (if copy pasted properly in the project).

Otherwise, let's make sure the code is properly formatted, get rid of global variables given they are used only once, properly update comments so they are more useful.

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,207 @@
...
+/// NEP-129 get information about this contract
+/// returns JSON string according to [NEP-129](https://github.com/nearprotocol/NEPs/pull/129)
+pub fn get_contract_info(&self) -> NEP129Response {
+    return NEP129Response {
+        dataVersion:1,
+        name: CONTRACT_NAME.into(),
+        version:CONTRACT_VERSION.into(),
+        developersAccountId:DEVELOPERS_ACCOUNT_ID.into(),
+        source:"https://github.com/Narwallets/diversifying-staking-pool".into(), 
+        standards:vec!("NEP-129".into()),
```
mfornet, line 122:
```suggestion
        standards:vec!["NEP-129".into()],  
```

specs/Standards/Contracts/SelfIdentification.md
```diff
@@ -0,0 +1,207 @@
...
+## Changelog
+
+### `0.1.0`
+
+- Heavy simplification. Only contract identification.
+- added webAppUrl per @vgrichina suggestion
+
+### `0.0.1`
+
+- Initial Proposal
```
mfornet, line 19: Remove. Add vgrichina as coauthor if necessary.

mfornet (changes requested):
Left some comments with regard to the format of the document (syntax only)

mfornet:
There are two alternative to this NEP: #154 and #275. Both of them are very similar, they propose to create an on-chain registry with the information of which NEP is supported by each contract.

Advantage of registry vs builtin self-identification:
- Add NEP discoverability for inmutable contract in hindsight.

Con:
- A trustless mechanism should be created to add and fetch elements to the registry. See approach suggested in #275 (using endorsements)

frol:
I am closing this NEP as it seems that it was superseded with NEP #330 which was later extended with #351. There are still some fields missing that were proposed in this NEP, so if anyone wants to extend NEP #330, feel free to create a NEP extension (PR with changes to the existing NEP document).

PR #132, opened 2020-11-04
recalibrate inflation

bowenwang1996:
# Summary

This proposal changes inflation calculation to be based on the during of the epoch in seconds instead of relying on number of blocks that are expected to be produced in a year.

# Motivation

Currently on mainnet blocks are produced usually at between 1.2 and 1.3 per second, which is great. However, when we calculate inflation per epoch, we use the following formula (note: this excludes the tokens burnt in transaction fees):
```
inflation = total_supply * yearly_inflate_rate * epoch_length / expected_num_blocks_per_year
```
where `expected_num_blocks_per_year` is a genesis parameter currently set to 31536000, i.e, a block every second. Since blocks are produced faster that, expected_num_blocks_per_year is smaller than the actual value, which means that, at the current rate, we will have ~20% more inflation than expected.

# Guide-level explanation

Calculating the inflation based on blocks isn’t the most reliably way to begin with. Even though currently epochs are shorter than expected, they could, in some cases, be longer than expected if there are a few validators offline for the entire duration of the epoch. To address this, I propose that we change the inflation calculation based on time elapsed instead of number of blocks. More specifically, this proposal only requires changing the inflation calculation to
```
inflation = total_supply * yearly_inflation_rate * duration_of_epoch_in_seconds / num_seconds_in_a_year
```
This is essentially still a first-order Taylor approximation of the target inflation rate, in the similar vein as the original calculation. However, it is more accurate and is less subject to the perturbation of the epoch time.

# Reference-level explanation

* Duration of an epoch can be easily calculated when we finalize the epoch by subtracting the timestamp of the first block of the epoch from that of the last block.
* Upgrade process is relatively straightforward. We simply switch to the new way of calculating inflation when the protocol version changes.
* For simplicity, we set `num_seconds_in_a_year` as a constant `31536000`. For leap year this will be not entirely accurate but the difference is negligibly small.


bowenwang1996:
@frol can we merge this one? Needs an approval

bucanero (approved):

frol (approved):

PR #158, opened 2021-02-18
Update Fungible Token entry to match NEP-141

mikedotexe:
This is the update to the storage management and account registration standards in addition to fungible tokens based on NEP-141.


oysterpack (changes requested):
some corrections are needed - see comments

specs/Standards/Storage.md
```diff
@@ -0,0 +1,206 @@
...
+Version `1.0.0`
+
+## Summary
+
+This standard handles state storage when accounts add and remove data stored in the state of a contract. It allows accounts to:
+
+1. Check an account's storage "balance."
+2. Determine the minimum storage needed to add account information such that the account can interact as expected in a contract.
+3. Add storage for an account; either their own or another account.
+4. Withdraw storage by removing associated account data from the contract.
```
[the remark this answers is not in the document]
chadoh: Updated to 

> Withdraw some or all of a storage deposit by removing associated account data from the contract and then making a call to remove unused deposit.

specs/Standards/Storage.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Motivation
+
+NEAR uses storage staking (different than state rent) as the mechanism to pay for storage on the blockchain. When storage is removed, the staked Ⓝ is released. Developers must take this into consideration when crafting smart contracts that store data on-chain. If no restrictions are enforced in a permissionless system, any account may add key/value pairs that bloat the storage. Since the account containing the smart contract must keep enough Ⓝ to pay for the storage, developers must ensure the state storage is "paid for" and "refunded" as the storage changes. This standard provides a generic approach to handle these concerns.
```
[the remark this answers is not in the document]
chadoh: Updated to

> developers must ensure the state storage is "paid for" as the storage increases, and can refund such storage deposits later if data is removed

specs/Standards/Storage.md
```diff
@@ -0,0 +1,203 @@
...
+
+- A previous fungible token standard ([NEP-21](https://github.com/near/NEPs/pull/21)) highlighting how [storage was paid](https://github.com/near/near-sdk-rs/blob/1d3535bd131b68f97a216e643ad1cba19e16dddf/examples/fungible-token/src/lib.rs#L92-L113) for when increasing the allowance of an escrow system.
+
+## Guide-level explanation
+
+We should be able to do the following:
+
+- Have an account sign a transaction, attaching a deposit of Ⓝ that will cover *their own* storage for a smart contract.
+- Have an account sign a transaction, attaching a deposit of Ⓝ that will cover the storage for another account to use a smart contract.
+- Allow an account to remove all their associated data stored on a contract and receive a refund for that storage in the form of Ⓝ.
```
chadoh, line 28: Updating this to:

> Allow an account to request refunds in Ⓝ for excess storage deposits on a given contract.

Does that seem right to you, @oysterpack?

specs/Standards/Storage.md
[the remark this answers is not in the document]
chadoh: I've also updated the "Future Possibilities" section with:

> - This standard avoids specifying how to close/unregister an account.
> - Instead of having developers use constants for the storage price per byte, perhaps this can be derived and instructions can be modified in this standard regarding how to calculate the value returned for `storage_minimum_balance`.
> - Ideally, contracts will update `available_balance` for all accounts every time the NEAR blockchain's configured storage-cost-per-byte is reduced.

specs/Standards/Storage.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Account withdraws storage deposit
+
+Alice and Bob decide to withdraw their storage deposit from the `mochi` contract.
+
+**Assumptions**
+
+- Alice's account is `alice`.
+- Bob's account is `bob`.
+- Alice and Bob's account balances are `0`.
```
chadoh, line 119: @oysterpack I've updated this whole example to avoid making it seem like this standard covers account removal/closure. Here's the updated version. Let me know if you have any feedback.

> #### Accounts withdraw excess storage deposit
> 
> Alice and Bob decide to withdraw some unused storage deposit from the `mochi` contract.
> 
> **Assumptions**
> 
> - Alice's account is `alice`.
> - Bob's account is `bob`.
> - Both Alice's and Bob's accounts have more deposited than they are using.
> 
> **High-level explanation**
> 
> 1. Knowing that she registered Bob's account and believing that this entitles her to receive a refund for that deposit, Alice views her & Bob's storage balances and sees each have extra.
> 2. Alice issues a transaction to withdraw her own excess deposit.
> 3. Alice attempts to withdraw Bob's excess deposit, but cannot.
> 4. Bob issues a transaction to withdraw his own excess deposit.
> 
> **Technical calls**
> 
> 1. Alice queries `mochi::storage_balance_of({ "account_id": "alice" })` and `mochi::storage_balance_of({ "account_id": "alice" })`.
> 
>    Checking her own account with NEAR CLI:
> 
>        near view mochi storage_balance_of '{"account_id": "alice"}'
> 
>    Response:
> 
>        View call: mochi.storage_balance_of({"account_id": "alice"})
>        {
>          total: '7050000000000000000000',
>          available: '2350000000000000000000'
>        }
> 
>    Checking Bob's account:
> 
>        near view mochi storage_balance_of '{"account_id": "bob"}'
> 
>    Response:
> 
>        View call: mochi.storage_balance_of({"account_id": "bob"})
>        {
>          total: '4700000000000000000000',
>          available: '2350000000000000000000'
>        }
> 
> 2. Alice calls `mochi::storage_withdraw({"amount": "2350000000000000000000"})` for her own account. NEAR CLI command:
> 
>        near call mochi storage_withdraw '{"amount": "2350000000000000000000"}' --accountId alice
> 
> 3. Alice realizes that `storage_withdraw` does not allow specifying the account to withdraw from. She has withdrawn all she can from her own account. When she re-checks `mochi::storage_balance_of({"account_id": "alice"})`, it indicates she has zero available balance. This is because storage withdrawal is only for the predecessor account that has signed the transaction. Alice cannot withdraw for Bob.
> 
> 4. Bob issues the same transaction as Alice did in step 2.
> 
>        near call mochi storage_withdraw '{"amount": "2350000000000000000000"}' --accountId bob

cc @mikedotexe

specs/Standards/Storage.md
```diff
@@ -0,0 +1,206 @@
...
+
+1. Alice calls `mochi::storage_withdraw({"amount": "2350000000000000000000"})`. NEAR CLI command:
+   
+   near call mochi storage_withdraw '{"amount": "2350000000000000000000"}' --accountId alice
+
+2. She suddenly remembers she paid for Bob's account registration, and checks to see if she can withdraw that balance as well, calling `mochi::storage_balance_of({"account_id": "alice"})` which indicates she has zero balance. This is because storage withdrawal is only for the predecessor account that has signed the transaction. Alice cannot withdraw for Bob.
+
+3. Bob issues the same transaction as Alice did in step 1.
+
+    near call mochi storage_withdraw '{"amount": "2350000000000000000000"}' --accountId bob
```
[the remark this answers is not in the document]
chadoh: See [updated example](https://github.com/near/NEPs/pull/158#pullrequestreview-595936016)

specs/Standards/Storage.md
```diff
@@ -0,0 +1,206 @@
...
+// * The account sending the transaction must have enough Ⓝ to cover the deposit.
+// If `account_id` is omitted, the sender's account will receive the attached deposit for storage. 
+// Returns the AccountStorageBalance structure
+function storage_deposit(
+    account_id: string|null
+): string {}
+
+// Withdraws a specified amount of Ⓝ for a given account that was previously reserved for storage.
+// Requirements:
+// * `amount` must be less than or equal to the storage balance for the sending account.
```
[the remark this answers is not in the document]
chadoh: Updated to: "`amount` must be less than or equal to the available storage balance for the sending account."

Given that this is a plain English description, I think "available storage balance" makes more sense, since "storage balance" is the concept being modified by "available." Feel free to unresolve this comment if you disagree.

specs/Standards/Storage.md
mikedotexe: The idea of withdrawing is removing all their data from the contract so they can wash their hands of any involvement in the contract.

specs/Standards/Storage.md
```diff
@@ -77,7 +77,7 @@ Alice registers her account on a fungible token contract.
 
        {
          total: '2350000000000000000000',
-         available: '2350000000000000000000'
+         available: '0'
```
mikedotexe, line 81: This is not the behavior I'm seeing. Just tried it (again) now.

```bash
mike@n  ~/near/near-sdk-rs/examples/fungible-token   hotfix/update-ft ± near delete 141.mike.testnet mike.testnet && near create-account 141.mike.testnet --masterAccount mike.testnet
Deleting account. Account id: 141.mike.testnet, node: https://rpc.testnet.near.org, helper: https://helper.testnet.near.org, beneficiary: mike.testnet
Transaction Id 2JqLijtspS5NxuYrwgq9B9e273XrXCirSYYm7FCaWX1z
To see the transaction in the transaction explorer, please open this url in your browser
https://explorer.testnet.near.org/transactions/2JqLijtspS5NxuYrwgq9B9e273XrXCirSYYm7FCaWX1z
Account 141.mike.testnet for network "default" was deleted.
Saving key to '/Users/mike/.near-credentials/default/141.mike.testnet.json'
Account 141.mike.testnet for network "default" was created.
 mike@n  ~/near/near-sdk-rs/examples/fungible-token   hotfix/update-ft ± near deploy 141.mike.testnet --wasmFile res/fungible_token.wasm --initFunction new --initArgs '{"owner_id": "141.mike.testnet", "total_supply": "1000000", "version": "1.0.0", "name": "Mochi Tokens", "symbol": "MOCHI", "reference": "https://example.com/mochi.json", "reference_hash": "7c879fa7b49901d0ecc6ff5d64d7f673da5e4a5eb52a8d50a214175760d8919a", "decimals": 19}'
Starting deployment. Account id: 141.mike.testnet, node: https://rpc.testnet.near.org, helper: https://helper.testnet.near.org, file: res/fungible_token.wasm
Transaction Id AZ3SxqswjfrPgP8tbr6ijjF1LWtSsYHiUzWtws6beXQs
To see the transaction in the transaction explorer, please open this url in your browser
https://explorer.testnet.near.org/transactions/AZ3SxqswjfrPgP8tbr6ijjF1LWtSsYHiUzWtws6beXQs
Done deploying and initializing 141.mike.testnet
 mike@n  ~/near/near-sdk-rs/examples/fungible-token   hotfix/update-ft ± near call 141.mike.testnet storage_deposit '' --amount '0.00235' --accountId mike.testnet
Scheduling a call: 141.mike.testnet.storage_deposit() with attached 0.00235 NEAR
Transaction Id HPFrXoxtUo8KMKVepnGjYpeU9aiEhwK2zfQaC5wMFNrv
To see the transaction in the transaction explorer, please open this url in your browser
https://explorer.testnet.near.org/transactions/HPFrXoxtUo8KMKVepnGjYpeU9aiEhwK2zfQaC5wMFNrv
{
  total: '2350000000000000000000',
  available: '2350000000000000000000'
}
```

specs/Standards/Tokens/FungibleTokenBase.md
```diff
@@ -0,0 +1,304 @@
+# Fungible Token ([NEP-141](https://github.com/near/NEPs/issues/141))
+
+See also the discussions:
```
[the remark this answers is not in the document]
mikedotexe: Some great points here. I think I will stick with `Metadata` though like how Wikipedia has it:
https://en.wikipedia.org/wiki/Metadata

specs/Standards/Tokens/FungibleTokenBase.md
```diff
@@ -0,0 +1,304 @@
...
+- **Total supply**: the total number of tokens in circulation.
+- **Balance owner**: an account ID that owns some amount of tokens.
+- **Balance**: an amount of tokens.
+- **Transfer**: an action that moves some amount from one account to another account, either an externally owned account or a contract account.
+- **Transfer and call**: an action that moves some amount from one account to a contract account where the receiver calls a method.
+- **Storage amount**: the amount of storage used for an account to be "registered" in the fungible token. This amount is denominated in Ⓝ, not bytes, and represents the [storage staked](https://docs.near.org/docs/concepts/storage-staking).
+
+Note, that the precision is not part of the default standard, since it's not required to perform actions. The minimum value is always 1 token.
+
+The standard acknowledges NEAR storage staking model and accounts for the difference in storage that can be introduced by actions on this contract. Since multiple users use the contract, the contract has to account for potential storage increase. Thus every change method of the contract that can change the amount of storage must be payable.
```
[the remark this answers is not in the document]
mikedotexe: good catch, at first blush that seemed okay to keep in, but you're right. That's some leftovers from the old standard's Nomicon text.

specs/Standards/Tokens/FungibleTokenBase.md
```diff
@@ -0,0 +1,304 @@
...
+
+When called, DEX makes 2 async transfers calls to exchange corresponding tokens.
+- DEX calls wLTC to transfer tokens DEX to Alex.
+- DEX calls wBTC to transfer tokens DEX to Charlie.
+
+**Technical calls**
+
+1. `alex` calls `wbtc::ft_transfer_call({"receiver_id": "dex", "amount": "8000000000", "msg": "deposit"})`.
+   1. makes async call `dex::ft_on_transfer({"sender_id": "alex", "amount": "8000000000", "msg": "deposit"})`.
+   2. attaches a callback `wbtc::ft_resolve_transfer({"sender_id": "alex", "amount": "8000000000"})`.
```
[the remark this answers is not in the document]
mikedotexe: you are the best eagle eye

specs/Standards/Storage.md
mikedotexe: I believe this does address account removal

specs/Standards/Storage.md
chadoh: Seems there is disagreement here, with @oysterpack and @evgenykuzyakov both stating that this standard is not meant to cover account closure/removal. Is there a recorded call where this was discussed and an agreement was reached for what is in and out of scope for the Account Storage standard?

specs/Standards/Storage.md
chadoh: Does that indicate an error in the standard-under-draft, or an error in this contract's adherence to it?

specs/Standards/Tokens/FungibleTokenMetadata.md
```diff
@@ -0,0 +1,82 @@
...
+- `version` is a string and is open for interpretation. That is to say, conventions will likely arise but are not enforced. It's not necessarily using semantic versioning. There may, for instance, be a benefit in having a version prefix/suffix specificially for wrapped or bridged tokens.
+- `name` is the human-readable name of the token.
+- `symbol` is the abbreviation, like wETH or AMPL.
+- `reference` is a link to a valid JSON file containing various keys offering supplementary details on the token. (For example: "/ipfs/QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm", "https://example.com/token.json", etc.)
+- `reference_hash` is the sha256 hash of the JSON file contained in the `reference` field. This is to guard against off-chain tampering.
+- `decimals` is used aid in the frontend showing the proper significant digits of a token. This concept is explained well in this [OpenZeppelin post](https://docs.openzeppelin.com/contracts/3.x/erc20#a-note-on-decimals).
+
+## Drawbacks
+
+- It could be argued that `symbol` and even `name` could belong as key/values in the `reference` JSON object.
```
vgrichina, line 76: I think given NEAR storage costs it makes sense to inline most of the metadata. Especially since `reference_hash` already disallows having mutable metadata (which would be a benefit of having separate reference), kinda not clear why force clients make another HTTP (or worse IPFS) request.

vgrichina: like your typical Rust contract is about 300kb+, metadata likely not gonna be more than single digit kb.

specs/Standards/Storage.md
oysterpack: `storage_withdraw` is only allowed to withdraw from the account's **available** storage balance. Thus, the total account storage balance cannot be withdrawn via `storage_withdraw` because the required minimum storage balance must be maintained.

To be clear,  this standard is not meant to cover account closure/removal - see [discussion][1] which documents the decision.

I would phrase as:
> Allow an account to withdraw Ⓝ from its storage available balance. The account storage balance is defined as the Ⓝ  amount in excess to the minimum account storage balance that is required by the contract.

and add an **Out of Scope** section for:
> - How to close the account and be able to withdraw all funds.
> - How the contract should account for changes in price for storage on the NEAR blockchain over time.

My [tutorial][2] also serves as a good reference.

[1]: https://github.com/near/NEPs/discussions/145#discussioncomment-361351
[2]: https://learn.figment.io/network-documentation/near/tutorials/1-project_overview/5-account-storage#quick-api-overview

specs/Standards/Storage.md
oysterpack: that is a bug in the contract - `available` balance should be 0. Once the required min storage balance is deposited, it cannot be withdrawn via the account storage API.

evgenykuzyakov: Then we should introduce an API to withdraw minimum storage balance somehow into the StorageManager standard. Because right now the minimum storage balance is non reclaimable.

mikedotexe (commented):
Can we please focus on getting this finalized by adding Account Registration?

specs/Standards/Tokens/FungibleTokenMetadata.md
```diff
@@ -53,23 +63,34 @@ A fungible token contract implementing the metadata standard shall contain a fie
...
-    spec: string;
-    name: string;
-    symbol: string;
     reference: string;
     reference_hash: string;
-    decimals: number;
+    decimals?: number;
+    icon?: string;
+    name?: string;
+    spec?: string;
```
evgenykuzyakov, line 73: `spec` should be required here. It indicates how you parse the remaining fields of the metadata object.

specs/Standards/AccountRegistration.md
```diff
@@ -0,0 +1,89 @@
...
+```ts
+/***************************************/
+/* CHANGE METHODS on Account Registrar */
+/***************************************/
+// Registers an account for using smart-contract. Must attach amount of NEAR given by `ar_registration_fee`.
+// Parameters:
+// * `account_id` - an account to register. Anyone can register any other
+//   account.
+// If the account is already registered then the function early returns and
+// refunds the attached NEAR.
```
evgenykuzyakov, line 51: Do we want to handle refund if the attached deposit is larger than the `ar_registration_fee`?

E.g. from current rust implementaiton:
```
 MUST refund the unused attached amount, e.g.
    - if the account already registered, it MUST return the full attached amount.
    - if the attached account is larger than the `ar_registration_fee` should refund extra.
```

specs/Standards/StorageManagement.md
```diff
@@ -0,0 +1,246 @@
...
+// Withdraws a specified amount of Ⓝ for a given account that was previously
+// reserved for storage.
+// Requirements:
+// * `amount` must be less than or equal to the available storage balance for
+//   the sending account. Is sent as a string but represents an unsigned
+//   128-bit integer.
+// * The calling account must already be registered with the contract
+// Returns the AccountStorageBalance structure
+function storage_withdraw(
+    amount: string
```
evgenykuzyakov, line 219: If `amount == null` then the all available storage balance should be withdrawn

```suggestion
    amount: string|null
```

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+// * Caller of the method must attach a deposit of 1 yoctoⓃ for security purposes
+// * Caller must have greater than or equal to the `amount` being requested
+// * The receiving contract must implement `ft_on_transfer` according to the standard.
+// * This contract must implement `ft_resolve_transfer` according to the standard.
+// `msg` is an argument that may specify any information expected by the receiving contract in order to properly handle the function. If the receiving contract needs something larger than a single string, `msg` could be a complex JSON structure, base64-encoded.
+function ft_transfer_call(
+   receiver_id: string,
+   amount: string,
+   msg: string,
+   memo: string|null
```
evgenykuzyakov, line 301: switch order
```suggestion
   memo: string|null,
   msg: string,
```

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+// This function is implemented on the receving contract.
+// As mentioned, the `msg` argument contains information necessary for the receiving contract to know how to process the request. This may include method names and/or arguments. 
+// Returns a value, or a promise which resolves with a value. The value is the
+// number of unused tokens. For instance, if `amount` is 10 but only 9 are
+// needed, it will return 1.
+function ft_on_transfer(
+    sender_id: string,
+    amount: string,
+    msg: string
+): PromiseOrValue<U128> {}
```
evgenykuzyakov, line 326: From Standard API it is still `string`
```suggestion
): string {}
```

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+): string {}
+
+/****************************************/
+/* CHANGE METHODS on receiving contract */
+/****************************************/
+
+// This function is implemented on the receving contract.
+// As mentioned, the `msg` argument contains information necessary for the receiving contract to know how to process the request. This may include method names and/or arguments. 
+// Returns a value, or a promise which resolves with a value. The value is the
+// number of unused tokens. For instance, if `amount` is 10 but only 9 are
```
evgenykuzyakov, line 320: It can probably updated to make it clear that the value is an integer encoded as a string, e.g. `"10"`

Unless we can make it clear for string balances.

specs/Standards/StorageManagement.md
```diff
@@ -0,0 +1,268 @@
...
+
+       near view social storage_minimum_balance
+
+   The response:
+
+       '100000000000000000000000'
+
+   This is 0.1Ⓝ specified in yoctoⓃ ([yocto](https://www.metricconversion.us/prefixes.htm) = 10<sup>-24</sup>).
+
+3. Alice deposits the proper amount in a transaction by calling `social::storage_deposit` with the attached deposit of '0.1'. Using NEAR CLI:
```
chadoh, line 78: The Storage Management standard's interplay with the Account Registration standard is not well-defined. The members of the Standards Working Group with whom I've spoken all have different ideas about how a Storage Manager contract should behave, with regards to Account Registration. An example scenario that generates many different opinions:

Given a contract that implements Storage Management at `contract`
And given a user `alice` who has not yet interacted with `contract`
When `alice` first interacts with `contract`
Does she need to first call `ar_register`, or can she call only `storage_deposit`?

Other points of disagreement:

* Is every Storage Management contract required to also implement all Account Registration functions?
* Does `storage_balance_of` display account registration fee as part of the `total`? 

Given that Storage Management is not needed for Fungible Token, I propose we split Storage Management into a separate Pull Request. This will allow us to finalize the core set of standards needed for Fungible Token with Metadata, while continuing to add clarity to the Storage Management standard.

mikedotexe: Sure, we can split the Storage Management standard into another pull request.

specs/Standards/Tokens/FungibleTokenCore.md
mikedotexe: pushed

specs/Standards/StorageManagement.md
mikedotexe: this is addressed now

specs/Standards/AccountRegistration.md
mikedotexe: we addressed this with min/max using bounds

specs/Standards/Tokens/FungibleTokenMetadata.md
mikedotexe: the current version has it as required

specs/Standards/Tokens/FungibleTokenMetadata.md
mikedotexe: hey, a lot has changed since this comment, but `reference` is not required. here's the current metadata. Sadly I don't think we can continue to iterate on this or else I will be decapitated. 😆 

```ts
type FungibleTokenMetadata = {
    spec: string;
    name: string;
    symbol: string;
    icon: string|null;
    reference: string|null;
    reference_hash: string|null;
    decimals: number;
}
```

specs/Standards/StorageManagement.md
```diff
@@ -0,0 +1,423 @@
...
+// The below defines the structure that will be returned for the methods:
+// * `storage_deposit`
+// * `storage_withdraw`
+// * `storage_balance_of`
+// The `total` and `available` values are string representations of unsigned
+// 128-bit integers.
+type StorageBalance = {
+   total: string;
+   available: string;
+}
```
evgenykuzyakov, line 327: Need better explanation what is `total` and `available`. Mention that they are the balance of tokens in yoctoNEAR.

specs/Standards/StorageManagement.md
```diff
@@ -0,0 +1,423 @@
...
+// If `account_id` is omitted, the deposit MUST go toward predecessor account.
+// If provided, deposit MUST go toward this account. If invalid, contract MUST
+// panic.
+//
+// If `registration_only=true`, contract MUST refund above the minimum balance
+// if the account wasn't registered and refund full deposit if already
+// registered.
+//
+// Any attached deposit in excess of `storage_balance_bounds.max` must be
+// refunded to predecessor account.
```
evgenykuzyakov, line 360: the `current_storage_balance + attached_deposit` in excess of `storage_balance_bounds.max` ...

specs/Standards/StorageManagement.md
```diff
@@ -0,0 +1,423 @@
...
+// Unregisters the predecessor account and returns the storage NEAR deposit.
+//
+// If the predecessor account is not registered, the function MUST return
+// `false` without panic.
+//
+// If `force=true` the function SHOULD ignore existing account data, such as
+// non-zero balances on an FT contract (that is, it should burn such balances),
+// and close the account. Otherwise, MUST panic if caller has existing account
+// data, such as a positive registered balance (eg token holdings) or if the
+// contract doesn't support forced unregistration.
```
evgenykuzyakov, line 393: Add note that if the contract implementation doesn't support forced closure of the account, the contract MAY panic. For example there are too much data to delete at once.

specs/Standards/StorageManagement.md
```diff
@@ -0,0 +1,423 @@
...
+//
+// MUST require exactly 1 yoctoNEAR attached balance to prevent restricted
+// function-call access-key call (UX wallet security)
+//
+// Returns `true` iff the account was unregistered.
```
evgenykuzyakov, line 398:
```suggestion
// Returns `true` iff the account was successfully unregistered.
```

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
+# Fungible Token ([NEP-141](https://github.com/near/NEPs/issues/141))
+
+Version `1.0.0`
+
+## Summary
+
+A standard interface for fungible tokens that allows for a normal transfer as well as a transfer and method call in a single transaction. The storage standard addresses the needs (and security) of storage staking. The [fungible token metadata standard](FungibleTokenMetadata.md) provides the fields needed for ergonomics across dApps and marketplaces.
```
evgenykuzyakov, line 7: Link storage standard

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+      }'
+
+   Then Alice (or the app she uses) will hold onto the result and use it in the next step. Let's say this result is `"c3dhcDpibm5hLDI="`, the base64-encoded version of the string "swap:bnna,2".
+
+2. Call `wnear::ft_on_transfer`. Using NEAR CLI:
+       near call wnear ft_transfer_call '{
+         "receiver_id": "amm",
+         "amount": "5000000000000000000000000",
+         "msg": "c3dhcDpibm5hLDI="
+       }' --accountId alice --amount .000000000000000000000001
```
evgenykuzyakov, line 161: Why encode this as base64? It complicates the example.

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+2. `bob` calls `dex::trade({"have": "wbtc", "have_amount": "8000000000", "want": "wltc", "want_amount": "900100000000"})`.
+3. `alice` calls `wltc::ft_transfer_call({"receiver_id": "dex", "amount": "100000000000000", "msg": "deposit"})`.
+   1. makes async call `dex::ft_on_transfer({"sender_id": "alice", "amount": "100000000000000", "msg": "deposit"})`.
+   2. attaches a callback `wltc::ft_resolve_transfer({"sender_id": "alice", "receiver_id": "dex", "amount": "100000000000000"})`.
+   3. the deposit completes on `dex`, all fungible tokens are used, so it returns `0` as the number of unused tokens back to `wltc`.
+   4. the `wltc::ft_resolve_transfer` function receives success as the execution outcome and `0` as the value. This means 100000000000000 will be subtracted from `bob`'s balance.   
+4. `alice` calls `dex::trade({"have": "wltc", "have_amount": "900100000000", "want": "wbtc", "want_amount": "8000000000"})`.
+   - `dex` calls `wbtc::transfer({"new_owner_id": "alice", "amount": "8000000000"})`
+   - `dex` calls `wltc::transfer({"new_owner_id": "bob", "amount": "900100000000"})`
```
evgenykuzyakov, line 258: Let's remove this example, since it talks about NEP-21

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+// Requirements:
+// * Caller of the method must attach a deposit of 1 yoctoⓃ for security purposes
+// * Caller must have greater than or equal to the `amount` being requested
+// * The receiving contract must implement `ft_on_transfer` according to the standard.
+// * This contract must implement `ft_resolve_transfer` according to the standard.
+// `msg` is an argument that may specify any information expected by the receiving contract in order to properly handle the function. If the receiving contract needs something larger than a single string, `msg` could be a complex JSON structure, base64-encoded.
+function ft_transfer_call(
+   receiver_id: string,
+   amount: string,
+   memo: string|null,
```
evgenykuzyakov, line 300: Need to explain arguments in the comment similar to `ft_transfer`

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+
+// Transfer tokens and call a method on a receiver contract. A successful workflow will end in a success execution outcome to the callback on the same contract at the method `ft_resolve_transfer`.
+//
+// You can think of this as being similar to attaching native NEAR tokens to a function call on some contract. It allows you to attach any Fungible Token in a call to a receiver contract, given by `receiver_id`, where `msg` gets passed to the receiver contract to indicate both a function to call and the parameters to that function.
+// Requirements:
+// * Caller of the method must attach a deposit of 1 yoctoⓃ for security purposes
+// * Caller must have greater than or equal to the `amount` being requested
+// * The receiving contract must implement `ft_on_transfer` according to the standard.
+// * This contract must implement `ft_resolve_transfer` according to the standard.
+// `msg` is an argument that may specify any information expected by the receiving contract in order to properly handle the function. If the receiving contract needs something larger than a single string, `msg` could be a complex JSON structure, base64-encoded.
```
evgenykuzyakov, line 296:
```suggestion
// `msg` is an argument that may specify any information expected by the receiving contract in order to properly handle the function. You can think about it as input arguments for the receiving contract to process this transfer. If the receiving contract needs something larger than a single string, `msg` could be a complex JSON structure, or a base64-encoded binary data.
```

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+): Promise {}
+
+// This function isn't called directly and shall implement logic ensuring it's called by "itself" as a callback to the promise sent to the receiver contract.
+// Here the `sender_id` will be the sender from the `ft_transfer_call` method call. The `receiver_id` and `amount` will be the same as the values provided to `ft_transfer_call`.
+// This returns a string representing a string version of an unsigned 128-bit integer of how many tokens were returned. (This value may be different than `amount`.)
+function ft_resolve_transfer(
+   sender_id: string,
+   receiver_id: string,
+   amount: string
+): string {}
```
evgenykuzyakov, line 311: This doesn't have to be part of the main interface. Let's move it below the main interface in a suggestion section

specs/Standards/Tokens/FungibleTokenCore.md
```diff
@@ -0,0 +1,356 @@
...
+
+// This function is implemented on the receving contract.
+// As mentioned, the `msg` argument contains information necessary for the receiving contract to know how to process the request. This may include method names and/or arguments. 
+// Returns a value, or a promise which resolves with a value. The value is the
+// number of unused tokens in string form. For instance, if `amount` is 10 but only 9 are
+// needed, it will return "1".
+function ft_on_transfer(
+    sender_id: string,
+    amount: string,
+    msg: string
```
evgenykuzyakov, line 325: Need to explain all arguments

specs/Standards/Tokens/FungibleTokenMetadata.md
```diff
@@ -0,0 +1,99 @@
...
+- Alice's account is `alice`.
+- The precision ("decimals" in this metadata standard) on wBTC contract is `10^8`.
+
+**High-level explanation**
+
+Alice issues a transaction to deploy and initialize the fungible token contract, providing arguments to the initialization function that set metadata fields.
+
+**Technical calls**
+
+1. `alice` deploys a contract and calls `wbtc::new({"owner_id": "wbtc", "total_supply": "100000000000000", "spec": "ft-1.0.0", "name": "Wrapped Bitcoin", "symbol": "WBTC", "reference": "https://example.com/wbtc.json", "reference_hash": "7c879fa7b49901d0ecc6ff5d64d7f673da5e4a5eb52a8d50a214175760d8919a", "decimals": 8})`.
```
evgenykuzyakov, line 43: reference hash has to be `base64` of 32 bytes, not hex. E.g. `AK3YRHqKhCJNmKfV6SrutnlWW/icN5J8NUPtKsNXR1M=`

specs/Standards/Tokens/FungibleTokenMetadata.md
```diff
@@ -0,0 +1,99 @@
...
+If this deploy and initialization were done using [NEAR CLI](https://docs.near.org/docs/tools/near-cli) the command would be:
+
+    near deploy wbtc --wasmFile res/ft.wasm --initFunction new --initArgs '{
+      "owner_id": "wbtc",
+      "total_supply": "100000000000000",
+      "spec": "ft-1.0.0",
+      "name": "Wrapped Bitcoin",
+      "symbol": "WBTC",
+      "reference": "https://example.com/wbtc.json",
+      "reference_hash": "7c879fa7b49901d0ecc6ff5d64d7f673da5e4a5eb52a8d50a214175760d8919a",
```
evgenykuzyakov, line 54:
```suggestion
      "reference_hash": "AK3YRHqKhCJNmKfV6SrutnlWW/icN5J8NUPtKsNXR1M=",
```

specs/Standards/Tokens/FungibleTokenMetadata.md
```diff
@@ -0,0 +1,99 @@
...
+- `decimals`: used in frontends to show the proper significant digits of a token. This concept is explained well in this [OpenZeppelin post](https://docs.openzeppelin.com/contracts/3.x/erc20#a-note-on-decimals).
+
+**An implementing contract MAY include the following fields on-chain**
+
+- `icon`: a small image associated with this token. Must be a [data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs), to help consumers display it quickly while protecting user data. Recommendation: use [optimized SVG](https://codepen.io/tigt/post/optimizing-svgs-in-data-uris), which can result in high-resolution images with only 100s of bytes of [storage cost](https://docs.near.org/docs/concepts/storage-staking). (Note that these storage costs are incurred to the token owner/deployer, but that querying these icons is a very cheap & cachable read operation for all consumers of the contract and the RPC nodes that serve the data.) Recommendation: create icons that will work well with both light-mode and dark-mode websites by either using middle-tone color schemes, or by [embedding `media` queries in the SVG](https://timkadlec.com/2013/04/media-queries-within-svg/).
```
evgenykuzyakov, line 87: I disagree that it has to be a data URL. I think an image URL is fine too

specs/Standards/Tokens/FungibleTokenMetadata.md
```diff
@@ -0,0 +1,99 @@
...
+- `reference`: a link to a valid JSON file containing various keys offering supplementary details on the token. Example: "/ipfs/QmdmQXB2mzChmMeKY47C43LxUdg1NDJ5MWcKMKxDu7RgQm", "https://example.com/token.json", etc. If the information given in this document conflicts with the on-chain attributes, the values in `reference` shall be considered the source of truth.
+- `reference_hash`: the base64-encoded sha256 hash of the JSON file contained in the `reference` field. This is to guard against off-chain tampering.
+
+## Drawbacks
+
+- Enforcement of `icon` to be a data URL rather than a link to an HTTP endpoint that could contain privacy-violating code cannot be done on deploy or update of contract metadata, and must be done on the consumer/app side when displaying token data.
```
evgenykuzyakov, line 93: I disagree with the icon has to be a data URL. You need to be experience web designed to craft a SVG icon. Safari still doesn't support `webp`, so your safest fall back is data URI encoded `png` or `gif` which is still a few KB of data on chain. 
Brave enforces `https` on the image urls, so you can't really mess the privacy. If you're fetching the reference, it's also violates privacy. The app may decide to use image proxy if needed.

specs/Standards/Tokens/FungibleTokenMetadata.md
chadoh: @vgrichina please weigh in

mikedotexe: Yesterday I was linked [a commit](https://github.com/near/NEPs/pull/158/commits/c234957ecdf25fdc2c37660e0f91a09bf52447ea) on this topic that only included a section in "Drawbacks" at the bottom. This data URI idea has not been discussed during the several weeks we've been working on this standard.

specs/Standards/Tokens/FungibleTokenMetadata.md
mikedotexe: Agreed, and we're doing it the way Eugene describes.

specs/Standards/StorageManagement.md
chadoh: Updated to  

```
// The `total` and `available` values are string representations of unsigned
// 128-bit integers showing the balance of a specific account in yocto
```

specs/Standards/StorageManagement.md
chadoh: updated to 

```
// The `storage_balance_of.total` + `attached_deposit` in excess of
// `storage_balance_bounds.max` must be refunded to predecessor account.
```

specs/Standards/StorageManagement.md
chadoh: Updated to

```
// If `force=true` the function SHOULD ignore existing account data, such as
// non-zero balances on an FT contract (that is, it should burn such balances),
// and close the account. Contract MAY panic if it doesn't support forced
// unregistration, or if it can't force unregister for the particular situation
// (example: too much data to delete at once).
//
// If `force=false` or `force` is omitted, the contract MUST panic if caller
// has existing account data, such as a positive registered balance (eg token
// holdings).
```

specs/Standards/Tokens/FungibleTokenCore.md
chadoh: I understood this to be our recommendation from discussion on Feb 26, and wanted to show an accurate example.

I have removed the "encode `msg` with base64` recommendation in 1cceedfdb856a3412845a6189b948fe0ea9efc36; does this seem correct to you?

specs/Standards/Tokens/FungibleTokenCore.md
chadoh: Removed in 42d1b8f1d7f1ae4f64e18706ee79ae0a0c5f8ade

specs/Standards/Tokens/FungibleTokenCore.md
chadoh: Updated in 34b68f1aa576657e8f50381b20c26c2f467a5852

specs/Standards/Tokens/FungibleTokenCore.md
chadoh: Handled an alternative way in https://github.com/near/NEPs/commit/34b68f1aa576657e8f50381b20c26c2f467a5852

specs/Standards/Tokens/FungibleTokenCore.md
chadoh: As in, the behavior is required, but the function name is not part of the standard?

specs/Standards/Tokens/FungibleTokenMetadata.md
chadoh: Originally changed here: 95e5101cb9e08ff68fd15d83aaba985aebb2f075

@vgrichina knows the security implications better than I do, but his concern is that if we allow HTTP endpoints in these images, then many projects will start including links to images that embed tracking logic (and maybe [other nasty stuff](https://thehackernews.com/2015/06/Stegosploit-malware.html)?). Vlad told me that data URLs cannot contain such privacy-violating software or exploits.

Additionally, if the point is to enable apps to display token information without querying a separate external endpoint for each token, then we should have the entire icon image stored on-chain.

To me it seems like forbidding HTTP endpoints is the safer starting point. If we decide to allow them in a future iteration, this will be an easy update that does not change the shape of the metadata. It could be `spec: "ft-1.0.1"`, for example.

I hope @vgrichina can weigh in soon, but I vote we keep it as-is for now.

specs/Standards/Tokens/FungibleTokenMetadata.md
chadoh: > If you're fetching the reference, it's also violates privacy. 

Yes, this is true, and something we should make application developers aware of.

> You need to be experience web designed to craft a SVG icon

I think we should set a high-but-reasonable bar in the recommendation. This was a concern that was brought up in the Feb 26 call. 

Contract development teams are free to turn PNG or GIF images into data URLs instead, if they can't afford to contract a designer for one asset and can't figure out how to use ImageMagick to turn their PNG into an SVG.

specs/Standards/Tokens/FungibleTokenCore.md
chadoh: Updated in https://github.com/near/NEPs/pull/158/commits/9b420e671d86c93119b581ef5e41b359b9262710

specs/Standards/Tokens/FungibleTokenCore.md
chadoh: Updated in https://github.com/near/NEPs/pull/158/commits/9b420e671d86c93119b581ef5e41b359b9262710

specs/Standards/Tokens/FungibleTokenMetadata.md
chadoh: Resolved in 6bea880bac30ffc9180f780cc11033609d5a0be1

specs/Standards/Tokens/FungibleTokenMetadata.md
```diff
@@ -90,8 +90,8 @@ type FungibleTokenMetadata = {
 
 ## Drawbacks
 
-- Enforcement of `icon` to be a data URL rather than a link to an HTTP endpoint that could contain privacy-violating code cannot be done on deploy or update of contract metadata, and must be done on the consumer/app side when displaying token data.
-- If on-chain icon uses a data URL or is not set but the document given by `reference` contains a privacy-violating `icon` URL, consumers & apps of this data should not naïvely display the `reference` version, but should prefer the safe version. This is technically a violation of the "`reference` setting wins" policy described above.
+- It could be argued that `symbol` and even `name` could belong as key/values in the `reference` JSON object.
+- It might make sense to add `icon_url` and `icon_hash` to the on-chain properties listed in `FungibleTokenMetadata`.
```
chadoh, line 90: These drawbacks don't seem to make sense with the current version of the spec. 

`icon` has been added, so we don't need a note about `icon_url`. Maybe we still need to mention `icon_hash`, especially if we are allowing `icon` to be an HTTP endpoint instead of a data URL. And if we're allowing an HTTP endpoint, then we should definitely mention the security & privacy implications in Drawbacks!

specs/Standards/Tokens/FungibleTokenMetadata.md
evgenykuzyakov: Might need to update where the metadata is stored in the contract in the token example, so we lazy load it in the contract. We may want to add a new collection. Like `Lazy<T: BorshSerialize+BorshDeserialize>`

specs/Standards/Tokens/FungibleTokenMetadata.md
mikedotexe: Yeah, this is easy to have an as additional metadata. We've had several discussions over weeks with community members and none of them have been brought into the conversation. We're not changing this at the last minute, sorry.

specs/Standards/Tokens/FungibleTokenMetadata.md
mikedotexe: Good point, this is from when I wrote all the standards on a Friday, before we added `icon`. I'll remove it, thanks.

specs/Standards/Tokens/FungibleTokenMetadata.md
vgrichina: Actually once I think about it: data URLs with SVG also likely not safe. As you can embed references to external stuff in SVG.

I think we need think about it seriously or we can end up having it very easy to track given user by sending them airdrop of your sketchy token.

vgrichina: Looks like potentially there is already protection against embedded references though https://stackoverflow.com/questions/59163530/can-i-use-an-svg-use-tag-with-an-external-reference-in-either-a-data-uri-or-vi but we need to double check.

vgrichina: @mikedotexe I don’t think “our community members didn’t notice security issue” is a sufficient argument for not addressing a security issue. 

If we keep icon in standard as is is, the likely outcome is that most good quality tools (including our wallet) would just have to ignore it altogether because of two reasons:
- privacy concerns
- absence of standard on the images themselves (and so likely very inconsistent icons between different tokens)

specs/Standards/Tokens/FungibleTokenMetadata.md
vgrichina: > I disagree with the icon has to be a data URL. You need to be experience web designed to craft a SVG icon. Safari still doesn't support `webp`, so your safest fall back is data URI encoded `png` or `gif` which is still a few KB of data on chain.
> Brave enforces `https` on the image urls, so you can't really mess the privacy. If you're fetching the reference, it's also violates privacy. The app may decide to use image proxy if needed.

Not sure how https helps here at all. The problem is privacy from the creators of tokens. Creators of token can track everybody who loads the icon if it’s web URL, no matter whether HTTP or HTTPS. 

like you can of course decide not to change standard. But in that case we not gonna use icon field in wallet at all.

vgrichina: > The app may decide to use image proxy if needed.

this doesn’t really help with airdropped tokens, as you can airdrop with custom contract and so icon URL for every user

specs/Standards/Tokens/FungibleTokenMetadata.md
vgrichina: > And if we're allowing an HTTP endpoint, then we should definitely mention the security & privacy implications in Drawbacks!

yeah, main implication gonna be NEAR Wallet not showing icon for your token if it requires querying untrusted server

specs/Standards/Tokens/FungibleTokenMetadata.md
chadoh: > We've had several discussions over weeks with community members and none of them have been brought into the conversation

Decision to add `icon` to spec was made [one week ago](https://youtu.be/MQWkK3j6CPM?t=2823).

chadoh: > If we keep icon in standard as is is, the likely outcome is that most good quality tools (including our wallet) would just have to ignore it altogether

@vgrichina is this true if we require a data URL? Seems like we might still have to ignore on-chain `icon` pending research into SVG including embedded references.

chadoh: I guess an alternative is that we don't add `icon` at all right now and for `ft-1.0.0` it will need to be linked from the `reference` file

specs/Standards/Tokens/FungibleTokenMetadata.md
mikedotexe: I'm now understanding what you're talking about. I think I needed an example for it to sink in. Some hacker could write a fungible token contract that has an icon that points to https://example.com/mike-purvis-opened-this.jpg and send me an email with a link to https://wallet.example.com/accounts/fake-token-182537893 and when I go there, the Wallet site requests mike-purvis-opened-this.jpg and the web server remembers this and then knows that I followed the link via email

specs/Standards/Tokens/FungibleTokenMetadata.md
mikedotexe: Thanks for all the discussion folks. I put a comment elsewhere with an example I came up with and finally understand the concern. Thanks for bearing with me.

evgenykuzyakov (approved):

chadoh (approved):

PR #162, opened 2021-02-22
feat: add secp validator key

bowenwang1996:
This NEP consolidates the discussion started in [this thread](https://gov.near.org/t/add-secp-validator-key-to-enable-pessimistic-bridge/401) to a more concrete proposal to add secp validator key to the protocol.

### DIFF
--- a/proposals/0000-template.md
+++ b/proposals/0000-template.md
@@ -1,0 +1,238 @@
+- Proposal Name: add-secp-validator-key
+- Start Date: 2021-02-20
+- NEP PR: [nearprotocol/neps#0000](https://github.com/nearprotocol/neps/pull/0000)
+- Issue(s): link to relevant issues in relevant repos (not required).
+
+# Summary
+[summary]: #summary
+
+We would like to introduce a sepc256k1 validator key in addition to the existing ed25519 validator key.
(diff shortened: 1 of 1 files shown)


proposals/0000-template.md
```diff
@@ -0,0 +1,238 @@
...
+    stake: &StakeAction,
+    last_block_hash: &CryptoHash,
+    epoch_info_provider: &dyn EpochInfoProvider,
+) -> Result<(), RuntimeError> {
+    if account.code_hash == STAKING_POOL_V1_HASH {
+        action_deploy_contract(new_staking_pool_contract);
+    }
+    // Then proceed to do what the function does now.
+}
+```
```
evgenykuzyakov, line 109: Why not replace the code by hash? Basically if we're trying to execute a code for a hash `123`, then we actually need to execute hardcoded other code.

bowenwang1996: I am not sure what you mean here. `action_deploy_contract(new_staking_pool_contract)` does replace the contract. Do you mean we shouldn't hardcode the contract binary? Then you still need to get the binary and store it in state.

bowenwang1996:
Obsolete. We decided to not implement this

Issue #141, opened 2021-01-08
[WIP] Fungible Token Core Standard


evgenykuzyakov:
# Fungible Token Core Standard

## Background

This standard is a subset of the proposed https://github.com/near/NEPs/issues/136 that only covers the core contract API.
Please read the linked standard.

It doesn't include token metadata standard that will be covered by a separate NEP, because the metadata may evolve.
It also doesn't include account registration standard that also should be covered by a separate NEP because it can be reused for other contract.

It's also not exact core API from the https://github.com/near/NEPs/issues/136
This standard allows to specify the amount of tokens used by the contract and refund a part. It also renames some arguments to
be more consistent with the original fungible token standard https://github.com/near/NEPs/issues/21

See https://github.com/near/NEPs/issues/122#issuecomment-755746724 for the rationale to move from https://github.com/near/NEPs/issues/122

### API

Token contract should implement the following Trait.

```rust
/// Core functionality of a Fungible Token contract
pub trait FungibleTokenCore {
    /// Simple transfer.
    /// Transfers positive `amount` of tokens from the `env::predecessor_account_id` to `receiver_id`.
    /// Both accounts should be registered with the contract for transfer to succeed.
    /// Method is required to be able to accept attached deposits - to not panic on attached deposit. See Security
    /// section of the standard.
    ///
    /// Arguments:
    /// - `receiver_id` - the account ID of the receiver.
    /// - `amount` - the amount of tokens to transfer. Should be a positive number in decimal string representation.
    /// - `memo` - an optional string field in a free form to associate a memo with this transfer.
    #[payable]
    fn ft_transfer(
        &mut self,
        receiver_id: ValidAccountId,
        amount: U128,
        memo: Option<String>
    );

    /// Transfer to a contract with a callback.
    /// Transfers positive `amount` of tokens from the `env::predecessor_account_id` to `receiver_id` account. Then
    /// calls `ft_on_transfer` method on `receiver_id` contract and attaches a callback to resolve this transfer.
    /// `ft_on_transfer` method should return the amount of tokens used by the receiver contract, the remaining tokens
    /// should be refunded to the `predecessor_account_id` at the resolve transfer callback.
    ///
    /// Token contract should pass all the remaining unused gas to the `ft_on_transfer` call.
    ///
    /// Malicious or invalid behavior by the receiver's contract:
    /// - If the receiver contract promise fails or returns invalid value, the full transfer amount should be refunded.
[...]

PR #168, opened 2021-03-06
github.com/near/neps/pull/141 is outdated

luciotato:
can confuse newcomers. Eugene could also update what's at http://github.com/near/neps/pull/141 to reflect the new standard


frol:
This has already been addressed, so I am closing this PR

PR #175, opened 2021-03-17
fix(standards): add ft_metadata method

chadoh:
The standard did not specify how a contract should actually return data


frol:
It seems it was addressed in another PR, so this PR can be safely closed.

PR #183, opened 2021-03-26
Add NFT Royalty Standard

mikedotexe:
This pull request represents discussion that are being had at: https://github.com/near/NEPs/discussions/182

This standard would benefit from following the pattern of the other standards, although not all sections are strictly required:
- Summary 
- Motivation
- Prior art/influence
- Interface
  - High-level examples
- Reference-level explanation
- Drawbacks
- Future possibilities


frol:
It seems it was prior art for NEP-199 (#224), so we can close it

PR #187, opened 2021-04-02
Update ApprovalManagement.md

thor314:


specs/Standards/NonFungibleToken/ApprovalManagement.md
```diff
@@ -319,8 +328,6 @@ The NFT contract must implement the following methods:
 // * Contract MAY require caller to attach larger deposit, to cover cost of
 //   storing approver data
 // * Contract MUST panic if called by someone other than token owner
-// * Contract MUST panic if addition would cause `nft_revoke_all` to exceed
-//   single-block gas limit
```
chadoh, line 323: This is referencing the hard cap on gas per block at the protocol layer, described here: https://docs.near.org/docs/concepts/gas#thinking-in-gas. This number will increase over time. Additionally, contract implementations may vary, meaning that some implementations will be able to support a larger number of approvals than others, even with the same maximum gas per block.

But every contract's `nft_revoke_all` would have _some_ limit. if adding this approval would make it impossible to call `nft_revoke_all`, then calling `nft_approve` should fail.

Maybe we can move this description elsewhere, since it's a lot of concepts to explain in this code comment.

chadoh: I've added this detail to the core standard in https://github.com/near/NEPs/pull/239. Please restore these two lines!

chadoh (changes requested):
See above request

jberrytech:
Sprint planning comment: we'll ping @thor314 for followup.

frol (approved):

PR #189, opened 2021-04-02
Update Core.md

thor314:
`receiver_id` makes no sense in the context given. `nft_resolve_transfer` does not call `nft_transfer_call` (the opposite is true).

### DIFF
--- a/specs/Standards/NonFungibleToken/Core.md
+++ b/specs/Standards/NonFungibleToken/Core.md
@@ -154,7 +154,6 @@ The following behavior is required, but contract authors may name this function 
 //
 // Arguments:
 // * `sender_id`: the sender of `ft_transfer_call`
-// * `receiver_id`: the `receiver_id` argument given to `ft_transfer_call`
 // * `token_id`: the `token_id` argument given to `ft_transfer_call`
 // * `approved_token_ids`: if using Approval Management, contract MUST provide
 //   set of original approved accounts in this argument, and restore these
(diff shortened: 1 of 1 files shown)


specs/Standards/NonFungibleToken/Core.md
```diff
@@ -163,7 +162,6 @@ The following behavior is required, but contract authors may name this function
 // Returns true if token was successfully transferred to `receiver_id`.
 function nft_resolve_transfer(
   owner_id: string,
-  receiver_id: string,
```
evgenykuzyakov, line 166: This is local callback, so it doesn't need to be here in the Standard.

evgenykuzyakov: We use `receiver_id`, because the `nft_resolve_transfer` has the ability to revert the transfer in case the `nft_on_transfer` panics or returns `false`. `receiver_id` is currently used to make sure we rollback from the correct receiver. So if the receiver has changed again, the rollback can't be completed.

thor314: So basically something like this

impl MintbaseStore {
  #[private]
  pub fn nft_resolve_transfer(
    &mut self,
    owner_id: String,
    token_id: u64,
    approvals: HashMap<AccountId, u64>,
    receiver_id: AccountId,
    memo: Option<String>,
  ) {
    assert_eq!(receiver_id, self.nft_owner(token_id.to_string()).unwrap());

PR #199, opened 2021-05-03
Payout NFT standard

thor314:


mattlockyer:
`max_len_payout` is to stop NFT transfer before Market receives something it can't use correct?

Otherwise looks fine to me.

thor314:
> `max_len_payout` is to stop NFT transfer before Market receives something it can't use correct?
> 
> Otherwise looks fine to me.

`max_len_payout` prevents the pathological situation where, if the market did NOT inform the NFT contract how many people it was willing to pay out, the following could occur; suppose for this example, the market will pay out at most 10 people.
1. The market asks the NFT contract for who to pay, and tells the NFT contract to transfer the token to Bob
2. The NFT contract transfers the token to Bob and gives the market a list of 12 people to pay out
3. The market refuses to pay out those 12 people, as the market will run out of gas; 

The people who should have gotten paid didn't get paid, and Bob gets ownership of the token.

We also don't want to live in a world where all markets are forced to support exactly 10 payouts, some might like to go higher. So to support both of these, `max_payout_len` solves both issues without risking annoying situations like the one above.

mattlockyer:
@mikedotexe @evgenykuzyakov this looks good to me. I like how `max_len_payout` helps to avoid NFT <> Market inconsistencies ahead of internal_transfer actually moving the NFT.

thor314:
adjusted the view method. It should probably include the max_len_payout parameter as well.

mattlockyer (approved):
This looks great. I have it working across multiple projects now. We should clean up the language (if any edits are needed for clarity) and merge into nomicon ASAP.

mikedotexe:
Moved to https://github.com/near/NEPs/pull/224

PR #203, opened 2021-05-16
fixes min/max typo and indented code blocks

anson-vandoren:


frol:
@chadoh Could you take a quick look here? It was your commit https://github.com/near/NEPs/commit/00326aeb412690c988015b20a47e4cb9e75682d6 that used `min_slip` instead of `max_slip`. I feel there is indeed a typo here, but I still wanted to double-check.

frol (approved):
I spent some time myself here and I believe this fix is correct, so I will merge this correction

PR #209, opened 2021-06-01
Document contract reward in Economics

Longarithm:


specs/Economics/README.md
```diff
@@ -247,3 +247,8 @@ def end_of_epoch(..., reward):
     # ...
     accounts[TREASURY_ACCOUNT_ID].amount = treasury_reward[t]
 ```
+
+## Contract Rewards
+
+Contract account is rewarded with 30% of gas burnt during execution of its functions, for the possibility to invoke them.
+Rewarding happens after applying the corresponding receipt with [`FunctionCallAction`](../RuntimeSpec/Actions.md#functioncallaction), gas is converted to tokens using gas price of the current block.
```
bowenwang1996, line 254: I think we should add a link to the documentation on receipt execution as well.

bowenwang1996 (approved):

PR #215, opened 2021-06-11
Update Account.md to document system account

mzhangmzz:


specs/DataStructures/Account.md
```diff
@@ -85,10 +85,13 @@ a..near          // Two dot separators in a row
 $$$              // Non alphanumeric characters are not allowed
 WAT              // Non lowercase characters are not allowed
 me@google.com    // @ is not allowed (it was allowed in the past)
+system           // cannot use the system account, see the section on System account below
 // TOO LONG:
 abcdefghijklmnopqrstuvwxyz.abcdefghijklmnopqrstuvwxyz.abcdefghijklmnopqrstuvwxyz
 ```
 
+## System account
+`system` is a special account that is used only for refund receipts. Users cannot create or access the `system` account. For refund receipts, we set the predecessor_id to be `system` to indicate that it is a refund receipt.
```
bowenwang1996, line 94: Let's also mention that this "account" in fact does not exist (i.e, it is not part of the state)

bowenwang1996 (approved):

PR #237, opened 2021-07-14
Fix account id specification

EgorKulikov:
As we actually do allow account ids of length 64 that are not hex, we should make it more clear in spec


bowenwang1996 (approved):
Thanks!

PR #238, opened 2021-07-14
standards/NFT: previous_owner_id

chadoh:
this was some work I had in progress on `master` a long time ago that I need to investigate

### DIFF
--- a/specs/Standards/Tokens/NonFungibleToken/Core.md
+++ b/specs/Standards/Tokens/NonFungibleToken/Core.md
@@ -82,8 +82,6 @@ function nft_transfer(
   memo: string|null,
 ) {}
 
-// Returns `true` if the token was transferred from the sender's account.
-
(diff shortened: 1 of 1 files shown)


specs/Standards/NonFungibleToken/Core.md
```diff
@@ -162,7 +160,7 @@ The following behavior is required, but contract authors may name this function
 //
 // Returns true if token was successfully transferred to `receiver_id`.
 function nft_resolve_transfer(
-  owner_id: string,
+  previous_owner_id: string,
```
frol, line 163: Given that this was [implemented and de-facto standardized via near-sdk-rs](https://github.com/near/near-sdk-rs/blob/a903f8c44a7be363d960838d92afdb22d1ce8b87/near-contract-standards/src/non_fungible_token/core/resolver.rs#L36), we should just merge it

frol (approved):

PR #241, opened 2021-07-19
NEPs on how to split states of shards for resharding

mzhangmzz:


mzhangmzz:
Most of the spec is done. Still need to fill some analysis sections like Rationales and alternatives, Unresolved questions, etc.

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,257 @@
...
+
+The proposal assumes that all validators track all shards and that challenges are not enabled.
+
+Suppose the new sharding assignment come into effect at epoch T. State migration is done at epoch T-1, when the validators for epoch T are catching up states for the next epoch. At the beginning of epoch T-1, they run state sync for the current shards if needed. From the existing states, they build states for the new shards, then apply changes to the new states when they process the blocks in epoch T-1.  This whole process runs off-chain as the new states will be not included in blocks at epoch T-1. At the beginning of epoch T, the new validators start to build blocks based on the new state roots.
+
+The change involves three parts. 
+
+## Dynamic Shards
+The first issue to address in splitting shards is the assumption that the current implmentation of chain and runtime makes that the number of shards never changes. This in turn involves two parts, how the validators know when sharding changes happen and how they store different states of shards from different epochs during the transition. The former is a protocol change and the latter only affects validators' internal states.
```
bowenwang1996, line 27:
```suggestion
The first issue to address in splitting shards is the assumption that the current implementation of chain and runtime assumes that the number of shards never changes. This in turn involves two parts, how the validators know when sharding changes happen and how they store different states of shards from different epochs during the transition. The former is a protocol change and the latter only affects validators' internal states.
```

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,257 @@
...
+
+### Protocol Change
+Sharding change will be triggered as protocol version changes. Everytime a resharding is scheduled, a new protocol version number is created for the switch to new sharding. Epochs with `protocol_version` larger or equal than the threshold number will take the new sharding assignment and epochs with `protocol_version` less will take the old sharding assignment. Since the `EpochInfo` of epoch T will be determined at the end of epoch T-2, the validators will have time to prepare for states for the new shards during epoch T-1.
+
+For example, for the Simple NightShade migration, the current protocol version is 42. If there is no other pending protcol update, we can schedule the SimpleNightShadeShardSplit to happen at protocol version 43. Then when the epoch's protocol version is updated to 43, the new sharding for Simple NightShade will be enabled.
```
bowenwang1996, line 32: I don't think the current protocol version is 42 :) http://rpc.mainnet.near.org/status

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,257 @@
...
+```
+converts `shard_id` to the `shard_ord` and `epoch_id` of a shard. `current_epoch_id` will be used as the starting search point, the function will only search for epoch T-1, T, T+1 if the current epoch is T. If it cannot find the `shard_id`, it returns an `EpochError`.
+#### `from_shard_ord_to_id`
+```rust
+from_shard_ord_to_id(&self, shard_ord: ShardOrd, epoch_id: &EpochId) -> Result<ShardId, EpochError>
+```
+converts `shard_ord` and `epoch_id` for a shard to its `shard_id`. It returns an EpochError if such shard does not exist.
+### EpochInfoV3
+Epoch info will include the `shard_id` of the shards in this epoch and a mappting from the current shards to their parent shards. We only allow adding new shards that are split from the existing shards, and if shard B and C are split from shard A, we call shard A the parent shard of shard B and C. For example, if epoch T-1 has shard with `shard_id` 0 and 1 and each of them will be split to two shards in epoch T, then the `EpochInfo` for epoch T will have `shards = [2, 3, 4, 5]` and `parent_shards = {2:0, 3:0, 4:1, 5:1}`.
```
bowenwang1996, line 71:
```suggestion
Epoch info will include the `shard_id` of the shards in this epoch and a mapping from the current shards to their parent shards. We only allow adding new shards that are split from the existing shards, and if shard B and C are split from shard A, we call shard A the parent shard of shard B and C. For example, if epoch T-1 has two shards with `shard_id` 0 and 1 and each of them will be split to two shards in epoch T, then the `EpochInfo` for epoch T will have `shards = [2, 3, 4, 5]` and `parent_shards = {2:0, 3:0, 4:1, 5:1}`.
```

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,257 @@
...
+returns the `shard_id`s of the shards in the next epoch
+
+## Build New States
+The following method in `Client` will be added or modified to split a shard's current state into multiple states.
+
+### `split_shards`
+```rust
+pub fn split_shards(me: &Option<AccountId>, sync_hash: CryptoHash, shard_id: ShardId)
+```
+builds states for the new shards that the shard `shard_id` will be split to. After this function is finished, the states for the new shards should be ready in `ShardTries` to be accessed.
```
bowenwang1996, line 128: What does this function do exactly? We cannot afford to synchronously compute the new states in this function. Otherwise the client actor will be blocked for a long time

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,257 @@
...
+pub struct EpochInfoV3 {
+    ... // All fields in EpochInfoV2
+    shards: Vec<ShardId>,
+    parent_shards: HashMap<ShardId, ShardId>,
+}
+```
+
+### ShardVersionManager
+A new struct `ShardVersionManager` will be created to manage sharding assignment when `ProtocolVersion` is changed. `EpochManager` will own `ShardVersionManager` that it can call to create shards for the `EpochInfo` of the next next epoch.
+
```
bowenwang1996, line 83: Given that this information is only needed to access state, I am not sure whether epoch manager is the best place to store it.

bowenwang1996 (commented):
Looks good. One small thing: we prefer to have [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line).

specs/Proposals/0040-split-states.md
mzhangmzz: I decided to put it inside EpochManager because shards information are saved as part of EpochInfo. EpochManager manages protocol version of epochs, and in this design, shard version is derived there, so it seems natural to put this information there.

specs/Proposals/0040-split-states.md
mzhangmzz: I assumed run_catchup is executed in its own thread, so I thought that wouldn't affect other activities of the client actor. I'll take a closer look

Longarithm (approved):
Thank you, it is highly readable!

specs/Proposals/0040-split-states.md
mzhangmzz: discussed offline

specs/Proposals/0040-split-states.md
mzhangmzz: discussed offline and agreed that we should move the catchup process to its own actor https://github.com/near/nearcore/issues/3201

specs/Proposals/0040-split-states.md
```diff
@@ -206,52 +271,46 @@ fn apply_chunks(...) -> Result<(), Error> {
...
+    ...
+
 }
 ```
 
 ## Garbage collection
 #TODO
 ???
+I think garbage collection should mostly work automatically? Need to verify that new ChunkExtra stored for these new shards will be garbage collected correctly.
+Otherwise I think it should be fine.
```
bowenwang1996: We need to handle garbage collection of the old state after it is split into multiple. Today garbage collection won't automatically clean that. This can be done later though

specs/Proposals/0040-split-states.md
```diff
@@ -206,52 +271,46 @@ fn apply_chunks(...) -> Result<(), Error> {
...
-NEP you are writing but otherwise related.
-
-If you have tried and cannot think of any future possibilities,
-you may simply state that you cannot think of anything.
-
-Note that having something written down in the future-possibilities section
-is not a reason to accept the current or a future NEP. Such notes should be
-in the section on motivation or rationale in this or subsequent NEPs.
-The section merely provides additional information.
+None
```
bowenwang1996, line 316: well we've discussed possibly switching to on-chain state updates once we introduce challenges...

EgorKulikov (approved):

nearmax:
Thank you for thorough NEP and following https://gov.near.org/t/quality-control-for-protocol-api-changes/1941#the-format-4 especially for talking about drawbacks and alternatives! Can we also add a list of affected projects (maybe @bowenwang1996 can help since he has a lot of context) and pre-mortem (imagine a most likely scenario of how this design change can lead to a failure). Thanks!

frol:
I am sorry for jumping late to the discussion, but reviewing https://github.com/near/nearcore/pull/4554, I feel like from the end-user point of view, they should care about some "shard id" (I feel it does not matter if it is unique in the epoch or unique across the epochs). Here are the questions I suggest we ask ourselves:

Why do you believe outside world should care about shard ord rather than unique shard id?
If outside world should care about ord, why should they care that it is “ord” rather than calling it “id”?

If we want some unique shard id, we may just introduce ShardUniqueId (ShardUid) instead of going through this massive renaming effort.

mzhangmzz:
> I am sorry for jumping late to the discussion, but reviewing [near/nearcore#4554](https://github.com/near/nearcore/pull/4554), I feel like from the end-user point of view, they should care about some "shard id" (I feel it does not matter if it is unique in the epoch or unique across the epochs). Here are the questions I suggest we ask ourselves:
> 
> Why do you believe outside world should care about shard ord rather than unique shard id?
> If outside world should care about ord, why should they care that it is “ord” rather than calling it “id”?
> 
> If we want some unique shard id, we may just introduce ShardUniqueId (ShardUid) instead of going through this massive renaming effort.

@frol thanks for the detailed comment! Regarding your questions
> Why do you believe outside world should care about shard ord rather than unique shard id?
That's a great question. I think unique shard id is a concept that closely tied to how we choose to represent shard internally in the state. Thus it may subject to change in the future if the implementation changes. By hiding it away from protocol level and the outside world, such change would not trigger protocol level upgrade in the future. 
If outside world should care about ord, why should they care that it is “ord” rather than calling it “id”?
> You are right on this. Initially I didn't think renaming would be such an invasive change and I wanted the naming to highlight the difference between "ord" and "id". If you think the renaming creates a lot of trouble on the API side, we can definitely not do it and just create a new name for the new id like you suggested.

bowenwang1996 (commented):
One question that occurred to me day: what happens to nonvalidator nodes which specify the shards they track? It seems that, according to what is proposed here, those nodes will start tracking all the shards that the original shard split into. The side effect is that what the node actually tracks will no longer be the same as what is specified in `config.json` and if a node stops and starts again after the split is done, it will stop tracking the shards whose ids are not in `tracked_shards`. This could be confusing for people who need to run nonvalidator nodes and we should at least document such peculiar behavior. It would certainly be better if we can make this experience more smooth for nonvalidator nodes as well.

mzhangmzz:
> One question that occurred to me day: what happens to nonvalidator nodes which specify the shards they track? It seems that, according to what is proposed here, those nodes will start tracking all the shards that the original shard split into. The side effect is that what the node actually tracks will no longer be the same as what is specified in `config.json` and if a node stops and starts again after the split is done, it will stop tracking the shards whose ids are not in `tracked_shards`. This could be confusing for people who need to run nonvalidator nodes and we should at least document such peculiar behavior. It would certainly be better if we can make this experience more smooth for nonvalidator nodes as well.

That's a good point. If we want to make the process smoother for validator nodes, when specifying shards to track, we can make them supply the epoch_id along with the shard_id as well, then at start, ShardTracker can figure out what shards these shards become in the current shards.

bowenwang1996:
> > One question that occurred to me day: what happens to nonvalidator nodes which specify the shards they track? It seems that, according to what is proposed here, those nodes will start tracking all the shards that the original shard split into. The side effect is that what the node actually tracks will no longer be the same as what is specified in `config.json` and if a node stops and starts again after the split is done, it will stop tracking the shards whose ids are not in `tracked_shards`. This could be confusing for people who need to run nonvalidator nodes and we should at least document such peculiar behavior. It would certainly be better if we can make this experience more smooth for nonvalidator nodes as well.
> 
> That's a good point. If we want to make the process smoother for validator nodes, when specifying shards to track, we can make them supply the epoch_id along with the shard_id as well, then at start, ShardTracker can figure out what shards these shards become in the current shards.

Well the main problem here is that let's say someone wants to track all shards. Today they will specify `tracked_shards: [0]` to do that. But after the split of shards is done and they stop and restart their node, they will stop tracking every shard other than shard 0, which is suboptimal. To combat this, I think we have the following options:
* Allow a node to specify that it wants to track all shards in config.json. Something like `tracked_shards: all`
* When a node stops, overwrite config.json with shards it actually tracks. This might not be ideal if a node only wants to track one shard even if it is split into two.

mzhangmzz:
> > > One question that occurred to me day: what happens to nonvalidator nodes which specify the shards they track? It seems that, according to what is proposed here, those nodes will start tracking all the shards that the original shard split into. The side effect is that what the node actually tracks will no longer be the same as what is specified in `config.json` and if a node stops and starts again after the split is done, it will stop tracking the shards whose ids are not in `tracked_shards`. This could be confusing for people who need to run nonvalidator nodes and we should at least document such peculiar behavior. It would certainly be better if we can make this experience more smooth for nonvalidator nodes as well.
> > 
> > 
> > That's a good point. If we want to make the process smoother for validator nodes, when specifying shards to track, we can make them supply the epoch_id along with the shard_id as well, then at start, ShardTracker can figure out what shards these shards become in the current shards.
> 
> Well the main problem here is that let's say someone wants to track all shards. Today they will specify `tracked_shards: [0]` to do that. But after the split of shards is done and they stop and restart their node, they will stop tracking every shard other than shard 0, which is suboptimal. To combat this, I think we have the following options:
> 
> * Allow a node to specify that it wants to track all shards in config.json. Something like `tracked_shards: all`
> * When a node stops, overwrite config.json with shards it actually tracks. This might not be ideal if a node only wants to track one shard even if it is split into two.

Yeah I understand what you meant. I think the issue again is that the meaning of shard id changes when resharding happens. One way to solve that is to be more specific than just providing the shard_id. The config file will include the epoch_id (or protocol version) along with the shard id. For example, let's say a validator starts with tracking shard 0 and they start from genesis. Then the config file will have tracked_shards: [0], tracked_shards_epoch: genesis_epoch_id. Then later when the validator restarts, the ShardTracker can know the shard `0` is not the current shard `0`, but the genesis shard `0`. It can derive which shards it should track now from the `EpochInfo`s stored in the database.

mzhangmzz:
As I am implementing the change regarding `ShardId` and `ShardOrd`, I do find that having two different ids referring to shards quite confusing and could be error prone. We need to extremely careful to use the correct type whenever we use `ShardId` or `ShardOrd`. Since `ShardId` occurs so many places in the code, it is painstaking to get it exactly right. It may also introduce problems in the future if we are not careful and mix up these two. Since they are the same type (u64), it is quite easy to do so. Thus, I am having second thought on whether we should keep them separate or just use `ShardId`( or `ShardUId` as @frol calls it) everywhere.

bowenwang1996:
@mzhangmzz using the same type does not really prevent us from mixing them up? It seems that we need some internal representation that is different anyways. So it seems that introducing a strictly different type may be helpful (this could be purely an implementation detail)

frol:
> Today they will specify `tracked_shards: [0]`

I suggest to take a step back and ask ourselves why would anyone care about shard number X? and how would they know that they *want* to track exactly the shard number X? Maybe they care about a set of specific accounts and we should track based on the list of account ids instead of the shard ord/id/uid?

mzhangmzz:
> > Today they will specify `tracked_shards: [0]`
> 
> I suggest to take a step back and ask ourselves why would anyone care about shard number X? and how would they know that they _want_ to track exactly the shard number X? Maybe they care about a set of specific accounts and we should track based on the list of account ids instead of the shard ord/id/uid?

I think that's a very good point. Maybe the outside world should not have information about the internal shard layout at all, since we want the sharding implementation to be as transient as possible. We can provide two options, tracking all shards, or tracking shards from accounts.

bowenwang1996:
> Maybe they care about a set of specific accounts and we should track based on the list of account ids instead of the shard ord/id/uid?

It seems that there are the following use cases:
* Tracking all shards.
* Tracking a set of accounts.

Today we already support tracking a set of accounts through `tracked_accounts`. We just need to make sure that this continues to work after having multiple shards. In addition, we need to make sure that people who want to track all shards have a way to do that.

specs/Proposals/0040-split-states.md
```diff
@@ -36,11 +36,21 @@ This in turn involves two parts, how the validators know when and how sharding c
 The former is a protocol change and the latter only affects validators' internal states.
 
 ### Protocol Change
-Sharding config for an epoch will be encapsulated in a struct `ShardLayout`, which not only contains number of shards, but also layout information to decide which account ids should be mapped to which shards. The `ShardLayout` information will be stored as part of `EpochConfig`. Right now, `EpochConfig` is stored in `EpochManager` and remains static across epochs. That will be changed in the new implementation so that `EpochConfig` can be changed according to protocol versions, similar to how `RuntimeConfig` is implemented right now.
+Sharding config for an epoch will be encapsulated in a struct `ShardLayout`, which not only contains number of shards, but also layout information to decide which account ids should be mapped to which shards.
```
bowenwang1996, line 39:
```suggestion
Sharding config for an epoch will be encapsulated in a struct `ShardLayout`, which not only contains the number of shards, but also layout information to decide which account ids should be mapped to which shards. 
```

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,381 @@
...
+    V0(ShardLayoutV0),
+    V1(ShardLayoutV1),
+}
+```
+ShardLayout is a versioned struct that contains all information needed to decide which accounts belong to which shards. Note that `ShardLayout` only contains information at the protocol level, so it uses `ShardOrd` instead of `ShardId`. 
+
+The API contains the following two functions.
+#### `parent_shards`
+```rust
+pub fn parent_shards(&self) -> Vec<ShardOrd>
```
bowenwang1996, line 120: We no longer use `ShardOrd` right?

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,381 @@
...
+maps account id to shard id given a shard layout
+
+#### `ShardLayoutV0`
+```rust
+pub struct ShardLayoutV0 {
+    /// map accounts evenly across all shards
+    num_shards: NumShards,
+}
+```
+A shard layout that maps accounts evenly across all shards. This is added to capture the current `account_id_to_shard_id` algorithm, to keep backward compatibility for some existing tests. `parent_shards` for `ShardLayoutV1` is always `None` and `version`is always `0`.
```
bowenwang1996, line 146: Probably worth explaining what "evenly" actually means

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,381 @@
...
+`ShardUId` is a unique identifier that a validator uses internally to identify shards from all epochs. It only exists inside a validator's internal state and can be different among validators, thus it should never be exposed to outside APIs.
+
+```rust
+pub struct ShardUId {
+    pub version: ShardVersion,
+    pub shard_id: u32,
+}
+```
+#### `TrieCachingStorage`
+Trie storage will be contruct database key from `ShardUId` and hash of the trie node.
```
bowenwang1996, line 201:
```suggestion
Trie storage will construct database key from `ShardUId` and hash of the trie node.
```

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,381 @@
...
+### `split_state_changes`
+```rust
+split_state_changes(shard_id: ShardId, state_changes: &Vec<RawStateChangesWithTrieKey>) -> HashMap<ShardId, Vec<RawStateChangesWithTrieKey>>
+```
+splits state changes to be made to a current shard to changes that should be applid to the new shards. Note that this function call can take a long time. To avoid blocking the client actor from processing and producing blocks for the current epoch, it should be called from a separate thread. Unfortunately, as of now, catching up states and catching up blocks are both run in client actor. They should be moved to a separate actor. However, that can be a separate project, although this NEP will depend on that project. In fact, the issue has already been discussed in [#3201](https://github.com/near/nearcore/issues/3201).
+
+### `apply_chunks`
+`apply_chunks` will be modified so that states of the new shards will be updated when processing chunks.
+ In `apply_chunks`, after processing each chunk, the state changes in `apply_results` are sorted into changes to new shards.
+At the end, we apply these changes to the new shards.
```
bowenwang1996, line 286: Ideally this should also be done in a different thread, but for this specific upgrade I think it is fine

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,381 @@
...
+## Garbage Collection
+The old states need to be garbage collected after the resharding finishes. The garbage collection algorithm today won't automatically handle that. (#TODO: why?)
+
+Althought we need to handle garbage collection eventually, it is not a pressing issue. Thus, we leave the discussion from this NEP for now and will add a detailed plan later.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The drawback of this approach is that it will not work when challenges are enabled since challenges to the transition to the new states will be too large to construct or verify.
+Thus, most of the change will likely be a one time use that only works for the Simple Nightshade transition, although part of the change involing `ShardId` may be reused in the future.
```
bowenwang1996, line 341:
```suggestion
Thus, most of the change will likely be a one time use that only works for the Simple Nightshade transition, although part of the change involving `ShardId` may be reused in the future.
```

specs/Proposals/0040-split-states.md
```diff
@@ -0,0 +1,381 @@
...
+- What is the impact of not doing this?
+  - The impact will be the delay of launching Simple Nightshade, or no launch at all.
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+- What parts of the design do you expect to resolve through the NEP process before this gets merged?
+  - Garbage collection
+  - State Sync?
+- What parts of the design do you expect to resolve through the implementation of this feature before stabilization?
+  - There might be small changes in the detailed implemenations or specifications of some of the functions described above, but the overall structure will not be changed.
```
bowenwang1996, line 360:
```suggestion
  - There might be small changes in the detailed implementations or specifications of some of the functions described above, but the overall structure will not be changed.
```

specs/Proposals/0040-split-states.md
mzhangmzz: Yes, sorry I haven't been keeping this NEP updated enough since there are too many changes. Just updated that.

specs/Proposals/0040-split-states.md
```diff
@@ -197,8 +206,17 @@ pub struct ShardUId {
...
 }
 ```
+
+`version` in `ShardUId` comes from the version of `ShardLayout` that this shard belongs. This way, different shards from different shard layout will have different `ShardUId`s.
+
+### Database storage
+The following database columns are stored with `ShardId` as part of the database key, it will be replaced by `ShardUId`
+- ColState
+- ColChunkExtra
+- ColTrieChanges
```
bowenwang1996, line 216: Nice!

specs/Proposals/0040-split-states.md
```diff
@@ -197,8 +206,17 @@ pub struct ShardUId {
...
+
 #### `TrieCachingStorage`
-Trie storage will be contruct database key from `ShardUId` and hash of the trie node.
+Trie storage will contruct database key from `ShardUId` and hash of the trie node.
```
bowenwang1996, line 219:
```suggestion
Trie storage will construct database key from `ShardUId` and hash of the trie node.
```

bowenwang1996 (commented):
Thanks for the update!

frol:
@mzhangmzz @bowenwang1996 Is it ready to be merged?

bowenwang1996:
This was implemented more than a year ago so let's merge the pull request. @mzhangmzz could you resolve conflicts and merge the PR? Thanks!

PR #255, opened 2021-09-07
fix: resolve inconsistent types

lilses:
Most NEPs have been written in Typescript interfaces and others Rust.

Since smart contracts in near are primarily written in Rust, it makes sense to use Rust. But definitely having 2 different language interfaces is confusing and has led to this pr.

We should have 1 language interface instead of 2. I hope this is the start of discussions that will lead to improved standards.


mikedotexe (changes requested):

frol:
It seems that this PR got outdated by this time. I am going to close it now, but if anyone wants to follow up, please feel free to create a new PR.

Issue #254, opened 2021-09-06
FT NEP-141, NFT NEP-171 require having minting/burning interface


telezhnaya:
Our users ask for supporting FT/NFT in the tools such as Explorer and Wallet. We want to show FT/NFT balance and the history of all FT/NFT movements.

To achieve that, we need to track all the FT/NFT balance-changing events. They include minting, transferring, and burning of the tokens.
For now, it's only possible to track transfers.

While digging into FT implementation, we've found [`internal_transfer`](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/fungible_token/core_impl.rs#L130) method, it looks like what we need. But it's the details of one exact implementation, and it's not possible to catch these calls, it's deep inside the contract. We can only see public interface calls such as `ft_transfer`.

It’s a critical moment for supporting FT/NFT in our and 3rd-party tools, our data would be incomplete in other cases.
We should include minting/burning in NEPs. Without that, we are blocked, we can't show valid FT/NFT balances.

Please share any thoughts, including other workarounds to solve the issue.

Previous issue #250 
Issue with collecting balance-changing events in Indexer for Explorer https://github.com/near/near-indexer-for-explorer/pull/153

@chadoh @mikedotexe @mattlockyer @MaximusHaximus @jberrytech @frol

telezhnaya:
FT NEP-141 https://nomicon.io/Standards/FungibleToken/Core.html
NFT NEP-171 https://nomicon.io/Standards/NonFungibleToken/Core.html

frol:
I want to highlight that currently Wallet uses some hacks that only allows it to find potential FT / NFT contracts that the account might be involved with and then tries to fetch the balance from there. There are several problems with that:

1. It is not reliable - if the person only mints the tokens, that asset won't be detected (e.g. wNEAR uses a method called `near_deposit` to mint tokens)
2. It is not efficient - Wallet scans JSONB column inside PostgreSQL database and with the grows of the number of transactions it will just timeout one day
3. It cannot get a history of operations and thus limited to only displaying the current balance

Explorer is also blocked on it since we want to display a feed of events instead of just pure transactions (e.g. when FT is transferred, the receiver account might not be even touched b native NEAR receipts and thus Explorer won't display those incoming transfers in the first place)

It is a complex topic, but we want to raise it sooner rather later to benefit across  the ecosystem and make NEAR user and developer friendly platform!

frol:
[...]

PR #245, opened 2021-07-28
feat:  Multi Token Standard Contract [NEP]

zcstarr:
This is a pr for the NEP for a new Multi Token Standard

PR #256, opened 2021-09-08
feat: add batch

evergreen-trading-systems:


frol:
Hey, I feel that we should define our standards to be as slim as possible. This standard extension still makes sense, but from what I have heard, there are too many ways of how tokens can be minted/burned, so it feels that those cannot be limited to just certain method names (it can be implemented that way, but it might just feel unnatural.

As per my https://github.com/near/NEPs/issues/254#issuecomment-914407269, I feel there are just 3 high-level scenarios that need to be standardized:

1. Observability of minting, transfer, burning events
2. Interface to transfer assets
3. Interface to get status, i.e. balance (for FT) and token-by-owner and owner-by-token (for NFT)

Batch operations are great, but I would keep them an extension. I find this current proposal to be quite complicated in terms of the interface, and at the same time, it might limit some other use-cases. Batch interfaces are not required by any of the listed scenarios, and while it might address my concerns about observability, I have heard that it is not realistic to expect people to create workarounds with cross-contract calls just to make sure to have a call with `nft_batch_*` as a method name. I feel that log-events is the way to go, and we can easily document that there can be more than a single mint/transfer/burn events in a single execution and that will completely solve (1) with just a minor tweak to the existing implementations and it does not interfere with this batch extension.

Thus, I suggest we discuss the observability in #254 and consider if these interfaces still need to be defined.

zcstarr:
just commenting here to ask if approval makes sense as an event @frol @evergreen-trading-systems, then downstream subscribers like app market places know when approval changes. They could just subscribe to the events when the approval happens. 

Otherwise they'd need to poll the chain. I think revocation isn't as important because if you're going to make tx assuming revocation, can just do an `is_approved` check before hand, no polling required.

The erc-721 event is :
```
    /// @dev This emits when the approved address for an NFT is changed or
    ///  reaffirmed. The zero address indicates there is no approved address.
    ///  When a Transfer event emits, this also indicates that the approved
    ///  address for that NFT (if any) is reset to none.
event Approval(address indexed _owner, address indexed _approved, uint256 indexed _tokenId);
```
I think we'd need approval_id, approved_account , owner, token_id . What do you all think? I think for #245 we're looking at the same thing.

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,92 @@
...
+events of these activities through logs.
+
+```ts
+// Interface to capture an event
+// and return formatted event string.
+interface EventJsonLog {
+    // Takes `EventLogData` and returns
+    // `EVENT_JSON: <EVENT_LOG_DATA>`
+    (e:EventLogData):string
+}
```
evergreen-trading-systems, line 41: This notifies implementors of their need to produce the final string compatible with the indexer. Look at line 39 `EVENT_JSON {"standard":"nep171", ...}`.

This is the response to https://github.com/near/NEPs/issues/254#issuecomment-928249174 to capture all kinds of events not just json.

evergreen-trading-systems:
@zcstarr Maybe no This whole need for event logs started because different people had different ways of doing the same thing and the wallet only displayed certain formats. The 3 main things every NFT marketplace is doing and already being logged are
- minting
- burning
- transfer

but only in a specific format so not every marketplace are seeing their transactions on the wallet.

That's why `approval` doesn't make sense first but after. I wish there was more than 1 wallet app today. When integration with metamask happens, issue like this won't move slow.

zcstarr:
> @zcstarr Maybe no This whole need for event logs started because different people had different ways of doing the same thing and the wallet only displayed certain formats. The 3 main things every NFT marketplace is doing and already being logged are

It's bigger than just wallet. Events help dapps and esp. server side dapps not have to poll the blockchain via calls. This enables those apps to be able to react to changes in state. 

If you look across the pond at ETH, alot of things are enabled by being able to wait on event and events.  It's just more efficient, you can simply check the incoming block data for events or filter the data to know about a change in state, that you might have to react to. 

I think approval probably is one of those changes in state. To flip it on it's head :)  @evergreen-trading-systems why would it not be important as a consuming app to know that the approval has changed?

evergreen-trading-systems:
I agree with you @zcstarr . Anything that is useful to index should have an event. Currently, the only implementors active in this conversation are the wallet developers. They are concerned with the 3 events like I said. Others may concerned with others.

I see the need for 2 prs
1. Standard structure that every event must adhere to e.g. `<event-type> <event-log-data>`
  a. `EVENT_JSON {...}`
  b. `EVENT_BINARY ...`
2. Types of events - `nft_mint`, `nft_burn`, `nft_transfer`, `nft_approval` etc

This separates the concern of events only being useful to wallet indexers and starts to be useful for all kinds of indexers. I will make the pr.

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,92 @@
...
+
+Version `1.0.0`
+
+## Summary
+
+Standard interfaces for NFT contract actions.
+
+## Motivation
+
+Nft driven apps such as marketplaces regularly perform many
```
mikedotexe, line 11:
```suggestion
NFT-driven apps such as marketplaces regularly perform many
```

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,92 @@
...
+of the same acions such as `minting`, `burning` and `transfering`.
```
mikedotexe, line 12:
```suggestion
of the same actions such as `minting`, `burning` and `transferring`.
```

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,92 @@
...
+Each app have their own way of performing these actions and it
+is difficult to consistently capture these events.
+This extension addresses that.
+
+For example, it's common in NFT marketplaces to have
+different methods of transfer a single token vs many 
+tokens. Other applications like wallets, will have
+a difficult time tracking consistently this information for many
+markets. This makes interactive with many nft driven apps
```
mikedotexe, line 21:
```suggestion
markets. This makes interactive with many NFT-driven apps 
```

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,92 @@
...
+For example, it's common in NFT marketplaces to have
+different methods of transfer a single token vs many 
+tokens. Other applications like wallets, will have
+a difficult time tracking consistently this information for many
+markets. This makes interactive with many nft driven apps 
+infeasible.
+
+Due to discussions here 
+https://github.com/near/NEPs/issues/254,
+a standard way to capture these events are needed.
```
mikedotexe, line 26:
```suggestion
a standard way to capture these events is needed.
```

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Interface
+
+Many apps may use different mechanisms to perform their batching
+requirements. This interface standardises this process by capturing
```
mikedotexe, line 31:
```suggestion
requirements. This interface standardizes this process by capturing
```

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,92 @@
...
+// * `sender_id`: "account.near"
+// * `receiver_id`: "receiver.near"
+// * `token_id`: "12345abc"
+interface NftTransferLog {
+    sender_id:string,
+    receiver_id:string,
+    token_id:string
+}
+```
+
```
mikedotexe, line 86:
```suggestion

## Further methods

Note that the example events covered above cover two different kinds of events:
1. Events that are not specified in the NFT Standard (`nft_mint`, `nft_burn`)
2. An event that is covered in the [NFT Core Standard](https://nomicon.io/Standards/NonFungibleToken/Core.html#nft-interface).

 This event standard also applies beyond the three events highlighted here, where future events follow the same convention of as the second type. For instance, if an NFT contract uses the [approval management standard](https://nomicon.io/Standards/NonFungibleToken/ApprovalManagement.html), it may emit an event for `nft_approve` if that's deemed as important by the developer community.
 
 Please feel free to open pull requests for extending the events standard detailed here as needs arise.
```

mikedotexe (changes requested):
Thanks for the patience, will rally other folks now.

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,101 @@
...
+    data: NftMintLog[]|NftTransferLog[]|NftBurnLog
+}
+
+// An event log to capture token minting
+// Arguments
+// * `account_id`: "account.near"
+// * `token_id`: "1"
+interface NftMintLog {
+    owner_id:string,
+    token_id:string
```
telezhnaya, line 63: We need to add an optional `memo` field here and to all other event log types.

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// An event log to capture token burning
+// Arguments
+// * `owner_id`: owner of tokens to burn
+// * `token_ids`: ["1","2"]
+interface NftBurnLog {
+    owner_id:string,
+    token_ids:string[]
```
telezhnaya, line 72: Are we sure we want to pass here the list? I think it's better to use here `token_id` as everywhere else not to mess up.

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,101 @@
...
+    receiver_id:string,
+    token_id:string
+}
+```
+
+## Further methods
+
+Note that the example events covered above cover two different kinds of events:
+1. Events that are not specified in the NFT Standard (`nft_mint`, `nft_burn`)
+2. An event that is covered in the [NFT Core Standard](https://nomicon.io/Standards/NonFungibleToken/Core.html#nft-interface).
```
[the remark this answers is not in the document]
mikedotexe: https://github.com/near/NEPs/pull/256/files#r726530405

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,101 @@
...
+// Arguments
+// * `standard`: name of standard e.g. nep171
+// * `version`: e.g. 1.0.0
+// * `event`: `nft_mint` | `nft_burn` | `nft_transfer`
+// * `data`: associate event data
+interface EventLogData {
+    standard:string,
+    version:string,
+    event:string,
+    data: NftMintLog[]|NftTransferLog[]|NftBurnLog
```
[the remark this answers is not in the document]
mikedotexe: https://github.com/near/NEPs/pull/256/files#r726534183

specs/Standards/NonFungibleToken/Event.md
mikedotexe: I think in the majority of cases for batch minting and burning, these will be from the same user account and this makes sense. Similar to suggestions I'm seeing in the [multi-token standard](https://github.com/near/NEPs/pull/245).

telezhnaya: Do we want to pass the list of ids in mint also?

Not suggesting doing that for transfer, it sounds weird. But I didn't know the case with burn, so I could be wrong here.

mikedotexe: Good point, I definitely think we should do that with mint

frol: @evergreen-trading-system @telezhnaya @mikedotexe I believe that if we go with batching all the way down, we should bake this support everywhere: mint, transfer, burn.

```js
JSON_EVENT {"standard": "nep171", "version": "1.0.0", "event": "mint": "data": [
  {"owner_id": "frol.near", "token_ids": ["1", "2"]}
]}

JSON_EVENT {"standard": "nep171", "version": "1.0.0", "event": "transfer": "data": [
  {"sender_id": "frol.near", "receiver_id": "olga.near", "token_ids": ["1", "2"]}
]}

JSON_EVENT {"standard": "nep171", "version": "1.0.0", "event": "burn": "data": [
  {"owner_id": "frol.near", "approval_id": "olga.near", "token_ids": ["1", "2"]},
  {"owner_id": "frol.near", "token_ids": ["3"]}
]}
```

frol: @mikedotexe Do you have a good naming for "an account that was approved to perform transfer/burn operations on behalf of the NFT owner"? I currently named it as "representative_id" just for the sake of no better name I could come up with. I feel it is similar to "Intermediate / Corresponding Bank", but I am really bad in that area of terminology.

specs/Standards/NonFungibleToken/Event.md
evergreen-trading-systems: What does `memo` represent?

I'm not aware of any `nft_mint` function that uses memo. For example https://github.com/near/near-sdk-rs/blob/master/examples/non-fungible-token/nft/src/lib.rs#L92

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,101 @@
...
+// * `account_id`: "account.near"
+// * `token_id`: "1"
+interface NftMintLog {
+    owner_id:string,
+    token_id:string
+}
+
+// An event log to capture token burning
+// Arguments
+// * `owner_id`: owner of tokens to burn
```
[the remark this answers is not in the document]
evergreen-trading-systems: Interesting - performing an action on behalf of the owner. Do we want this information in logs? Because all actions could be performed on behalf of owners and we'll need `approval_id` in all events. 

The more information we add, the more chance of hitting the 16kb limitation is. I based these events on how they are actually being used today:

https://github.com/zavodil/pluminite-ui/blob/master/contract/src/burn.rs#L7
https://github.com/near-apps/uhhm-market/blob/main/contracts/nft-simple/src/burn.rs#L7
https://github.com/ParasHQ/paras-nft-contract/blob/main/paras-nft-contract/src/lib.rs#L579

If this is not priority now, in the next version we can add this.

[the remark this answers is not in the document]
mikedotexe: Agreed to not call it `approval_id` as that's a number and we're only wanting to emit info on the account that burned it. Still needs to add some name for that.
TODO good naming for this.

evergreen-trading-systems:
Only thing pending https://github.com/near/NEPs/pull/256#discussion_r728170028

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,105 @@
...
+// * `standard`: name of standard e.g. nep171
+// * `version`: e.g. 1.0.0
+// * `event`: `nft_mint` | `nft_burn` | `nft_transfer`
+// * `memo`: optional message
+// * `data`: associate event data
+interface EventLogData {
+    standard:string,
+    version:string,
+    event:string,
+    memo?:string,
```
telezhnaya, line 51:
```suggestion
```

evergreen-trading-systems:
Are we adding `ApprovalEventLog` https://github.com/near/NEPs/pull/256#discussion_r725971483 and are we keeping all events in one file or separating per event?

frol:
> Are we adding `ApprovalEventLog` and are we keeping all events in one file or separating per event?

I suggest we strike to arrive at a state where all the relevant pieces are right at the places where we describe the concept. Thus, I vote for having:

1. a separate NEP for generic events (where we only define events as logs with `EVENT_JSON` prefix and `{"standard": string, "version": string, "event": string, "data": any}` basic structure)
2. mint/transfer/burn events in the core document of NFT standard
3. define other events (e.g. approval) in the documents that describe the given concept

That said, given I don't want us to extend the scope of this PR, let's proceed as was decided originally, but given the discussion about `ApprovalEventLog`, I suggest we start with a "refactoring" (extract the generic events into a separate NEP) and then define `ApprovalEventLog` referencing the generic events NEP

mikedotexe:
> Are we adding `ApprovalEventLog` [#256 (comment)](https://github.com/near/NEPs/pull/256#discussion_r725971483) and are we keeping all events in one file or separating per event?

As discussed in the meeting, it's like Frol said. There will be a broader standard (similar to the Storage Management standard) that describes events using logs. Then, as folks from the community discuss how this applies to various standards — like we're doing for the NFTs — will have a separate section under that standard. This is exactly how we have this PR currently.

> I'm wary about approvals for burn and mint. I'm not aware of approvers minting on behalf of minters and burners burning on behalf of owners. Actually it's not used at all in that way today.
>As a first release, we should prioritise for known use cases that are already in operation. Other useful but lesser known features can be added in subsequent versions.

No one has suggested approvals for minting. That doesn't make sense because a token has to be created first to have approvals. As pointed out in the meeting, burning is already how it works in other ecosystems like Ethereum as shown by the OpenZeppelin implementations. 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e2876b947db9ad79f0bdb0d8b044de3803cd96e9/contracts/token/ERC721/ERC721Burnable.sol#L22

Feels like we had a very solid agreement at the end of the meeting, so I'm surprised by these new suggetions.

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,108 @@
...
+    (e:EventLogData):string
+}
+
+// Interface to capture data 
+// about an event
+// Arguments
+// * `standard`: name of standard e.g. nep171
+// * `version`: e.g. 1.0.0
+// * `event`: `nft_mint` | `nft_burn` | `nft_transfer`
+// * `memo`: optional message
```
mikedotexe, line 45:
```suggestion
```
Think this was an accidental copy/paste

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,101 @@
...
+}
+```
+
+## Further methods
+
+Note that the example events covered above cover two different kinds of events:
+1. Events that are not specified in the NFT Standard (`nft_mint`, `nft_burn`)
+2. An event that is covered in the [NFT Core Standard](https://nomicon.io/Standards/NonFungibleToken/Core.html#nft-interface).
+
+ This event standard also applies beyond the three events highlighted here, where future events follow the same convention of as the second type. For instance, if an NFT contract uses the [approval management standard](https://nomicon.io/Standards/NonFungibleToken/ApprovalManagement.html), it may emit an event for `nft_approve` if that's deemed as important by the developer community.
```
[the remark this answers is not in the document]
evergreen-trading-systems: So going to make separate pr for approval events

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,107 @@
...
+// An event log to capture token transfer
+// Arguments
+// * `authorised_id`: approved account to transfer
+// * `sender_id`: "account.near"
+// * `receiver_id`: "receiver.near"
+// * `token_ids`: ["1", "12345abc"]
+// * `memo`: optional message
+interface NftTransferLog {
+    authorised_id:string,
+    sender_id:string,
```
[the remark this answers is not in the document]
frol: I think @telezhnaya suggested:

```js
{
owner_id:"owner.near",
authorised_id:"authorised.sender.near",
receiver_id:"receiver.near",
token_ids: ["1"]
}
```

Though, if we change sender to owner we get a different sort of confusion, when `authorised_id` is omitted:

```js
{
owner_id:"owner.near",
receiver_id:"receiver.near",
token_ids: ["1"]
}
```

I don't have a suggestion so far, but I agree with (sender, authorised, receiver) being confusing since sender is not actually sending in this case, and (owner, receiver) is not immediately clear as well.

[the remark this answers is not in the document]
evergreen-trading-systems:
```ts
{
old_owner_id:"owner.near",
authorised_id:"authorised.sender.near",
new_owner_id:"receiver.near",
token_ids: ["1"]
}
```

Makes sense. Though it won't have symmetrical interface with others. Is this ok?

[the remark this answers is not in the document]
evergreen-trading-systems: Can't answer on how `authorised_id` will be used but I can say all useful events in general e.g. `ApprovalEvent` you mentioned is part of the plan - standard nft events. 

I will be creating another pr shortly regarding and I invite to mention as many events you think should be standardised.

frol (approved):
I believe this document is good enough to be merged. It addresses the observability problem for Indexers to pick up the information about common NFT operations.

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,107 @@
...
+This interface standardizes that process.
+It captures these actions through logs.
+
+```ts
+// Interface to capture an event
+// and return formatted event string.
+interface EventJsonLog {
+    // Takes `EventLogData` and returns
+    // `EVENT_JSON: <EVENT_LOG_DATA>`
+    (e:EventLogData):string
```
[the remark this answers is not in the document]
telezhnaya: Are we sure that we don't want to put everything into json with the structure like
```
{
"type": "event_json",
"data": [...] // here could be all events
}
```
?

It simplifies everything, we don't need to overcomplicate this doc, Indexer does not need the logic with cutting/trimming.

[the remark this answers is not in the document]
evergreen-trading-systems: @mikedotexe

specs/Standards/NonFungibleToken/Event.md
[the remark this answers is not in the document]
zcstarr: thanks @frol makes sense!

specs/Standards/NonFungibleToken/Event.md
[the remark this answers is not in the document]
frol: @telezhnaya Individual calls to `near_sdk::env::log` are recorded as individual strings in the `logs` array of the execution outcome, so we don't care about new lines inside. We only want to specify that each event should be a separate call to the log host function.

Proof: https://explorer.testnet.near.org/transactions/326HmzvgRYND7skVCdTpCnHMji6qiU8JFhQMNuvS2b74
image

Contract:

```rust
use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
use near_sdk::{near_bindgen, setup_alloc};

setup_alloc!();

#[near_bindgen]
#[derive(BorshDeserialize, BorshSerialize, Default)]
pub struct Test {}

#[near_bindgen]
impl Test {
    pub fn test_log(&mut self) {
        near_sdk::env::log(r#"EVENT_JSON:{
            "standard": "nep",
            "version": "1.0.0",
            "event": "mint",
            "data": [
                {"owner_id": "frol.near", "token_ids": ["house"]}
            ]
        }"#.as_bytes());
        near_sdk::env::log("EVENT_JSON:{\"standard\": \"nep\", \n \"version\": \"1.0.0\",    \n   \"event\": \"burn\", \"data\": [{\"owner_id\": \"frol.near\", \"token_ids\": [\"house\"]}]}".as_bytes());
    }
}
```

frol:
@evergreen-trading-systems Please, review my latest commit (I wanted to make a proposed change, but it was really hard to read, so I decided to commit it (if that is too much, we can always revert the commit)

evergreen-trading-systems:
Thanks @frol your changes made this NEP much more concise. 

1 question. Is it true you can have a valid event log with no data i.e. `EVENT_JSON:{"standard": "nepXXX", "version": "1.0.0", "event": "xyz_is_triggered"}`

evergreen-trading-systems:
Cool makes sense. There is no strong need for it to be now so agree with case-by-case.

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,200 @@
...
+interface NftMintLog {
+    owner_id: string,
+    token_ids: string[],
+    memo?: string
+}
+
+// An event log to capture token burning
+// Arguments
+// * `owner_id`: owner of tokens to burn
+// * `authorised_id`: approved account to burn, if applicable
```
mikedotexe, line 107:
```suggestion
// * `authorized_id`: approved account to burn, if applicable
```
Hate to be nit-picky about this, but I do see other references in the NEP to the word "authorize" and "authorized" and think this will be an introduction of a word spelling that is inconsistent with numerous cases in our docs, NEAR CLI, nearcore, Rust SDK, and other NEPs.

So I think we'll have to stick with the spelling we have in these other places for consistently, lest we cause confusion for end users.


Screen Shot 2021-10-21 at 2 25 37 PM
Screen Shot 2021-10-21 at 2 23 23 PM
Screen Shot 2021-10-21 at 2 22 05 PM
Screen Shot 2021-10-21 at 2 20 42 PM
Screen Shot 2021-10-21 at 2 19 42 PM

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,200 @@
...
+// * `token_ids`: ["1","2"]
+// * `memo`: optional message
+interface NftBurnLog {
+    owner_id: string,
+    authorised_id?: string,
```
mikedotexe, line 112:
```suggestion
    authorized_id?: string,
```

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,200 @@
...
+    token_ids: string[],
+    memo?: string
+}
+
+// An event log to capture token transfer
+// Arguments
+// * `authorised_id`: approved account to transfer
```
mikedotexe, line 119:
```suggestion
// * `authorized_id`: approved account to transfer
```

specs/Standards/NonFungibleToken/Event.md
```diff
@@ -0,0 +1,200 @@
...
+// * `old_owner_id`: "owner.near"
+// * `new_owner_id`: "receiver.near"
+// * `token_ids`: ["1", "12345abc"]
+// * `memo`: optional message
+interface NftTransferLog {
+    authorised_id?: string,
```
mikedotexe, line 125:
```suggestion
    authorized_id?: string,
```

mikedotexe (approved):
Great work everyone

PR #263, opened 2021-09-27
Update description to mark approval_id as optional

danielwpz:


frol (approved):
As a NEP moderator, I approve this PR given that it does not change the behavior and only clarifies the text descriptions

PR #268, opened 2021-10-10
feat: add event-standard

lilses:
Based on https://github.com/near/NEPs/pull/256#issuecomment-935191107

Need to separate standards of event vs. the information it captures.

### Event Standard

This pr says every event should adhere to common format - json, binary, etc.

### DIFF
--- a/specs/Standards/Events/Core.md
+++ b/specs/Standards/Events/Core.md
@@ -1,0 +1,60 @@
+# Event Logs
+
+Version `1.0.0`
+
+## Summary
+
+A standard interface for logging events including
+JSON, Binary and future formats. This addresses
+of the problem of duplicate events in inconsistent
+formats needing different logic to parse.
+
+## Motivation
+
+Near apps commonly log data on chain to be indexed
+off chain. For every kind of dApp, there are generic sets
+of information that are useful to index:
+- Wallet like Apps
+    - minting events
+    - transfer events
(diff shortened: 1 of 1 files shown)


evergreen-trading-systems:
wasn't thinking of that but yes, do you mean this - https://github.com/near/NEPs/pull/268/files#diff-54ad87ad1f7568238db42e0751f9ceecdcf416d189dc7896e6a565aee0bc2140R51 

As this standard evolves and reaches a more stable version e.g. >= 1.0, we can expect less disruptive changes for implementors. Of course it shouldn't be enforced so if v0.5 works for consumer A then potentially still be able to parse v1.0 events but some fields would be null.

evergreen-trading-systems:
Yes the limitation is 16kb string. It is already limited e.g. a batch with very large number is not possible. Hoping for a good baseline but is a known drawback.

While it is a good idea to version events themselves, soon will end up something like this

`NEP1_JSON {"version":"1.0.0", "standard":"nep1"}`
`NEP2_JSON {"version":"1.0.0", "standard" :"nep1"}`

This could be seen as unnecessary duplication or a very complex versioning system.

We want the events to be generic and not only for NFTs. Most of the specificities are in the `data` field, the rest is generic.

frol:
This PR was taken over in #298 and resulted in NEP-297.

PR #273, opened 2021-11-02
fix(nft-payouts): Missing memo argument.

willemneal:
As the title says the `memo` argument in the `nft_transfer_payout` was missing.

However, after reviewing the details of this standard I also have several questions.

1.  Why is a the payouts a struct and not a type alias. e.g.
```rust
type Payouts = HashMap<AccountIds, U128>;
```

(update: This is how Paras defines the type).

2. Where did the max length of 10 come from?  Since the consumer is going to iterate over the entries would a vector of tuples decrease this cost?

3. If the contract call would fail with too many payouts anyway, why introduce this implementation detail into the API?

4. As the `nft_transfer_payout` default implementation shows, it is made up of calls to external APIs.  So why does the contract need to implement it?

### DIFF
--- a/specs/Standards/NonFungibleToken/Payout.md
+++ b/specs/Standards/NonFungibleToken/Payout.md
@@ -87,7 +87,7 @@ pub trait Payouts {
     receiver_id: AccountId,
     token_id: String,
     approval_id: u64,
-    memo: Option<u64>
+    memo: Option<u64>,
     balance: U128,
     max_len_payout: u32,
   ) -> Payout {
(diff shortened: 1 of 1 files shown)


specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -80,12 +87,13 @@ pub trait Payouts{
     receiver_id: AccountId,
     token_id: String,
     approval_id: u64,
+    memo: Option<u64>
```
willemneal, line 90:
```suggestion
    memo: Option<u64>,
```

PR #276, opened 2021-11-17
feat(NFT): Add option types and add missing memo parameter

willemneal:
This adds the missing memo argument needed by `nft_transfer` and changes types  inspired by Paras' usage. Notably using a type alias for `Payout` of a hashmap.  Also adds more options so that the API is less strict.

https://github.com/ParasHQ/paras-marketplace-contract/blob/ac35d84a1af7ea369c3916db32af0dd1e5a43024/paras-marketplace-contract/src/external.rs#L7-L14

```rust
 fn nft_transfer_payout(
        &mut self,
        receiver_id: AccountId,
        token_id: TokenId,
        approval_id: Option<u64>,
        memo: Option<String>,  // This is ignored by Paras
        balance: U128,
        max_len_payout: Option<u32>,
    );
```

- [x] Need to specify that a `nft_transfer_payout` that returns a bad "payout" should still payout to old token owner


emarai:
> I'm unsure why the balance would be optional, so if they could explain their choice that would be great.

Hi @willemneal , regarding why the balance is optional, we picked it up from near-examples/nft-market. But it seems it has been changed to be no longer optional (https://github.com/near-examples/nft-market/commit/ba2b5ce07043061059a5482e815433101d4c455f#diff-bd87736f4e1006e27d541b8adc5bc34e439b4c1ce8c234cefc752408d01eb369L130)

microchipgnu:
I agree with removing the wrapper on Payout. Don't see its utility.

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -80,18 +83,27 @@ pub trait Payouts{
...
+    self.nft_transfer(receiver_id, token_id, approval_id, memo);
     payout
   }
 }
 ```
 
-Note that NFT and financial contracts will vary in implementation. This means that some extra CPU cycles may occur in one NFT contract and not another. Furthermore, a financial contract may accept fungible tokens, native NEAR, or another entity as payment. Transferring native NEAR tokens is less expensive in gas than sending fungible tokens. For these reasons, the maximum length of payouts may vary according to the customization of the smart contracts.
+## Fallback on error
+
+In the case where either the `max_len_payout` causes a panic, or a malformed `Payout` is returned, the caller contract should transfer all funds to the token owner.
```
BenKurrek, line 100: Example scenario background information:
- Benji puts token A for sale for 100 NEAR. 
- token A has a payout object which dictates that Mike should get 20 NEAR and Benji should get 80 NEAR. 
- Willem comes along and purchases the NFT for 100 NEAR on the marketplace.

Logic: 
- Marketplace contract calls `nft_transfer_payout` on the NFT contract which will transfer the NFT to willem and send back a payout object. 
- That payout object as per the current standard was wrapped in a struct. Your new changes propose it should just be a type (just a hashmap). 

In your new proposal: 
- If the NFT contract sends back a struct: 
```
{
  "payout": {
    "mike": "20000000000000000000000000",
    "benji": "80000000000000000000000000",
  }
}
```
This should transfer all the funds (100 NEAR) to the token owner, which is Willem in this case. That means Willem got the NFT for free essentially (which is fine because so sorry Benji, but your NFT was minted on an improper contract that didn't implement the standards correctly). 

- If the NFT contract sends back a payout object: 
```
{
    "mike": "20000000000000000000000000",
    "benji": "80000000000000000000000000",
}
```
Everything would be fine and the marketplace should pay mike 20 NEAR and benji 80 NEAR.

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -80,18 +83,27 @@ pub trait Payouts{
...
+
+## Potential pitfalls
+
+The payout must include all accounts that should receive funds. Thus the token owner must be included.
```
BenKurrek, line 104: I wouldn't force the token owner in here. What if an NFT contract wanted an NFT where the owner didn't get any funds?

BenKurrek (changes requested):

specs/Standards/NonFungibleToken/Payout.md
willemneal: This should be reworded.  The point is that it's a pitfall to assume that you only need to specify those who get royalties and the market will send the rest to the owner.  I agree with the payout being the list of all who should receive funds.

specs/Standards/NonFungibleToken/Payout.md
willemneal: I'll change it to original `token_owner`.  The seller should be the default.

willemneal, line 100:
```suggestion
In the case where either the `max_len_payout` causes a panic, or a malformed `Payout` is returned, the caller contract should transfer all funds to the original token owner selling the token.
```

specs/Standards/NonFungibleToken/Payout.md
willemneal, line 104:
```suggestion
The payout must include all accounts that should receive funds. Thus it is a mistake to assume that the original token owner will receive funds if included.
```

BenKurrek (approved):

willemneal:
I reverted the payout type from this PR.  It can be discussed in the future.  However, this PR does fix several issues with the current standard, which shouldn't be breaking changes to those that have implemented it.  For example, an Option type is encoded the same way as a normal one just isn't a required parameter.  Furthermore the memo parameter was missing which is required for `nft_transfer`.

@mikedotexe @BenKurrek

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -54,44 +60,53 @@ Financial contracts MAY take a cut of the NFT sale price as commission, subtract
...
+  fn nft_payout(&self, token_id: String, balance: U128, max_len_payout: Option<u32>) -> Payout;
   /// Given a `token_id` and NEAR-denominated balance, transfer the token
   /// and return the `Payout` struct for the given token. Panic if the
   /// length of the payout exceeds `max_len_payout.`
-  #[payable]
   fn nft_transfer_payout(
     &mut self,
     receiver_id: AccountId,
     token_id: String,
     approval_id: u64,
```
mikedotexe, line 88: was this supposed to be optional as well? as I see in this implementation:
https://github.com/TENK-DAO/tenk/blame/d65e15c5fdc4b70d6aed0c8181bce33ea0a16e68/contracts/tenk/src/payout.rs#L73

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -1,25 +1,30 @@
 # Standard for a Multiple-Recipient-Payout mechanic on NFT Contracts (NEP-199)
+
 Version `2.0.0`.
```
mikedotexe, line 3: feels like we should bump this a major version?

specs/Standards/NonFungibleToken/Payout.md
willemneal:
```suggestion
    approval_id: Option<u64>,
```

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -54,44 +60,53 @@ Financial contracts MAY take a cut of the NFT sale price as commission, subtract
...
 ## Reference-level explanation
+
 ```rust
 /// A mapping of NEAR accounts to the amount each should be paid out, in
 /// the event of a token-sale. The payout mapping MUST be shorter than the
 /// maximum length specified by the financial contract obtaining this
 /// payout data. Any mapping of length 10 or less MUST be accepted by
 /// financial contracts, so 10 is a safe upper limit.
-#[derive(Serialize, Deserialize)]
+#[Derive(Serialize, Deserialize)]
```
willemneal, line 70:
```suggestion
#[derive(Serialize, Deserialize)]
```

specs/Standards/NonFungibleToken/Payout.md
willemneal: This shouldn't be a breaking change.
```suggestion
Version `2.1.0`.
```

willemneal:
@mikedotexe Who should I contact on the Dev Platform team?  Is there a Dev Platform team?

willemneal:
@mikedotexe

willemneal:
@mikedotexe

willemneal:
@mikedotexe

willemneal:
@mikedotexe @BenKurrek @mattlockyer

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -107,4 +121,4 @@ In the future, the NFT contract itself may be able to place an NFT transfer is a
 
 ## Errata
```
BenKurrek, line 123: Can we add this to the errata?

BenKurrek (changes requested):

specs/Standards/NonFungibleToken/Payout.md
willemneal, line 123: - Version `2.1.0` adds the `memo` parameter, which while used by `nft_transfer` in implementations that follow `2.0.0`, would have encoded used `None`. It also changes `max_len_payout` to an option type.

willemneal, line 123:
```suggestion

- Version `2.1.0` adds a memo parameter to `nft_transfer_payout`, which previously forced implementers of `2.0.0` to pass `None` to the inner `nft_transfer`. Also refactors `max_len_payout` to be an option type.
```

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -107,4 +121,5 @@ In the future, the NFT contract itself may be able to place an NFT transfer is a
...
 
-Version `2.0.0` contains the intended `approval_id` of `u64` instead of the stringified `U64` version. This was an oversight, but since the standard was live for a few months before noticing, the team thought it best to bump the major version.
\ No newline at end of file
+- Version `2.1.0` adds a memo parameter to `nft_transfer_payout`, which previously forced implementers of `2.0.0` to pass `None` to the inner `nft_transfer`. Also refactors `max_len_payout` to be an option type.
+Version `2.0.0` contains the intended `approval_id` of `u64` instead of the stringified `U64` version. This was an oversight, but since the standard was live for a few months before noticing, the team thought it best to bump the major version.
```
willemneal, line 125:
```suggestion
- Version `2.0.0` contains the intended `approval_id` of `u64` instead of the stringified `U64` version. This was an oversight, but since the standard was live for a few months before noticing, the team thought it best to bump the major version.
```

BenKurrek (approved):

mattlockyer (approved):
Looks good to me.

mikedotexe (approved):
actually maybe this is fine as a minor release, after thinking last night.

willemneal:
@mikedotexe any blockers on merging this.

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -2,27 +2,31 @@
 
 ## (NEP-199)
 
-Version `2.0.0`.
+Version `2.1.0`.
```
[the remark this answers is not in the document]
frol: Optional parameters are not serialized to JSON, and thus old front-ends will be still compatible with the new contracts if they don't specify optional parameters, also new front-ends will be also compatible with the old contracts as extra JSON fields are just silently ignored. @mikedotexe Am I missing something?

frol (approved):

frol:
@near/wg-contract-standards I feel this NEP is well-received, so we can cast a vote here.

mfornet (approved):
LGTM.

frol:
We are reaching a consensus, but there is still a chance to influence the decision. I have transitioned this proposal to the final-comment period, and we will make the final vote on the next Contract Standards Working Group call, which will be the last chance to speak before we merge this PR.

ori-near:
@near/wg-contract-standards – As @frol mentioned above, we would like to schedule a working group meeting to review this NEP. Before we proceed, we would like to get closer to a consensus on this NEP.

@abacabadabacaba Could you please fully read this NEP and comment in the thread if you are leaning with approving or rejecting it? Please make sure to include your rationale and any feedback that you have for the author. Thank you.

abacabadabacaba:
I lean with approving this NEP. Adding a memo parameter seems useful given that the underlying transfer function takes it, and making `max_len_payout` optional also makes sense.

frol:
Just to resolve ambiguity, I want to re-emphasize that I lean towards approving this NEP extension as it brings useful features in a backward-compatible way, so there is no risk of breaking existing ecosystem.

ori-near:
As the moderator, I would like to thank the author @willemneal for submitting this NEP extension, and for the @near/wg-contract-standards working group members for your review. Based on the voting indications above, it seems like this proposal is close to a decision. We are therefore scheduling the second Contract Standards Working group meeting next week, where this NEP can enter the final voting stage. We encourage anyone who is interested to join the meeting.

Meeting Info
📅 Date: Friday, Nov 18, 4pm UTC
🗒️ Topic: NEP-276 Voting
✏️ [Register](https://near-org.zoom.us/webinar/register/4016682147474/WN_7fK7eqIyQmiD46n1jFHX2A)

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -66,35 +72,43 @@ Financial contracts MAY take a cut of the NFT sale price as commission, subtract
...
+pub trait Payouts {
   /// Given a `token_id` and NEAR-denominated balance, return the `Payout`.
   /// struct for the given token. Panic if the length of the payout exceeds
   /// `max_len_payout.`
-  fn nft_payout(&self, token_id: String, balance: U128, max_len_payout: u32) -> Payout;
+  fn nft_payout(&self, token_id: String, balance: U128, max_len_payout: Option<u32>) -> Payout;
   /// Given a `token_id` and NEAR-denominated balance, transfer the token
   /// and return the `Payout` struct for the given token. Panic if the
   /// length of the payout exceeds `max_len_payout.`
-  #[payable]
```
frol, line 80: > Also the `payable` decorator is removed since the market place handles the transfers.

I don't understand this statement. `payable` attribute guards against using function-call-only access keys, and given that `nft_transfer_payout` is a sibling to `nft_transfer` which is guarded, I believe it is a mistake to drop `payable` here.

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -66,35 +72,43 @@ Financial contracts MAY take a cut of the NFT sale price as commission, subtract
...
     token_id: String,
-    approval_id: u64,
+    approval_id: Option<u64>,
+    memo: Option<String>,
     balance: U128,
-    max_len_payout: u32,
-  ) -> Payout{
+    max_len_payout: Option<u32>,
+  ) -> Payout {
     assert_one_yocto();
```
frol, line 95: This can be dropped as it does exactly the same as the `payable` attribute.

frol (changes requested):

specs/Standards/NonFungibleToken/Payout.md
willemneal:
```suggestion
```

specs/Standards/NonFungibleToken/Payout.md
```diff
@@ -66,35 +72,43 @@ Financial contracts MAY take a cut of the NFT sale price as commission, subtract
...
   /// Given a `token_id` and NEAR-denominated balance, return the `Payout`.
   /// struct for the given token. Panic if the length of the payout exceeds
   /// `max_len_payout.`
-  fn nft_payout(&self, token_id: String, balance: U128, max_len_payout: u32) -> Payout;
+  fn nft_payout(&self, token_id: String, balance: U128, max_len_payout: Option<u32>) -> Payout;
   /// Given a `token_id` and NEAR-denominated balance, transfer the token
   /// and return the `Payout` struct for the given token. Panic if the
   /// length of the payout exceeds `max_len_payout.`
-  #[payable]
   fn nft_transfer_payout(
```
willemneal, line 86:
```suggestion
 #[payable]
  fn nft_transfer_payout(
```

specs/Standards/NonFungibleToken/Payout.md
willemneal: Ah good point! I think at the time I thought it was because the caller should handle the payout and that the caller would always be another contract, but this isn't true.

frol (approved):

ori-near:
Thank you to everyone who attended the second Contract Standards Working Group meeting today! The working group members reviewed the NEP and reached the following consensus:

**Status:** Approved

- [x] @frol https://github.com/near/NEPs/pull/276#issuecomment-1311708759
- [x] @abacabadabacaba https://github.com/near/NEPs/pull/276#issuecomment-1306166486
- [x] @mfornet https://github.com/near/NEPs/pull/276#pullrequestreview-1134378267

Meeting Recording: 
[Coming Soon]

@willemneal Thank you for authoring this NEP!

**Next Steps:**
- [x] @ori-near to merge NEP

**Implementation Status:**
NEAR SDK Rust: PR is in the final stage waiting the author to address the final comments (potentially needs to be forked to run the last mile) https://github.com/near/near-sdk-rs/pull/778
NEAR SDK JavaScript: Awaiting community contribution

Community contributions to the SDKs are welcome!

PR #278, opened 2021-11-24
Modenize language

fulldecent:


frol:
@ori-near We will need to work on some policy, so we don't spend our time debating the choice of words. It is a quite low priority compared to all the other tasks we currently have on our endless pile of tasks.

PR #293, opened 2021-12-09
doc(nft): add example to events with authorized_id

telezhnaya:
We've already collected some information about NFTs in Indexer DB.

Some of the entries do not follow our standard fully: `authorized_id` is filled by ids from internal contract implementation (examples: 1, 2, 7).
I feel it's our fault; we do not communicate this part properly. I've added one more example to the doc.

Another part of the issue is to share it with the contract creators.
For now, we have only one contract that used `authorized_id` field
https://explorer.near.org/accounts/misfits.tenk.near
It's not hard to google them
https://paras.id/collection/misfits.tenk.near

Does anyone know how to contact them?

### DIFF
--- a/specs/Standards/NonFungibleToken/Event.md
+++ b/specs/Standards/NonFungibleToken/Event.md
@@ -104,7 +104,7 @@ interface NftMintLog {
 // An event log to capture token burning
 // Arguments
 // * `owner_id`: owner of tokens to burn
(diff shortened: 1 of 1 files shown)


frol (approved):

willemneal:
@telezhnaya it was me.  I fixed it and will add it to the events PR for the sdk to ensure others don't make the same mistake.

PR #300, opened 2021-12-15
doc(nep141): add FT NEP-141 events section

### LINKED ISSUE #271: FT NEP-141 require having minting/burning interface
We finished the work on support of NFT https://github.com/near/NEPs/issues/254 https://github.com/near/NEPs/issues/256 :tada:
It's better to read [NFT events standard](https://nomicon.io/Standards/NonFungibleToken/Event.html) at first, since we want to do the pretty same thing for FTs.
Current FT doc: https://nomicon.io/Standards/FungibleToken/README.html

Some initial thoughts:
- The work here should go much easier, we can take the same approach as in FTs: we use logs to imitate the events.
- For one smart contract, we have at most one FT associated with it. Is it true? If so, we don't need to pass `token_id`. UPD: no, it's not true for MultiToken standard. So we have to pass `token_id` even in a simple FT contract.
- We have the new field: `amount`, it should be provided in the highest precision for each smart contract.
- We need to re-check that we don't have any fees in FT. If Alice transfers 1 FT to Bob, it means that Alice spent exactly 1 FT (and the fee in NEAR, we don't need to think about it, it's not on the smart contract abstraction layer), Bob received exactly 1 FT.
- We still have 3 types of actions: `mint`, `burn` and `transfer`.
- We still have optional `memo` field.
- We don't need to add `msg` field, it's technical field for supporting cross-contract calls, we don't care about that.
- We don't have the support of approval management for FT ([approvals in NFT](https://nomicon.io/Standards/NonFungibleToken/ApprovalManagement.html)), we even don't have the issue about that, but I've heard we want such functionality. Should we think about approvals at first?
- Do we need batch events? Should we use batch syntax (read: arrays in all places where possible) just because we can?

@mikedotexe I know it's not your war anymore, but if you are ready to participate in any role, you are more than welcome ❤️ 

@jberrytech @frol @MaximusHaximus 
Please tag anyone else who could be interested. Especially I want FT contract developers to be involved

telezhnaya:
Closes #271

Please add here FT contract creators that you know

### DIFF
--- a/specs/SUMMARY.md
+++ b/specs/SUMMARY.md
@@ -54,6 +54,7 @@
   - [Fungible Token](Standards/FungibleToken/README.md)
       - [Core Standard](Standards/FungibleToken/Core.md)
       - [Metadata](Standards/FungibleToken/Metadata.md)
+      - [Events](Standards/FungibleToken/Event.md)
   - [Non-Fungible Token](Standards/NonFungibleToken/README.md)
       - [Core Standard](Standards/NonFungibleToken/Core.md)
       - [Metadata](Standards/NonFungibleToken/Metadata.md)
--- a/specs/Standards/FungibleToken/Event.md
+++ b/specs/Standards/FungibleToken/Event.md
@@ -76,12 +76,12 @@ Batch mint:
 
 ```js
 EVENT_JSON:{
-  "standard": "nep141",
-  "version": "1.0.0",
-  "event": "ft_mint",
-  "data": [
-    {"owner_id": "foundation.near", "amount": "500"}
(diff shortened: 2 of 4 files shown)


frol (approved):

frol:
@mikedotexe Do you know the stakeholders for Events support in FT standard? Can you invite them?

@telezhnaya Great job! We also need to add the links to this section on other pages in the nomicon, so it can be discovered

ilblackdragon (approved):

mikedotexe (approved):
Thank you for doing this!

PR #303, opened 2021-12-20
NFT Resolve Transfer Approved Account Parameter Patch Fix

### LINKED ISSUE #301: NFT Core Standard Approved Account IDs Type [NEP]
It was brought to my attention that the [core standard implementation](https://github.com/near/near-sdk-rs/blob/ddc5b142130e69dd9aa2cd264ba316282273af35/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L451) has a parameter `approved_account_ids` which is of type `Option<HashMap<AccountId, u64>>`. In the [core standards](https://nomicon.io/Standards/NonFungibleToken/Core.html), however, the `approved_account_ids` is of type `null|string[]`. 

I'm wondering if this is an error in the [core standards](https://nomicon.io/Standards/NonFungibleToken/Core.html) for the following reason. If the transfer needs to be reverted and the token should be sent back to its original owner, the contract must **also** revert the token's approvals map. By passing in a `HashMap<AccountId, u64>` like the [core standard implementation](https://github.com/near/near-sdk-rs/blob/ddc5b142130e69dd9aa2cd264ba316282273af35/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L451), you can easily revert the token's approvals list as seen both [here](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/non_fungible_token/core/core_impl.rs#L500) and [here](https://github.com/near-examples/nft-tutorial/blob/49145ad6a6976b8decd9a799d880561af10d51c9/nft-contract/src/nft_core.rs#L278). 

I'm wondering how you can revert the token's approvals list if you're only passing in an array of accounts to the function. Please let me know if I'm missing something. It seems like all the contracts I've stumbled across (including the core standard implementation) uses a HashMap and not an array. 

Example: 
Benji owns a token with the following map: 
```
Token {
  owner_id: Benji
  approvals: {
    marketplace A: 0
    marketplace B: 1
  }
}
```

Benji then calls `nft_transfer_call` to transfer the token to Mike. The step by step is as follows: 
1. NFT gets transferred to Mike. This means the state of the token is as follows:

```
Token {
  owner_id: Mike
  approvals: {}
}
```
With the next approval ID to be given out being `2`. 

2. NFT contract does a cross contract call to Mike's contract. Let's say it returns true meaning that the token **should be sent back to  Benji**: 

3. Resolve transfer gets called. At this point we've got two scenarios based on the conflicting standards. 

A) Passing in the hash map of approved account IDs which is the following in our scenario:
```
approvals: {
    marketplace A: 0
    marketplace B: 1
  }
```
- The contract will do its logic and refund the token's approvals (which is empty rn). It will then transfer the token back to Benji. This should, as far as my understanding goes, revert the token's approvals back to it's old state (which is what is passed into the function). **This includes the approval IDs given out as well as the accounts themsevles**: 

This means that the state of the token will be the following after transfer: 
```
Token {
  owner_id: Benji
  approvals: {
    marketplace A: 0
    marketplace B: 1
  }
}
```
This makes sense and is what a bunch of contracts are doing.

B) If you just pass in an array, you would pass in the following in our scenario: 
`approvals = [marketplace A, marketplace B]`
- How do you revert the token object back to it's old approvals map if you don't have their original approval IDs? You **could** use the incrementing approval ID and just reassign the accounts new approval IDs to put in the map but this doesn't seem right to me as it would render all marketplaces useless when it comes to this token since they wouldn't be able to transfer the token.

BenKurrek:
Changed the approved account IDs parameter in nft_resolve_transfer to be type Record<string, number> instead of string[]. 

Fixes #301

### DIFF
--- a/specs/Standards/NonFungibleToken/Core.md
+++ b/specs/Standards/NonFungibleToken/Core.md
@@ -157,15 +157,15 @@ The following behavior is required, but contract authors may name this function 
 // * `receiver_id`: the `receiver_id` argument given to `nft_transfer_call`
 // * `token_id`: the `token_id` argument given to `nft_transfer_call`
 // * `approved_token_ids`: if using Approval Management, contract MUST provide
-//   set of original approved accounts in this argument, and restore these
-//   approved accounts in case of revert.
+//   record of original approved accounts in this argument, and restore these
+//   approved accounts and their approval IDs in case of revert.
 //
 // Returns true if token was successfully transferred to `receiver_id`.
 function nft_resolve_transfer(
   owner_id: string,
   receiver_id: string,
   token_id: string,
-  approved_account_ids: null|string[],
+  approved_account_ids: null|Record<string, number>,
 ): boolean {}
 ```
 
@@ -200,6 +200,7 @@ function nft_on_transfer(
 
 ## Errata
 
+* **2021-12-20**: updated `nft_resolve_transfer` argument `approved_account_ids` to be type `null|Record<string, number>` instead of `null|string[]`. This gives contracts a way to restore the original approved accounts and their approval IDs. More information can be found in [this](https://github.com/near/NEPs/issues/301) discussion.
 * **2021-07-16**: updated `nft_transfer_call` argument `approval_id` to be type `number|null` instead of `string|null`. As stated, approval IDs are not expected to exceed the JSON limit of 2^53.
 
   [ERC-721]: https://eips.ethereum.org/EIPS/eip-721
--- a/specs/Standards/NonFungibleToken/Core.md
+++ b/specs/Standards/NonFungibleToken/Core.md
@@ -150,10 +150,10 @@ The following behavior is required, but contract authors may name this function 
 // * Contract MUST forbid calls to this function by any account except self
 // * If promise chain failed, contract MUST revert token transfer
 // * If promise chain resolves with `true`, contract MUST return token to
-//   `sender_id`
+//   `owner_id`
 //
 // Arguments:
-// * `sender_id`: the sender of `nft_transfer_call`
+// * `owner_id`: the original owner of the NFT.
 // * `receiver_id`: the `receiver_id` argument given to `nft_transfer_call`
 // * `token_id`: the `token_id` argument given to `nft_transfer_call`
 // * `approved_account_ids `: if using Approval Management, contract MUST provide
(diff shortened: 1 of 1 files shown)


specs/Standards/NonFungibleToken/Core.md
```diff
@@ -157,15 +157,15 @@ The following behavior is required, but contract authors may name this function
...
 // * `token_id`: the `token_id` argument given to `nft_transfer_call`
 // * `approved_token_ids`: if using Approval Management, contract MUST provide
-//   set of original approved accounts in this argument, and restore these
-//   approved accounts in case of revert.
+//   record of original approved accounts in this argument, and restore these
+//   approved accounts and their approval IDs in case of revert.
 //
 // Returns true if token was successfully transferred to `receiver_id`.
 function nft_resolve_transfer(
   owner_id: string,
```
mikedotexe, line 165: This is one line higher than I can easily make a commit suggestion, would you mind changing this line to be `owner__id`, as it currently says `sender_id` (in the comments)
https://github.com/near/NEPs/pull/303/files#diff-9132c1e8e04717bd55ad822800135f4094e4814a1d5fdb50e86d1204681dbf8eR156

specs/Standards/NonFungibleToken/Core.md
```diff
@@ -157,15 +157,15 @@ The following behavior is required, but contract authors may name this function
 // * `receiver_id`: the `receiver_id` argument given to `nft_transfer_call`
 // * `token_id`: the `token_id` argument given to `nft_transfer_call`
 // * `approved_token_ids`: if using Approval Management, contract MUST provide
```
mikedotexe, line 159:
```suggestion
// * `approved_account_ids `: if using Approval Management, contract MUST provide
```

mikedotexe (approved):
Sounds good to me, a couple of minor fixes could be made in the comments section

mikedotexe (changes requested):
Actually, can we please bump the patch version? And by this I mean at the top of this Core page, have it be Version 1.0.1

mikedotexe (approved):
Actually, can we please bump the patch version? And by this I mean at the top of this Core page, have it be Version 1.0.1

austinabell (approved):
Yeah, this seems correct to me, I thought this had already been updated. Maybe it's worth documenting what the approval IDs would be without the approval standard? I'm not even sure, and seems like the values can be anything, but not sure if this is intentional?

PR #304, opened 2022-01-04
feat(NFT): add memo and authorized_id to `nft_resolve_transfer`

willemneal:
In the case where a `nft_transfer_call` is successful the event from the transaction should be emitted by `nft_resolve_transfer`.  If the event is emitted by the initial call to `nft_transfer_call` and the promise of `nft_on_transfer` fails, then an additional log must be emitted when `nft_resolve_transfer` rolls the transaction back.

Thus it should be the case that `nft_relove_transfer` emits the log only if the transaction was successful.  However, it is missing access to the `memo` parameter  and the `authorized_id` that both need to be emitted which this PR adds.

### DIFF
--- a/specs/Standards/NonFungibleToken/Core.md
+++ b/specs/Standards/NonFungibleToken/Core.md
@@ -166,6 +166,7 @@ function nft_resolve_transfer(
   owner_id: string,
   receiver_id: string,
   token_id: string,
+  authorized_id: null|string,
   approved_account_ids: null|Record<string, number>,
   memo: null|string
 ): boolean {}
--- a/specs/Standards/NonFungibleToken/Core.md
+++ b/specs/Standards/NonFungibleToken/Core.md
@@ -159,6 +159,7 @@ The following behavior is required, but contract authors may name this function 
 // * `approved_account_ids `: if using Approval Management, contract MUST provide
 //   record of original approved accounts in this argument, and restore these
 //   approved accounts and their approval IDs in case of revert.
+// * `memo` (optional): to be included in event log if transfer is sucessful.
 //
 // Returns true if token was successfully transferred to `receiver_id`.
 function nft_resolve_transfer(
@@ -166,6 +167,7 @@ function nft_resolve_transfer(
   receiver_id: string,
   token_id: string,
   approved_account_ids: null|Record<string, number>,
+  memo: null|string
 ): boolean {}
 ```
 


BenKurrek:
@willemneal 

If `nft_transfer_call` is successful, shouldn't the log be emitted during the `internal_transfer` that happens before the cross contract call? I would think that the `nft_resolve_transfer` should **only** log if the transfer should be reverted. It would make more sense to log exactly when the transfers are happening no? The two possibilities are: 

A. successful transaction
`nft_transfer_call` --> `internal_transfer` (LOG HERE) --> `nft_on_transfer` --> `nft_resolve_transfer`

B. unsuccessful transaction
`nft_transfer_call` --> `internal_transfer` (LOG HERE) --> `nft_on_transfer` --> `nft_resolve_transfer` (Revert the transfer therefore log here as well)

In order to combat times where not enough GAS is attached to the call to complete the cross contract call, a require should be used to make sure the user never runs into this scenario as it will lead to a log being fired even though the logic is reverted.

specs/Standards/NonFungibleToken/Core.md
```diff
@@ -159,13 +159,16 @@ The following behavior is required, but contract authors may name this function
 // * `approved_account_ids `: if using Approval Management, contract MUST provide
 //   record of original approved accounts in this argument, and restore these
 //   approved accounts and their approval IDs in case of revert.
+// * `memo` (optional): to be included in event log if transfer is sucessful.
```
BenKurrek, line 162: `authorized_id` should be added here

BenKurrek (changes requested):

specs/Standards/NonFungibleToken/Core.md
willemneal:
```suggestion
// * `authorized_id` (optional): contract that initiated transfer to be included in event log if transfer is sucessful.
// * `memo` (optional): to be included in event log if transfer is sucessful.
```

willemneal: Like that?

willemneal:
@BenKurrek, which is better: no logging on an error or an extra log?  I'd argue that for the sake of indexers and gas, only logging in the successful case is the better route.  

In my view a transfer only happens if it isn't reverted in the final callback.  Which is why as you point out it is crucial that there be a check in `nft_transfer_call` to ensure that there is enough gas to revert the transaction.

Also `internal_transfer` isn't part of the standard, but rather the question of which part of the interface would be responsible for logging.

BenKurrek:
> @BenKurrek, which is better: no logging on an error or an extra log? I'd argue that for the sake of indexers and gas, only logging in the successful case is the better route.
> 
> In my view a transfer only happens if it isn't reverted in the final callback. Which is why as you point out it is crucial that there be a check in `nft_transfer_call` to ensure that there is enough gas to revert the transaction.
> 
> Also `internal_transfer` isn't part of the standard, but rather the question of which part of the interface would be responsible for logging.

Yea for the sake of gas and indexers, only logging the successful transfer would do the trick. I was talking with @mikedotexe and we had come to the conclusion that logging **exactly** what's actually happening might be better. 

The NFT **is** being transferred to the receiver and then **another** transfer happens if `nft_on_transfer` returns true. If you only log the successful cases, you miss all that (sure it doesn't actually affect anything since not logging and logging it would have the same outcome as from the indexer's perspective, both scenarios lead to the original owner having possession over the NFT). 

I'm relatively indifferent but would like to know other people's opinions on this.

willemneal:
@mikedotexe @telezhnaya @frol thoughts here.  I still contend that having fewer emitted events the better.  While I do see the benefit that if an incorrect implementation didn't revert properly then the log for the initial transfer would be valid. However, in that case the contract isn't correctly following the standard.

mikedotexe:
Hey all, while I do care deeply about the standards, I haven't been on the Dev Platform team in a few months and would want to defer to them. My opinion on the matter is that logs are meant to broadcast "what's happening" on the blockchain, and people can take that info and make it into whatever human-readable form they like.

So if I'm doing transfer-and-call of a fungible token, strictly speaking it _did_ get transferred before the cross-contract call where the receiver does the `*_on_transfer` logic, and if it ends up getting reverted it technically _does_ get transferred back. I believe the indexers will have all the information they need to say, "wait, is that from a `*_resolve_transfer` returning the same amount? okay, well in our frontend we're going to just show it in a simpler way" and we're not stopping folks ingesting the events from organizing it however they want.

All that to say I'm more in favor of being quite literal and emitting the event in the `internal_transfer` or wherever the implementor actually changes the state, even if it's a momentary state change.

frol (changes requested):
@willemneal Emitted logs should match the exact state even if intermediate, so we never arrive to a misalignment even for a single block. Given that the callbacks might fail, and also it is absolutely legal to view the state on a specific block, we should emit logs on actual transfer as @BenKurrek suggested (`nft_transfer_call` -> `internal_transfer` and, if necessary, `nft_resolve_transfer` -> `internal_transfer`).

Thus, I vote for closing this PR since the transfer log with the memo should only be emitted in the `nft_transfer_call`

willemneal:
Sounds good!

PR #314, opened 2022-01-29
fix: update storage rent to storage staking scheme

amgando:


bowenwang1996 (commented):
> Any account with less than this minimum amount will not be able to maintain a responsive contract and will, instead, return an error related to this mismatch in storage vs minimum account balance

This is not precise. What happens is that if you try to do any action that results in a contract going below the threshold, the action would fail with the error mentioned here

frol (approved):

victorchimakanu:
Hello everyone, I'd like to understand the current status of this NEP? 

@amgando @frol

PR #315, opened 2022-02-03
Consistent NFT Standard Token Struct Field Names

BenKurrek:
The token struct field names are not consistent with current [implementations](https://github.com/near/near-sdk-rs/blob/6596dc311036fe51d94358ac8f6497ef6e5a7cfc/near-contract-standards/src/non_fungible_token/token.rs#L12) and the rust [docs](https://docs.rs/near-contract-standards/latest/near_contract_standards/non_fungible_token/struct.Token.html). I've changed the token struct field names to be consistent:
- `id` changed to `token_id`
- `approvals` changed to `approved_account_ids`

### DIFF
--- a/specs/Standards/NonFungibleToken/ApprovalManagement.md
+++ b/specs/Standards/NonFungibleToken/ApprovalManagement.md
@@ -2,7 +2,7 @@
 
 ## [NEP-178](https://github.com/near/NEPs/discussions/178)
 
-Version `1.0.1`
+Version `1.1.0`
 
 ## Summary
 
--- a/specs/Standards/NonFungibleToken/ApprovalManagement.md
+++ b/specs/Standards/NonFungibleToken/ApprovalManagement.md
@@ -2,7 +2,7 @@
 
 ## [NEP-178](https://github.com/near/NEPs/discussions/178)
 
-Version `1.0.0`
+Version `1.0.1`
 
 ## Summary
 
@@ -21,7 +21,7 @@ However, some Non-Fungible Token developers, marketplaces, dApps, or artists may
 Prior art:
 
 - Ethereum's [ERC-721]
-- [NEP-4](https://github.com/near/NEPs/pull/4), NEAR's old NFT standard that does not include approvals per token ID
+- [NEP-4](https://github.com/near/NEPs/pull/4), NEAR's old NFT standard that does not include approved_account_ids per token ID
 
 ## Example Scenarios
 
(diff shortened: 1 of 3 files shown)


willemneal:
@BenKurrek  @austinabell  Seems to be at odds with https://github.com/near/near-sdk-rs/pull/727?notification_referrer_id=NT_kwDOABah7LIzMDUzMTMyNDgwOjE0ODMyNDQ#event-6005212068

willemneal:
I say it's good as is and the sdk should change.  This is really just how the token gets serialized from the contract so it shouldn't break state to update it.

austinabell:
> I say it's good as is and the sdk should change. This is really just how the token gets serialized from the contract so it shouldn't break state to update it.

yeah but would break existing tooling and contracts that expect a certain format for the result (like marketplaces or things like this). The decision should be based on what the community is using.

BenKurrek:
> > I say it's good as is and the sdk should change. This is really just how the token gets serialized from the contract so it shouldn't break state to update it.
> 
> yeah but would break existing tooling and contracts that expect a certain format for the result (like marketplaces or things like this). The decision should be based on what the community is using.

Yea the main reason I'm proposing this change is because pretty much all the contracts I've seen are using this naming scheme.

BenKurrek:
Bumping this PR @mikedotexe @austinabell

specs/Standards/NonFungibleToken/ApprovalManagement.md
```diff
@@ -2,7 +2,7 @@
 
 ## [NEP-178](https://github.com/near/NEPs/discussions/178)
 
-Version `1.0.0`
+Version `1.0.1`
```
austinabell, line 5: What semver rules are we using? I'm for this change but a patch version for a breaking change is a bit odd

austinabell (approved):
LGTM but we need to see and notify anyone directly following this standard, also would be nice to get buy-in from a few others.

specs/Standards/NonFungibleToken/ApprovalManagement.md
BenKurrek: Yea I could change this to 2.0.0 - i'm not sure what the convention is.

BenKurrek: I'll go with this - I couldn't find anything either.

BenKurrek:
@mattlockyer do you have any thoughts on this?

mattlockyer:
I definitely lean in favor of more explicit naming for standards e.g. token_id, approved_account_ids

We'd have to stick to version bumping and clients will have to adjust accordingly. Thankfully, for cross contract calls, the interfaces still seem fine.

willemneal:
@mikedotexe

mikedotexe (approved):

PR #323, opened 2022-02-25
Update Core.md

Doc-Failure:
if you try to implement the NEP141 methods as string, the near wallet will give you the error: "Error: [big.js] Invalid number".
You have to implement those methods as u128/U128

### DIFF
--- a/specs/Standards/FungibleToken/Core.md
+++ b/specs/Standards/FungibleToken/Core.md
@@ -252,7 +252,7 @@ function ft_on_transfer(
     sender_id: string,
     amount: string,
     msg: string
-): string {}
+): u128 {}
 
 /****************/
 /* VIEW METHODS */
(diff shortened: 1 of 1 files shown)


willemneal:
There is no standard representation of `u128`, which is why there is a comment above the functions explaining that the string is a u128 int. You'll find this across the standards: https://github.com/near/NEPs/blob/master/specs/Standards/NonFungibleToken/Enumeration.md#interface

However, the standard itself isn't consistent:
https://github.com/near/NEPs/blob/master/specs/Standards/NonFungibleToken/Payout.md#reference-level-explanation

I'm a fan of either all `U128` or `string` with a comment, but definitely not `u128` as that will lead people to implement it with an int representation.

frol:
I tend to agree with what @willemneal said. The standard has the following notes:

> - All amounts, balances and allowance are limited by `U128` (max value `2**128 - 1`).
> - Amounts in arguments and results have are serialized as Base-10 strings, e.g. `"100"`. This is done to avoid JSON limitation of max integer value of `2**53`.

@Doc-Failure Can you elaborate more on:

> if you try to implement the NEP141 methods as string, the near wallet will give you the error: "Error: [big.js] Invalid number". You have to implement those methods as u128/U128

My expectation is definitely that you need to implement NEP141 with the string-encoded decimals.

frol:
I believe it will be more productive to define the `JsonU128` type (a "string-encoded decimal number") and refactor all the NEPs to use it. I am closing this PR in favor of #404.

PR #330, opened 2022-03-03
feat: Add Contract Metadata Standard

BenKurrek:
This is based on the [discussion](https://github.com/near/NEPs/discussions/329) held outlining a way for auditing and viewing source code for a deployed smart contract.

There is no trivial way of finding the source code or author of a deployed smart contract. By having a standard that outlines how to view the source code of an arbitrary smart contract, an environment of openness and collaboration is created.


specs/Standards/ContractMetadata.md
```diff
@@ -0,0 +1,49 @@
...
+```ts
+function contract_metadata(): ContractMetadata {}
+```
+
+A new function for setting the contract metadata in case of updates to the currently deployed wasm or link to the open source code may optionally be implemented:
+
+```ts
+function set_contract_metadata(
+  metadata: ContractMetadata,
+): ContractMetadata {}
```
austinabell, line 39: why do you need this? wouldn't you just statically update what is returned when you upgrade the contract code? Or is this to be able to update the link when you don't want to upgrade the contract?

BenKurrek: That's a good point - you could just have everything as static variables in the contract but I kept it open in case people wanted to bump versions without having to redeploy the contract (for organizational purposes).

austinabell: but then if it's optional, maybe it shouldn't be a part of the standard? Kinda like mint for nfts?

BenKurrek: That's a good point - I'll change it and leave it up to the author to figure out they want to change the metadata.

agileurbanite:
👍

specs/Standards/ContractMetadata.md
```diff
@@ -0,0 +1,43 @@
+# Contract Metadata
+
+## [NEP-330](https://github.com/near/NEPs/discussions/329)
+
+Version `1.0.0`
+
+## Summary
+
+The contract metadata is a standard interface to allow for auditing and viewing source code for a deployed smart contract. Implementation of this standard is purely optional but is recommended for developers whose contracts are open source.
```
bucanero, line 9:
```suggestion
The contract metadata is a standard interface that allows auditing and viewing source code for a deployed smart contract. Implementation of this standard is purely optional but is recommended for developers whose contracts are open source.
```

specs/Standards/ContractMetadata.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Motivation
+
+There is no trivial way of finding the source code or author of a deployed smart contract. By having a standard that outlines how to view the source code of an arbitrary smart contract, an environment of openness and collaboration is created.
```
bucanero, line 13:
```suggestion
There is no trivial way of finding the source code or author of a deployed smart contract. Having a standard that outlines how to view the source code of an arbitrary smart contract creates an environment of openness and collaboration.
```

specs/Standards/ContractMetadata.md
```diff
@@ -0,0 +1,43 @@
...
+```ts
+function contract_metadata(): ContractMetadata {}
+```
+
+It is up to the author of the contract to keep the version and link up to date when new code is deployed. They can choose to update the metadata with a setter, have it static on the contract, or any other way of their choosing.
+
+### An implementing contract MAY include the following fields on-chain
+
+- `version`: a string that references the specific commit hash or version of the code that is currently deployed on-chain. This can be included regardless of whether or not the contract is open-sourced and can also be used for organizational purposes.
+- `link`: a string that references the link to the open source code. This can be anything such as Github or a CID to somewhere on IPFS.
```
bucanero, line 39:
```suggestion
- `link`: a string that references the link to the open-source code. This can be anything such as Github or a CID to somewhere on IPFS.
```

bucanero (approved):

specs/Standards/ContractMetadata.md
```diff
@@ -0,0 +1,43 @@
...
+The initial discussion can be found [here](https://github.com/near/NEPs/discussions/329).
+
+## Interface
+
+Metadata applies at the contract level (`ContractMetadata`):
+
+```ts
+type ContractMetadata = {
+  version: string|null, // optional, commit hash being used for the currently deployed wasm. If the contract is not open-sourced, this could also be a numbering system for internal organization / tracking such as "1.0.0" and "2.1.0".
+  link: string|null, //optional,  link to open source code such as a Github repository or a CID to somewhere on IPFS.
```
agileurbanite, line 24: @BenKurrek @austinabell @MaximusHaximus what are your thoughts of adding a system field so we can get some info on what system helped generate the wasm as well.  This way we can try to rebuild the wasm and compare the hashes for auditing purposes.

agileurbanite: `system: "local"|string,`

agileurbanite:
Reading up on how others handle this can we also add:
```
linkReference: deployed address of address of other smart contracts on which the current smart contract has a dependency.
object: Current smart contract bytecode
opcodes: Operation codes that are human-readable low-level instructions.
sourceMap: source map is to match each contract instruction to the section of the source code from which it was generated.
```

specs/Standards/ContractMetadata.md
austinabell: I mean, you would need architecture and toolchain version at least. Would the purpose of this be human-readable so someone could try to replicate it? You would also need some commit or version since the link doesn't specify repo links to commit that it was built with

agileurbanite: yeah ideally i want to be able to reference the source code and source dependencies to help with auditing the contract.  Here's what polygon does which is pretty cool: https://polygonscan.com/address/0x217cF04C783818E5b15Ae0723b22Ee2415Ab5fe3#code

austinabell:
> Reading up on how others handle this can we also add:
> 
> ```
> linkReference: deployed address of address of other smart contracts on which the current smart contract has a dependency.
> object: Current smart contract bytecode
> opcodes: Operation codes that are human-readable low-level instructions.
> sourceMap: source map is to match each contract instruction to the section of the source code from which it was generated.
> ```

address and contract code can all be pulled without this. (This method is suggesting as part of the contract). This would only be needed if this info was stored externally

opcodes doesn't apply. Assume you're comparing with solidity, but the wasm instructions aren't important and the imported functions can just be pulled from the wasm blob, so this is redundant.

source map there is no use for this currently, because they can be linked through just analyzing the wasm binary. Doesn't seem that useful in our case to add bloat to the standard

agileurbanite:
> > Reading up on how others handle this can we also add:
> > ```
> > linkReference: deployed address of address of other smart contracts on which the current smart contract has a dependency.
> > object: Current smart contract bytecode
> > opcodes: Operation codes that are human-readable low-level instructions.
> > sourceMap: source map is to match each contract instruction to the section of the source code from which it was generated.
> > ```
> 
> address and contract code can all be pulled without this. (This method is suggesting as part of the contract). This would only be needed if this info was stored externally
> 
> opcodes doesn't apply. Assume you're comparing with solidity, but the wasm instructions aren't important and the imported functions can just be pulled from the wasm blob, so this is redundant.
> 
> source map there is no use for this currently, because they can be linked through just analyzing the wasm binary. Doesn't seem that useful in our case to add bloat to the standard

I'm mainly looking at this from an auditing and reference standpoint. Also having a sourceMap would be useful for debugging imo but can be wrong.

austinabell:
> I'm mainly looking at this from an auditing and reference standpoint. Also having a sourceMap would be useful for debugging imo but can be wrong.

You can get this information from the wasm binary by just following the function export, so I don't see how any format would be more useful than this. Esp since if auditing the compiled code one would convert the wasm file into wat or use some tool that does this mapping for you

specs/Standards/ContractMetadata.md
BenKurrek: I was assuming that the toolchain version would be kept in the Github repo.  I don't really see the need to add a system field but would love other people's opinions

mfornet, line 24: What about adding some sort of organization contact link as well that points to the website of the application, or the developers; potentially with more contact information!

The difference between it and `link` is that one is to discover the source code, and the other is to discover what is the preferred way to reach developers/community behind this smart contract.

BenKurrek: I think this information could be found by utilizing the `link` since you can open issues or PRs on Github and follow or view the developers' public profiles. The case where I think this might be useful, is if the link is a CID to somewhere on IPFS since you don't have the same ability to reach out or view the developers as you do using Github.

thoughts? @austinabell @agileurbanite

specs/Standards/ContractMetadata.md
```diff
@@ -0,0 +1,43 @@
...
+
+Version `1.0.0`
+
+## Summary
+
+The contract metadata is a standard interface that allows auditing and viewing source code for a deployed smart contract. Implementation of this standard is purely optional but is recommended for developers whose contracts are open source.
+
+## Motivation
+
+There is no trivial way of finding the source code or author of a deployed smart contract. Having a standard that outlines how to view the source code of an arbitrary smart contract creates an environment of openness and collaboration.
```
mfornet, line 13: One of the main use cases of having this standard, is if we are able to reproduce the same binary after we have access to the source code. I'm wondering if as part of it we should add a proposal or advise about how we can achieve byte for byte reproducibility. 

Having a standard for reproducibility can be useful for external tools to automatically verify the code.

This is hard in general, mostly because there is more than one stack that can be used to generate the contract binary (i.e today we support rust and assembly script). Relevant discussion on how to make this possible on rust: https://github.com/rust-lang/cargo/issues/9506

specs/Standards/ContractMetadata.md
mfornet: I would actually go even further and provide three different fields (similar to [crates.io](https://crates.io/crates/rand)): `Homepage`, `Documentation` and `Repository` + `version` (to pin the exact version and allow reproducing the exact same binary).

Some services like [npm](https://www.npmjs.com/package/near-api-js) only have `Homepage` and `Repository`. (Maybe `documentation` is a bit too much).

specs/Standards/ContractMetadata.md
willemneal: A link to a docker image could be sufficient.

jlogelin (changes requested):
Great reviews here - it looks like it's conceptually ready for Draft. @BenKurrek could you please reformat this spec under neps/nep-0330.mddirectory following the new structure and [template](https://github.com/near/NEPs/blob/master/nep-0000-template.md) described [here](https://github.com/near/NEPs/blob/master/neps/nep-0001.md). You can leave the explanation in Docusaurus, preferably describing the spec at a high level and perhaps usage, and linking to the spec itself.

jlogelin (changes requested):
@BenKurrek This is nice work. We now also need a formatted standard document as part of the new process. In this case, the standard would be `neps/nep-0330.md`. See [here](https://github.com/near/NEPs#process), [template](https://github.com/near/NEPs/blob/master/nep-0000-template.md) and [nep success criteria](https://github.com/near/NEPs/blob/master/neps/nep-0001.md#what-does-a-successful-nep-look-like) for more details.

agileurbanite:
@BenKurrek mind taking a look at @jlogelin comments above?

BenKurrek:
Hey @jlogelin @agileurbanite been crazy busy - i'll take a look. Thanks so much!

BenKurrek:
@jlogelin @agileurbanite @austinabell I just pushed the changes to adhere to the new process for adding new NEPs as per Jay's suggestions. Take a look and let me know what you think.

neps/nep-0330.md
```diff
@@ -0,0 +1,75 @@
...
+
+In order to view this information, contracts must include a getter which will return the struct.
+
+```ts
+function contract_metadata(): ContractMetadata {}
+```
+
+## Reference Implementation
+
+As an example, say there was an NFT contract deployed on-chain which was currently using the commit hash `39f2d2646f2f60e18ab53337501370dc02a5661c` and had its open source code located at `https://github.com/near-examples/nft-tutorial`. This contract would then include a struct which has the following fields:
```
jlogelin, line 49: Ideally, we want to demonstrate at least a minimal implementation of the interface functions in NEAR contract code in this section. Could you link to the code that shows the use of `contract_metadata()`

jlogelin (changes requested):

jlogelin (approved):

PR #352, opened 2022-04-14
NFT metadata events (draft)

tifrel:


tifrel:
Opened up to soon, I will re-open once this is more refined.

PR #358, opened 2022-05-17
Expose Chain Id and Protocol Version

mfornet:


ori-near:
@mfornet Thank you for submitting this NEP. In order to move this Protocol NEP to the review stage, can you please add a reference implementation?

ori-near:

PR #360, opened 2022-05-26
NEP-360: Reduce the minimum gas price

0xbosnic:
This is a proposal based on the community discussion from the NEAR forums: https://gov.near.org/t/proposal-reduce-the-gas-floor-price-on-the-near-network/20980

### DIFF
--- a/neps/nep-0360.md
+++ b/neps/nep-0360.md
@@ -1,0 +1,70 @@
+---
+NEP: 360
+Title: Reduce minimum gas price
+Author: Marko Bosnic <marko@tonic.foundation>
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/0360
+Status: Draft
+Type: Standards Track
+Category: Contract
+Created: 26-May-2022
+---
+
+## Summary
+
+Gas fees on NEAR are artificially high. The solution is to reduce the minimum gas price (also known as floor price).
+NEAR’s gas floor price is currently 0.0001 NEAR/Tgas. ([explorer](https://explorer.near.org/), [nearcore](https://github.com/near/nearcore/blob/f951d6ed8e491ef842da6bd9448d3242472c126b/nearcore/res/mainnet_genesis.json#L16)). This means that a user must pay at least that gas price for any transaction, regardless of how many transactions are in a block. While a fee market is necessary for efficient use of network resources (see Solana’s [recent congestion issues](https://github.com/solana-labs/solana/issues/22820) for an example of issues caused by not having a fee market), a high floor price does not create a real fee market. Lowering the floor price does not preclude the ability for a stable fee market to form when the network is close to maximum capacity.
+
+## Motivation
+### NEAR is well below capacity
+NEAR had ~640k transactions in the last 24 hours according to the [block explorer](https://explorer.near.org/). This is probably a slight overestimate since it includes failed transactions from Aurora, the most active contract on the network. See the contract [transaction history](https://explorer.near.org/accounts/aurora) to see a sample of recent failures.
+
+Let’s be generous and say the network averages 8 successful transactions per second on average. An FT transfer is 8 tgas, while a swap on Ref Finance (with an existing exchange balance) is 14 tgas. The gas limit per block is 1000 Tgas [source](https://github.com/near/nearcore/blob/f951d6ed8e491ef842da6bd9448d3242472c126b/nearcore/res/mainnet_genesis.json#L15). Suppose the average smart contract transaction uses 10 tgas. This implies ~100 transactions per block per shard, assuming no other bottlenecks. With 4 shards on mainnet that means the network is operating at approximately 2% capacity.
+
+NEAR is operating nowhere close to capacity and currently has a fraction of the activity of the major L1s - it has the highest mcap/TVL ratio of any chain in the top 30 on [Defi Llama](https://defillama.com/chains). At $10, a fungible token transaction is 0.8 cents. On Solana, a network with at least 100x the current transaction volumes of NEAR, an SPL token transfer is 0.000005 SOL (~0.02 cents, 97% cheaper).
+
+## Rationale and alternatives
+### Prior Art
+The last major discussion around NEAR's gas prices was in 2020 [NEP-92](https://github.com/near/NEPs/issues/92). The original intent was to leave room to reduce fees later. The increase from NEP-92 was done shortly after the mainnet launch, and NEAR was 90%+ cheaper at the time.
+
+### Benefits and Ecosystem Impact
+NEAR is a volatile asset and the price increases faster than network activity. Increased gas costs when there is no shortage of block space is effectively a tax paid by users due to speculative market activity.
+
+While paying a cent for an average transaction might not seem like an issue for an infrequent user, these costs add up for liquidiy providers/arbitrageurs who improve the efficiency of markets on-chain. This makes NEAR look even worse relatively if there is another rally in the token price without corresponding network growth.
+
+Cutting fees will make NEAR even more competitive with top L1s. The ecosystem has had considerable momentum in the last few months, and NEAR’s sharding model will continue to generate interest among developers. A reduction in fees will enable more applications to build on NEAR in an economically viable way, which will drive consumer adoption.
+
+## Specification
+This proposal is to reduce the minimum price of gas from 0.0001 N/Tgas to 0.00001 N/Tgas. The total cost of a transaction is gas price * gas amount.
+
+## Reference Implementation (Required for Contract category, optional for other categories)
+
+The implementation of this proposal is submitted as a PR to nearcore - [pull/6869](https://github.com/near/nearcore/pull/6869).
+
+This is the technical portion of the NEP. Explain the design in sufficient detail that:
+
+- Its interaction with other features is clear.
+- Where possible, include a `Minimum Viable Interface` subsection expressing the required behavior and types in a target Near Contract language. (ie. traits and structs for rust, interfaces and classes for javascript, function signatures and structs for c, etc.)
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
+## Security Implications (Optional)
+Reducing the gas floor price will reduce the theoretical cost of a short DDOS attack on the network. Note that the dollar cost for an attack has dropped >50% during several price drops without a corresponding increase in attempts to flood the network.
+
+
+## Drawbacks (Optional)
+- Decreased protocol revenue: Relative to the market cap protocol revenue is currently negligible. NEAR trades at > 900 P/E ratio according to TokenTerminal [7].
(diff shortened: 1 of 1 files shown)


neps/nep-0360.md
```diff
@@ -0,0 +1,70 @@
...
+NEAR’s gas floor price is currently 0.0001 NEAR/Tgas. ([explorer](https://explorer.near.org/), [nearcore](https://github.com/near/nearcore/blob/f951d6ed8e491ef842da6bd9448d3242472c126b/nearcore/res/mainnet_genesis.json#L16)). This means that a user must pay at least that gas price for any transaction, regardless of how many transactions are in a block. While a fee market is necessary for efficient use of network resources (see Solana’s [recent congestion issues](https://github.com/solana-labs/solana/issues/22820) for an example of issues caused by not having a fee market), a high floor price does not create a real fee market. Lowering the floor price does not preclude the ability for a stable fee market to form when the network is close to maximum capacity.
+
+## Motivation
+### NEAR is well below capacity
+NEAR had ~640k transactions in the last 24 hours according to the [block explorer](https://explorer.near.org/). This is probably a slight overestimate since it includes failed transactions from Aurora, the most active contract on the network. See the contract [transaction history](https://explorer.near.org/accounts/aurora) to see a sample of recent failures.
```
willemneal, line 19: Perhaps put the date here.

neps/nep-0360.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Benefits and Ecosystem Impact
+NEAR is a volatile asset and the price increases faster than network activity. Increased gas costs when there is no shortage of block space is effectively a tax paid by users due to speculative market activity.
+
+While paying a cent for an average transaction might not seem like an issue for an infrequent user, these costs add up for liquidiy providers/arbitrageurs who improve the efficiency of markets on-chain. This makes NEAR look even worse relatively if there is another rally in the token price without corresponding network growth.
+
+Cutting fees will make NEAR even more competitive with top L1s. The ecosystem has had considerable momentum in the last few months, and NEAR’s sharding model will continue to generate interest among developers. A reduction in fees will enable more applications to build on NEAR in an economically viable way, which will drive consumer adoption.
+
+## Specification
+This proposal is to reduce the minimum price of gas from 0.0001 N/Tgas to 0.00001 N/Tgas. The total cost of a transaction is gas price * gas amount.
```
willemneal, line 37: What is the motivation for this particular price? Just returning to the 90% lower fee?

mfornet: Moreover, something that is not mentioned is why not reduce the minimum gas price by (say) 100x. [This comment](https://gov.near.org/t/proposal-reduce-the-gas-floor-price-on-the-near-network/20980/14) on the forum is answering the question, but I think we need to provide a more concise reason.

Some thoughts here: What will happen today (with no dynamic resharding) if the minimum gas price is `1yoctoNEAR/Gas`? I guess the gas price will be much lower than today and more unstable. (How unstable?)

What will happen when dynamic resharding is added? There should be a clear strategy about how to create new shards based on network activity/capacity, which will be directly related to the current gas price. One simple approach is to add/remove shards based on an upper/lower threshold disconnected from the minimum gas price. Minimum gas price will be relevant mostly when there is a single shard due to low activity.

0xbosnic: How are full blocks handled currently? [The documentation](https://docs.near.org/concepts/basics/transactions/gas#gas-price) mentions that gas prices can increase per block but only by 1%. Without dynamic resharding there needs to be more flexibility in the fee market.

> One simple approach is to add/remove shards based on an upper/lower threshold disconnected from the minimum gas price. Minimum gas price will be relevant mostly when there is a single shard due to low activity.

This is a reasonable approach. If the market-clearing gas price exceeds some multiple of the floor for more than a specified number of blocks it makes sense to re-shard. What level of gas price volatility is acceptable depends on the performance implications of resharding. I'm not familiar enough with the proposed implementation to make a judgement there.

akhi3030:
Thanks for the proposal and the NEP.  I have the following questions please:

- In [this post](https://gov.near.org/t/proposal-reduce-the-gas-floor-price-on-the-near-network/20980/14?u=akhi3030), a question was raised about how reducing the fees will mean that we will burn even fewer tokens which will mean that it will be even harder for the network to become sustainable.  Has this issue been discussed somewhere already?  Can you please provide any resolutions for this please?
- [Currently](https://github.com/near/nearcore/blob/26d1d7884a78f4e3759295559c0390f21f8aebac/chain/chain/src/types.rs#L205), the max gas price is defined in terms of the min gas price.  Is the plan to lower the max price as well or to keep it as it?

0xbosnic:
>a question was raised about how reducing the fees will mean that we will burn even fewer tokens which will mean that it will be even harder for the network to become sustainable.
This assumes
1) Transaction volumes remain the same
2) Average cost = floor price

If transactions become cheaper more transactions become economically viable. Some proportion of the lost revenue will be made up with the additional volume. If network demand stays low then revenue may be lower than today, but a network with no growth would not be sustainable in any case.

>Currently, he max gas price is defined in terms of the min gas price. Is the plan to lower the max price as well or to keep it as it?

This depends as I noted above: if there is a low performance overhead to dynamic sharding then it makes more sense to reshard rather than allow the maximum gas price to increase past a point. If resharding is undesirable for whatever reason, it does not make sense to have a ceiling price on gas.

akhi3030:
> This assumes
> 
>     1. Transaction volumes remain the same
> 
>     2. Average cost = floor price
> 
> 
> If transactions become cheaper more transactions become economically viable. Some proportion of the lost revenue will be made up with the additional volume. If network demand stays low then revenue may be lower than today, but a network with no growth would not be sustainable in any case.

Do you have any thoughts on this calculation from the post?

> At the current gas price floor, 45 shards filled 50% on average would just about equalize this. (Why 45? Factoring in that only 70% of gas is burnt, 30% is contract reward. Thus, 1.584Ⓝ/s / ( 1000Tgas/s * 0.0001Ⓝ/Tgas * 0.5 * 0.7) = 45.26) Decreasing the floor by 10x means we need 450 shards instead.

> 
> > Currently, he max gas price is defined in terms of the min gas price. Is the plan to lower the max price as well or to keep it as it?
> 
> This depends as I noted above: if there is a low performance overhead to dynamic sharding then it makes more sense to reshard rather than allow the maximum gas price to increase past a point. If resharding is undesirable for whatever reason, it does not make sense to have a ceiling price on gas.

Given that we do not have dynamic resharding right now, do you have thoughts on how we should set the ceiling price in the interim?

mzhangmzz (commented):
Technically this NEP is easy to implement. But economically, I'm not sure if this is the right way to go. I share the same concern as @jakmeier mentioned in his [comment](https://gov.near.org/t/proposal-reduce-the-gas-floor-price-on-the-near-network/20980/14), there are disadvantages of reducing gas price. I think it's trade-off between lowering the price for users and potentially bring more users vs having unstable gas price. It's not clear to me how much more users we can bring to the system, so it's hard to make a judgement here. Maybe It would be helpful to know some use cases that are limited by gas price.

0xbosnic:
A few use cases:

- on-chain market makers: quoting prices on chain has a cost due to gas. In order to compensate for this risk, spreads are wider. Cheaper gas would lead to tighter spreads -> more liquidity.
- arbitrage bots: cheaper gas -> more viable arbitrage opportunities/efficient pricing
- oracles: cheaper to provide more frequent updates

The assumption that gas prices will be more volatile is only true if there is a massive increase in demand. NEAR has had ~452k transactions in the last 24h according to the block explorer, with an unknown percentage that were successful. Assuming 100% are successful, that is an average of ~5tps. If each shard can do ~100tps, that's 1.25% utilization.

Gas prices would only become more volatile if we had a _sustained_ 80x increase in transaction volumes and 0 new shards. I don't know what the schedule is to increase the number of shards on mainnet, but it seems more likely that will occur before an 80x increase in transactions.

0xhipo:
As a HFT protocol, I've to say gas fee is a very important concern of running strategies (arbitrage/market making) on NEAR. With 10x fewer tps, the high performance (fast confirmation, swift blocktime) would has less impact.  A lower gas fee, could also incorporate a bidding mechanism, would be crucial to thrive a defi eco.

jakmeier:
Thanks for the example use cases, this is really helpful! Even better if you could back it up with some numbers what current spending for such a use case is, why that is too much, and what it would have to be to make it (more) viable. Decreasing cost by 10x still feels a bit arbitrary and we have to make sure it actually helps in the way we expect it to.

For example, with assumed 10x cheaper gas costs, is there any oracle service that would use that to increase data update frequency by 10x? Is cost really their bottleneck right now, or are there technical limitations and they don't even care about cost? Is gas cost the main contributor to cost as opposed to operating costs etc? Would 10x higher frequency be useful to anyone?

I would really like to see some calculations that explain why 10x cheaper makes a real impact.

> The assumption that gas prices will be more volatile is only true if there is a massive increase in demand. NEAR has had ~452k transactions in the last 24h according to the block explorer, with an unknown percentage that were successful. Assuming 100% are successful, that is an average of ~5tps. If each shard can do ~100tps, that's 1.25% utilization.
> 
> Gas prices would only become more volatile if we had a sustained 80x increase in transaction volumes and 0 new shards.

That's not quite correct. The gas price increases as soon as a block has more than 50% of the gas limit. But your point still stands with 40x. (Not sure where 100tps comes from but this would obviously depend on how much gas each tx uses.)

Also, if we reach 100% "utilization", this is really bad congestion of the network. It means you have to get lucky to get your tx processed quickly, if you can get it in at all. Because NEAR was built without gas price bidding to favor predictable costs, you cannot even pay more to get in. Adding price bidding is of course possible but a different discussion.

I think for NEAR protocol today, a healthy network should rarely have full blocks. Most blocks should be below 50% gas limit, to ensure availability to users at all times. If it goes above 50%, the price goes up exponentially. Not necessarily to find a free-market price but rather in the hopes that who ever spammed the network causing full blocks will stop doing so and make room for other users again. At least that's how I think about the gas price in NEAR. But of course this is not the only view and it can always change.

bowenwang1996:
@evgenykuzyakov do you mind reviewing this NEP as a subject matter expert? Thanks!

evgenykuzyakov:
Without getting too deep into potential gas vulnerabilities due to potential shard congestions. I suggest the following:
- Drop minimum gas price 10X
- Increase the gas price growth due to congestion 10X (from 1% to 10%), so a congestion attack due to spam may be stopped faster.
- Change the way gas price is increased from MEAN gas consumption to MAX gas consumption across shards.
- Pessimistic gas price prepayment for function calls may probably stay the same (something like 6.6X of max).

ori-near:
@0xbosnic Based on our offline conversation, my understanding is that you do not have the time to contribute the reference implementation for the proposed design changes that @evgenykuzyakov highlighted above. I surfaced this to the Protocol Working Group (@near/wg-protocol) to see what could be the next step.

ori-near:
As a moderator, I am closing this NEP given the concerns that design concerns that @evgenykuzyakov mentioned above and  a confirmation from the author (@0xbosnic) that he does not have capacity to update the implementation reference. If anyone is interested in addressing the concerns above, please submit a new NEP.

PR #361, opened 2022-05-28
Fixed Error in Access Key Docs

BenKurrek:
Fixed the access key docs which previously mixed up function call and full access key permissions.

### DIFF
--- a/specs/DataStructures/AccessKey.md
+++ b/specs/DataStructures/AccessKey.md
@@ -15,7 +15,7 @@ pub struct AccessKey {
 }
 ```
 
-There are 2 types of `AccessKeyPermission` in Near currently: `FullAccess` and `FunctionCall`. `FullAccess` grants permission to request any action on an account like [DeployContract](Transaction.md#DeployContract), [Transfer](Transaction.md#Transfer) tokens to other account, calling functions on smart contracts [FunctionCall](Transaction.md#FunctionCall), [Stake](Transaction.md#Stake) and even delete accounts [DeleteAccountAction](Transaction.md#DeleteAccountAction). `FullAccess` also allows managing access keys. `AccessKeyPermission::FunctionCall` is limited to only having permission to call non-payable methods on contracts.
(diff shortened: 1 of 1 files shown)


specs/DataStructures/AccessKey.md
```diff
@@ -15,7 +15,7 @@ pub struct AccessKey {
...
+There are 2 types of `AccessKeyPermission` in NEAR currently: `FullAccess` and `FunctionCall`. `FullAccess` grants permissions to issue any action on the account. This includes [DeployContract](Transaction#DeployContract), [Transfer](Transaction#Transfer) tokens, call functions [FunctionCall](Transaction#FunctionCall), [Stake](Transaction#Stake) and even permission to delete the account [DeleteAccountAction](Transaction#DeleteAccountAction). `FunctionCall` on the other hand, can **only** be used to call a specific set of predefined methods on one given contract. It has an allowance of $NEAR that can be spent on **GAS and transaction fees only**. Function call access keys **cannot** be used to transfer $NEAR.
```
joshmuente, line 18: Afaik the FunctionCall can also be used to give permissions for all methods not just specific.
Also the build checks are failing because the links are not working... I guess they should point to Actions instead?

```suggestion
There are 2 types of `AccessKeyPermission` in NEAR currently: `FullAccess` and `FunctionCall`. `FullAccess` grants permissions to issue any action on the account. This includes [DeployContract](Actions#DeployContract), [Transfer](Transaction#Transfer) tokens, call functions [FunctionCall](Transaction#FunctionCall), [Stake](Actions#Stake) and even permission to delete the account [DeleteAccountAction](Actions#DeleteAccountAction). `FunctionCall` on the other hand, **only** grants permission for all or a specific set of methods on one given contract. It has an allowance of $NEAR that can be spent on **GAS and transaction fees only**. Function call access keys **cannot** be used to transfer $NEAR.
```

BenKurrek: You're correct in that function call access keys can be used to call any method (if specified). I'll rewrite to make it more explicit. I'll also double check the links.

frol (approved):

PR #362, opened 2022-06-06
Update urls on NEPs

flmel:
I've noticed several urls were broken on the NEPs so I had a go at all of them This PR updates the urls  and moves them as references at the bottom of the md Files. 
This improves on DRY and will help updating them in future. 

Few questions: 
Where shall https://github.com/near/NEPs/blob/master/neps/nep-0021.md#reference-implementation url point to ?
Shall we move all mentioned urls as refs at the bottom of the page (currently the NEPs use mixed approach, this pr makes all urls pointing to local(to repository) urls as refs)?

### DIFF
--- a/neps/nep-0297.md
+++ b/neps/nep-0297.md
@@ -60,7 +60,7 @@ interface EventLogData {
 }
 ```
 
-Thus, to emit an event, you only need to log a string following the rules above. Here is a bare bones example using Rust SDK `near_sdk::log!` macro (security note: prefer using `serde_json` or alternatives to serialize the JSON string to avoid potential injections and corrupted events):
+Thus, to emit an event, you only need to log a string following the rules above. Here is a bare-bones example using Rust SDK `near_sdk::log!` macro (security note: prefer using `serde_json` or alternatives to serialize the JSON string to avoid potential injections and corrupted events):
 ```rust
 use near_sdk::log;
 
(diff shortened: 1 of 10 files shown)


bucanero (approved):

bucanero:
>  Shall we move all mentioned urls as refs at the bottom of the page (currently the NEPs use mixed approach, this pr makes all urls pointing to local(to repository) urls as refs)?

I personally prefer to have the link next to the text itself (instead of having it at the bottom) so I know where the hyperlink is going to, but since in these files it was already mixed, I think it's ok.

neps/nep-0297.md
```diff
@@ -60,7 +60,7 @@ interface EventLogData {
 }
 ```
 
-Thus, to emit an event, you only need to log a string following the rules above. Here is a barebones example using Rust SDK `near_sdk::log!` macro (security note: prefer using `serde_json` or alternatives to serialize the JSON string to avoid potential injections and corrupted events):
+Thus, to emit an event, you only need to log a string following the rules above. Here is a bare bones example using Rust SDK `near_sdk::log!` macro (security note: prefer using `serde_json` or alternatives to serialize the JSON string to avoid potential injections and corrupted events):
```
bucanero, line 63:
```suggestion
Thus, to emit an event, you only need to log a string following the rules above. Here is a bare-bones example using Rust SDK `near_sdk::log!` macro (security note: prefer using `serde_json` or alternatives to serialize the JSON string to avoid potential injections and corrupted events):
```

bucanero (approved):

PR #363, opened 2022-06-07
Typo in TokenMetadata

devey:
Fixing this typo cause it's a bit confusing as it might lead to thinking that `title` prop has been renamed to `name` in new specs, but it seems to be indeed just a typo.


frol (approved):

frol:
@devey Thanks for catching it!

PR #364, opened 2022-06-15
NEP-364: Efficient signature verification host functions

blasrodri:
Efficient signature verification and hashing precompile functions

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,22 +10,23 @@ Changes to the protocol specification and standards are called NEAR Enhancement 
 
 ## NEPs
 
-| NEP #                                                             | Title                                                             | Author                                      | Status |
-| ----------------------------------------------------------------- | ----------------------------------------------------------------- | ------------------------------------------- | ------ |
-| [0001](https://github.com/near/NEPs/blob/master/neps/nep-0001.md) | NEP Purpose and Guidelines                                        | @jlogelin                                   | Living |
-| [0021](https://github.com/near/NEPs/blob/master/neps/nep-0021.md) | Fungible Token Standard (Deprecated)                              | @evgenykuzyakov                             | Final  |
-| [0141](https://github.com/near/NEPs/blob/master/neps/nep-0141.md) | Fungible Token Standard                                           | @evgenykuzyakov @oysterpack                 | Final  |
-| [0145](https://github.com/near/NEPs/blob/master/neps/nep-0145.md) | Storage Management                                                | @evgenykuzyakov                             | Final  |
-| [0148](https://github.com/near/NEPs/blob/master/neps/nep-0148.md) | Fungible Token Metadata                                           | @robert-zaremba @evgenykuzyakov @oysterpack | Final  |
-| [0171](https://github.com/near/NEPs/blob/master/neps/nep-0171.md) | Non Fungible Token Standard                                       | @mikedotexe @evgenykuzyakov @oysterpack     | Final  |
-| [0177](https://github.com/near/NEPs/blob/master/neps/nep-0177.md) | Non Fungible Token Metadata                                       | @chadoh @mikedotexe                         | Final  |
-| [0178](https://github.com/near/NEPs/blob/master/neps/nep-0178.md) | Non Fungible Token Approval Management                            | @chadoh @thor314                            | Final  |
-| [0181](https://github.com/near/NEPs/blob/master/neps/nep-0181.md) | Non Fungible Token Enumeration                                    | @chadoh @thor314                            | Final  |
-| [0199](https://github.com/near/NEPs/blob/master/neps/nep-0199.md) | Non Fungible Token Royalties and Payouts                          | @thor314 @mattlockyer                       | Final  |
-| [0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md) | Multi Token Standard                                              | @zcstarr @riqi @jriemann @marcos.sun        | Review |
-| [0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md) | Events Standard                                                   | @telezhnaya                                 | Final  |
-| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Source Metadata                                                   | @BenKurrek                                  | Review |
-| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Efficient signature verification and hashing host functions |                                             | Draft  |
+|NEP #   | Title  | Author  | Status  |
+|---|---|---|---|
+|[0001](https://github.com/near/NEPs/blob/master/neps/nep-0001.md)   | NEP Purpose and Guidelines  | @jlogelin  | Living  |
+|[0021](https://github.com/near/NEPs/blob/master/neps/nep-0021.md)   | Fungible Token Standard (Deprecated) | @evgenykuzyakov | Final |
+|[0141](https://github.com/near/NEPs/blob/master/neps/nep-0141.md)   | Fungible Token Standard | @evgenykuzyakov @oysterpack | Final |
+|[0145](https://github.com/near/NEPs/blob/master/neps/nep-0145.md)   | Storage Management | @evgenykuzyakov | Final |
+|[0148](https://github.com/near/NEPs/blob/master/neps/nep-0148.md)   | Fungible Token Metadata | @robert-zaremba @evgenykuzyakov @oysterpack | Final |
+|[0171](https://github.com/near/NEPs/blob/master/neps/nep-0171.md)   | Non Fungible Token Standard | @mikedotexe @evgenykuzyakov @oysterpack | Final |
+|[0177](https://github.com/near/NEPs/blob/master/neps/nep-0177.md)   | Non Fungible Token Metadata | @chadoh @mikedotexe | Final |
+|[0178](https://github.com/near/NEPs/blob/master/neps/nep-0178.md)   | Non Fungible Token Approval Management | @chadoh @thor314 | Final |
+|[0181](https://github.com/near/NEPs/blob/master/neps/nep-0181.md)   | Non Fungible Token Enumeration | @chadoh @thor314 | Final |
+|[0199](https://github.com/near/NEPs/blob/master/neps/nep-0199.md)   | Non Fungible Token Royalties and Payouts | @thor314 @mattlockyer | Final |
+|[0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md)   | Multi Token Standard | @zcstarr @riqi @jriemann @marcos.sun | Review |
+|[0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md)   | Events Standard | @telezhnaya | Final |
+|[0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md)   | Source Metadata | @BenKurrek | Review |
+
+
 
 ## Specification
 
@@ -53,26 +54,26 @@ Spec changes are ultimately done via pull requests to this repository (formalize
 2. The forum has comment threading which allows the community and NEAR Collective to ideate, ask questions, wrestle with approaches, etc. If more immediate responses are desired, consider bringing the conversation to [Zulip](https://near.zulipchat.com/#narrow/stream/320497-nep-standards).
 3. When the governance conversations have reached a point where a clear plan is evident, create a pull request, using the instructions below.
 
-   - Clone this repository and create a branch with "my-feature".
-   - Update relevant content in the current specification that are affected by the proposal.
-   - Create a Pull request, using [nep-0000-template.md](nep-0000-template.md) to describe motivation and details of the new Contract or Protocol specification. In the document header, ensure the `Status` is marked as `Draft`, and any relevant discussion links are added to the `DiscussionsTo` section.
-     Use the pull request number padded with zeroes. For instance, the pull request `219` should be created as `neps/nep-0219.md`.
-   - Add your Draft standard to the `NEPs` section of this README.md. This helps advertise your standard via github.
-   - Update Docusaurus documentation under the `specs/Standards` to describe the contract standard at a high level, how to integrate it into a Dapp, and a link to the standard document (ie. `neps/nep-0123.md`). This helps advertise your standard via [nomicon](https://nomicon.io/). Any related nomicon sections should be prefixed and styled using the following snippet:
+   * Clone this repository and create a branch with "my-feature".
+   * Update relevant content in the current specification that are affected by the proposal.
+   * Create a Pull request, using [nep-0000-template.md](nep-0000-template.md) to describe motivation and details of the new Contract or Protocol specification. In the document header, ensure the `Status` is marked as `Draft`, and any relevant discussion links are added to the `DiscussionsTo` section.
+   Use the pull request number padded with zeroes. For instance, the pull request `219` should be created as `neps/nep-0219.md`.
+   * Add your Draft standard to the `NEPs` section of this README.md. This helps advertise your standard via github.
+   * Update Docusaurus documentation under the `specs/Standards` to describe the contract standard at a high level, how to integrate it into a Dapp, and a link to the standard document (ie. `neps/nep-0123.md`). This helps advertise your standard via [nomicon](https://nomicon.io/). Any related nomicon sections should be prefixed and styled using the following snippet:
 
-   ```
-   :::caution
-   This is part of proposed spec [NEP-123](https://github.com/near/NEPs/blob/master/neps/nep-0123.md) and subject to change.
-   :::
-   ```
+    ```
+    :::caution
+    This is part of proposed spec [NEP-123](https://github.com/near/NEPs/blob/master/neps/nep-0123.md) and subject to change.
+    :::
+    ```
 
-   - Once complete, submit the pull request for editor review.
-
-   - The formalization dance begins:
-     - NEP Editors, who are unopinionated shepherds of the process, check document formatting, completeness and adherence to [NEP-0001](neps/nep-0001.md) and approve the pull request.
-     - Once ready, the author updates the NEP status to `Review` allowing further community participation, to address any gaps or clarifications, normally part of the Review PR.
-     - NEP Editors mark the NEP as `Last Call`, allowing a 14 day grace period for any final community feedback. Any unresolved show stoppers roll the state back to `Review`.
-     - NEP Editors mark the NEP as `Final`, marking the standard as complete. The standard should only be updated to correct errata and add non-normative clarifications.
+   * Once complete, submit the pull request for editor review.
+  
+   * The formalization dance begins:
+     * NEP Editors, who are unopinionated shepherds of the process, check document formatting, completeness and adherence to [NEP-0001](neps/nep-0001.md) and approve the pull request.
+     * Once ready, the author updates the NEP status to `Review` allowing further community participation, to address any gaps or clarifications, normally part of the Review PR.
+     * NEP Editors mark the NEP as `Last Call`, allowing a 14 day grace period for any final community feedback. Any unresolved show stoppers roll the state back to `Review`.
+     * NEP Editors mark the NEP as `Final`, marking the standard as complete. The standard should only be updated to correct errata and add non-normative clarifications.
 
 Tip: build consensus and integrate feedback. NEPs that have broad support are much more likely to make progress than those that don't receive any comments. Feel free to reach out to the NEP assignee in particular to get help identify stakeholders and obstacles.
 
@@ -91,19 +92,19 @@ This repository uses [Docusaurus](https://docusaurus.io/) for the [Nomicon websi
 
    - Run the local docs development server
 
-     ```sh
-     # Start the site
-     yarn start
-     ```
+      ```sh
+      # Start the site
+      yarn start
+      ```
 
-     _Expected Output_
+      _Expected Output_
 
-     ```sh
-     # Website with live reload is started
-     Docusaurus server started on port 3000
-     ```
+      ```sh
+      # Website with live reload is started
+      Docusaurus server started on port 3000
+      ```
 
-     The website for docs will open your browser locally to port `3000`
+      The website for docs will open your browser locally to port `3000`
 
 2. Make changes to the docs
 
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Changes to the protocol specification and standards are called NEAR Enhancement 
 | [0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md) | Multi Token Standard                                              | @zcstarr @riqi @jriemann @marcos.sun        | Review |
 | [0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md) | Events Standard                                                   | @telezhnaya                                 | Final  |
 | [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Source Metadata                                                   | @BenKurrek                                  | Review |
-| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Efficient signature verification and hashing precompile functions |                                             | Draft  |
+| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Efficient signature verification and hashing host functions |                                             | Draft  |
 
 ## Specification
 
--- a/README.md
+++ b/README.md
@@ -10,23 +10,22 @@ Changes to the protocol specification and standards are called NEAR Enhancement 
 
 ## NEPs
 
-|NEP #   | Title  | Author  | Status  |
-|---|---|---|---|
-|[0001](https://github.com/near/NEPs/blob/master/neps/nep-0001.md)   | NEP Purpose and Guidelines  | @jlogelin  | Living  |
-|[0021](https://github.com/near/NEPs/blob/master/neps/nep-0021.md)   | Fungible Token Standard (Deprecated) | @evgenykuzyakov | Final |
-|[0141](https://github.com/near/NEPs/blob/master/neps/nep-0141.md)   | Fungible Token Standard | @evgenykuzyakov @oysterpack | Final |
-|[0145](https://github.com/near/NEPs/blob/master/neps/nep-0145.md)   | Storage Management | @evgenykuzyakov | Final |
-|[0148](https://github.com/near/NEPs/blob/master/neps/nep-0148.md)   | Fungible Token Metadata | @robert-zaremba @evgenykuzyakov @oysterpack | Final |
-|[0171](https://github.com/near/NEPs/blob/master/neps/nep-0171.md)   | Non Fungible Token Standard | @mikedotexe @evgenykuzyakov @oysterpack | Final |
-|[0177](https://github.com/near/NEPs/blob/master/neps/nep-0177.md)   | Non Fungible Token Metadata | @chadoh @mikedotexe | Final |
-|[0178](https://github.com/near/NEPs/blob/master/neps/nep-0178.md)   | Non Fungible Token Approval Management | @chadoh @thor314 | Final |
-|[0181](https://github.com/near/NEPs/blob/master/neps/nep-0181.md)   | Non Fungible Token Enumeration | @chadoh @thor314 | Final |
-|[0199](https://github.com/near/NEPs/blob/master/neps/nep-0199.md)   | Non Fungible Token Royalties and Payouts | @thor314 @mattlockyer | Final |
-|[0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md)   | Multi Token Standard | @zcstarr @riqi @jriemann @marcos.sun | Review |
-|[0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md)   | Events Standard | @telezhnaya | Final |
-|[0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md)   | Source Metadata | @BenKurrek | Review |
-
-
+| NEP #                                                             | Title                                                             | Author                                      | Status |
+| ----------------------------------------------------------------- | ----------------------------------------------------------------- | ------------------------------------------- | ------ |
+| [0001](https://github.com/near/NEPs/blob/master/neps/nep-0001.md) | NEP Purpose and Guidelines                                        | @jlogelin                                   | Living |
+| [0021](https://github.com/near/NEPs/blob/master/neps/nep-0021.md) | Fungible Token Standard (Deprecated)                              | @evgenykuzyakov                             | Final  |
+| [0141](https://github.com/near/NEPs/blob/master/neps/nep-0141.md) | Fungible Token Standard                                           | @evgenykuzyakov @oysterpack                 | Final  |
+| [0145](https://github.com/near/NEPs/blob/master/neps/nep-0145.md) | Storage Management                                                | @evgenykuzyakov                             | Final  |
+| [0148](https://github.com/near/NEPs/blob/master/neps/nep-0148.md) | Fungible Token Metadata                                           | @robert-zaremba @evgenykuzyakov @oysterpack | Final  |
+| [0171](https://github.com/near/NEPs/blob/master/neps/nep-0171.md) | Non Fungible Token Standard                                       | @mikedotexe @evgenykuzyakov @oysterpack     | Final  |
+| [0177](https://github.com/near/NEPs/blob/master/neps/nep-0177.md) | Non Fungible Token Metadata                                       | @chadoh @mikedotexe                         | Final  |
+| [0178](https://github.com/near/NEPs/blob/master/neps/nep-0178.md) | Non Fungible Token Approval Management                            | @chadoh @thor314                            | Final  |
+| [0181](https://github.com/near/NEPs/blob/master/neps/nep-0181.md) | Non Fungible Token Enumeration                                    | @chadoh @thor314                            | Final  |
+| [0199](https://github.com/near/NEPs/blob/master/neps/nep-0199.md) | Non Fungible Token Royalties and Payouts                          | @thor314 @mattlockyer                       | Final  |
+| [0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md) | Multi Token Standard                                              | @zcstarr @riqi @jriemann @marcos.sun        | Review |
+| [0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md) | Events Standard                                                   | @telezhnaya                                 | Final  |
+| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Source Metadata                                                   | @BenKurrek                                  | Review |
+| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Efficient signature verification and hashing precompile functions |                                             | Draft  |
 
 ## Specification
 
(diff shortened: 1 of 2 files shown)


mikedotexe:
bump 😇

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+
+With regards to hashing, there are already several hashing functions implemented natively.
+For our use case, we need to be able to use `blake2b` has function, which is not available.
+
+## Rationale and alternatives
+
+Add
+
+```rust
+pub fn ed25519_verify(sig: &ed25519::Signature, msg: &[u8], pub_key: &ed25519::Public) -> bool
```
sascha1337, line 36: 👀

README.md
```diff
@@ -10,23 +10,22 @@ Changes to the protocol specification and standards are called NEAR Enhancement
...
+| [0181](https://github.com/near/NEPs/blob/master/neps/nep-0181.md) | Non Fungible Token Enumeration                                    | @chadoh @thor314                            | Final  |
+| [0199](https://github.com/near/NEPs/blob/master/neps/nep-0199.md) | Non Fungible Token Royalties and Payouts                          | @thor314 @mattlockyer                       | Final  |
+| [0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md) | Multi Token Standard                                              | @zcstarr @riqi @jriemann @marcos.sun        | Review |
+| [0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md) | Events Standard                                                   | @telezhnaya                                 | Final  |
+| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Source Metadata                                                   | @BenKurrek                                  | Review |
+| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Efficient signature verification and hashing precompile functions |                                             | Draft  |
```
mfornet, line 28:
```suggestion
| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Efficient signature verification and hashing host functions |                                             | Draft  |
```

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+especially in PoS consensus mechanisms. Our rough numbers say that we’ll be
+verifying ~200 signatures every minute (Polkadot’s authority set is 300 signers).
+
+Therefore, we need a mechanism to perform these operations in a cost-effective way
+in terms of gas and speed. Currently, NEAR does not have any native signature verification toolbox.
+This implies that a light client operating insid NEAR will have to import a library
+compiled to WASM as mentioned in [Zulip](https://near.zulipchat.com/#narrow/stream/295302-general/topic/light_client).
+
+With regards to hashing, there are already several hashing functions implemented natively.
+For our use case, we need to be able to use `blake2b` has function, which is not available.
```
nagisa, line 29:
```suggestion
For our use case, we need to be able to use the `blake2b` hash function, which is not currently available.
```

neps/nep-0364.md
nagisa: This addition does not obviously follow from the text in the motivation section above. The motivation mentioned blake2b only. It would help if the motivation described what the signature verification algorithm looks like, at least in somewhat broad terms, so that an informed decision can be made which of these functions are sufficiently high-value additions.

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+
+- Why is this design the best in the space of possible designs?
+  Adding exisiting proved and vetted crypto crates into the runtime is a safe workaround. It will boost performance
+  between 20-25x according to our benchmarks. This will both reduce operation costs significantly and will also
+  enable the contract to verify all the signatures in one transaction, which will simplify the contract design.
+
+- What other designs have been considered and what is the rationale for not choosing them?
+  This has been answered before.
+
+## Specification
```
nagisa, line 130: The list of native host functions being added should probably be moved from the section above to this section. This section ought to describe what the protocol looks like after the change is made and the section above should be limited to describing _why_ the change like this is being made.

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+
+Overall: the difference between the two versions (native vs wasi + singlepass are)
+
+```
+ed25519: 25.64x slower
+ecdsa: 20.75x slower
+shnorrkel: 24.86x slower
+```
+
+- What is the impact of not doing this?
```
nagisa, line 117:
```suggestion
### What is the impact of not doing this?
```

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+  Costs of running IBC compatible trustless bridges would be very high. Plus, the fact that signature verification
+  is such an expensive operation will force the contract to split the process of batch verification of signatures
+  into multiple transactions.
+
+- Why is this design the best in the space of possible designs?
```
nagisa, line 122:
```suggestion
### Why is this design the best in the space of possible designs?
```

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+  Adding exisiting proved and vetted crypto crates into the runtime is a safe workaround. It will boost performance
+  between 20-25x according to our benchmarks. This will both reduce operation costs significantly and will also
+  enable the contract to verify all the signatures in one transaction, which will simplify the contract design.
+
+- What other designs have been considered and what is the rationale for not choosing them?
+  This has been answered before.
```
nagisa, line 128:
```suggestion
### What other designs have been considered and what is the rationale for not choosing them?
  <replicate the text from the answer above>
```

nagisa (commented):
Largely just wording and structure nits from me.

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+## Summary
+
+This NEP introduces the request of adding into the NEAR runtime a set of pre-compile
+functions that can help IBC compatible light clients run on-chain.
+
+## Motivation
+
+Signature verification and hashing are ubiquitous operations in light clients,
+especially in PoS consensus mechanisms. Our rough numbers say that we’ll be
+verifying ~200 signatures every minute (Polkadot’s authority set is 300 signers).
```
bowenwang1996, line 21: What does "we" here refer to? How is the number 200 calculated?

neps/nep-0364.md
bowenwang1996: Agree. For each host function, please explain why exactly they are needed and their specification

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+# transaction id DpWpK2jnyBRgZ7fScaBKULikJyvmMjDHx7AyWMrJL5VB
+
+near call sitoula-test.testnet verify_ecdsa '{"signature_p1": [231, 117, 17, 89, 49, 142, 111, 201, 161, 107, 167, 147, 215, 167, 196, 226, 200, 176, 184, 62, 196, 240, 210, 137, 77, 198, 90, 97, 201, 212, 96, 229], "signature_p2": [1, 31, 7, 121, 178, 247, 150, 131, 108, 250, 173, 71, 100, 192, 83, 64, 145, 85, 254, 69, 176, 7, 114, 89, 64, 205, 30, 243, 193, 78, 142, 27], "msg": [107,97,106,100,108,102,107,106,97,108,107,102,106,97,107,108,102,106,100,107,108,97,100,106,102,107,108,106,97,100,115,107], "iterations": 10}' --accountId sitoula-test.testnet --gas 300000000000000
+# transaction id AKFwCuQ8g2a8t9SX7vcve348BvwoU3M42MSef7gy1kPA
+```
+
+With `iterations = 130` **all these calls return ExecutionError**: `'Exceeded the maximum amount of gas allowed to burn per contract.'`
+With iterations = 50 these are the results:
```
bowenwang1996, line 60: Is one iteration one signature verification?

neps/nep-0364.md
```diff
@@ -0,0 +1,153 @@
...
+
+- What is the impact of not doing this?
+  Costs of running IBC compatible trustless bridges would be very high. Plus, the fact that signature verification
+  is such an expensive operation will force the contract to split the process of batch verification of signatures
+  into multiple transactions.
+
+- Why is this design the best in the space of possible designs?
+  Adding exisiting proved and vetted crypto crates into the runtime is a safe workaround. It will boost performance
+  between 20-25x according to our benchmarks. This will both reduce operation costs significantly and will also
+  enable the contract to verify all the signatures in one transaction, which will simplify the contract design.
```
bowenwang1996, line 125: I don't necessarily agree that it is a safe workaround. The cost of introducing a new host function is actually quite high:
* A bug in the host function code would cause the entire network to go down
* Once the host function is introduced, one cannot remove it as that could break contracts which already rely on the behavior of the host functions.

bowenwang1996 (commented):
This NEP seems to imply that compiling these functions to wasm is actually feasible, though not desirable, for IBC use cases. I'd like to see a more detailed discussion on the alternatives and cost of not introducing this change. In general, we should be very carefully with adding more host functions.

blasrodri:
Thanks for the feedback. I've incorporated your comments.

I've removed the hashing function from this NEP so that it only focuses on signature verification. We'll likely need another one for introducing new hashes that aren't supported.

neps/nep-0364.md
blasrodri: Maybe "safe workaround" isn't the right way of putting it. However, these are highly adopted libraries with operations that are part of multiple services and blockchains. In that way, I was trying to convey the fact that these are no new functions that haven't been battle-tested. If you have any suggestions on how to communicate it differently I'd be happy to update it.

neps/nep-0364.md
blasrodri: Yes

blasrodri: > Agree. For each host function, please explain why exactly they are needed and their specification

I've added a reference in the motivation section. Would you like it to be more specific, being included in the comment of each host function?

neps/nep-0364.md
blasrodri: I've moved it, thanks

neps/nep-0364.md
blasrodri: Removed blake2b from this NEP. Will likely create another one that requests for all the hashing functions that we need.

neps/nep-0364.md
```diff
@@ -0,0 +1,171 @@
...
+
+This NEP aims to introduce the following host functions:
+
+```rust
+pub fn ed25519_verify(sig: &ed25519::Signature, msg: &[u8], pub_key: &ed25519::Public) -> bool
+pub fn sr25519_verify(sig: &sr25519::Signature, msg: &[u8], pub_key: &sr25519::Public) -> bool
+pub fn ecdsa_verify(sig: &ecdsa::Signature, msg: &[u8], pub_key: &ecdsa::Public) -> bool
+```
+
+Once this NEP is approved and integrated, these four functions will be available in the `near_sdk` crate in the
```
nagisa, line 150:
```suggestion
Once this NEP is approved and integrated, these functions will be available in the `near_sdk` crate in the
```

neps/nep-0364.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Specification
+
+This NEP aims to introduce the following host functions:
+
+```rust
+pub fn ed25519_verify(sig: &ed25519::Signature, msg: &[u8], pub_key: &ed25519::Public) -> bool
+pub fn sr25519_verify(sig: &sr25519::Signature, msg: &[u8], pub_key: &sr25519::Public) -> bool
+pub fn ecdsa_verify(sig: &ecdsa::Signature, msg: &[u8], pub_key: &ecdsa::Public) -> bool
+```
```
nagisa, line 148:
```suggestion
pub fn ed25519_verify(sig: &ed25519::Signature, msg: &[u8], pub_key: &ed25519::Public) -> bool;
pub fn sr25519_verify(sig: &sr25519::Signature, msg: &[u8], pub_key: &sr25519::Public) -> bool;
pub fn ecdsa_verify(sig: &ecdsa::Signature, msg: &[u8], pub_key: &ecdsa::Public) -> bool;
```
```

File: neps/nep-0364.md [hunk 13]
```diff
@@ -0,0 +1,171 @@
...
+- SR25519 signature verification function
+- ECDSA signature verification function
+
+using `ed25519_dalek`, `schnorrkel` and `secp256k1` crates into NEAR runtime as pre-compiled functions.
+
+Note: seems like `ecdsa_verify` is implemented, but called `erecover` in https://github.com/near/nearcore/blob/master/runtime/near-vm-logic/src/logic.rs#L1049
+Upon confirmation, we can proceed to remove it.
+
+Benchmarks were run using a signature verifier smart contract on-chain importing the aforementioned functions from
+widespread used crypto Rust crates. The biggest pitfall of that solution
```
nagisa, line 50: What solution is “that solution” referring to here? I would imagine computing these signatures in WASM? Please make the wording more precise.

neps/nep-0364.md
```diff
@@ -0,0 +1,171 @@
...
+
+Adding exisiting proved and vetted crypto crates into the runtime is a safe workaround. It will boost performance
+between 20-25x according to our benchmarks. This will both reduce operation costs significantly and will also
+enable the contract to verify all the signatures in one transaction, which will simplify the contract design.
+
+### What other designs have been considered and what is the rationale for not choosing them?
+
+Adding exisiting proved and vetted crypto crates into the runtime is a safe workaround. It will boost performance
+between 20-25x according to our benchmarks. This will both reduce operation costs significantly and will also
+enable the contract to verify all the signatures in one transaction, which will simplify the contract design.
```
nagisa, line 138: This is just a copy of the text from a subsection above, isn’t it?

One alternative I can think of is improving the runtime implementation sufficiently that it compiles WASM code to a sufficiently fast machine code in the first place. It won’t be as fast as LLVM produced code due to the need for gas accounting and such, and it would take a ton of effort, compared to adding a host function, but it is an alternative worth spelling out.

Another alternative is to deal with high cost of computing/verifying these signatures in some other manner – such as decreasing the overall cost of gas and increasing the limits of gas available to attach to the contract. Again, not as straightforward as a host function to implement and is less localized of a change.

blasrodri:
> This NEP seems to imply that compiling these functions to wasm is actually feasible, though not desirable, for IBC use cases. I'd like to see a more detailed discussion on the alternatives and cost of not introducing this change. In general, we should be very carefully with adding more host functions.

I've mentioned the size of the validator set, and how there are plans to increase it. With the current gas costs and performance of the wasm code, we can only verify 130 signatures on one transaction using the most expensive scheme. Based on this [discussed proposal](https://polkadot.polkassembly.io/referendum/16) the validator set can be increased to 1000, which will require verifying at least 667 signatures. 

Based on the performance improvement of 20-25x, doing this natively gives us sufficient breathing room for these numbers and makes the infrastructure a bit more "future-proof".

abacabadabacaba (changes requested):
I'm totally fine with adding Ed25519 (specifically, the variant that NEAR is already using), and we already have a host function for secp256k1. However, I am against the idea of adding Sr25519.

In general, I think that we should strive to reduce the number of host functions that we have, and only add new ones if there is a compelling need. Currently, almost all of the Math API host functions implement functionality that is available in the EVM. EVM is used in many blockchains, and is de facto a standard. Ed25519 is also widely used in NEAR and elsewhere. Sr25519, however, has very little usage. We don't want to add every cryptographic function existing somewhere, as this will increase our maintenance burden too much. In fact, I wasn't even able to find a specification for this function, and we certainly don't want to add a function without a specification.

I would recommend @blasrodri to explore the avenues for optimizing a pure-WASM implementation instead. It seems that Sr25519 implements some sort of aggregate signatures, which could be verified more quickly than verifying each signature individually. Also, it is probably possible to do batch verification, similar to Ed25519. I couldn't find it implemented though, which is yet another sign that it is not mature and/or popular enough.

blasrodri:
> I'm totally fine with adding Ed25519 (specifically, the variant that NEAR is already using), and we already have a host function for secp256k1. However, I am against the idea of adding Sr25519.
> 
> In general, I think that we should strive to reduce the number of host functions that we have, and only add new ones if there is a compelling need. Currently, almost all of the Math API host functions implement functionality that is available in the EVM. EVM is used in many blockchains, and is de facto a standard. Ed25519 is also widely used in NEAR and elsewhere. Sr25519, however, has very little usage. We don't want to add every cryptographic function existing somewhere, as this will increase our maintenance burden too much. In fact, I wasn't even able to find a specification for this function, and we certainly don't want to add a function without a specification.
> 
> I would recommend @blasrodri to explore the avenues for optimizing a pure-WASM implementation instead. It seems that Sr25519 implements some sort of aggregate signatures, which could be verified more quickly than verifying each signature individually. Also, it is probably possible to do batch verification, similar to Ed25519. I couldn't find it implemented though, which is yet another sign that it is not mature and/or popular enough.

Thanks for your input. I've cleaned up and left only `Ed25519`. We'll figure out `Sr25519` if we need to do so. And, as you mentioned, we can use `erecover` for ECDSA. Please let me know if the current update is in line with what you want.

abacabadabacaba:
@blasrodri The current update is good in terms of the functionality. There are some technical changes that are needed, though. In particular, the type of the arguments and the return value of host functions must be `u32` or `u64` for technical reasons (near-sdk can provide a wrapper function with a more convenient signature).

blasrodri:
> @blasrodri The current update is good in terms of the functionality. There are some technical changes that are needed, though. In particular, the type of the arguments and the return value of host functions must be `u32` or `u64` for technical reasons (near-sdk can provide a wrapper function with a more convenient signature).

Thanks! I've added both the spec for the host function itself, and a `rust-sdk` version of it.

neps/nep-0364.md
```diff
@@ -0,0 +1,181 @@
...
+/// `base + write_register_base + write_register_byte * num_bytes + ed25519_verify_base + ed25519_verify_byte * num_bytes`
+pub fn ed25519_verify(
+    &mut self,
+    sig_len: u64,
+    sig_ptr: u64,
+    msg_len: u64,
+    msg_ptr: u64,
+    pub_key_len: u64,
+    pub_key_ptr: u64,
+    register_id: u64,
```
abacabadabacaba, line 143: What is `register_id` for? The result is just a boolean, it can be returned as `u64`.

blasrodri: I'll update it then!

blasrodri:
@abacabadabacaba anything missing?

bowenwang1996 (commented):
@blasrodri is there an implementation of this proposal?

neps/nep-0364.md
```diff
@@ -0,0 +1,180 @@
...
+## Security Implications (Optional)
+
+We have chosen these crates because they are the ones already integrated into the
+[sp_io](https://docs.rs/sp-io/latest/sp_io/)](https://docs.rs/sp-io/latest/sp_io/)
+crate heavily used in the [substrate](https://github.com/paritytech/substrate) repository.
+
+## Unresolved Issues (Optional)
+
+- What parts of the design do you expect to resolve through the NEP process before this gets merged?
+  Both the function signatures and crates are up for discussion.
```
bowenwang1996, line 166: I suggest that we use the library that nearcore already uses for ed25519 signatures. This would avoid introducing new dependencies and ensure that NEAR signatures themselves can be 100% verified in smart contracts

blasrodri: As far as I can see the proposal is the same as the one being used today
https://github.com/near/nearcore/blob/master/core/crypto/src/signature.rs#L729-L732
https://github.com/near/NEPs/pull/364/files#diff-1d8966e006ab1c29466f5c4d5ce49a965477f992e7d16c1d032d5b0164096459R38

blasrodri:
> @blasrodri is there an implementation of this proposal?

https://github.com/near/nearcore/pull/7165

blasrodri:
> > @blasrodri is there an implementation of this proposal?
> 
> [near/nearcore#7165](https://github.com/near/nearcore/pull/7165)

If I get the green light, I can start focusing on the gas costs calculations for this.

bowenwang1996:
From my perspective the proposal is good to go. @abacabadabacaba any other concerns from your side?

neps/nep-0364.md
```diff
@@ -0,0 +1,180 @@
...
+
+```rust
+/// Ed25519 is a public-key signature system with several attractive features
+///
+/// Proof of Stake Validator sets can contain different signature schemes.
+/// Ed25519 is one of the most used ones across blockchains, and hence it's importance to be added.
+/// For further reference, visit: https://ed25519.cr.yp.to
+/// # Cost
+///
+/// `base + write_register_base + write_register_byte * num_bytes + ed25519_verify_base + ed25519_verify_byte * num_bytes`
```
jakmeier, line 134: Where did these gas costs come from? Please see my comment on the implementation PR and update accordingly. https://github.com/near/nearcore/pull/7165#discussion_r942131909

blasrodri: Thanks! Will update this as soon as the PR reflects the right costs.

ori-near:
Discussing and making decisions on NEPs is crucial for helping the protocol evolve. As @bowenwang1996 discussed in the NEAR Public Protocol Meeting today, the NEP process introduces Working Groups as a way to operationalize the review of NEPs that builds community support. I propose that we use this NEP as the first one to help bootstrap the Protocol work group.

@marmichalski @mfornet – Could you please fully read this NEP and comment in the thread with any feedback/concerns, and specifically if you would approve this NEP? If not, please specify why and what it would take for you to approve it. I can schedule the first working group meeting after your review. Thank you!

abacabadabacaba (approved):
LGTM in terms of the functionality. It is still needed to specify the exact behavior on all possible inputs (including invalid ones), as well as the precise behavior of the Ed25519 verification (`ed25519-dalek` crate supports several different behaviors, depending on the function used as well as the feature flags). Also, some grammar fixes are still needed.

blasrodri:
> LGTM in terms of the functionality. It is still needed to specify the exact behavior on all possible inputs (including invalid ones), as well as the precise behavior of the Ed25519 verification (`ed25519-dalek` crate supports several different behaviors, depending on the function used as well as the feature flags). Also, some grammar fixes are still needed.

Can you please guide me on how to create these tests? Not sure how to construct the input.

bowenwang1996:
I lean towards accepting this proposal based on the following:
* The proposal addresses an important need in the ecosystem. It would enable more bridges to be built and allow transaction signature verification within a smart contract.
* The proposal is not too invasive. We already use ed25519 signatures extensively in the protocol and the implementation of this new host function does not create a lot more additional complexity in the protocol itself.
* The [implementation](https://github.com/near/nearcore/pull/7165) is almost complete and shows the feasibility of the proposal.

bowenwang1996 (approved):

neps/nep-0364.md
```diff
@@ -0,0 +1,188 @@
...
+
+Costs of running IBC-compatible trustless bridges would be very high. Plus, the fact that signature verification
+is such an expensive operation will force the contract to split the process of batch verification of signatures
+into multiple transactions.
+
+### Why is this design the best in the space of possible designs?
+
+Adding existing proved and vetted crypto crates into the runtime is a safe workaround. It will boost performance
+between 20-25x according to our benchmarks. This will both reduce operating costs significantly and will also
+enable the contract to verify all the signatures in one transaction, which will simplify the contract design.
```
mm-near, line 107: If Polkadot uses 3 different types of keys - are you still able to verify all the signatures in one transaction ? (for example in a scenario, where all the polkadot validators are using sr25519 ? )

mm-near:
First - thanks a lot @blasrodri for the PR with the implementation.

I'm a little bit undecided here:

**On the pro-side:** there are the things that @bowenwang1996 mentioned.

**On the con-side:** seems that the main reason was to save gas and simplify contract code (to handle all verification in one transaction). But with Polkadot having 3 different type of keys, the contract still has to handle the scenario where all of them are non-ed25519. I'd like to understand if this NEP helps in such scenario.

neps/nep-0364.md
```diff
@@ -0,0 +1,188 @@
...
+## Specification
+
+This NEP aims to introduce the following host function:
+
+```rust
+/// Ed25519 is a public-key signature system with several attractive features
+///
+/// Proof of Stake Validator sets can contain different signature schemes.
+/// Ed25519 is one of the most used ones across blockchains, and hence it's importance to be added.
+/// For further reference, visit: https://ed25519.cr.yp.to
```
mfornet, line 131: I wonder if we should add more inline information about how `ed25519` works, in particular how it handles some of the edge cases. An alternative to add this information inline would be to add this information to [Nomicon](https://nomicon.io/), given `ed25519` is already a cyptographic primitive used in NEAR before this NEP.

The motivation is to allow potential new client implementations to be able to build the client by only reading Nomicon and NEPs.

mfornet (approved):
I lean towards approving this proposal.

This is a step forward to enable new kind of Light Clients, and this could help developing some kind of privacy solutions on chain as well.

---

In particular, main motivation is to support a Light Client with Polkadot ecosystem, but adding this single host function doesn't seem enough, in case the other types of signatures are used. This means that it is expected more host functions of this kind will be continued to be proposed. For example it seems just as relevant to add BLS primitives to support a Light Client with Ethereum PoS.

Supporting this particular proposal is easier giving ED25519 is extensively used in NEAR client already, however it is not clear it will be as easy for different types of crypto-primitives.

ori-near:
Thank you @bowenwang1996, @mfornet, and @mm-near for reviewing this NEP! Based on your comments, it seems like this proposal is close to a decision. We are therefore scheduling the first Protocol Standards Working group meeting next week, where this NEP can enter the final voting stage. Please register below.

Meeting Info
📅 Date: Tuesday, Sep 6, 4pm UTC
🗒️ Topic: NEP-364
✏️ [Register](https://near-org.zoom.us/webinar/register/5416617893723/WN_V7GuJyh6T96eEUjulochaQ)

What are Working Groups?
An initiative to bootstrap a way to review NEPs that builds community support and contributes to the NEAR protocol and standards. You can learn more about NEPs and the working groups here: https://near.org/working-groups/.

bowenwang1996:
> First - thanks a lot @blasrodri for the PR with the implementation.
> 
> I'm a little bit undecided here:
> 
> **On the pro-side:** there are the things that @bowenwang1996 mentioned.
> 
> **On the con-side:** seems that the main reason was to save gas and simplify contract code (to handle all verification in one transaction). But with Polkadot having 3 different type of keys, the contract still has to handle the scenario where all of them are non-ed25519. I'd like to understand if this NEP helps in such scenario.

We already have secp signature verification through ecrecover and my understanding is that sr25519 isn't widely used. cc @blasrodri Also, this proposal is not specific to bridge to polkadot. It allows other type of bridges to be built (for example between different NEAR chains).

mm-near:
I lean towards approving this proposal.

README.md
```diff
@@ -10,23 +10,22 @@ Changes to the protocol specification and standards are called NEAR Enhancement
...
+| [0181](https://github.com/near/NEPs/blob/master/neps/nep-0181.md) | Non Fungible Token Enumeration                                    | @chadoh @thor314                            | Final  |
+| [0199](https://github.com/near/NEPs/blob/master/neps/nep-0199.md) | Non Fungible Token Royalties and Payouts                          | @thor314 @mattlockyer                       | Final  |
+| [0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md) | Multi Token Standard                                              | @zcstarr @riqi @jriemann @marcos.sun        | Review |
+| [0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md) | Events Standard                                                   | @telezhnaya                                 | Final  |
+| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Source Metadata                                                   | @BenKurrek                                  | Review |
+| [0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) | Efficient signature verification and hashing host functions |                                             | Draft  |
```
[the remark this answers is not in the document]
matklad: unrelated to this specific PR, but I see a process issue (so, cc @frol): our process for NEPs should not require modification of shared files (like readme.md), as that is very prone to conflicts.

matklad:
One interesting unresolved question [came up during implementation](https://github.com/near/nearcore/pull/7165#discussion_r961800700): what's the proper error handling for invalid keys and signatures? 

Not every byte sequence is a syntactically valid key/sig, so we have to do *something* about that. 

Our choices are: 

* return `false` (ie, just consider this to be invalid signature for the message)
* return some kind of error (so, the return would be `ok | invalid key | invalid sig | wrong sig`)
* consider this to be a programming error, and terminate contract execution. 

There are two reasons why a key or a signature might be invalid.

_First_, the byte length could be wrong. I think it's pretty clear that this should just be programming error -- the contract should pass `[u8; 32]` to the most function call. 

_Second_, even if the length is correct, the key or signature might still be considered invalid due to this conditions:

* https://docs.rs/ed25519-dalek/latest/src/ed25519_dalek/public.rs.html#142
* https://docs.rs/ed25519/latest/src/ed25519/lib.rs.html#302

In the case of invalid signature, I am somewhat confident that the right approach is just `return false`. It seems logical that, if the signature is trivially invalid, than it's just an invalid signature. 


I don't know what's the right answer for invalid public key. On the one hand, public keys is usually something you get from the store, so you can consider them valid. On the other hand, if someone *creates* a public key, you'd have to have some mechanism to ensuring that it is in fact valid. I *think* `return false` is probably the right strategy here as well, but I'd love to hear from people more knowledgeable than me :)

In any case, error handling behavior should be documented as a part of the nep.

frol:
@blasrodri Please, revert the changes to the README.md file. We will automate the process of listing NEPs, but for now, I will update it manually after we merge the NEP.

frol (approved):
Thank you to everyone who attended the first Protocol working group meeting today! The working group members reviewed the NEP and reached the following consensus:

Status: Approved with a request to address remaining comments (see below)
* [x] @bowenwang1996 
* [x] @mfornet 
* [x] @mm-near 

(Meeting Recording is coming soon)

Remaining requests to @blasrodri:

* [ ] Address the comment from @matklad: https://github.com/near/NEPs/pull/364#issuecomment-1238147491 (specify the invalid input behavior)
* [ ] Address the comment from @abacabadabacaba: https://github.com/near/NEPs/pull/364#pullrequestreview-1074852114 (specify the exact ED25519 mode that is expected to be used)

matklad:
Implementaiton tracking issue: https://github.com/near/nearcore/issues/7567

blasrodri:
@matklad @abacabadabacaba I've addressed your comments on commit 
bdf64d0
Feel free to let me know if there's anything that can be improved.

neps/nep-0364.md
```diff
@@ -164,6 +173,8 @@ And a `rust-sdk` possible implementation could look like this:
 pub fn ed25519_verify(sig: &ed25519::Signature, msg: &[u8], pub_key: &ed25519::Public) -> bool;
 
 ```
+The current implementation is imported from the crate `ed25519-dalek`, version 1. It uses no feature flags
+except from the default one. This is the exact same setup used in the crate `near-crypto`.
```
frol, line 177: @abacabadabacaba Can you maybe help to identify how to state this requirement in a crate-agnostic way, so we don't get tied to the exact implementation?

neps/nep-0364.md
```diff
@@ -146,6 +146,15 @@ This NEP aims to introduce the following host function:
...
 ///  ed25519_verify_base + ed25519_verify_byte * (num_bytes_signature + num_bytes_message)`
+///
+/// # Errors
+///
+/// The signature size is fixed and equal to SIGNATURE_LENGTH (64 bytes). In case the length of signature input is not equal to
+/// SIGNATURE_LENGTH, then the function returns HostError::Ed25519VerifyInvalidInput with the message "invalid signature length".
+/// 
+/// A similar case occurs with the public key. Its size is known and its equal to PUBLIC_KEY_LENGTH (32 bytes). In case the
+/// input length provided for the publc key doesn't match PUBLIC_KEY_LENGTH, the function will return the following error:
+/// HostError::Ed25519VerifyInvalidInput with the message "invalid public key length".
```
frol, line 157: @matklad Do you think it is a good direction to go with extending `HostError` type or should we keep the scope of potential errors minimal?

matklad: I think this doesn't matter much: from the protocol perspective, how exactly `HostError` looks is unobservable. Implementation wise, adding a new variant there is OK.


But this actually points to the problem with wording of this NEP: what is written here is how the function would look inside `logic.rs`, but that's an impl detail which is completely irrelevant from the specification point of view. 

NEPs should say how the WASM import looks, not how it is implemented. For the time being, I'd rephrase this section as:

== Specification

The following host function is added

```rust
extern "C"{

/// Verify an ED25519 signature given a message and a public key.
/// # Returns
/// - 1 if the signature was properly verified
/// - 0 if the signature failed to be verified
///
/// # Cost
///
/// Each input can either be in memory or in a register. Set the length of the input to `u64::MAX`
/// to declare that the input is a register number and not a pointer.
/// Each input has a gas cost input_cost(num_bytes) that depends on whether it is from memory
/// or from a register. It is either read_memory_base + num_bytes * read_memory_byte in the
/// former case or read_register_base + num_bytes * read_register_byte in the latter. This function
/// is labeled as `input_cost` below.
///
/// `input_cost(num_bytes_signature + num_bytes_message + num_bytes_public_key) +
///  ed25519_verify_base + ed25519_verify_byte * num_bytes_message`
///
/// # Errors
///
/// If the signature size is not equal to 64 bytes, or public key length is not equal to 32 bytes, contract execution is terminated with an error. 
  fn ed25519_verify(
    sig_len: u64,
    sig_ptr: u64,
    msg_len: u64,
    msg_ptr: u64,
    pub_key_len: u64,
    pub_key_ptr: u64,
  ) -> u64;
}
```

That is: 

* write the exten function from perspecive of the contract
* don't mention HostError, as that is unobservable to contracts (it is more or less like panic)

We could perhaps consider using proper wasm for the spec here

```wasm
(module 
  (import "env" "ed25519_verify" (func (param 64) ... (result i64)))
)
```

but I feel that Rust-syntax to experss that is more accessible, and better matches what we've been doing historically

neps/nep-0364.md
```diff
@@ -0,0 +1,199 @@
...
+///
+/// Each input can either be in memory or in a register. Set the length of the input to `u64::MAX`
+/// to declare that the input is a register number and not a pointer.
+/// Each input has a gas cost input_cost(num_bytes) that depends on whether it is from memory
+/// or from a register. It is either read_memory_base + num_bytes * read_memory_byte in the
+/// former case or read_register_base + num_bytes * read_register_byte in the latter. This function
+/// is labeled as `input_cost` below.
+///
+/// `input_cost(num_bytes_signature + num_bytes_message, num_bytes_public_key) +
+///  ed25519_verify_base + ed25519_verify_byte * (num_bytes_signature + num_bytes_message)`
```
matklad, line 148: Double check that the cost matches what we actually charge. In partiuclar, I don't think we count signature lenght as a part of this component?

blasrodri: [in the implementation](https://github.com/near/nearcore/pull/7165/files#diff-8df9ae8ecbf1224f7f000dd60280838a7399a49e00af2754cc41f5a4b1ac2640R1162) there's only variable cost for the message length. I can adjust it to match that.

blasrodri: I've updated it

neps/nep-0364.md
blasrodri: Updated it

neps/nep-0364.md
abacabadabacaba: @frol It is difficult to describe the specific behavior concisely without referring to a specific implementation. [This blog post](https://hdevalence.ca/blog/2020-10-04-its-25519am) describes the various ways in which the existing Ed25519 implementations differ in practice. The behavior that we are using, which is shared by Go `crypto/ed25519`, Rust `ed25519-dalek` (using `verify` function with `legacy_compatibility` feature turned off) and several others, makes the following decisions:

* The encoding of the values $R$ and $s$ must be canonical, while the encoding of $A$ doesn't need to.
* The verification equation is $R=[s]B-[k]A$.
* No additional checks are performed. In particular, the points outside of the order-<span></span>$l$ subgroup are accepted, as are the points in the torsion subgroup.

While the above hopefully describes the behavior that we seek in an unambiguous manner, we may still want to point to a specific implementation to refer to in case of an ambiguity.

frol: @abacabadabacaba Thank you!

@blasrodri Please, include the proposed details into the NEP

blasrodri: @frol updated it

abacabadabacaba: @blasrodri Probably Security Implications is not the most appropriate place for this, this should probably rather go just above this thread. Also, whenever the `ed25519-dalek` crate is mentioned, it should be specified that `verify` function should be used. This is because there are other verification functions in this crate (`verify_strict`, `verify_batch`) that behave differently.

frol:
@blasrodri It seems there is one last remaining suggestion from @abacabadabacaba.

@matklad Do you feel all your comments were addressed and we are good to merge this NEP?

matklad:
Yeah, this looks good to me.

blasrodri:
> @blasrodri It seems there is one last remaining suggestion from @abacabadabacaba.
> 
> @matklad Do you feel all your comments were addressed and we are good to merge this NEP?

thanks for following up. I've updated @abacabadabacaba comments. Hopefully that's what he meant.

neps/nep-0364.md
```diff
@@ -0,0 +1,203 @@
...
+
+And a `rust-sdk` possible implementation could look like this:
+
+```rs
+
+pub fn ed25519_verify(sig: &ed25519::Signature, msg: &[u8], pub_key: &ed25519::Public) -> bool;
+
+```
+The current implementation is imported from the crate `ed25519-dalek`, version 1. It uses no feature flags
+except from the default one. This is the exact same setup used in the crate `near-crypto`.
```
abacabadabacaba, line 172: This sentence now seems redundant, given that the expected behavior is described below more thoroughly.

abacabadabacaba (approved):
LGTM

frol (approved):
Congrats everyone! I am starting the merge process now

@blasrodri Thanks for championing this effort!

Thanks to all the reviewers for your thorough work!

jakmeier:
Stabilization has just been merged: https://github.com/near/nearcore/pull/8098

Expect this to land in testnet with the next testnet release.

PR #365, opened 2022-06-21
(draft) move RPC API docs

bucanero:


bucanero:
> @thisisjoshford @bucanero Is the intent to move the RPC API reference from the docs to nomicon? It is still going to happen, or should we close it now?

hello @frol 👋 

This was mostly a proof-of-concept, and we decided to keep the RPC API reference under `docs.near.org`. Let's close this draft PR as it's no longer needed. 👍

PR #370, opened 2022-07-11
Injected Wallet Standard

lewis-sqa:
This NEP has been re-authored in a new PR, so it could be adopted from the original author for some tweaks to be made -- thanks so much for your time and efforts to get it this far, @lewis-sqa  :)

See: [NEP-408](https://github.com/near/NEPs/pull/408)


specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+
+Although we had some technical challenges, the decision to store the key pairs in the dApp means users own them and gas-only intensive `FunctionCall` dApps work seamlessly with wallets that aren't always available. The expectation for `near-api-js` to become wallet-agnostic will reduce the complexity shifted onto the dApp as it can abstract away the logic required for handling the key pairs and only defer to the wallet when further permission is needed.
+
+A side effect to this approach (coupled with multiple accounts) means we must generate the key pairs ahead of time and pass each public key to the `signIn` method as a list of `accounts`. The introduction of the `connect` and `disconnect` methods allows dApps to request visibility for a subset of accounts imported in the wallet to help populate this list and sign transactions.
+
+## Specification
+
+Injected wallets are browser extensions that implement the `Wallet` API (see below) on the `window` object. To avoid namespace collisions and easily detect when they're available, wallets will mount under their own key within `window.near` (e.g. `window.near.sender`).
```
lostpebble, line 46: I think this may open up the door to some potential issues. Sender is already using `window.near` to host their wallet API- but besides that, to assign deeply (2 levels) into the window object might make people inadvertently remove previous assignments:

```ts
window.near = {}; // This will remove all other implementations
window.near.sender = { ...senderApi };
```

A way to get around it would be:

```ts
if (window.near == null) {
  window.near = {};
}

window.near.sender = { ...senderApi };
```

Hopefully people implement their extensions like this- but having this kind of deep assignment does open to door to people getting it wrong and inadvertently voiding others' APIs.

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+// Retrieve the list of accounts we have visibility of.
+const { accounts } = window.near.wallet;
+
+// Request FunctionCall access to the 'guest-book.testnet' smart contract for each account.
+await window.near.wallet.signIn({
+  permission: {
+    receiverId: "guest-book.testnet",
+    methodNames: [],
+  },
+  accounts: accounts.map(({ accountId }) => {
```
lostpebble, line 303: So is it expected that `connect()` and `signIn()` are now separate actions that the user needs to respond to in the wallet ?

It's mentioned that during `connect()` the user must select which accounts they want to "expose" to the dapp. And then the `signIn()` action is separate to allow a `FunctionCall` access on the accounts... This feels like a lot of steps which could be combined into one? (`connect()` could optionally pass `FunctionCall` parameters if the dapp wants method access on the selected accounts)

I also feel it's a little strange connecting with multiple accounts, seems like it might be confusing UX for users and requires more management on the dapp side to make sure account selection is well implemented for user actions. Whenever I've connected to a dapp- its usually with the expectation of a single account.

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+
+type Unsubscribe = () => void;
+
+interface Wallet {
+  id: string;
+  connected: boolean;
+  network: Network;
+  accounts: Array<Account>;
+  
+  connect(): Promise<Array<Account>>;
```
lostpebble, line 101: Should there not be some way for the dapp to indicate which network they are using? Right now, a `testnet` dapp might open the wallet with `connect()` and the user might select accounts on the wrong network.

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: Yes - `connect()` is separate from `signIn()` for a few reasons.

**From the dApp side:**
`connect()` means 'wallet, please tell me what account(s) I can request signing requests for'
`signIn()` means 'wallet, please add this access key to one of the accounts that was exposed when I `connect`'d to the wallet, so that I (the dapp) can submit transactions using that key instead of having to request explicit signing via the wallet even for simple and 'safe' transactions.

**From the wallet side:**
`connect()` means 'dApp, here are the accounts that the user would like to be able to use to interact with you'
`signIn()` means 'dApp, I have added an access key to the requested account so that you can submit limited transactions directly on behalf of the user without needing explicit Wallet interactions for each transaction'

Because `connect()` does not involve a blockchain transaction and is _only_ an interaction between the dApp and the wallet, its failure modes are specific to wallet interactions and will never include failures such as 'transaction failed' etc.

Because `signIn()` involves adding an access key to the requested account, its failure mode is specific to those modes encountered when a blockchain transaction fails for whatever reason.

The separation between `connect()` and `signIn()` becomes more clear when we look forward to off-chain account verification.  

For example, lets say you want to have someone 'log in' to your off-chain forum site by using their NEAR account as their identity.  

Currently, you would need to have the user `signIn()` so that you could verify that they own a particular account by looking on-chain and verifying that they added the access key you requested to the account.  Although this approach is secure (we know that someone owns a particular account if they can add access keys to it), it is sub-optimal in cases where the requesting app never intends to submit **any** transactions using that limited access key; essentially we are having the user mutate their on-chain account to prove they own it, but in a way that costs them on-chain storage and isn't actually useful after they have added the key, thus proving they own the account.

Note that this can be the case even for dApps that do interact with the blockchain! Some dApps require that all transactions they submit must be signed using a Full Access Key, so the 'sign in' process that creates a limited access key is superfluous to them.

So the idea is that while you will always `connect()` a dApp to a wallet instance, depending on the dApp type, it may request a `signIn()` if the dApp needs/wants a limited access key in its local scope, but alternatively it may request something like e.g. `verifyAccountOwner()` instead, if the dApp only needs to be absolutely sure that the requesting user controls a particular on-chain account.  Note that the failure mode of this new `verifyAccountOwner()` method will be different than either `connect()` or `signIn()` because it will not involve submitting an on-chain transaction; by keeping these behaviours separate, we allow for more flexible usage patterns and simplify each of the APIs involved and their possible failure modes.

lostpebble: > **From the dApp side:**
> connect() means 'wallet, please tell me what account(s) I can request signing requests for'

> **From the wallet side:**
> connect() means 'dApp, here are the accounts that the user would like to be able to use to interact with you'

I just feel like this is going to be a confusing UX for the user, as well as for wallets- now we have to provide an interface for users to select multiple accounts on a `connect()` request ? I don't think I've ever "logged in" to a website with multiple accounts before and I don't really see the benefit ? Especially since we can easily swap and change between accounts by just making a new connect request to the wallet (potentially there could be a `swapAccount()` method or something of the like, which opens the wallet to select a new account, even if we are currently "logged in" with an account). What I was trying to say is that I feel this is going to cause more complexity for both dapps and wallets (and potentially confuse users). More direct action paths are always better in my view.

Also dapps are now going to have to know for sure which one of the multiple selected wallets the user wanted to complete an action with. Does this mean its expected of all dapps to provide a kind of account dropdown list (or similar UI) for "currently active" account? You wouldn't want to do a forum action with the wrong account by mistake... now that we have "connected" with multiple accounts we require additional layers to prevent such mistakes.

If you are hard set on this interface, then I would request that you allow Dapps to provide an extra parameter to `connect()` - something like `{ occupancy?: "multple" | "single" | number }` - that would allow those Dapps which only expect to work with single accounts the option to build their UI around that.

Lastly, I do hope that the `connect()` method is going to provide even more context to the Wallet. Especially `network`, as I mentioned in another comment. Connecting with a wallet in an incorrect network is only going to get caught later on the Dapp side and could be a surprising error for users. Better to have these things determined from the get go.

specs/Standards/Wallets/InjectedWallets.md
esaminu: Perhaps the `Account` instances should have that information:

```typescript
interface Account {
  accountId: string;
  publicKey: utils.PublicKey;
  networkId: string;
}
```
I don't think we should implicitly rely on `accountId`s having that information due to implicit accounts and TLAs

esaminu: This would not be necessary actually since the `Wallet` interface has the `network` property and the dapp would only call `connect` on wallets with a `networkId` relevant to it.

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+  console.log("Network Changed", network);
+});
+```
+
+## Drawbacks
+
+There are cases when a dApp wants to sign a transaction with a signer that's yet to be determined (e.g. "Donate" button). Using the `Transaction` instance from `near-api-js` sadly limits this possibility as it requires the details of a `signerId` and `publicKey` upfront. Below are suggested solutions:
+
+- A workaround that maintains the `Transaction` instance might involve creating transactions with empty an `signerId` and `publicKey` to signal to wallets during validation that a prompt is required to select an account (similar to `connect`) before reconstructing the transaction.
+- Consider an intermediate abstraction between the dApp and wallet to defer some parameters for transactions (or at least make them optional). This gives wallets an opportunity to prompt for missing data and in some cases reduce complexity for dApps as the `nonce` and `blockHash` can be handled internally.
```
esaminu, line 369: I agree with this. Could we use something similar to the [`SignAndSendTransactionOptions`](https://github.com/near/near-api-js/blob/master/packages/near-api-js/src/account.ts#L66) type used in [near-api-js](https://github.com/near/near-api-js) ?  This is currently used in `Account` and [`ConnectedWalletAccount`](https://github.com/near/near-api-js/blob/master/packages/near-api-js/src/wallet-account.ts#L298). We could modify it slightly to include an optional `signerId` property:

```typescript
interface TransactionOptions {
    receiverId: string;
    actions: Array<transactions.Action>;
    signerId?: string;
}

interface SignTransactionParams {
  transaction: TransactionOptions;
}

interface SignTransactionsParams {
  transactions: Array<TransactionOptions>;
}

signTransaction(params: SignTransactionParams): Promise<transactions.SignedTransaction>;
signTransactions(params: SignTransactionsParams): Promise<Array<transactions.SignedTransaction>>;
```

This has the added benefit of moving [the work of constructing a transaction](https://docs.near.org/integrator/create-transactions#transaction-requirements) to the wallet side including having to fetch and increment the key nonce and decoding the `blockHash` using `borsh`.

Also, this way the dapp does not have to call `connect` if it just wants a transaction signed by an arbitrary `accountId`.

esaminu:
Would signing arbitrary messages be in scope for this NEP? I could see use cases where dapps need arbitrary messages signed with a FAK which would be stored in the wallet instead of the LAKs in the dapp.

specs/Standards/Wallets/InjectedWallets.md
lostpebble: I don't really understand? For instance, our wallet allows both testnet and mainnet connections- the same goes for most other wallets as well I think?

Where is the network set by the dapp, basically? Since they would need to indicate that somewhere.

Right now this Wallet Interface is the entire wallet object set on the window object... Are we expected to set multiple, one for each network? That would not be a great solution imo.

ori-near:
Discussing and making decisions on NEPs is crucial for evolving the NEAR platform. The NEP process introduces Working Groups as a way to operationalize the review of NEPs that builds community support. I propose that we use this NEP as the first one to help bootstrap the Wallet Standards Working Group.

@cameron-NEAR, @MaximusHaximus, @esaminu – Could you please fully read this NEP and comment in the thread with any feedback/concerns by September 30, and specifically if you would approve this NEP? If not, please specify reasoning and what it would take for you to approve it. I can schedule the first working group meeting after your review in early October. Thank you!

specs/Standards/Wallets/InjectedWallets.md
esaminu: I think the `Wallet` instance is a lot simpler and more extensible when it targets one `nodeUrl` and therefore `networkId`. It's analogous to a provider in this way.

There is no functionality specified here that interacts with the node but if it's added by the wallet or specified in future we'd have to keep solving the problem of which node we are talking to. Wallets can set multiple `Wallet` instances if they support multiple networks at the same time but generally injected wallets support 1 network at a time and will remove one of the `Wallet` instances when the user toggles between them.

lostpebble: > but generally injected wallets support 1 network at a time and will remove one of the Wallet instances when the user toggles between them.

How would this work? Where is this "toggle" happening?

This is what I'm trying to get at- we need to have some form of communication from the dapp side as to which wallet instance is required (i.e. which network the dapp is targeting).

> I think the Wallet instance is a lot simpler and more extensible when it targets one nodeUrl and therefore networkId. It's analogous to a provider in this way.

I can agree, that once the network we are using for interaction is established- having the wallet act as a single network "provider" could be useful and extensible. But first we need a way to communicate the required network, where is that done?

esaminu: > How would this work? Where is this "toggle" happening?

This happens on the wallet UI, e.g. metamask has a dropdown on top right

> we need a way to communicate the required network, where is that done?

The dapp would iterate through the wallets in window until it finds a wallet that matches the `networkId` it's looking to interact with

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+
+Standard interface for injected wallets.
+
+## Motivation
+
+dApps are finding it increasingly difficult to support the ever expanding choice of wallets due to their wildly different implementations. While projects such as [Wallet Selector](https://github.com/near/wallet-selector) attempt to mask this problem, it's clear the ecosystem requires a standard that will not only benefit dApps but make it easier for established wallets to support NEAR.
+
+## Rationale and alternatives
+
+At its most basic, a wallet contains key pairs required to sign messages. This standard aims to define an API (based on our learning from [Wallet Selector](https://github.com/near/wallet-selector)) that achieves this requirement through a number of methods exposed on the `window` object.
```
esaminu, line 13: @lewis-sqa [Ethereum's JSON-RPC Methods](https://docs.metamask.io/guide/rpc-api.html#ethereum-json-rpc-methods) are generalizable to web wallets as well e.g. [portis wallet](https://docs.portis.io/#/tips-and-tricks?id=custom-eth_sendtransaction-parameters), do we have to make this NEP extension wallet specific by specifying that this API should be exposed on the `window` object and where?

esaminu:
I notice that wallets are not interacting directly with the `nodeUrl`  e.g. to send transactions. Is the dapp is expected to submit the signed transaction? Should we add other methods that have the wallet submit the transactions to rpc after signing? This is easier for dapps since that way we don't have 2 redundant `nodeUrl`s and the dapp can delegate all rpc interactions to the wallet if it prefers.

esaminu:
I notice that wallets are not interacting with the `nodeUrl`  to send transactions. Is the dapp is expected to submit the signed transaction? Should we add other methods that have the wallet submit the transactions to rpc after signing? This is easier for dapps since that way we don't have 2 redundant `nodeUrl`s and the dapp can delegate all rpc interactions to the wallet if it prefers.

specs/Standards/Wallets/InjectedWallets.md
lostpebble: > This happens on the wallet UI, e.g. metamask has a dropdown on top right

Are we not talking about a stage that is long before that? We are on a dapp site- the user may not have interacted with their wallet for a while now, or maybe recently had interacted with a testnet site. But now a mainnet dapp is going to ignore their wallet completely because it just wasn't recently "set" to the right network.

I feel like this is going to be a pain going forward and create more UX issues between dapps and users. And the way to avoid this would just be a simple parameter which the dapp can send to the injected wallet object. Again, most wallets are multi-tenancy, they can handle both cases with ease. Its really not that hard to set an internal parameter for the network and create a specific "provider" of wallet functionality.

Another issue- what if the user was signed into a testnet dapp, but then opened their wallet to check out a mainnet account for some reason, send some funds etc. And then close the wallet again, with the full intention of still interacting with the current dapp where they left off. With the current spec, this is going to create a bit of an ambiguous situation in the dapp now. These are the kinds of headaches that could easily be solved by just being more deterministic, as much as possible, from both sides.

PinkiNice: I agree with idea that API should be as clear as possible. Favoring being deterministic instead of "convenience" feels like the right choice for me.

So having `window.near.wallet.connect({ networkId: 'mainnet' })` instead of `window.near.wallet.connect()`
And to be clear we could also introduce a `window.near.wallet.supportedNetworks()`

There are multiple ways how this API may look and even a solution which still keeps `Wallet` tied to a single network may be used:
`const testnetWallet = window.near.wallet.getWallet('testnet')`

MaximusHaximus:
Hello folks,
Lewis is no longer working on this task, so I am taking over the NEP as author and am creating a new PR based on a branch that I control; I'll be pushing updates to that PR in a few minutes to address some of the questions/concerns mentioned in these discussions.  Thanks so much for the input and engagement, this stuff is important and we want to make sure we're solving the right problems :). Please come along to the new PR at https://github.com/near/NEPs/pull/408 :). I'll ping there as soon as new commits are pushed.

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: :+1: good point -- I've added some text to warn wallet authors against overwriting the object

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: :+1: Making this adjustment as suggested - thanks for the discussion folks!

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: This NEP outlines APIs that don't have direct parallels with a web wallet; event emitters and methods that return promises that resolve to the results of direct calls to an injected wallet can (for example) throw errors directly to the dApp from the wallet, whereas a web wallet running in another tab cannot.  I'm all for making this a more generally applicable API if that makes sense, but I don't think we need to block this landing to accomplish that in the future if desirable.

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: It does make things more complex to allow users to select more than one account to be shared with the dApp from a wallet -- but this is actually a very often requested feature that we hope to support as a first-class concept in the API.  Note that although `connect()` can return multiple accounts, _only one of those accounts generally would be considered 'active' at once in the dApp context_ -- and generally that account would be set after either a `verifyOwner()` or `signIn()` call has been completed for the account that the user wants to be their _active_ account.

Although each dApp author could write their own code to handle interacting with wallets and prompting the user to select the 'active' account, we don't think it's a good use of their time to do so (this is the equivalent of boiler-plate that distracts the dApp developer from solving the real problems they are trying to solve).  This aligns with the philosophy that integrating with many different wallets and providing the functionality to select which of those wallets to use should also be something that we provide tools for the dApp author to use. We will be adding this functionality to the wallet selector package / modal UIs, so that the dApp developer can use the modal UI it provides to allow the user to see the list of accounts, and to select which account should be the 'active' account for signing in, verifying owner, or signing requests.

If a dApp author wants to write their own account switcher/selector UI, they could do so by piggy-backing on the core package which exposes events, accounts, and a concept of 'active account' for the dApp developer -- or they could just use the UI components that we publish from the wallet-selector and be up-and-running immediately.

Note that it currently stands, we do not dictate that a wallet _must_ return multiple accounts from a `connect()` call, _even if there are multiple loaded into that wallet_.  The only requirement is that it returns an array with _at least_ 1 account in it -- which allows wallet authors to make the UI for the `connect()` method simple (select one) or more complex (select multiple) as they see fit for their wallet design, while ensuring that our API is also compatible with more than one being returned.

@lostpebble Does this address your concerns here, or do you think that adding `occupancy` as an argument to the `connect()` method is still a requirement for this to be reasonable, keeping in mind that if we do add that option, wallet providers implementation and UI will become more complex.

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: @lostpebble @PinkiNice Thanks for the discussion, we appreciate the insights.  The idea currently discussed here was to drive the network and accounts list _from the wallet to the dApp_, using events (`networkChanged` and `accountsChanged`).  It does create an issue if the dApp does not change to reflect the new network or accounts list where the wallet UI chosen network may not match the dApp's chosen network.

I've updated the spec to eliminate the event emitted for `networkChanged` from the wallet to the dApp context, and instead inverted the control so that the dApp requests the wallet for a particular network ID during `connect()`, and there is an implicit guarantee that the wallet instance on the window that we reference from then on will never reference any other network.

Thanks for the discussion points!  If I have missed the mark on this at all, please follow up on #408 :)

specs/Standards/Wallets/InjectedWallets.md
lostpebble: Hi @MaximusHaximus - thanks for taking Network into consideration with the `connect()` method.

> @lostpebble Does this address your concerns here, or do you think that adding occupancy as an argument to the connect() method is still a requirement for this to be reasonable, keeping in mind that if we do add that option, wallet providers implementation and UI will become more complex.

I think its okay to leave it out- as long as its always okay for a user to only select a single account to connect with. If you think certain dApps might require the user to connect with multiple (or a certain amount) of accounts, for whatever reason, then having an `occupancy` option might be a good idea so the wallet can show the relevant UI. But I don't think should ever be the case?

This "multi-account" concept does complicate the Wallet UI a bit though. I guess in our case, when a Dapp calls the `connect({ network: "testnet" })` method, we'll show the user an interface that allows them to do one of the following:

* Select a single testnet account
* Select multiple specific testnet accounts
* Connect with all known wallet testnet accounts

We'll have to think of ways of streamlining such an interface.

And after that, I would suppose that future calls to `verifyOwner()` do not require any further prompts from the user (since they've already allowed the connection). But `signIn()` would run the specific sign-in flow for the account the dApp has chosen to use.

PR #366, opened 2022-06-23
NEP-366: Meta transactions

ilblackdragon:
## Summary

In-protocol meta transactions allowing for third-party account to initiate and pay transaction fees on behalf of the account.

## Motivation

NEAR has been designed with simplify of onboarding in mind. One of the large hurdles right now is that after creating an implicit or even named account user doesn't have NEAR to pay gas fees to interact with apps.

For example, apps that pay user for doing work (like NEARCrowd or Sweatcoin) or free-to-play games.

[Aurora Plus](https://aurora.plus) has shown viability of the relayers that can offer some number of free transactions and a subscription model. Shifting the complexity of dealing with fees to the infrastructure from the user space.

## Rationale and alternatives

Proposed here design provides the easiest for users and developers way to onboard and pay for user transactions.
There is no requirement on the user account, including user account may not even exist on chain and implicit account can be used.

An alterantive is a proxy contracts deployed on the user account. 
This design has severe limitations as it requires user to deploy such contract and additional costs for storage.

## Specification

Next changes to protocol are required:
 - New Action kind: `DelegateAction(receiver_id, Action, signer_pk, signature)`
 - On conversion to Receipt, such action is sent as `DelegateAction` to the `receiver_id`.
 - On processing Receipt, it verifies the `signature` over the given account and `signer_pk`. Unwraps into `receiver_id: AccountId`, `action: Action` inside into new Receipt with given `receiver_id` and `action`. This means that `predeccessor_id` of such action has been overriden.

See the [DelegateAction specification](specs/RuntimeSepc/Actions.md#DelegateAction) for details.

## Security Implications

Delegate actions do not override `signer_pk`, leaving that to the original signer that initiated transaction (eg relayer in meta transaction case). Although it is possible to override `signer_pk` in the context with one from `DelegateAction`, there is no clear value to do that.

See ***Validation*** section in [DelegateAction specification](specs/RuntimeSepc/Actions.md#DelegateAction) for security considerations around what user signs and validation of actions with different permissions.

## Drawbacks

Increases complexity of the NEAR's transactional model.

Meta transactions take an extra block to execute, as they first need to be included by the originating account, then routed to the delegate account and only after that to the real destination.

Delegate actions are not programmable as they require having signatures. Original proposal contained a new `AccessKey` kind that would support programmable delegated actions. On the other hand, that would be limiting without progamability of smart contracts, hence that idea evolved into [Account Extensions](https://github.com/nearprotocol/neps/pull/0000).

## Future possibilities

[...]

PR #342, opened 2022-03-27
feat(runtime): Add delegate keys

ilblackdragon:
## Summary

Delegate keys allow to assign permissions to another account instead of a key. 
This allows to delegate ownership of an account by giving set of permissions that are usually restricted only to FullAccessKey.

## Motivation

Issue that became evident trying to develop upgradability patterns for contracts is complexity of going from full access key on the contract to a contract managed by multisig or a DAO.

Right now it is required to implement a complex logic of upgrading contract and managing assets/control inside the contract in the first version to allow “owner” account to call these methods.

## Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

## Specification

See updates to AccessKey.md and Actions.md

## Reference Implementation

TBD

## Drawbacks

This adds complexity to the account model.

## Copyright
[copyright]: #copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

PR #371, opened 2022-07-13
Exposing Function Call Access Key Info At Runtime

BenKurrek:
This PR is based on the gov post found [here](https://gov.near.org/t/expose-gas-price-and-contract-access-key-info-within-runtime/24788) that talks about introducing a way to query for function call access key information for the given contract at runtime.

### DIFF
--- a/neps/nep-0371.md
+++ b/neps/nep-0371.md
@@ -1,6 +1,6 @@
 ---
 NEP: 371
-Title: Runtime Access Key Info
+Title: Runtime Function Call Access Key Info
 Author: Ben Kurrek <ben.kurrek@near.org>
 DiscussionsTo: https://gov.near.org/t/expose-gas-price-and-contract-access-key-info-within-runtime/24788
 Status: Draft
@@ -11,56 +11,43 @@ Created: 13-Jul-2022
 
 ## Summary
 
-There is currently no way to query for information about access keys stored on your contract at runtime.
+There is currently no way to query for information about function call access keys stored on your contract at runtime.
 
 ## Motivation
 
-As contracts start to use access keys in more creative ways, developers will want to know information such as the access key's allowance, nonce, receiver and more. Since contracts are often built for scalability, Gas and storage costs are often delegated to the end user. If you charge the user for an access key and its allowance, when the key is deleted or used up, that allowance should be refunded to the purchaser. Currently, there is no way to query for the leftover allowance before a key is deleted. This information can tell you how much Gas was spent, and allows for a slew of different scenarios to arise.
+As contracts start to use access keys in more creative ways, developers will want to know information such as the access key's allowance, receiver and more. Since contracts are often built for scalability, Gas and storage costs are often delegated to the end user. If you charge the user for an access key and its allowance, when the key is deleted or used up, that allowance should be refunded to the purchaser. Currently, there is no way to query for the leftover allowance before a key is deleted. This information can tell you how much Gas was spent, and allows for a slew of different scenarios to arise.
 
 ## Rationale and alternatives
 
-Without creating a promise, we can only query for information stored on the same shard. This is why in this proposal, the access key information at runtime will only come from keys stored on the current contract. We had also investigated the possibility of introducing pagination for all keys stored on the contract but this is infeasible at the moment due to the significant redesign requirement of the VM logic.
+Without creating a promise, we can only query for information stored on the same shard. This is why in this proposal, the access key information at runtime will only come from keys stored on the current contract. We had also investigated the possibility of introducing pagination for all keys stored on the contract but this is infeasible at the moment due to the significant redesign requirement of the VM logic. In addition, we have thought about introducing a way to query for full access key information but this didn't seem useful and we can always add it in another NEP.
 
 ## Specification
 
-In the runtime, key information will be queryable by passing in the desired public key. The key information will be of the following form.
+In the runtime, function call access key information will be queryable via a function by passing in the desired public key. The key information will be of the following form.
 
 ```js
-enum KeyPermission {
-    FullAccess,
-    FunctionCall(FunctionCall),
-}
-
-type KeyInformation = {
+type FunctionCallKeyInfo = {
     public_key: string,
-    access_key: {
-        nonce: number,
-        permission: KeyPermission
-    }
-}
-
-type FunctionCall = {
+    nonce: number,
     receiver_id: string,
     allowance: string,
     method_names: Array<string>,
 }
 ```
 
-The key information will expose the public key and access key data. If the key is a full access key, the permission will simply be a string `"FullAccess"`. If the key is a function call access key, the permission will be of type `FunctionCall`.
-
-This information should be queryable via a new function called `key_info_by_public_key_for_current_account` which takes a public key as a parameter.
+This information should be queryable via a new function called `function_call_key_info_for_current_contract` which takes a public key as a parameter.
 
 ```ts
-function key_info_by_public_key_for_current_account(
+function function_call_key_info_for_current_contract(
   public_key: &PublicKey,
-): KeyInformation | null
+): FunctionCallKeyInfo | null
 ```
 
-This function should be exposed in the environment and callable using `env::key_info_by_public_key_for_current_account();`. An example of this can be seen below.
+This function should be exposed in the environment and callable using `env::function_call_key_info_for_current_contract();`. An example of this can be seen below.
 
 ```rs
 pub fn check_key_exists(&self, pk: PublicKey) {
-    let key_info: KeyInformation = env::key_info_by_public_key_for_current_account(&pk);
+    let key_info: FunctionCallKeyInfo = env::function_call_key_info_for_current_contract(&pk);
 
     if let Some(info) = key_info {
         near_sdk::log!("key info found");
@@ -74,20 +61,21 @@ An example of returning the allowance of the key can be seen below.
 
 ```rs
 pub fn check_key_allowance(&self, pk: PublicKey) {
-    let key_info: KeyInformation = env::key_info_by_public_key_for_current_account(&pk).unwrap();
+    let key_info: FunctionCallKeyInfo = env::function_call_key_info_for_current_contract(&pk).unwrap();
 
-    if let KeyPermission::FunctionCall(fc) = key_info.access_key.permission {
-        let allowance = fc.allowance;
-        near_sdk::log!("key allowance: {}", allowance);
-    } else {
-        near_sdk::log!("key has no allowance as it's a full access key");
-    }
+    let allowance = fc.allowance;
+    near_sdk::log!("key allowance: {}", allowance);
 }
 ```
 
 ## Future possibilities
 
-Some of the future possibilities for the standard could be to return a vector of `KeyInformation` by paginating through the keys on the contract. In addition, we could add a `key_total_supply` function that returns the number of access keys stored on the contract. One could also expand and introduce the ability to query for access key info from other contracts by using a promise.
+There were several future possibilities we had discussed and decided it would be best to start small and expand in the future if there was public interest. One of which was the ability to query for full access key information as well. In addition, adding the ability to check the type of key as either full access or function call access would allow for an interesting slew of use-cases.
+
+We had also discussed the idea of paginating through keys on the contract and returning a vector of `FunctionCallKeyInfo`, or full access key info. In addition, we could add a `key_total_supply` function that returns the number of access keys stored on the contract. We had also thought of allowing 
+
+
+One could also expand and introduce the ability to query for access key info from other contracts by using a promise.
 
 ## Copyright
 [copyright]: #copyright
--- a/neps/nep-0371.md
+++ b/neps/nep-0371.md
@@ -27,15 +27,24 @@ The contract runtime provides one host function called `access_key_info` which i
 
 ```rust
 fn access_key_info (
-    public_key_len: u64,                // in:  number of bytes of the public key to query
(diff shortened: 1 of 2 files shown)


neps/nep-0371.md
```diff
@@ -0,0 +1,149 @@
...
+Created: 13-Jul-2022
+---
+
+## Summary
+
+There is currently no way to query for information about access keys stored on a contract at runtime. Adding the host function `access_key_info` to the runtime specification will allow to query this for the current account an easy way. This fully covers the most important use cases that are currently known for querying access keys at runtime.
+
+## Motivation
+
+As contracts start to use access keys in more creative ways, developers will want to know information such as the access key's allowance, receiver and more. Since contracts are often built for scalability, Gas and storage costs are often delegated to the end user. If you charge the user for an access key and its allowance, when the key is deleted or used up, that allowance should be refunded to the purchaser. Currently, there is no way to query for the leftover allowance before a key is deleted. This information can tell you how much Gas was spent, and allows for a slew of different scenarios to arise.
```
akhi3030, line 18: > Since contracts are often built for scalability, Gas and storage costs are often delegated to the end user.

I am not sure I understand what scalability has to do with gas and storage costs being delegated to the end user.

neps/nep-0371.md
```diff
@@ -0,0 +1,149 @@
...
+## Rationale and alternatives
+
+Without creating a promise, we can only query for information stored on the same shard. This is why in this proposal, the access key information at runtime will only come from keys stored on the current contract. We had also investigated the possibility of introducing pagination for all keys stored on the contract but this is infeasible at the moment due to the significant redesign requirement of the VM logic.
+
+## Runtime specification
+
+The contract runtime provides one host function called `access_key_info` which is callable in regular calls and views.
+
+```rust
+fn access_key_info (
```
akhi3030, line 29: So this API can only be used to access info about a known key?  What if a contract wants to know the actual set of keys associated with it?  Maybe there is already an API for that or is that not an interesting use case?

neps/nep-0371.md
```diff
@@ -0,0 +1,149 @@
...
+- Returns 1 if the given public key is a full access key to the current account. The memory location at `nonce_ptr` contains the current nonce value of the access key. The memory location at `allowance_ptr` is not changed by the host. The register content of `account_id_register_id` and `method_names_register_id` are not changed, either.
+- Returns 2 if the given public key is a function call access key to the current account. The memory location at `nonce_ptr` contains the current nonce value of the access key. The memory location at `allowance_ptr` contains a `u128` value that is the remaining allowance in yoctoNEAR. The register `account_id_register_id` contains the account name of the access key holder. The register `method_names_register_id` contains a comma separated list of all method names that the access key is valid for.
+
+### Panics
+
+- `InvalidPublicKey`: Calling `access_key_info` with an invalid public key.
```
akhi3030, line 48: Just for my understanding, how would this error be raised?  I would've expected that the implementation will iterate over the list of keys registered against the contract and compare them against the one that the contract passed in.  This error seems to suggest that we need to do more processing on the key that is passed in.

neps/nep-0371.md
```diff
@@ -0,0 +1,149 @@
...
+The contract runtime provides one host function called `access_key_info` which is callable in regular calls and views.
+
+```rust
+fn access_key_info (
+    public_key_len: u64,                // in:  number of bytes of the public key to query
+    public_key_ptr: u64,                // in:  pointer to the public key to query
+    nonce_ptr: u64                      // out: pointer to where a u64 can be written by the host
+    allowance_ptr: u64,                 // out: pointer to where a u128 can be written by the host
+    account_id_register_id: u64,        // out: register to write the access key holder account id
+    method_names_register_id: u64,      // out: register to write the method names as a comma separated list
```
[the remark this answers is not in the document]
jakmeier: Register API is specified here: https://nomicon.io/RuntimeSpec/Components/BindingsSpec/RegistersAPI

It looks like it is not documented / specified but the upper limit of a register is set by a runtime parameter, `max_register_size: 104_857_600`. Plus, the sum of all register sizes is limited by `registers_memory_limit: 1_073_741_824`. But both of these are currently only theoretical limits, as writing 1GiB into registers would cost around 2925 Tgas and writing 100MiB costs around 285 Tgas. (But I should still include it as a failure mode in the specs, thanks for pointing it out!)

The upper limit on the total length of all method names is limited by the runtime parameter `max_number_bytes_method_names: 2_000`, which is enforced when creating an access key. Those 2000 bytes can be split up into method names and separators arbitrarily. (Each valid method name is also limited to 256 bytes but that's not really relevant here, since no validity checks are made on the method names of an access key.)

neps/nep-0371.md
jakmeier: It could interesting to paginate keys, we just haven't had anyone coming up with an actual use case, yet. :)

And because the API for that would be significantly trickier to implement,  we opted it out, for now.

I expect most calls to either look up the current access key (`signer_account_pk()` already exists for that) or otherwise take an argument from outside that specifies which access keys to check. Listing all keys through rpc REST API has been supported for a long time, so the signer side should always be able to provide the list if necessary.

neps/nep-0371.md
jakmeier: This is essentially a deserialization error for a `PublicKey` struct. Simplest example would be an invalid length.

Regarding implementation, we will not iterate the list of keys. Each access key is stored in the trie at a specific trie key. The trie key is defined as `col::ACCESS_KEY ++ account_id ++ col::ACCESS_KEY ++ public_key`. So we will directly look up this node in the trie.

neps/nep-0371.md
```diff
@@ -0,0 +1,170 @@
...
+    public_key_len: u64,           
+    /// in:  pointer OR register ID to the public key to query
+    public_key_ptr: u64,           
+    /// out: pointer to where a u64 can be written by the host
+    nonce_ptr: u64                 
+    /// out: pointer to where a u128 can be written by the host
+    allowance_ptr: u64,            
+    /// out: register to write the access key holder account id
+    account_id_register_id: u64,   
+    /// out: register to write the method names as a comma separated list
```
matklad, line 40: >as a comma separated list

This is problematic: `,` is a valid character in a method name, so this API creates possibility for ambiguity.

neps/nep-0371.md
matklad: >And because the API for that would be significantly trickier to implement

To expand on this, we don't *really* have a way to iterate keys in our runtime.

neps/nep-0371.md
BenKurrek: If you're building for scalability, you generally want your users to front these costs as to not drain your contract.

matklad:
The issue with method_names seems significant to me. It seems like there's no way out of serializing variable number of variable-length keys. My gut feeling is that we should probably bite the bullet and just return borsh of the whole access key.

jakmeier:
> The issue with method_names seems significant to me. It seems like there's no way out of serializing variable number of variable-length keys. My gut feeling is that we should probably bite the bullet and just return borsh of the whole access key.

@matklad I guess you are right. Do you think we can sneak it in as part of this NEP, making the precedence for borsh being used in the runtime API? Or should this go through a separate NEP with full discussion and getting everyone with a stake involved? I was hoping to avoid too much overhead here but it feels like adding borsh to the spec crosses a line that we should not rush over too quickly.

bowenwang1996 (commented):
Thanks for the proposal. I would like to understand the following:
* The motivation talks about making developer experience better. Could you provide a specific example of why this host function is needed?
* What happens if we change `AccessKey` in the future? (for example we implement https://github.com/near/NEPs/pull/342) How does that affect this host function?

BenKurrek:
> Thanks for the proposal. I would like to understand the following:
> 
> * The motivation talks about making developer experience better. Could you provide a specific example of why this host function is needed?
> * What happens if we change `AccessKey` in the future? (for example we implement #342) How does that affect this host function?

@bowenwang1996 I can speak to the first point. As stated in the [governance](https://gov.near.org/t/expose-gas-price-and-contract-access-key-info-within-runtime/24788) post:

For a real world example of why this would help, i’ll explain a use case. Imagine you have a contract that allows users to “purchase” function call access keys that can do different things. The user must pay for the access key allowance when the key is created. An example of this is the linkdrop contract.

When the key is used up / deleted, dApps might want to refund users for any allowance that was unspent. Currently, there is no way to do this. In addition, when creating the key, often times, the dev will know how much GAS will be attached to the function calls and they might want to dynamically calculate the pessimistic allowance that the key must have and to do this, they need the current GAS price.

In addition, @mattlockyer had shared his opinions in the post as well: 

An abstract but fundamental use case would be some form of meta transaction.

I want Alice to fund an access key, that is added to a contract she doesn’t control, such that Bob can call a method 10 times.
On the 10th call, the key is deleted and any unused allowance from that key is returned to Alice.

Currently, there is no way to check the unused allowance of the key from inside the contract runtime, AND there is no way to check the current gas price, so crude approximations have to be used that often overestimate how much gas was actually spent.

Panics and reversions further complicate this scenario, since gas is used, but no state is changed. This forces the developer to never panic on an assertion, and instead write a crude approximate of used gas to state and return false, before exiting the method.

If the protocol allowed us to see the allowance, method_names and receiver_id of access keys, there could be more use cases opened up.

bowenwang1996:
> I want Alice to fund an access key, that is added to a contract she doesn’t control, such that Bob can call a method 10 times.
> On the 10th call, the key is deleted and any unused allowance from that key is returned to Alice.

I see. So basically you want Alice to pay for Bob's transactions. There is a [separate proposal](https://github.com/near/NEPs/pull/366) on meta transactions, which could provide in my opinion a better solution to this problem. Are there any other examples or use cases for this feature?

jakmeier:
> * What happens if we change `AccessKey` in the future? (for example we implement #342) How does that affect this host function?

@bowenwang1996 Valid question, I see two directions to deal with such cases.

1. Add different functions for each permission type. E.g. `full_access_key_info()`, `function_access_key_info()`, `delegation_access_key_info()`
2. Make `access_key_info()` more flexible.

I am torn between the two. The first option may lead to inconsistent API, as we add more and more functions over time. The second option tries predict the future, which usually doesn't work.

To make it more tangible, the second option could roughly look like this: (Also incorporating @akhi3030 suggestion to paginate methods.)
- `access_key_info` queries by PK, maybe rename it to `access_key_info_by_pk()`. It returns 0: key doesn't exist, 1: key is full access key, 2: key is function access key. Other return values are reserved for future use. All output parameters are removed.
- For #342, we could add  `access_key_info_by_account_id()`.
- `access_key_iter_next` is used for paginating over key information, returned to registers in a pre-defined order that depends on the key type.
    - For function access key, this could be [`nonce`, `public_key`, `allowance`, `method[0]`, `method[1]`, ..., ].
    - For a delegation access key, it could be [`nonce`, `account_id`, `action[0]`, `action[1]`, ..., ]

I *think* this would allow to cleanly add all kinds of new access keys. But who knows what the future brings.

BenKurrek:
Here is another user asking for this functionality this morning: 
https://stackoverflow.com/questions/73277923/near-sdk-rs-view-account-keys

ori-near:
Hey @BenKurrek – thank you for submitting this NEP. In order to move this NEP to the review stage, can you please add a reference implementation?

ori-near:
Hi @BenKurrek @jakmeier (or anyone else) – are you still interested in moving this NEP forward? If so, can you please submit a reference implementation?

BenKurrek:
I'm definitely interested in moving it forward but don't have time to put out a reference implementation at the moment. Perhaps I can revisit when I have some time.

mattlockyer:
@ori-near can you please provide some links to other reference implementations e.g. a couple of closed PRs to nearcore that we can look at?

I don't think this should be incredibly difficult.

We know everyone would like to keep the API surface small for packages like near-sys, however getting access key information about the account your contract is currently executing on is fairly critical for some use cases.

jakmeier:
@mattlockyer here is the complete reference implementation for another host function: https://github.com/near/nearcore/pull/7165 (some follow-up work was necessary but for a reference implementation this is more than needed)

Regarding this feature here, I agree it is fairly basic and probably should be there. The main questions/concerns for me are:
1. Is it actually making new use cases possible, or is it just making hacky solutions slightly less hacky? If it is the latter, we should consider making changes that solve the  use cases cleanly. (This is why I deprioritized working on this to focus on things like supporting meta transactions instead.)
2. What would happen to `fn access_key_info()` if we make drastic changes to how keys work in the future? There are a bunch of ideas flying around here and we would add complexity to each of those ideas if we also need to consider how `fn access_key_info()` is affected.

But if you think this is a valuable addition to for smart contract devs and you can provide a reference implementation, then we can certainly move it forward. Me personally, I don't think I  have the capacity to implement it in the near future.

nearbuild:
@mattlockyer @BenKurrek  wanted to reping on this since progress on https://github.com/near/NEPs/pull/452 moves forward

walnut-the-cat:
Hello! Protocol NEP moderator here. it seems there hasn't been any activity on this NEP for quite some time. Is it still on going? Or shall we close NEP for now? @BenKurrek, @jakmeier

walnut-the-cat:
Another ping on status of this NEP. As more than two months have passed since the last activity, if we don't hear back from the author by December 4th, we will mark the NEP as retracted for now. Thanks-

walnut-the-cat:
I will assume this NEP is inactive and label as retracted for now. If anyone picks the NEP up again, please ping NEP moderators to make it active.

garikbesson:

PR #372, opened 2022-07-14
Gas price host function

BenKurrek:
Introduces a way to query for the current price of Gas at runtime from within smart contracts.

### DIFF
--- a/neps/nep-0372.md
+++ b/neps/nep-0372.md
@@ -1,5 +1,5 @@
 ---
-NEP: 9999
+NEP: 372
 Title: Runtime Gas Price
 Author: Ben Kurrek <ben.kurrek@near.org>
 DiscussionsTo: https://gov.near.org/t/expose-gas-price-and-contract-access-key-info-within-runtime/24788
--- a/neps/nep-0372.md
+++ b/neps/nep-0372.md
@@ -23,23 +23,37 @@ There doesn't seem to be many alternative solutions other than the one proposed.
 
 ## Specification
 
+Two new host functions are added to the runtime. One function is for the gas price at the current block height. The other is for the pessimistic gas price that has been used to purchase the attached gas to the currently executing receipt.
+
+Both function return a gas price measured in yoctoNEAR per unit of Gas.
+
 ```rust
+/// Gas price at the current block height.
 fn current_gas_price () -> u128
+
+/// Gas priced paid to purchase gas for the current receipt. The difference
+/// between this and the current gas price will be refunded for all the burnt
+/// gas, while unused gas will be refunded at the full pessimistic price.
+/// This function can be used to calculate how much balance is left in the
+/// account after refunds, or how much allowance will be left in an access key.
+fn pessimistic_receipt_gas_price () -> u128
 ```
 
-The function returns the effective gas price in yoctoNear / gas for the current block.
+Successful execution of both functions is guaranteed, as long as there is sufficient gas.
 
-Successful execution of this function is guaranteed, as long as there is sufficient gas.
-
-The gas cost for this call will be the base host function call`wasm_base` without anything extra. This is in-line with similar host functions like `prepaid_gas` that also query a single value that is readily available at the host.
+The gas cost for both these calls will be the base host function call `wasm_base` without anything extra. This is in-line with similar host functions like `prepaid_gas` that also query a single value that is readily available at the host.
 
 ## User perspective
 
-In the SDK's environment, a new function will be exposed which is queryable at runtime. This function will return the amount of yoctoNEAR it costs to purchase 1 unit of Gas.
+In the SDK's environment, new functions can be exposed which are queryable at runtime. This function will return the amount of yoctoNEAR it costs to purchase 1 unit of Gas.
 
 ```js
-/// Returns the amount of yoctoNEAR it costs to purchase 1 unit of Gas
+/// Returns the amount of yoctoNEAR it costs to purchase 1 unit of Gas.
 function current_gas_price(): Balance
+
+/// Returns the amount of yoctoNEAR/gas that has been paid, pessimistically,
+/// when creating the current receipt. This is useful for calculating refunds.
+function pessimistic_receipt_gas_price(): Balance
 ```
 
 The value returned may change between blocks due to an increase in the price. A common scenario to get the amount of yoctoNEAR based on the attached Gas can be seen below.
--- a/neps/nep-0372.md
+++ b/neps/nep-0372.md
@@ -25,23 +25,23 @@ There doesn't seem to be many alternative solutions other than the one proposed.
 
 Two new host functions are added to the runtime. One function is for the gas price at the current block height. The other is for the pessimistic gas price that has been used to purchase the attached gas to the currently executing receipt.
 
-Both function return a gas price measured in yoctoNEAR per unit of Gas.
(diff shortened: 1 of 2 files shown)


neps/nep-0372.md
```diff
@@ -0,0 +1,77 @@
...
+/// between this and the current gas price will be refunded for all the burnt
+/// gas, while unused gas will be refunded at the full pessimistic price.
+/// This function can be used to calculate how much balance is left in the
+/// account after refunds, or how much allowance will be left in an access key.
+fn pessimistic_receipt_gas_price (gas_price_ptr: u64);
+```
+
+The gas cost for both these calls will be the base host function call `wasm_base` + `wasm_write_memory_base` + 16 * `wasm_write_memory_byte`. This is in-line with similar host functions like `account_balance` that also query a single `u128` value that is readily available at the host.
+
+Both functions must fail with `MemoryAccessViolation` if [`gas_price_ptr`, `gas_price_ptr` + 16) is not completely within guest memory.
```
akhi3030, line 44: Should we specify if this can be called during views or only during regular calls.

jakmeier: That's a very good point. I think both functions should not be callable in views.
- `current_gas_price` would _maybe_ make sense but it would be a bit tricky to implement. Right now, view calls are oblivious of the actual as price, they just burn gas at a price of 0 NEAR/gas.
- `pessimistic_receipt_gas_price` in a view call just makes no sense, there is no receipt in a view call.

neps/nep-0372.md
```diff
@@ -0,0 +1,79 @@
...
+```rust
+/// Gas price at the current block height.
+fn current_gas_price (gas_price_ptr: u64);
+
+/// Gas priced paid to purchase gas for the current receipt. The difference
+/// between this and the current gas price will be refunded for all the burnt
+/// gas, while unused gas will be refunded at the full pessimistic price.
+/// This function can be used to calculate how much balance is left in the
+/// account after refunds, or how much allowance will be left in an access key.
+fn pessimistic_receipt_gas_price (gas_price_ptr: u64);
```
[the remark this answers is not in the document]
jakmeier: I like the suggestion, I will update the NEP to that if nobody has other opinions

neps/nep-0372.md
```diff
@@ -0,0 +1,79 @@
...
+Created: 14-Jul-2022
+---
+
+## Summary
+
+There is currently no way to query for the current Gas price at runtime from within smart contracts.
+
+## Motivation
+
+As the network becomes more popular, Gas prices may differ between blocks. By adding the ability to query for the price of Gas, it opens many possibilities for developers. Some use cases could include a contract that knows how much Gas is attached to the call or how much Gas has been used so far using the current `prepaid_gas` and `used_gas` functions but they want to convert that to a $NEAR value using the price of Gas.
```
bowenwang1996, line 18: Please mention explicit use cases. The text here does not provide sufficient justification for why we need to add this host function

bowenwang1996 (commented):
I would like to understand better the motivation behind exposing gas price and how the use cases would work if we decide to introduce some form of gas auction in the future. I would like to avoid introducing some host function that people start using in their smart contracts only to realize later that the semantics of the host function may change

jakmeier:
> I would like to understand better the motivation behind exposing gas price and how the use cases would work if we decide to introduce some form of gas auction in the future. I would like to avoid introducing some host function that people start using in their smart contracts only to realize later that the semantics of the host function may change

I hope someone else can write about actual practical use cases. But from a technical side, I see two new abilities this grants to users:
1. Decide on-chain whether it makes financial sense to do an action given the current price. (E.g. decide that a trade is only worth it if price < X)
2. Calculate how many tokens will be in the account (or allowance in an access key) after refunds of the current receipt.

With regard to changes in semantic, complete overhauls of the gas model could lead to deprecation of for example `current_gas_price`, I grant you that. But assuming new features are only additive to the current gas model, I thought about it and concluded that we are mostly fine, if we are careful in the specification. I suggest this:

- rename `current_gas_price` to `burn_gas_price` which returns the yoctoNEAR/gas rate at which gas is being burnt during this receipt execution
- `purchased_gas_price` returns the yoctoNEAR/gas rate that has been paid for the receipt earlier (unchanged)

Note that this changes nothing but the name of the first function under the current gas model.
But if, for example, we introduce auctions as a way to get priority at times of congestion, it could be that the base price at the current height of the block is not the same as the `burn_gas_price`.

With my suggestion, users could rely on the following semantics:
- `burn_gas_price` is the relevant gas price for financial decisions right now
- `purchased_gas_price` * `unused_gas` is the refund for unspent gas
- (`purchased_gas_price` -`burn_gas_price`) * `burnt_gas` is the refund for pessimistic inflation

If we were to stick with `current_gas_price` instead of `burn_gas_price`, users would do the same calculations as above, but if the gas price of the block height can ever be different from what gas price a specific user pays right now, it might break their formula.

neps/nep-0372.md
```diff
@@ -0,0 +1,79 @@
+---
+NEP: 372
+Title: Runtime Gas Price
+Author: Ben Kurrek <ben.kurrek@near.org>, Jakob Meier <jakob@near.org>
+DiscussionsTo: https://gov.near.org/t/expose-gas-price-and-contract-access-key-info-within-runtime/24788
+Status: Review
+Type: Standards Track
```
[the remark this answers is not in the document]
jakmeier: What track would you suggest? I though new runtime features go to this track.

> [NEP Types](https://github.com/near/NEPs/blob/master/neps/nep-0001.md#nep-types)
> 
>    A Standards Track NEP describes a new feature or implementation for the Near protocol or Near Smart Contract implementation standards.
>
>    An Informational NEP describes a Near design issue, or provides general guidelines or information to the Near community, but does not propose a new feature. Informational NEPs do not necessarily represent a Near community consensus or recommendation, so users and implementers can ignore Informational NEPs or follow their advice.
>
>    A Process NEP describes a process surrounding Near, or proposes a change to (or an event in) a process. Process NEPs are like Standards Track NEPs but apply to areas other than the Near platform itself. They may propose an implementation, but not to Near's codebase; they often require community consensus; unlike Informational NEPs, they are more than recommendations, and users are typically not free to ignore them. Examples include procedures, guidelines, changes to the decision-making process, and changes to the tools or environment used in Near development. Any meta-NEP is also considered a Process NEP.

jakmeier:
_Re-posting what I wrote on the draft implementation:_

Not enough interest from the community has been shown. So we dropped the ball on this and focus on other things.

There is also reason to wait. Paying for others using access keys was one use case here. But this can also be addressed with meta transactions. So we also want to wait and see how that will end up being implemented.

Personally, I think this could still be valuable. Using prepaid access keys have shorter latency than meta transaction. And link drops are harder to pull of with meta transactions compared to access keys. And I would imagine some users also want to stop doing certain actions if the gas price is too high.

But it's up to smart contract devs that need to raise their voice if they want this feature.

jakmeier:
@ori-near  Let's give some time, maybe a week, for objections but close this NEP otherwise. I still believe it _probably_ makes sense to add it, but there seems to be no demand right now and I don't want this to be hanging forever. (We can always reopen a new NEP if the demand arises in the future.)

cc @BenKurrek I think you would be the main person that might have objections and/or bring forward new arguments for why we need this.

jakmeier:
closing now that a week has passed with still no activity

PR #384, opened 2022-08-27
NEP-384: Host function for fetching block hashes in contract runtime

seunlanlege:
[rendered](https://github.com/near/NEPs/blob/031bde348517a55f94185af98d5a4683dd606f05/neps/nep-0384.md)

### DIFF
--- a/neps/nep-0384.md
+++ b/neps/nep-0384.md
@@ -107,6 +107,37 @@ Based on NEAR's architecture these block hashes will have to be stored in ecah c
 Implementation consideration: Extra work to maintain block hashes is required to be done "for free". This has to be efficient.
 All 256 block hashes should always stay in memory for that.
 
+### Storage management
+
+Each chunk should keep track of the latest 256 block hashes, mapped to their respective block heights.
+
+There are (at least) two alternatives. One is to store the entire number of hashes in a queue. In this case,
+calling the function `block_hash` would imply in the worst case, to iterate over the entire array.
+```
+key: "latest_block_hashes"
+type: Queue<Option<u8; 32>, 256>
+```
+Alternatively, it would be possible to store a mapping that indicates which blocks are available, and
+an array that stores the hashes themselves.
+
+```
+key: "mapping_existing_block_hashes"
+type: HashMap<block_height, ([u8; 32], IndexInArray)>
+```
+
+```
+key: "latest_block_hashes"
+type: [Option<u8; 32>, 256]
+```
+
+where `IndexInArray` tells where this block hash is stored in the `latest_block_hashes` array. This alternative
+has the advantage of not needing to iterate over the entire array (or queue) for finding the block hash, but instead
+having to maintain two datastructures.
+
+### Missing chunks
+
+If a chunk is skipped, the shard will not contain an updated state trie. That is, the next block will not have access to the latest
+256 block hashes. In order to deal with this case, ... TODO
 
 
 ## Copyright
--- a/neps/nep-0384.md
+++ b/neps/nep-0384.md
@@ -1,8 +1,8 @@
 ---
-NEP: xxxx
+NEP: 384
 Title: Host function for fetching block hashes in contract runtime
 Author: Seun Lanlege <seunlanlege@gmail.com>
-DiscussionsTo: https://github.com/near/NEPs/pull/9999
+DiscussionsTo: https://github.com/near/NEPs/pull/384
 Status: Draft
 Type: Runtime Spec
 Category: Contract
--- a/neps/nep-0384.md
+++ b/neps/nep-0384.md
@@ -11,7 +11,7 @@ Created: 27-Aug-2022
 
 ## Summary
 
-A new host function `env::block_hash` which will allow contract authors access the blockhashes of 256 of the most reccent blocks. This is akin to the [`blockhash`](https://docs.soliditylang.org/en/v0.8.15/units-and-global-variables.html#block-and-transaction-properties) api in ethereum.
+A new sdk api `env::block_hash` which will allow contract authors access the blockhashes of 256 of the most reccent blocks. This is akin to the [`blockhash`](https://docs.soliditylang.org/en/v0.8.15/units-and-global-variables.html#block-and-transaction-properties) api in ethereum.
 
 
 ## Motivation
@@ -98,8 +98,10 @@ A few other designs were considered such as a function `env::block_header` which
 
 ```rust
 /// This function should return a blockhash for 256 of the most reccent blocks
-fn block_hash(number: u32) -> Option<[u8; 32]>;
+fn block_hash(block_height: u64) -> Option<[u8; 32]>;
 ```
+
+This api will require that the map of block heights to hashes are stored in each shard's state trie pending the implementation of a global state trie.
 
 ## Copyright
 
--- a/neps/nep-0384.md
+++ b/neps/nep-0384.md
@@ -101,7 +101,13 @@ A few other designs were considered such as a function `env::block_header` which
 fn block_hash(block_height: u64) -> Option<[u8; 32]>;
 ```
 
-This api will require that the map of block heights to hashes are stored in each shard's state trie pending the implementation of a global state trie.
+This API will require that the map of block heights to hashes are stored in each shard's state trie pending the implementation of a global state trie.
+Based on NEAR's architecture these block hashes will have to be stored in ecah chunk state trie.
+
+Implementation consideration: Extra work to maintain block hashes is required to be done "for free". This has to be efficient.
+All 256 block hashes should always stay in memory for that.
+
+
 
 ## Copyright
 
--- a/neps/nep-xxxx.md
+++ b/neps/nep-xxxx.md
@@ -1,0 +1,108 @@
+---
+NEP: xxxx
+Title: Host function for fetching block hashes in contract runtime
+Author: Seun Lanlege <seunlanlege@gmail.com>
+DiscussionsTo: https://github.com/near/NEPs/pull/9999
+Status: Draft
+Type: Runtime Spec
+Category: Contract
+Created: 27-Aug-2022
+---
+
+## Summary
+
+A new host function `env::block_hash` which will allow contract authors access the blockhashes of 256 of the most reccent blocks. This is akin to the [`blockhash`](https://docs.soliditylang.org/en/v0.8.15/units-and-global-variables.html#block-and-transaction-properties) api in ethereum.
+
+
+## Motivation
+
+The IBC protocol requires that blockchains (aka host state machines) are able to introspect their own `ConsensusState` (aka Header). This is stated, amongst other requirements in their [IBC host requrements](https://github.com/cosmos/ibc/blob/main/spec/core/ics-024-host-requirements/README.md#consensus-state-introspection). This api is what allows trustless IBC connection handshakes to happen: 
+
+```ts
+// ConnOpenTry relays notice of a connection attempt on chain A to chain B (this code is executed on chain B).
+function connOpenTry(
+  counterpartyConnectionIdentifier: Identifier,
+  counterpartyPrefix: CommitmentPrefix,
+  counterpartyClientIdentifier: Identifier,
+  clientIdentifier: Identifier,
+  clientState: ClientState,
+  counterpartyVersions: string[],
+  delayPeriodTime: uint64,
+  delayPeriodBlocks: uint64,
+  proofInit: CommitmentProof,
+  proofClient: CommitmentProof,
+  proofConsensus: CommitmentProof,
+  proofHeight: Height,
+  consensusHeight: Height) {
+    // generate a new identifier
+    identifier = generateIdentifier()
+    
+    abortTransactionUnless(validateSelfClient(clientState))
+    abortTransactionUnless(consensusHeight < getCurrentHeight())
+    expectedConsensusState = getHostConsensusState(consensusHeight) // we need to return a near header at this height.
+    expectedConnectionEnd = ConnectionEnd{INIT, "", getCommitmentPrefix(), counterpartyClientIdentifier,
+                             clientIdentifier, counterpartyVersions, delayPeriodTime, delayPeriodBlocks}
+
+    versionsIntersection = intersection(counterpartyVersions, getCompatibleVersions())
+    version = pickVersion(versionsIntersection)
+
+    connection = ConnectionEnd{TRYOPEN, counterpartyConnectionIdentifier, counterpartyPrefix,
+                               clientIdentifier, counterpartyClientIdentifier, version, delayPeriodTime, delayPeriodBlocks}
+    abortTransactionUnless(connection.verifyConnectionState(proofHeight, proofInit, counterpartyConnectionIdentifier, expectedConnectionEnd))
+    abortTransactionUnless(connection.verifyClientState(proofHeight, proofClient, clientState))
+    abortTransactionUnless(connection.verifyClientConsensusState(
+      proofHeight, proofConsensus, counterpartyClientIdentifier, consensusHeight, expectedConsensusState))
+
+    provableStore.set(connectionPath(identifier), connection)
+    addConnectionToClient(clientIdentifier, identifier)
+}
+
+// ConnOpenAck relays acceptance of a connection open attempt from chain B back to chain A (this code is executed on chain A).
+function connOpenAck(
+  identifier: Identifier,
+  clientState: ClientState,
+  version: string,
+  counterpartyIdentifier: Identifier,
+  proofTry: CommitmentProof,
+  proofClient: CommitmentProof,
+  proofConsensus: CommitmentProof,
+  proofHeight: Height,
+  consensusHeight: Height) {
+    abortTransactionUnless(consensusHeight < getCurrentHeight())
+    abortTransactionUnless(validateSelfClient(clientState))
+    connection = provableStore.get(connectionPath(identifier))
+    abortTransactionUnless((connection.state === INIT && connection.version.indexOf(version) !== -1)
+    expectedConsensusState = getHostConsensusState(consensusHeight) // we need to return a near header at this height.
+    expectedConnectionEnd = ConnectionEnd{TRYOPEN, identifier, getCommitmentPrefix(),
+                             connection.counterpartyClientIdentifier, connection.clientIdentifier,
+                             version, connection.delayPeriodTime, connection.delayPeriodBlocks}
+    abortTransactionUnless(connection.verifyConnectionState(proofHeight, proofTry, counterpartyIdentifier, expectedConnectionEnd)) // so we can verify that the counterparty chain has the correct header.
+    abortTransactionUnless(connection.verifyClientState(proofHeight, proofClient, clientState))
+    abortTransactionUnless(connection.verifyClientConsensusState(
+      proofHeight, proofConsensus, connection.counterpartyClientIdentifier, consensusHeight, expectedConsensusState))
+    connection.state = OPEN
+    connection.version = version
+    connection.counterpartyConnectionIdentifier = counterpartyIdentifier
+    provableStore.set(connectionPath(identifier), connection)
+}
+```
+
+The api `env::block_hash` allows us validate the full header submitted via contract calldata by comparing hashes.
+
+## Rationale and alternatives
+
+A few other designs were considered such as a function `env::block_header` which would return the full header at a given height, but was discouraged against a result of expensive db calls. This mapping of block numbers to headers are cheap enough to be stored on-chain.
+
+
+## Specification
+
+```rust
+/// This function should return a blockhash for 256 of the most reccent blocks
+fn block_hash(number: u32) -> Option<[u8; 32]>;
+```
+
+## Copyright
+
+[copyright]: #copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


mfornet:
I want to highlight that this is possible today by building a LC from the chain to itself, and making a proof of the current blockhash. I understand using this approach is significantly most expensive, if not completely prohibitive (due to the number of signature verifications required). It should be cheaper after #364 is added. We should consider this as an alternative to this NEP.

seunlanlege:
> I want to highlight that this is possible today by building a LC from the chain to itself, and making a proof of the current blockhash. I understand using this approach is significantly most expensive, if not completely prohibitive (due to the number of signature verifications required). It should be cheaper after https://github.com/near/NEPs/pull/364 is added. We should consider this as an alternative to this NEP.

So just to be clear, you're asking us to run a light client of near, on near? Even with native signature verification this is still prohibitively expensive. I'm not sure how this could be considered as an alternative.

neps/nep-0384.md
```diff
@@ -0,0 +1,108 @@
+---
+NEP: 384
+Title: Host function for fetching block hashes in contract runtime
+Author: Seun Lanlege <seunlanlege@gmail.com>
+DiscussionsTo: https://github.com/near/NEPs/pull/384
+Status: Draft
+Type: Runtime Spec
```
ori-near, line 7: Can we change this to Protocol?

ori-near:
Thank you @seunlanlege for submitting this NEP. In order to move this Protocol NEP to the review stage, can you please add a reference implementation?

blasrodri:
> Thank you @seunlanlege for submitting this NEP. In order to move this Protocol NEP to the review stage, can you please add a reference implementation?

We won't be able to add a reference implementation without having some pointers from the core dev team on how to do so. Happy to contribute on this, but please give us a hand :)

frol:
@blasrodri Please, try starting a conversation on Zulip as it is unclear how to move forward from here to us (NEP moderators): https://near.zulipchat.com/#narrow/stream/295306-pagoda.2Fcontract-runtime

blasrodri:
> @blasrodri Please, try starting a conversation on Zulip as it is unclear how to move forward from here to us (NEP moderators): https://near.zulipchat.com/#narrow/stream/295306-pagoda.2Fcontract-runtime

Will do! Thanks

neps/nep-0384.md
```diff
@@ -0,0 +1,108 @@
...
+Category: Contract
+Created: 27-Aug-2022
+---
+
+## Summary
+
+A new host function `env::block_hash` which will allow contract authors access the blockhashes of 256 of the most reccent blocks. This is akin to the [`blockhash`](https://docs.soliditylang.org/en/v0.8.15/units-and-global-variables.html#block-and-transaction-properties) api in ethereum.
```
akhi3030, line 14: One epoch is 43_200 blocks and RPC nodes are configured to keep a 5 epochs worth of data before garbage collecting.  So I would imagine that a RPC node can easily return information about more than just 256 recent blocks.  Should this number be increased or specified in a different way to make the feature more generic?

neps/nep-0384.md
```diff
@@ -0,0 +1,108 @@
...
+## Rationale and alternatives
+
+A few other designs were considered such as a function `env::block_header` which would return the full header at a given height, but was discouraged against a result of expensive db calls. This mapping of block numbers to headers are cheap enough to be stored on-chain.
+
+
+## Specification
+
+```rust
+/// This function should return a blockhash for 256 of the most reccent blocks
+fn block_hash(number: u32) -> Option<[u8; 32]>;
```
akhi3030, line 101:
```suggestion
/// Returns a blockhash for the `number` block if available.
fn block_hash(number: u32) -> Option<[u8; 32]>;
```

akhi3030, line 101: I believe that block heights are u64 in NEAR.

mina86: If this returns multiple block hashes than this should be `Vec<[u8; 32]>`.

neps/nep-0384.md
```diff
@@ -0,0 +1,108 @@
...
+    connection.counterpartyConnectionIdentifier = counterpartyIdentifier
+    provableStore.set(connectionPath(identifier), connection)
+}
+```
+
+The api `env::block_hash` allows us validate the full header submitted via contract calldata by comparing hashes.
+
+## Rationale and alternatives
+
+A few other designs were considered such as a function `env::block_header` which would return the full header at a given height, but was discouraged against a result of expensive db calls. This mapping of block numbers to headers are cheap enough to be stored on-chain.
```
mina86, line 94: > This mapping of block numbers to headers are cheap enough to be stored on-chain.

What do you mean by that? I don’t think we want to store that on-chain in the sense that it’s stored in state of the latest block.

neps/nep-0384.md
seunlanlege: This should return a single block hash

neps/nep-0384.md
seunlanlege: > So I would imagine that a RPC node can easily return information about more than just 256 recent blocks.

All nodes in the network must have access to those values, otherwise any block containing transactions that access much older block hashes would panic

neps/nep-0384.md
seunlanlege: > What do you mean by that?

Yes i did assume that they would be stored on-chain, would that not be the case? I worry about consensus failures if that is not the case.

neps/nep-0384.md
mina86: So `number` is block height?

neps/nep-0384.md
mina86: We might have different understanding of ‘on-chain’.  The data is on-chain in the sense that everyone agrees that if you’re in a branch whose head is block X than that uniquely determines hashes of all previous blocks.  However, it’s not stored ‘on-chain’ in the sense that this in formation isn’t immediately available in block X.  To get that information you have to fetch X’s parent, that block’s parent, that block’s parent etc.

neps/nep-0384.md
mina86: > One epoch is 43_200 blocks and RPC nodes are configured to keep a 5 epochs worth of data before garbage collecting.

I don’t think we can relay on this at protocol lever though.  For example, RPC nodes can be configured to store just three epochs worth of data.  Even that is partially because we settled on arbitrary limit of at least three epochs.  In practice RPC node could work with a shorter horizon I think.  I’d rather have this function be limited to as few blocks as possible so we don’t encode current client configuration in the protocol.

The question here is, where does the 256 comes from?

seunlanlege: > 256

This will have to be enforced at the protocol level, as minimum number of stored blocks to validate inncoming blocks

neps/nep-0384.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+## Specification
+
+```rust
+/// This function should return a blockhash for 256 of the most reccent blocks
```
jakmeier, line 100: > 256 of the most reccent blocks

What *exactly* does that mean?

1. every block in the range of `current block height - 256` up to `current block height - 1`?
2. the last 256 blocks that are not missing?
3. the last 256 final blocks?

Option 3 would make the implementation much easier and consequently the host function call cheaper. But I suspect you want 1. or 2.

jakmeier (commented):
Sorry I am not familiar with IBC. To understand the use case better, can you explain how often this would have to be called roughly? (If it happens only once a month to establish a new connection or something like that, we can design it quite differently than if it is called 10 times in each block.)

neps/nep-0384.md
seunlanlege: 1. Is the intended case

seunlanlege:
> Sorry I am not familiar with IBC. To understand the use case better, can you explain how often this would have to be called roughly? (If it happens only once a month to establish a new connection or something like that, we can design it quite differently than if it is called 10 times in each block.)

It can be called as many times in a block.

blasrodri:
@seunlanlege we're discussing some details here: https://github.com/near/nearcore/pull/7792#pullrequestreview-1139661632

can you please clarify the implementation? please provide some examples/scenarios.

neps/nep-0384.md
blasrodri:
```suggestion
/// This function should return a blockhash for 256 of the most reccent blocks
fn block_hash(number: u64) -> Option<[u8; 32]>;
```

blasrodri:
> @seunlanlege we're discussing some details here: [near/nearcore#7792 (review)](https://github.com/near/nearcore/pull/7792#pullrequestreview-1139661632)
> 
> can you please clarify the implementation? please provide some examples/scenarios.

What we are looking for is a host function that allows us to retrieve a block hash
by specifying a block height (if available) from within the latest 256 blocks.

Let's assume we want the block hash for height `101`. In thata case, we'd pass:
`block_hash(101)` and it will return:
    - `Some(hash)` of the block in height `101` if it's within the 256 window block.
    - `None` if the height does not exist
    - `None` if the height does not belong to the 256 window of blocks from `[block(current_height)...block(255 heights before)]`.

blasrodri:
@ori-near  @mina86 @jakmeier I've provided an implementation here:
https://github.com/near/nearcore/pull/7792

What would be the next steps in order to move this NEP forward? 
@seunlanlege can we make it ready for review?

blasrodri:
@seunlanlege relaying a question from @matklad 
> One extra question I have here is what guarantees that we actually do have 256 blocks in storage? As in, what if a new node joins the network and syncs up just to a very recent block, skipping its parents

seunlanlege:
> @seunlanlege relaying a question from @matklad
> 
> > One extra question I have here is what guarantees that we actually do have 256 blocks in storage? As in, what if a new node joins the network and syncs up just to a very recent block, skipping its parents

So this is why in the spec, i specify that this data must be stored on-chain. i.e part of the state trie. So that this data is present to all nodes, irrespective of their pruning configuration:

https://github.com/near/NEPs/blob/031bde348517a55f94185af98d5a4683dd606f05/neps/nep-0384.md#rationale-and-alternatives

also see my comment: https://github.com/near/NEPs/pull/384#discussion_r989411434

jakmeier:
@seunlanlege Do you suggest to include the hashes of up to 256 blocks as new nodes in the trie? If so, the NEP specification would have to include what the trie keys for that would be and how the set of hashes is represented. It also must be made very clear in the NEP which exact hashes are included and which are removed from `State`. All of this becomes part of the state root computation, which makes this a significantly more invasive protocol change than just adding a host function.

@mm-near what do you think about this suggestion? I imagine this makes things conceptually easier for challenges and view clients, right? Do you see any problems with this suggestion?

seunlanlege:
> If so, the NEP specification would have to include what the trie keys for that would be and how the set of hashes is represented.

This isn't neccessary contract authors shouldn't be reading the hashes from the state trie, there should be an ~host~ in-built function they can call with a block number that'll return the hash if it exists.

jakmeier:
> > If so, the NEP specification would have to include what the trie keys for that would be and how the set of hashes is represented.
> 
> This isn't necessary, contract authors shouldn't be reading the hashes from the state trie, there should be an ~host~ in-built function they can call with a block number that'll return the hash if it exists.

Sure, it doesn't need to be defined on the host function interface. But it must be defined for the protocol specification. NEPs should specify protocol changes they make.

mm-near:
@jakmeier - yes, this helps a lot (makes the implementation of this host function a lot easier) - but it moves some complexity to other spots:

* we'd have to keep this data (block hash) in the state of each shard (as host function can only access the current shard's state)
* we'd have to define where exactly in trie do we keep it (that should be the easy one) - and for how long (are we sure that we want 256? or longer?)
* the chunk producers will be responsible for adding this to state (and we'd need to change the validators / apply_chunk functions too)

neps/nep-0384.md
mm-near: ok - but this means that there might be situations where there are no blocks within the last '256' heights (if network is in trouble and cannot gather consensus etc)  - and we're ok with that?

neps/nep-0384.md
mm-near: +1 - we should call it block_height (as number is ambiguous)

jakmeier:
> @jakmeier - yes, this helps a lot (makes the implementation of this host function a lot easier) - but it moves some complexity to other spots:
> 
>     * we'd have to keep this data (block hash) in the state of each shard (as host function can only access the current shard's state)
> 
>     * we'd have to define where exactly in trie do we keep it (that should be the easy one) - and for how long (are we sure that we want 256? or longer?)
> 
>     * the chunk producers will be responsible for adding this to state (and we'd need to change the validators / apply_chunk functions too)

Thanks!

@seunlanlege do you think you can come up with a design that addresses these points? Let us know if you need help understanding details.

And cc @blasrodri, I think once the design is clearly defined, this will add a whole new set of work for the reference implementation.

mm-near:
And a side note -- if we add this data to the state - we'll have to be careful when we do resharding (as we want a copy of this data to end up on each shard afterwards). So this is slowly moving us towards potentially designing the 'global storage' (but I don't think we have a NEP/discussion on that yet)

blasrodri:
> > @jakmeier - yes, this helps a lot (makes the implementation of this host function a lot easier) - but it moves some complexity to other spots:
> > ```
> > * we'd have to keep this data (block hash) in the state of each shard (as host function can only access the current shard's state)
> > 
> > * we'd have to define where exactly in trie do we keep it (that should be the easy one) - and for how long (are we sure that we want 256? or longer?)
> > 
> > * the chunk producers will be responsible for adding this to state (and we'd need to change the validators / apply_chunk functions too)
> > ```
> 
> Thanks!
> 
> @seunlanlege do you think you can come up with a design that addresses these points? Let us know if you need help understanding details.
> 
> And cc @blasrodri, I think once the design is clearly defined, this will add a whole new set of work for the reference implementation.

@seunlanlege and I would like to request a call to discuss the points mentioned above, and to resolve all remaining doubts in general. Perhaps @jakmeier/@mm-near/@mina86 could hop in? I believe that a sync session between us could potentially unlock is and help us move faster. What do you guys think?

cc: @ori-near

ori-near:
Hi @blasrodri – great idea to jump on a synchronous call to address the unresolved doubts. Happy to help schedule a call.

@seunlanlege @jakmeier @mm-near @mina86 – can you all please confirm your availability using this link?
https://calendly.com/d/gsj-vwj-vp2/nep-384-discussion

jakmeier:
Yes, I'm happy to jump in a call. I've selected all suggested time slots that are not already in conflict with other meetings, unfortunately there were not many.

Btw, I wouldn't count on @mina86 joining, as they recently left their position at Pagoda. But of course, you are more than welcome if you are getting bored already :)

ori-near:
Great, thank you for everyone who confirmed their availability! We scheduled the meeting for Wednesday, November 16 at 17:30 UTC. @blasrodri Please designate a person to post the outcome of the discussion in this thread.

mina86:
> Btw, I wouldn't count on @mina86 joining, as they recently left their position at Pagoda. But of course, you are more than welcome if you are getting bored already :)

Yeah, don’t count on me.  I successfully fight my boredom by finally finishing Pillars of Eternity. ;)

Good luck with the NEP!

blasrodri:
> Great, thank you for everyone who confirmed their availability! We scheduled the meeting for Wednesday, November 16 at 17:30 UTC. @blasrodri Please designate a person to post the outcome of the discussion in this thread.

Yes! @akc2267 will be helping out

blasrodri:
```
NEP-384 Notes
- Post outcome of discussion on github issue and tag Ori
- modify block authorship process to have a final insert hash of previous block
- NEAR has state trie for each shard.
- block contains block header and chunk hashes of each shard
    - chunks are mini-blocks and include state root
    - chunks can be missed if miners take too long to produce a chunk
    - if you know the block hash, you can verify each state piece
- for time being, each account is stuck on a single shard
    - in future an account can be migrated to a different shard max every ~12h
    - doing account to account txs are assumed to be across different shards - will always be across multiple blocks
- we need to grab block hash
    - it has timestamp with it (at block level, not chunk)
    - needs header data with merkle tree to all chunk hashes
    - this needs to be stored in each shard state
- Can we just check the chunk where the contract account lives?
    - nope, need all chunks
- we should put state in every shard chunk
    - update state when you apply a chunk at the end
        - also at end of every block
        - reviewers will probably be annoyed that this consumes gas without paying fees.
        - can do it all in memory if we do it efficiently
            - time bets are off if we need to offload on disk
        - 1 teragas is 1ms
        - in worst case, amount of time <= gas paid
- Trie Key NEP to store key for block hashes & 256 hash blob
    - small enough to rotate in memory
    - pointer to shift it around
- slashing and challenges coming next year
    - need to ensure we future proof our implementation for this
    - Murchin? is the tech lead for this decision
```

akc2267:
@ori-near below are some notes from today's meeting. Main takeaway is @seunlanlege and @blasrodri will modify this ticket to take into account new info on block hashes, placing state in each chunk, and gas efficiency requirements
`- modify block authorship process to have a final insert hash of previous block
- NEAR has state trie for each shard.
- block contains block header and chunk hashes of each shard
    - chunks are mini-blocks and include state root
    - chunks can be missed if miners take too long to produce a chunk
    - if you know the block hash, you can verify each state piece
- for time being, each account is stuck on a single shard
    - in future an account can be migrated to a different shard max every ~12h
    - doing account to account txs are assumed to be across different shards - will always be across multiple blocks
- we need to grab block hash
    - it has timestamp with it (at block level, not chunk)
    - needs header data with merkle tree to all chunk hashes
    - this needs to be stored in each shard state
- Can we just check the chunk where the contract account lives?
    - nope, need all chunks
- we should put state in every shard chunk
    - update state when you apply a chunk at the end
        - also at end of every block
        - reviewers will probably be annoyed that this consumes gas without paying fees.
        - can do it all in memory if we do it efficiently
            - time bets are off if we need to offload on disk
        - 1 teragas is 1ms
        - in worst case, amount of time <= gas paid
- Trie Key NEP to store key for block hashes & 256 hash blob
    - small enough to rotate in memory
    - pointer to shift it around
- slashing and challenges coming next year
    - need to ensure we future proof our implementation for this
    - Murchin? is the tech lead for this decision
`

jakmeier:
Thanks for the detailed notes! To summarize the next steps, for those with no second to spare:
- The NEP will be updated by @seunlanlege to include the block hashes in each chunk state trie.
- Implementation needs to be updated afterwards by @blasrodri.
- For the implementation: Extra work to maintain block hashes is required to be done "for free". This has to be efficient. All 256 block hashes should always stay in memory for that. (note @blasrodri this could be tricky because the runtime is stateless in general, we can discuss it once you get there)

cc @ori-near

neps/nep-0384.md
```diff
@@ -0,0 +1,116 @@
...
+fn block_hash(block_height: u64) -> Option<[u8; 32]>;
+```
+
+This API will require that the map of block heights to hashes are stored in each shard's state trie pending the implementation of a global state trie.
```
jakmeier, line 104: Just to clarify: The status on global state is not even close to committed, at this point it is just a high-level idea of potential work for the mid to far future. So we really need a solid design for duplicating the state on all shards, with specified binary layout etc.

blasrodri: Will this design be part of the NEP?

jakmeier: The design should be described in the NEP with reasonable details.

I would expect to see at least a definition of the key-value pairs involved in the trie and what encoding shall be used. Other details, such as padding, can probably be discussed later or left as implementation detail entirely.

Also, the NEP should describe how missing chunks will be handled. If a chunk is skipped, that shard will not have an updated state trie, so it will not be possible to have the latest block hashes. The NEP should define what the block hashes will be available during and after such a scenario.

blasrodri: I've started drafting these points you've mentioned. I am clueless on how to deal with the skipped chunks. Can you help me out there?

blasrodri: @jakmeier

jakmeier: Thanks for the ping.

Yeah, it's a bit tricky. The basic problem is that a shard does not have an updated state root if no new chunk has been produced. Getting the latest block hash is impossible if we want to retrieve it from the state root.

However, do we really need the latest block hash? Maybe a solution would be to define, in the specification, that the host function will always return the block hash of the latest block in which the shard's chunk is present.

My reasoning for why it would be okay: If shard 1 skipped a few chunks, then a contract in shard 1 should not really care about state changes in these blocks. All observable state for that shard is still the same as it was before the skipped chunks.

Counter point: When building bridges, there is an off-chain party calling into the contract that is on an arbitrary shard. This party cares about the global state, not about the arbitrary shard's view of the state.

I don't know enough about the use cases to judge if the counter point is a valid concern. I suspect it might. But I don't really see a solution here otherwise. We would probably have to give up on the idea of storing the block hashes in the state trie and search for more complex alternatives. Such as using some form of global state, as we have also discussed before, I think. (Of course, this would hugely increase the scope.)

blasrodri: IBC works by having a light client of the counterparty chain running on NEAR (in this case). This means that we should be able to create this light client once (initialize it), and then maintain its state periodically, to ensure that it's a reasonable up to date representation of what's happening on the other chain. This Light Client, needs to be able to talk to a NEAR light client living on the counterparty chain.

There could be scenario in which the light client in NEAR receives a packet from the other chain, in which it needs to verify a certain state for a block number for which there's no reference on that specific shard. In that case, the request will fail. Even when the execution failed, that shard would get its local state updated for that specific block height, but unfortunately it won't match the block number from the packet (it will be higher).

One potential option could be to retry until the execution of the contract is mapped with the shard that actually contains the right state. However, that's not guaranteed to happen within a certain time frame, and I believe it's not even guaranteed to happen at all, right?

Based on this, seems like we need one design in which either:

- all shards have  access to a global state
- each shard gets its state updated on each block regardless if there was or not a contract being executed
on that specific block (by executing a dummy contract on each shard on every block, for instance)

jakmeier: Thanks for the explanation! 

I think there is a misunderstanding for why empty chunks occur. It has nothing to do with whether a contract was executed or not. It's actually because sometimes chunk-producers may miss their chunk, for various reasons. We don't stall the entire chain in that case, instead we let other shards move on and submit their chunk anyway. In which case there will be a block with a missing chunk.

> One potential option could be to retry until the execution of the contract is mapped with the shard that actually contains the right state. However, that's not guaranteed to happen within a certain time frame, and I believe it's not even guaranteed to happen at all, right?

We have a guarantee that a chunk will **eventually** be produced, as long as 2/3 of chunk and block producer work properly. Otherwise the liveness property of the protocol would be violated.
For the exact timing, I'm not sure if we can quantify it. But here we are basically worrying about the case where one shard makes no progress for a long time. No matter how this NEP is implemented, the shard making progress is a pre-condition for the bridge to work anyway. Therefore, I would say that problem is out of scope for this NEP. 

So I think retrying would work and wouldn't slow things down more than fundamentally possible. It doesn't sound super super pretty but maybe it's the best solution, as otherwise, it seems we would require some form of global state to be implemented first. (Which would almost certainly also only see an update on shard x after shard x has produced a chunk.)

blasrodri: I'm happy to retry :). I'll update the NEP accordingly then

ori-near:
Hi @seunlanlege / @blasrodri / @jakmeier (or anyone who is interested in championing this NEP forward), we are cleaning the NEP backlog and noticed that this PR still has some activity, so we assume you are still actively working on it. 

Please ping the @near/nep-moderators once you are ready for us to review it. We will review it again in early March, unless we hear from you sooner. We typically close NEPs that are inactive for more than two months, so please let us know if you need more time.

blasrodri:
> Hi @seunlanlege / @blasrodri / @jakmeier (or anyone who is interested in championing this NEP forward), we are cleaning the NEP backlog and noticed that this PR still has some activity, so we assume you are still actively working on it.
> 
> Please ping the @near/nep-moderators once you are ready for us to review it. We will review it again in early March, unless we hear from you sooner. We typically close NEPs that are inactive for more than two months, so please let us know if you need more time.

Thanks @ori-near !
I’ll be updating the NEP based on the latest input and start working on the implementation too.

walnut-the-cat:
Hello! Protocol NEP moderator here. it seems there hasn't been any activity on this NEP for quite some time. Is it still on going? Or shall we close NEP for now? cc. @blasrodri

blasrodri:
> Hello! Protocol NEP moderator here. it seems there hasn't been any activity on this NEP for quite some time. Is it still on going? Or shall we close NEP for now? cc. @blasrodri

Please close :)

Issue #136, opened 2020-12-21
Interactive Fungible Token


robert-zaremba:
Following early discussion  about NEP-21, the work on [NEARswap](https://github.com/near-clp/smart-contracts#our-protocols) and author work on security tokens in Ethereum ecosystem, we propose a new fungible token standard.

* &#x1F449; &nbsp;  **[Smart contract interface](https://github.com/robert-zaremba/nep-136-fungible-token/blob/main/src/interface.rs)** (Rust code).
* Details and analysis: https://github.com/robert-zaremba/nep-136-fungible-token 

## Rationale

The only approved token standard in NEAR ecosystem is [NEP-21](https://github.com/near/NEPs/blob/master/specs/Standards/Tokens/FungibleToken.md).
It's an adaptation of [ERC-20](https://eips.ethereum.org/EIPS/eip-20) token standard from Ethereum. Both NEP-21 and ERC-20 are designed to be minimalistic and functional: provide clear interface for token transfers, allowance management (give access to token transfers to other entities or smart-contract - this is useful if other smart contract wants to withdraw it's allowance for things we buy). The ERC-20 standard is the first standard Ethereum fungible token standard and created lot of legacy. All early tooling (wallets, explorers, token templates) were build with ERC-20 in mind. Over time, the ecosystem listed many problems related to ERC-20 design:
1. `decimals` should be required, rather than optional. This is essential to user-friendly token amount handling.
2. Lack of standarized metdata reference. Token should provide a resource which describes it's metadata (name, description, etc...) - this can be an URL, CID, or directly written in the contract code.
3. `transfer` and `transferFrom` are lacking a reference (memo) argument - which is essential for compliance reasons. We need to be able to be able to link the transfer to an an external document (eg: invoice), order ID or other on-chain transaction or simply set a transfer reason.
4. Direct `transfers` to smart-contract in general is an error and should be protected. Both in Ethereum in NEAR, smart-contracts are NOT notified if someone is sending them tokens. This causes funds lost, token locked and many critical misbehavior.
5. Avoid problems mentioned in the previous point, all transfers should be done through `approve` (allowance creation) and `transferFrom`, which is less intuitive and makes UX more complex: not only we need to create and keep track of right allowance (with all edge cases: user creates allowance, but token is not calling `transferFrom` and makes user to create another allowance).
6. Fees calculation. With `approve` + `transferFrom`, the business provider has to make an additional transaction (transferFrom) and calculate it in the operation cost.

[...]

PR #394, opened 2022-09-15
Updating NEP-141 authors

robert-zaremba:
NEP-141 came through my proposal https://github.com/near/NEPs/issues/136 as a more robust Interface for Fungible Tokens and reflection from the previous NEP-122. 
Prior to that we were discussing the pitfalls for ERC20 and a better solution that would serve NEAR ecosystem needs. See Rationale described in https://github.com/near/NEPs/issues/136. I was part of the core working group (together with Evgeny, @oysterpack, @luciotato, @mikedotexe  ) designing the NEP-141, where we hardened the NEP-136 design by analyzing the storage use cases and metadata management.

### DIFF
--- a/neps/nep-0141.md
+++ b/neps/nep-0141.md
@@ -1,7 +1,7 @@
 ---
 NEP: 141
 Title: Fungible Token Standard
-Author: Evgeny Kuzyakov <ek@near.org>, @oysterpack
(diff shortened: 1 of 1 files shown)


robert-zaremba:
cc @mikedotexe

frol (approved):
This is fair. Thanks for fixing it

PR #400, opened 2022-10-03
Clarify reference and draft implementation

akhi3030:
I believe this document uses the term "reference implementation" to mean "draft implementation".  Given that we say that `nearcore` is the reference implementation of the protocol, this PR rewords "reference implementation" to "draft implementation" to clarify the intend.

### DIFF
--- a/neps/nep-0001.md
+++ b/neps/nep-0001.md
@@ -99,8 +99,7 @@ The NEP process has various roles and responsibilities.
 - Submits proposal with prototype when ready for review and changes the status to “Draft”
 - Addresses comments
 - Presents to working group and incorporates necessary changes
-- Executes implementation once approved
-- Provides reference implementation and documentation once approved
+- Provides final implementation with sufficient testing and documentation once approved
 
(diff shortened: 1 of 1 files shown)


frol:
@akhi3030 It was actually the Protocol Working Group's requirement that they only approve a NEP when there is a reference implementation added to nearcore as a PR. Otherwise, we might have a pile of useless NEPs that nobody will implement. I don't think we should downgrade "reference implementation" to "draft implementation" in the section of "What does a successful NEP look like?". Do you still believe otherwise?

akhi3030:
CC: @ori-near and @bowenwang1996.  Based on a brief slack chat we had, my understanding is that "reference implementation" above should mean "draft implementation".  I don't have a strong opinion one way or the other, just wanted to clarify the intend of the document if we are all in agreement.

bowenwang1996:
> my understanding is that "reference implementation" above should mean "draft implementation"

Yeah I agree. The implementation should demonstrate the feasibility of the proposal, but it is not expected to be perfectly polished and ready to be merged before the SMEs start reviewing the proposal.

frol (approved):

PR #406, opened 2022-10-10
Typo fixed

zlog-in:
`receipt_id` should be [`Receipt.receipt_id`]


frol (approved):

frol:
@zlog-in Good catch, thanks!

PR #408, opened 2022-10-10
NEP-408: Injected Wallet API

MaximusHaximus:
## Summary

This is a proposal for a new NEP regarding an Injected Wallet Standard. Adopted from https://github.com/near/NEPs/pull/370 as the original author is no longer working on this task; see #370 for prior discussion.

## Demo

You can find a demo of this standard using the project detailed below:

  - Implementation: https://github.com/lewis-sqa/near-injected-wallet-poc
  - Running the project:
    - Execute `yarn install && yarn start` from root.
    - Navigate to http://localhost:1234.


frol:
This NEP is formatted according to NEP guidelines, so I am passing it to the review stage.

@MaximusHaximus do you need to address outstanding comments or is it ready for final review?

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+
+```ts
+window.near.wallet.on("networkChanged", ({ network }) => {
+  console.log("Network Changed", network);
+});
+```
+
+## Drawbacks
+
+There are cases when a dApp wants to sign a transaction with a signer that's yet to be determined (e.g. "Donate" button). Using the `Transaction` instance from `near-api-js` sadly limits this possibility as it requires the details of a `signerId` and `publicKey` upfront. Below are suggested solutions:
```
andy-haynes, line 366: > Using the `Transaction` instance from `near-api-js` sadly limits this possibility as it requires the details of a `signerId` and `publicKey` upfront.

Reads more succinct as

> The `Transaction` class from `near-api-js` precludes this behavior as it must be instantiated with both `signerId` and `publicKey`.

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+
+```ts
+const { accounts } = window.near.wallet;
+
+console.log(accounts) // [{ accountId: "test.testnet", publicKey: PublicKey }]
+```
+
+#### Methods
+
+##### `connect`
```
andy-haynes, line 162: What should happen when another method is called before the first invocation of `connect`?

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+  networkChanged: { network: Network };
+}
+
+type Unsubscribe = () => void;
+
+interface Wallet {
+  id: string;
+  connected: boolean;
+  network: Network;
+  accounts: Array<Account>;
```
andy-haynes, line 99: Why use an array over an object hash here?

andy-haynes (changes requested):
There are some grammatical issues that would ideally be resolved before publishing, hopefully not too late in the process 🙂 

Also left some clarifying questions that I didn't see addressed on the original PR

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+- Although risk is limited with `FunctionCall` access keys, storing key pairs in the dApp means a level of trust is required to ensure they're kept securely to avoid being compromised.
+- Signing transactions that match the `FunctionCall` access key stored in the dApp means we don't require connection to a wallet. This is particularly useful for wallets such as Ledger and WalletConnect where it isn't always available. The drawback to this approach is the dApp must handle much of the logic found already in the wallet such as storing key pairs and matching applicable access keys during signing.
+
+Although we had some technical challenges, the decision to store the key pairs in the dApp means users own them and gas-only intensive `FunctionCall` dApps work seamlessly with wallets that aren't always available. The expectation for `near-api-js` to become wallet-agnostic will reduce the complexity shifted onto the dApp as it can abstract away the logic required for handling the key pairs and only defer to the wallet when further permission is needed.
```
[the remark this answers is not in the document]
MaximusHaximus: :+1: agreed! editing.

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+
+A side effect to this approach (coupled with multiple accounts) means we must generate the key pairs ahead of time and pass each public key to the `signIn` method as a list of `accounts`. The introduction of the `connect` and `disconnect` methods allows dApps to request visibility for a subset of accounts imported in the wallet to help populate this list and sign transactions.
```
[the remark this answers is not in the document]
MaximusHaximus: Re-wording this paragraph :)

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+
+## Specification
+
+Injected wallets are browser extensions that implement the `Wallet` API (see below) on the `window` object. To avoid namespace collisions and easily detect when they're available, wallets will mount under their own key within `window.near` (e.g. `window.near.sender`).
```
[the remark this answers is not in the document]
MaximusHaximus: Yep! I'll clarify that wording to be more explicit

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: Since the spec requires an account ID be provided for other actions such as `signIn` and `signTransaction`, the dApp / consumer of this API would typically not know which account ID to use for the payloads for those methods unless they have called `connect()` previously to identify which account IDs are supported by the currently selected  wallet.

We do not explicitly _require_ that `connect` is _always_ called before the other methods to facilitate cases where the dApp might use cached data from a prior wallet connection call, or data from another external source to compose payloads for the other methods.

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: It was the simplest way to allow wallet implementors to be able to control which account would be 'first' in the list of accounts to sign in as or sign transactions for :)

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: :+1: Good call! :)

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+
+The introduction of this standard makes it possible for `near-api-js` to become wallet-agnostic and eventually move away from the high amount of coupling with NEAR Wallet. It simplifies projects such as [Wallet Selector](https://github.com/near/wallet-selector) that must implement various abstractions to normalise the different APIs before it can display a modal for selecting a wallet.
+
+This standard takes a different approach to a wallet API than other blockchains such as [Ethereum's JSON-RPC Methods](https://docs.metamask.io/guide/rpc-api.html#ethereum-json-rpc-methods). Mainly, it rejects the `request` abstraction that feels unnecessary and only adds to the complexity both in terms of implementation and types. Instead, it exposes various methods directly on the top-level object that also improves discoverability.
+
+There has been many iterations of this standard to help inform what we consider the "best" approach right now for NEAR. Below is a summary of the key design choices:
```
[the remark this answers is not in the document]
MaximusHaximus: :+1: fixed!

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,369 @@
...
+- Access to multiple accounts allow dApps more freedom to improve UX as users can seamlessly switch between accounts.
+- Aligns with WalletConnect via the [Bridge Wallet Standard](./BridgeWallets.md).
+
+### Store key pairs of FunctionCall access keys in dApp vs. wallet
+
+NEAR's unique concept of `FunctionCall` access keys makes it possible for dApps to execute transactions without having to prompt the user each time. This limited access means they can be shared with relatively low risk, provided gas allowance and method names are appropriately defined.
+
+There has been mixed views on exactly who should store these key pairs as there are trade-off to be considered with either option:
+
+- Storing key pairs in the wallet means users have a contextual view of which dApps have limited access to their accounts. However, this can be solved through improvements to the `view_access_key` endpoints to include a description and/or url that helps users make a more informed decision on whether they want to remove it.
```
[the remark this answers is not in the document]
MaximusHaximus: Good question... this paragraph was a bit confusing and the reference to metadata for access keys isn't directly relevant to this NEP.  Adding metadata to access keys would be a core chain NEP proposal which we would then integrate with this NEP where applicable.  This was alluding to the fact that _if_ the wallet held on to both types of keys, the wallet could show the user a list of keys that were used to sign in -- but that is possible _anyway_ without the wallet holding on to the private key in the pair -- for example, the near wallet UI displays all FunctionCall keys on an account without holding any of their private keys, and could track and display any metadata it wanted about keys that were adding by using the near wallet UI without ever knowing the private keys involved in each sign-in.  I'm re-writing this to be clearer :+1: thanks for the feedback

MaximusHaximus:
Hello everyone,

I've made some changes based on discussion in #370 and other feedback received.  Thank you all so much for your input; we really appreciate it.  @ori-near I believe this is ready to be scheduled for a vote at this time! 

@frol thanks for your moderation :)

ori-near:
Thank you @MaximusHaximus for reviewing this NEP! Based on your comment, it seems like this proposal is close to a decision. We are therefore scheduling the first Wallet Standards Working group meeting this week, where this NEP can enter the final voting stage. 

Next Step
@near/wg-wallet-standards Could you please fully read this NEP and comment in the thread with any feedback/concerns by tomorrow? Please specify if you are leaning with approving this NEP. If not, please specify why and what it would take for you to approve it. Thank you! 

Meeting Info
📅 Date: Friday, Oct 21, 5pm UTC
🗒️ Topic: NEP-408 Voting
✏️ [Register](https://near-org.zoom.us/webinar/register/5416617893723/WN_9zRso2njRN2wiv8iVx4I9g)

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,356 @@
...
+      [transactions.functionCall(
+        "addMessage",
+        { text: "Hello World! (2/2)" },
+        utils.format.parseNearAmount("0.00000000003"),
+        utils.format.parseNearAmount("0.01")
+      )],
+      utils.serialize.base_decode(block.header.hash)
+    )
+  ]
+});
```
esaminu, line 273:
```suggestion
import { transactions, providers, utils } from "near-api-js";

// Retrieve accounts (assuming already connected) and current network.
const { network, accounts } = window.near.wallet;

// Setup RPC to retrieve transaction-related prerequisites.
const provider = new providers.JsonRpcProvider({ url: network.nodeUrl });

const signedTxs = await window.near.wallet.signTransactions({
  transactions: [
    {
        signerId: accounts[0].accountId,
        receiverId: "guest-book.testnet",
        actions: [transactions.functionCall(
            "addMessage",
            { text: "Hello World! (1/2)" },
            utils.format.parseNearAmount("0.00000000003"),
            utils.format.parseNearAmount("0.01")
        )]
    },
    {
        signerId: accounts[0].accountId,
        receiverId: "guest-book.testnet",
        actions: [transactions.functionCall(
            "addMessage",
            { text: "Hello World! (2/2)" },
            utils.format.parseNearAmount("0.00000000003"),
            utils.format.parseNearAmount("0.01")
        )]
    }
  ]
}); 
```

esaminu (commented):
Updated examples and copy to match changed  `signTransaction` and `signTransactions` interfaces

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,356 @@
...
+
+type Unsubscribe = () => void;
+
+interface Wallet {
+  id: string;
+  connected: boolean;
+  network: Network;
+  accounts: Array<Account>;
+  
+  connect(params: ConnectParams): Promise<Array<Account>>;
```
lostpebble, line 105: _I'll copy my message from the previous issue ( https://github.com/near/NEPs/pull/370#discussion_r996990360 )- so we can keep the discussion here now._

Hi @MaximusHaximus - thanks for taking Network into consideration with the `connect()` method.

> @lostpebble Does this address your concerns here, or do you think that adding occupancy as an argument to the connect() method is still a requirement for this to be reasonable, keeping in mind that if we do add that option, wallet providers implementation and UI will become more complex.

I think its okay to leave it out- as long as its always okay for a user to only select a single account to connect with. If you think certain dApps might require the user to connect with multiple (or a certain amount) of accounts, for whatever reason, then having an `occupancy` option might be a good idea so the wallet can show the relevant UI. But I don't think should ever be the case?

This "multi-account" concept does complicate the Wallet UI a bit though. I guess in our case, when a Dapp calls the `connect({ network: "testnet" })` method, we'll show the user an interface that allows them to do one of the following:

* Select a single testnet account
* Select multiple specific testnet accounts
* Connect with all known wallet testnet accounts

We'll have to think of ways of streamlining such an interface.

And after that, I would suppose that future calls to `verifyOwner()` do not require any further prompts from the user (since they've already allowed the connection). But `signIn()` would run the specific sign-in flow for the account the dApp has chosen to use.

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,356 @@
...
+  accountsChanged: { accounts: Array<Account> };
+}
+
+interface ConnectParams {
+    networkId: string;
+}
+
+type Unsubscribe = () => void;
+
+interface Wallet {
```
lostpebble, line 99: As @PinkiNice and I mentioned in the previous NEP version (https://github.com/near/NEPs/pull/370#discussion_r987196550):

I think it could be good to also allow wallets to list their supported Networks. As this would allow dApps to avoid trying to `connect()` to wallets that don't support their app network.

It could be good to include a property in the Wallet interface such as:

```ts
interface Wallet {
  supportedNetworkIds: string[];
}
```

Or maybe something easier for the dApp to consume, like a function they can call:

```ts
type SupportsNetwork = (networkId: string) => boolean;

interface Wallet {
  // ...
  supportsNetwork: SupportsNetwork ;
}
```

esaminu:
With the examples updated per https://github.com/near/NEPs/pull/408#pullrequestreview-1144636376, I lean towards approving this proposal because it will allow dapps and tools like [near-api-js ](https://github.com/near/near-api-js) to make assumptions about extension wallet implementations and not have to support multiple different implementations at once. Especially because extension wallets have emerged as a popular browser wallet type but I'd love to see this generalized to other wallet types as well.

I'd love to see https://github.com/near/NEPs/pull/370#issuecomment-1254067800 discussed due to the url redundancy and also convenience aspect but I don't see it as a blocker

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,356 @@
...
+
+## Specification
+
+Injected wallets are typically browser extensions that implement the `Wallet` API (see below).  References to the currently available wallets are tracked on the `window` object. To avoid namespace collisions and easily detect when they're available, wallets must mount under their own key of the object `window.near` (e.g. `window.near.sender`).
+**NOTE: Do not replace the entire `window.near` object with your wallet implementation, or add any objects as properties of the `window.near` object that do not conform to the Injected Wallet Standard**
+
+At the core of a wallet are [`signTransaction`](#signtransaction) and [`signTransactions`](#signtransactions). These methods, when given a [`Transaction`](https://nomicon.io/RuntimeSpec/Transactions) instance from [`near-api-js`](https://github.com/near/near-api-js), will prompt the user to sign with a key pair previously imported (with the assumption it has [`FullAccess`](https://nomicon.io/DataStructures/AccessKey) permission).
```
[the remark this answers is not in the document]
MaximusHaximus: Good catch! <3

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: :+1: Thanks for the heads up -- I've added the `supportsNetwork` function to the standard; I like the idea of a function in this case rather than a static array as it feels more flexible and the dApp is probably going to know what networkId they are checking for support of.

I defined it as an async method that resolved to a boolean rather than the sync method you had sketched out, for flexibility on the wallet side for how it determines if a passed networkId is supported.  This should insulate us from needing to make a breaking API change or fragment the API if a wallet implementor needs to do something async 'behind the scenes' in order to to determine the result of this call.

specs/Standards/Wallets/InjectedWallets.md
MaximusHaximus: @lostpebble Yup -- It's totally fine for the wallet side implementor to just create a UI that only allows a single account to be selected -- they just need to return that single account back to the caller in an array so that the datatype is consistent and forward-compatible with the idea of multiple accounts being connected to a dApp from a single wallet instance.

If you want some inspiration as far as how this looks in action in a PoC, take a look at the WalletConnect v2 PoC that @lewis-sqa created.

He implemented a [basic 'wallet' UI](https://github.com/lewis-sqa/web-examples/tree/SQC-170/bridge-wallet-poc) and [associated guestbook dApp](https://github.com/near/wallet-selector/tree/SQC-170/bridge-wallet-poc) that supports multiple accounts on the connect() call (this is how WalletConnect works in all cases).

I'll mark this one resolved -- thanks so much for your feedback & opinions <3 👍

cjd425:
After speaking with many wallet builders across the ecosystem, I lean towards approving this proposal because: 

1. They would benefit from a generalized approach to connect many wallets to a given app.

2. It will give users more confidence in the security of extension wallets.

3. Teams will have more bandwidth to focus on differentiating their products since they wouldn't need to take the time to support multiple different implementations at once.

4. This NEP aligns with the transition to Wallet Selector.

5. It's in line with onboarding Ethereum developers who are used to multiple connected wallets.

6. There have been issues with dApps using near-api-js-sdk and which wallets they support as the transition to wallet selector becomes more prevalent.

Cameron-Banyan:
After speaking with many wallet builders across the ecosystem, I lean towards approving this proposal because: 

1. They would benefit from a generalized approach to connect many wallets to a given app.

2. It will give me more confidence in the security of extension wallets.

3. Teams will have more bandwidth to focus on differentiating their products since they wouldn't need to take the time to support multiple different implementations at once.

4. This NEP aligns with the transition to Wallet Selector.

5. It's in line with onboarding Ethereum developers who are used to multiple connected wallets.

6. There have been issues with dApps using near-api-js-sdk and which wallets they support as the transition to wallet selector becomes more prevalent.

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,325 @@
...
+
+Remove visibility of all accounts from the wallet.
+
+```ts
+await window.near.wallet.disconnect();
+```
+
+##### `signIn`
+
+Add `FunctionCall` access key(s) for one or more accounts. This request should require explicit approval from the user.
```
think-in-universe, line 262: How does the application get access to the created function call key after `signIn`?

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,325 @@
...
+        publicKey: keyPair.getPublicKey()
+      };
+    })
+  ),
+});
+```
+
+#### Events
+
+##### `accountsChanged`
```
think-in-universe, line 317: We may want to include more events such as `networkChanged`.

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -0,0 +1,325 @@
...
+- Generally, it is recommended to only keep `FullAccess` keys in wallet scope and hidden from the dApp consumer. `FunctionCall` type keys should be generated and owned by the dApp, and requested to be added using the `signIn` method. They should **not** be 'hidden' inside the wallet in the way that `FullAccess` type keys are.
+
+## Specification
+
+Injected wallets are typically browser extensions that implement the `Wallet` API (see below).  References to the currently available wallets are tracked on the `window` object. To avoid namespace collisions and easily detect when they're available, wallets must mount under their own key of the object `window.near` (e.g. `window.near.sender`).
+**NOTE: Do not replace the entire `window.near` object with your wallet implementation, or add any objects as properties of the `window.near` object that do not conform to the Injected Wallet Standard**
```
think-in-universe, line 41: What's advantage and disadvantage of mounting the `Wallet` object at `window.near` here?

I think developers who're familiar with Ethereum may prefer to use `window.near` as the root of the Wallet API, as long as they can switch among the providers.

specs/Standards/Wallets/InjectedWallets.md
lostpebble: I was wondering the same- but after looking through the interface, the dapp key's `publicKey` is set inside the `Account` data structure which is passed to the `signIn()` method.

frol (approved):
Thank you to everyone who attended the first Wallet Standards Working Group meeting today! The working group members reviewed the NEP and reached the following consensus:

**Status: Approved**

* [x] @Cameron-Banyan https://github.com/near/NEPs/pull/408#issuecomment-1286082231
* [x] @esaminu https://github.com/near/NEPs/pull/408#issuecomment-1284546176
* [x] @MaximusHaximus 

Meeting Recording: TBD

@lewis-sqa Thank you for championing the original NEP #370!

@MaximusHaximus Thank you for walking the last mile and bringing the NEP to the final line!

@lostpebble @andy-haynes Thank you for all the valuable inputs and reviews of this NEP!

Next steps:

* [x] @frol should merge NEP
* [ ] Align the implementation in the [Wallet Selector](https://github.com/near/wallet-selector) to match this NEP

andy-haynes (approved):

PR #411, opened 2022-10-11
Replace a noun with its verb

zlog-in:


frol:
@zlog-in Could you prepare a single PR with all the typos fixed, please? It is quite time-consuming to go through one fix at a time

PR #414, opened 2022-10-12
Nomicon: Change `state rent` to `storage staking`

idea404:
This PR changes the outdated concept in `FinancialTransaction.md`. 

Mentioned: https://github.com/near/docs/issues/345

### DIFF
--- a/specs/RuntimeSpec/Scenarios/FinancialTransaction.md
+++ b/specs/RuntimeSpec/Scenarios/FinancialTransaction.md
@@ -92,7 +92,7 @@ Specifically it does the following checks:
 - Retrieves the state of the access key of that `alice_near` used to sign the transaction;
 - Checks that transaction nonce is greater than the nonce of the latest transaction executed with that access key;
 - Subtracts the `total_cost` of the transaction from the account balance, or throws `InvalidTxError::NotEnoughBalance`. If the transaction is part of a transaction by a FunctionCall Access Key, subtracts the `total_cost` from the `allowance` or throws `InvalidAccessKeyError::NotEnoughAllowance`;
-- Checks whether the `signer` account has insufficient balance for storage staking and throws `InvalidTxError::LackBalanceForState` if so
+- Checks whether the `signer` account has insufficient balance for the storage deposit and throws `InvalidTxError::LackBalanceForState` if so
 - If the transaction is part of a transaction by a FunctionCall Access Key, throws `InvalidAccessKeyError::RequiresFullAccess`;
 - Updates the `alice_near` account with the new balance and the used access key with the new nonce;
 
@@ -111,4 +111,4 @@ It will check that this receipt does not have data dependencies (which is only t
 - Computes how much reward should be burnt;
 - Checks if `bob_near` still exists and if it does, deposits the transferred tokens to `bob_near`;
 - Checks if `bob_near` still exists and if so, deposits the gas reward to the `bob_near` account;
-- Checks if `bob_near` has enough balance for storage staking, the transaction's temporary state is dropped/rolled back if there is not enough balance;
+- Checks if `bob_near` has enough balance for the storage deposit, the transaction's temporary state is dropped/rolled back if there is not enough balance;


specs/RuntimeSpec/Scenarios/FinancialTransaction.md
```diff
@@ -110,10 +110,11 @@ If any of the above operations fail all of the changes will be reverted.
 The receipt created in the previous section will eventually arrive to a runtime on the shard that hosts `bob_near` account.
 Again, it will be processed by `Runtime::apply` which will immediately call `Runtime::process_receipt`.
 It will check that this receipt does not have data dependencies (which is only the case of function calls) and will then call `Runtime::apply_action_receipt` on `TransferAction`.
+
 `Runtime::apply_action_receipt` will perform the following checks:
 
 - Retrieves the state of `bob_near` account, if it still exists (it is possible that Bob has deleted his account concurrently with the transfer transaction);
-- Applies the rent to Bob's account;
+- Checks if `bob_near` has enough balance for storage staking;
```
jakmeier, line 117: this sounds a bit confusing to me
1. there is no change in storage, so the check is not needed for the example (of course it is still done in reality, just not very relevant to explain how a NEAR token transfer works)
2. In the code, the check is done after everything has been applied to the temporary state. The temporary state is dropped / rolled back if there is not enough balance. Maybe this level of detail is not relevant but still I would consider switching the order in this list, moving the storage staking check at the end.

specs/RuntimeSpec/Scenarios/FinancialTransaction.md
```diff
@@ -110,10 +110,11 @@ If any of the above operations fail all of the changes will be reverted.
...
 - Computes the cost of processing a receipt and a transfer action;
-- Checks if `bob_near` still exists and if it is deposits the transferred tokens;
-- Computes how much reward should be paid to the validators from the burnt gas.
+- Computes how much reward should be burnt as a means to indirectly reward validators
```
jakmeier, line 119: Not sure if "indirectly reward validators" is helpful here. It hints at a much larger topic and the terminology of "indirectly rewarding validators" is not well established AFAIK. I think just saying that we burn it is better here.

Also, you removed "deposits the transferred tokens" from the list, why is that? Seems like the most important step of a NEAR transfer action. :)

specs/RuntimeSpec/Scenarios/FinancialTransaction.md
```diff
@@ -110,10 +110,11 @@ If any of the above operations fail all of the changes will be reverted.
...
+- Checks if `bob_near` still exists and if so, deposits the gas reward to the `bob_near` account;
```
jakmeier, line 120: yes :+1:

jakmeier (commented):
Nice, important to clean up our naming here!

Can you do the same changes to the other list, that in the section **Transaction to receipt**?

> - Applies the storage rent to the alice_near account;
> ...
> - Checks whether after all these operations account has enough balance to passively pay for the rent for the next several blocks (an economical constant defined by Near Protocol). Otherwise account will be open for an immediate deletion, which we do not want;
> ...
> - Computes how much reward should be paid to the validators from the burnt gas.
>

Also, consider rephrasing this section a bit:
> Pre-requisites
>
> ...
> We also need to assume that both Alice and Bob have some number of tokens on their accounts. Alice needs >100 tokens on the account so that she could transfer 100 tokens to Bob, but also Alice and Bob need to have some tokens to pay for the rent of their account -- which is essentially the cost of the storage occupied by the account in the Near Protocol network.

I think *rent* can still be a good analogy. But maybe we should mention here that we call it "storage staking" because we are not actually taking away any tokens, we only keep the staking rewards they missed out on.

idea404:
> I think rent can still be a good analogy. But maybe we should mention here that we call it "storage staking" because we are not actually taking away any tokens, we only keep the staking rewards they missed out on.

I've updated the section mentioning the cost is refundable. I understood yday from Akhi the storage itself once reserved is not being staked. I find that saying the amount is being staked would thus be misleading. `Refundable deposit` would also work by this matter

specs/RuntimeSpec/Scenarios/FinancialTransaction.md
idea404: have updated section, I like to include detail myself, especially here considering it's Nomicon, but if you still feel this does not belong in this run-through I'm happy to remove!

specs/RuntimeSpec/Scenarios/FinancialTransaction.md
idea404: Missed that! Ty!

specs/RuntimeSpec/Scenarios/FinancialTransaction.md
jakmeier: yeah with this item at the end, it's probably good to keep it in

jakmeier:
> > I think rent can still be a good analogy. But maybe we should mention here that we call it "storage staking" because we are not actually taking away any tokens, we only keep the staking rewards they missed out on.
> 
> I've updated the section mentioning the cost is refundable. I understood yday from Akhi the storage itself once reserved is not being staked. I find that saying the amount is being staked would thus be misleading. `Refundable deposit` would also work by this matter

True, we are not staking it in the sense that it is being used for consensus. Somehow we still ended up using "storage staking" as the term for users paying for storage by locking up their tokens. But I agree with your reasoning.

In that sense, maybe we should not call it storage staking at all in this document? Maybe replacing the two occurrences with something like  "reserved balance for storage" would be clearer? But no strong opinion from my side, it's fine to keep it as it is or to change it.

frol (approved):
I rely on @jakmeier's review here, as he is code-owner for nearcore, so as a moderator, I am approving this change.

@idea404 Please, merge it once you are ready.

PR #417, opened 2022-10-19
NEP-417: Meta transactions

e-uleyskiy:
# Summary
This is a fork of [NEP-366](https://github.com/near/NEPs/pull/366). Main changes:
1. The implicit account case is described according to the comments [one](https://github.com/near/NEPs/pull/366#issuecomment-1278931208) and [two]( https://github.com/near/NEPs/pull/366#issuecomment-1279333217)
2. `DelegateAction` and `DelegateActionMessage` formats are changed.

*I will attach a diagram a little bit later.*

### DIFF
--- a/neps/nep-0417.md
+++ b/neps/nep-0417.md
@@ -1,8 +1,8 @@
 ---
-NEP: 366
+NEP: 417
 Title: Meta Transactions
 Author: Illia Polosukhin <ilblackdragon@gmail.com>, Egor Uleyskiy (egor.ulieiskii@gmail.com), Alexander Fadeev (fadeevab.com@gmail.com)
-DiscussionsTo: https://github.com/nearprotocol/neps/pull/366
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/417
 Status: Draft
 Type: Protocol Track
 Category: Runtime
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Changes to the protocol specification and standards are called NEAR Enhancement 
 |[0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md)   | Multi Token Standard | @zcstarr @riqi @jriemann @marcos.sun | Review |
 |[0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md)   | Events Standard | @telezhnaya | Final |
 |[0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md)   | Source Metadata | @BenKurrek | Review |
-|[0366](https://github.com/near/NEPs/blob/master/neps/nep-0000.md) | Meta Transactions | @ilblackdragon | Draft |
+|[0366](https://github.com/near/NEPs/blob/master/neps/nep-0366.md)   | Meta Transactions | @ilblackdragon @e-uleyskiy @fadeevab | Draft |
 
 
 
@@ -69,7 +69,7 @@ Spec changes are ultimately done via pull requests to this repository (formalize
     ```
 
    * Once complete, submit the pull request for editor review.
-  
+
    * The formalization dance begins:
      * NEP Editors, who are unopinionated shepherds of the process, check document formatting, completeness and adherence to [NEP-0001](neps/nep-0001.md) and approve the pull request.
      * Once ready, the author updates the NEP status to `Review` allowing further community participation, to address any gaps or clarifications, normally part of the Review PR.
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Changes to the protocol specification and standards are called NEAR Enhancement 
 |[0245](https://github.com/near/NEPs/blob/master/neps/nep-0245.md)   | Multi Token Standard | @zcstarr @riqi @jriemann @marcos.sun | Review |
 |[0297](https://github.com/near/NEPs/blob/master/neps/nep-0297.md)   | Events Standard | @telezhnaya | Final |
 |[0330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md)   | Source Metadata | @BenKurrek | Review |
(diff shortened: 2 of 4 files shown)


neps/nep-0417.md
```diff
@@ -0,0 +1,202 @@
...
+There are no requirements on the user account. The user account may not even exist on the chain and an implicit account can be used instead.
+
+An alternative is to a proxy contracts deployed on the user account.
+This design has severe limitations as it requires the user to deploy such contract and incur additional costs for storage.
+
+## Specification
+
+The main flow of the meta transaction will be as follows:
+ - User requests the relayer for its access key (public key) which the relayer are going to use. The user specifies this key as `relayer_pub_key` in `DelegateAction`.
+ - User fetches the relayer's access key `nonce`, increments it and assigns the value to `relayer_nonce` in `DelegateAction`: `relayer_nonce = nonce + 1`.
```
akhi3030, line 36: Will the relayer have one nonce per user?  If not, then a relayer can only work with one user at a time.  This will have lower throughput and a potential attack vector where the user can request the nonce from the relayer and never send the subsequent transaction.

neps/nep-0417.md
```diff
@@ -0,0 +1,202 @@
...
+ - User specifies `sender_id` (the user's account id) and `relayer_id` in `DelegateAction`.
+ - User signs `DelegateAction` specifying the set of actions that they need to be executed.
+ - User forms `SignedDelegateAction` with the `DelegateAction` and the `DelegateAction` signature.
+ - User forms `DelegateActionMessage` with the `SignedDelegateAction`.
+ - User sends `DelegateActionMessage` data to the relayer.
+ - Relayer verifies actions specified in `DelegateAction`: the total cost and whether the user included the reward for the relayer.
```
akhi3030, line 42: - Should we clarify how a user may include the reward given that the user may not have any NEAR tokens yet.
- Is there an attack vector where the relayer can keep the reward and not finish the transaction?  Will the user have any recourse there or will users generally have to trust relayers.

neps/nep-0417.md
```diff
@@ -0,0 +1,202 @@
...
+ - Relayer forms a `Transaction` with `receiver_id` equals to `delegate_action.sender_id` and `actions: [SignedDelegateAction { ... }]`. Signs it with its key. Note that such transactions can contain other actions toward user's account (for example calling a function).
+ - This transaction is processed normally. A `Receipt` is created with a copy of the actions in the transaction.
+ - When processing a `SignedDelegateAction`, a number of checks are done (see below), mainly a check to ensure that the `signature` matches the user account's key.
+ - When a `Receipt` with a valid `SignedDelegateAction` in actions arrives at the user's account, it gets executed. Execution means creation of a new Receipt with `receiver_id: AccountId` and `actions: Action` matching `receiver_id` and `actions` in the `DelegateAction`.
```
akhi3030, line 46: Should we clarify what we mean by "user's account" here?  If I understand things properly, users may not have accounts yet.

neps/nep-0417.md
```diff
@@ -0,0 +1,202 @@
...
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/417
+Status: Draft
+Type: Protocol Track
+Category: Runtime
+Created: 19-Oct-2022
+---
+
+## Summary
+
+In-protocol meta transactions allowing for third-party account to initiate and pay transaction fees on behalf of the account.
```
akhi3030, line 14: nit: I wonder if this feature should be called delegate transactions instead of meta transactions.  I feel like that is a more precise name.

akhi3030 (commented):
Thanks @e-uleyskiy.  This looks like a great start.  My comments are generally minor.  They should help someone less familiar with the proposal understand it better.

neps/nep-0417.md
e-uleyskiy: Actually the implementation of the relayer node is up to developers, I think. But if you want me to describe it in this NEP, I will do. Bellow are my thoughts:

A relayer should have one nonce per access key. And one access key per user. A relayer can import a lot of access keys to its account and schedule them between users.

> the user can request the nonce from the relayer and never send the subsequent transaction

The relayer can reserve an access key for some time. When this time ends, other users can use this key.
And one important thing, the user requests *access key/public key*, not nonce.

akhi3030: I agree that the implementation of the relayer node will be up to the developers.  And I suppose a NEP is not the right place to provide best practices for developers either.  However, I felt that fellow developers who are interested in building a relayer would be curious about this point.  For me personally, just this conversation should suffice and I don't think we need to add more text to the NEP.

fadeevab:
@e-uleyskiy Thanks for the NEP!

The following parts should be added into the NEP:
1. Security control: `signer_id` of implicit account **must be equal** to `delegate_action.public_key`.
2. **Use cases** into the Overview.
3. A **sequence diagram** describing the **Specification** chapter.
4. **Refund cases** for failed transactions.

ori-near:
Hey @e-uleyskiy, thank you for submitting this NEP. Since this is a continuation of https://github.com/near/NEPs/pull/366, could you please submit a pull request to change 366 and close 417? This way we can keep the content and discussion centralized in one NEP. Thank you very much.

e-uleyskiy:
@ori-near
Ok, when I fix all issues the reviewers have found, I will submit a pull request to the meta-tx branch.

e-uleyskiy:
Close this pull request. Need to discuss more about this approach.
https://github.com/near/NEPs/pull/366#issuecomment-1285406908
https://github.com/near/NEPs/pull/366#issuecomment-1285518017

ori-near:
Moved this to Rejected since this is a duplicate/continuation of https://github.com/near/NEPs/pull/366.

PR #418, opened 2022-10-19
Proposal to allow attached_deposit in view contexts

austinabell:
A small UX improvement for contract developers. Markdown https://github.com/near/NEPs/blob/austin/view_attached_deposit/neps/nep-0418.md

Original discussion: https://near.zulipchat.com/#narrow/stream/295306-pagoda.2Fcontract-runtime/topic/attached_deposit.20view.20error

### DIFF
--- a/neps/nep-0418.md
+++ b/neps/nep-0418.md
@@ -31,7 +31,7 @@ The alternative of returning some special value, say `u128::MAX`, is that it wou
 
 After the protocol version, the error inside `attached_deposit` will be skipped, and for all view calls, `0u128` will be set at the pointer passed in.
 
-## Reference Implementation (Required for Protocol Working Group proposals, optional for other categories)
+## Reference Implementation
 
 
 Currently, the implementation for `attached_deposit` is as follows:
--- a/neps/nep-0418.md
+++ b/neps/nep-0418.md
@@ -78,6 +78,26 @@ This has a breaking change of the functionality of `attached_deposit` and affect
 - The Rust SDK, as well as other SDKs, can add the `attached_deposit() == 0` check by default to all methods for safety of use.
 - Potentially, other host functions can be allowed where reasonable values can be inferred. For example, `prepaid_gas`, `used_gas` could return 0.
 
+## Decision Context
+
+### 1.0.0 - Initial Version
+
+The initial version of NEP-418 was approved by Tools Working Group members on January 19, 2023 ([meeting recording](https://youtu.be/poVmblmc3L4)).
+
+#### Benefits
+
+- This will allow contract SDK frameworks to add the `attached_deposit == 0` assertion for every function on a contract by default.
+- This behavior matches the Solidity/Eth payable modifier and will ensure that funds aren't sent accidentally to a contract in more cases than currently possible.
+- Given that there is no way of checking if a function call is within view context to call `attached_deposit` conditionally, this NEP only changes a small surface of the API instead of introducing a new host function.
+
+#### Concerns
+
+| # | Concern | Resolution | Status |
+| - | - | - | - |
+| 1 | Proposal potentially triggers the protocol version change | It does not trigger the protocol version change. Current update could be considered a client-breaking change update. | Resolved |
+| 2 | The contract can assume that `attached_deposit` will panic in view contexts. | The possibility that this is done _and_ has some value connected with a view call result off-chain seems extremely unlikely. | Won't Fix |
+| 3 | Can we assume that in all view calls, the `attached_deposit` in the VMContext always zero? | Yes, there is no way to set `attached_deposit` in view calls context | Resolved |
+
 ## Copyright
 [copyright]: #copyright
 
--- a/neps/nep-0418.md
+++ b/neps/nep-0418.md
@@ -39,15 +39,15 @@ The error inside `attached_deposit` for view calls will be removed, and for all 
 Currently, the implementation for `attached_deposit` is as follows:
 ```rust
 pub fn attached_deposit(&mut self, balance_ptr: u64) -> Result<()> {
-	self.gas_counter.pay_base(base)?;
+    self.gas_counter.pay_base(base)?;
 
-	if self.context.is_view() {
-		return Err(HostError::ProhibitedInView {
-			method_name: "attached_deposit".to_string(),
-		}
-		.into());
-	}
-	self.memory_set_u128(balance_ptr, self.context.attached_deposit)
+    if self.context.is_view() {
+        return Err(HostError::ProhibitedInView {
+            method_name: "attached_deposit".to_string(),
+        }
+        .into());
+    }
+    self.memory_set_u128(balance_ptr, self.context.attached_deposit)
 }
 ```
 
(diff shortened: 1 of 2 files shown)


mfornet:
I support this proposal. It can probably be extended to other methods like `predecessor_account_id,` where some sensible value can be returned. We were having [a similar issue in aurora-engine](https://github.com/aurora-is-near/aurora-engine/pull/477).

Btw, I don't think this requires a protocol version upgrade, given this only changes the behavior of view methods which can be done today by some clients without any impact on the network.

neps/nep-0418.md
```diff
@@ -0,0 +1,90 @@
...
+}
+```
+
+Which would just have to add the check:
+
+```rust
+pub fn attached_deposit(&mut self, balance_ptr: u64) -> Result<()> {
+	self.gas_counter.pay_base(base)?;
+
+	if self.context.protocol_version < VIEW_ATTACHED_DEPOSIT_PROTOCOL_VERSION && self.context.is_view() {
```
frol, line 58: As @mfornet mentioned, it seems that this proposal does not necessarily trigger a protocol change, which lowers the barrier for this NEP to be approved, I guess.

frol:
@near/wg-protocol I reviewed this NEP as a moderator and it looks good. Please, review and cast your decision in the comments, and let @near/nep-moderators when you want to move forward to the voting stage.

neps/nep-0418.md
```diff
@@ -0,0 +1,90 @@
...
+
+The rationale for assigning `0u128` to the pointer (`u64`) passed into `attached_deposit` is that it's the least breaking change. 
+
+The alternative of returning some special value, say `u128::MAX`, is that it would cause some unintended side effects for view calls using the `attached_deposit`. For example, if `attached_deposit` is called within a function, older versions of a contract that do not check the special value will return a result assuming that the attached deposit is `u128::MAX`. This is not a large concern since it would just be a view call, but it might be a bad UX in some edge cases, where returning 0 wouldn't be an issue.
+
+## Specification
+
+After the protocol version, the error inside `attached_deposit` will be skipped, and for all view calls, `0u128` will be set at the pointer passed in.
+
+## Reference Implementation (Required for Protocol Working Group proposals, optional for other categories)
```
frol, line 34:
```suggestion
## Reference Implementation
```

frol:
@mfornet 

>  It can probably be extended to other methods like predecessor_account_id, where some sensible value can be returned.

I cannot think of a sensible value for `predecessor_account_id` given it should return a valid AccountId.

neps/nep-0418.md
austinabell: Is that not a side effect we want to avoid? ie someone querying a function at a historical block and it giving inconsistent results back? Do we not have any guarantees about this?

mfornet:
> @mfornet
> 
> > It can probably be extended to other methods like predecessor_account_id, where some sensible value can be returned.
> 
> I cannot think of a sensible value for `predecessor_account_id` given it should return a valid AccountId.

One suggestion is to use `system`, and the other is to allow the user to pass the `predecessor_account_id` as an argument to the view call (no signatures required).

Outside of the scope of this PR but still relevant:

Ideally, we should be able to do dry runs of `change-calls` using `view-call`, where you can mimic every field (attached-deposit / attached-gas / predecessor-account-id / signer-account-id), and it should work even with storage mutation (just that mutations should not be stored). Allowing cross-contract calls in view mode would be good, but it is more challenging. It would require the client to be tracking shards from all touched contracts (plus, it requires extra logic to mimic block delays between cross-contract calls, probably).

neps/nep-0418.md
mfornet: This could be considered a client-breaking change update, but it isn't a protocol upgrade. If there were more than one client out there, it could be possible that some clients would implement these features in different ways.

I understand that making this change could break some assumptions about the code that has been written already that relies on current behavior. But even if you do a Protocol Upgrade, it won't solve the issue for them (users can't pin an older protocol version). 

IMO we should use the proposed behavior for legacy blocks as well, but I understand that if this means breaking too many apps, it is unacceptable. In that case, we can apply this behavior after some blocks. Even if the choice is to use this strategy after some fixed block, I still think no Protocol Version update is required (just the client version).

austinabell: hmm, I guess my intuition was that `view` semantics were built into the protocol. I guess that's technically not the case.

I'll change the proposal to not change behaviour based on network upgrade, and we can evaluate and try to minimize the risk of this known unknown and think of other potential risks.

ori-near:
@near/wg-protocol – As @frol mentioned above, we would like to move this NEP closer to consensus so we can schedule a working group meeting to make a decision.

@bowenwang1996 @mm-near @mfornet Could you please fully read this NEP and comment in the thread if you are leaning with approving or rejecting it? Please make sure to include your rationale and any feedback that you have for the author. Thank you.

ori-near:
As @frol mentioned above, we would like to schedule a working group meeting to review this NEP. Before we proceed, we would like to get closer to a consensus on this NEP.

@volovyks @DavidM-D @rsmith-osc Could you please fully read this NEP and comment in the thread if you are leaning with approving or rejecting it? Please make sure to include your rationale and any feedback that you have for the author. Thank you.

ailisp:
As a working group member, I lean towards approving this NEP based on the rationale and security concerns. Based on my knowledge on protocol and contract tooling, proposed solution is reasonably good addition to the tooling with trivial security implications to the protocol. The drawbacks is acceptable. The unsolved issue is also not a big problem.

volovyks:
As a working group member, I lean towards approving this proposal based on existing discussions in this PR, Zullip and rationale from @austinabell in the NEP itself. The breaking change created by this change is extremely unlikly to affect real contracts. At the same time I agree that it will improve DevEx.

Other proposals, like `predecessor_acount_id` should be discussed separately, since they can affect more users.

DavidM-D:
I also approve this proposal. It seems sensible and well thought through.

ori-near:
As the moderator, I would like to thank the author @austinabell for submitting this NEP, and for the @near/wg-tooling working group members for your review. Based on the voting indications above, it seems like this proposal is close to a decision. We are therefore scheduling the first Tools Working group meeting this week, where this NEP can enter the final voting stage. Anyone can discuss the technical details by adding your comments to this discussion. And join the call to learn about the final decision and how to get more involved.

Meeting Info
📅 Date: Thursday, Jan 19, 8 AM UTC
✏️ [Register](http://bit.ly/3W5abDF)

ori-near:
Thank you to those who attended the first Tools Working Group meeting yesterday! The working group members reviewed the NEP and reached the following consensus:

Status: Approved

-  @ailisp https://github.com/near/NEPs/pull/418#issuecomment-1370662862
-  @volovyks https://github.com/near/NEPs/pull/418#issuecomment-1372324830
-  @DavidM-D https://github.com/near/NEPs/pull/418#issuecomment-1375550347

Meeting Recording:
https://youtu.be/poVmblmc3L4

@austinabell Thank you for authoring this NEP!

Next Steps:
- [ ] @austinabell OR @near/nep-moderators To embed the discussed benefits & Concerns section into the NEP document
- [ ] @near/nep-moderators To approve and merge the NEP
- [ ] @near/nep-moderators To post on the NEAR Gigs Board a call for champions who are interested in implementing this NEP in the tooling

austinabell:
I updated the branch for https://github.com/near/nearcore/pull/7936#issuecomment-1402259938 but just can't push to the PR branch as it's a nearcore repo, and I don't have access. Not sure if anything else is blocking that coming in now that this seems "accepted"

frol (approved):

frol:
@austinabell Thanks for working on the NEP and PR to nearcore!

Congrats everyone with one more NEP being merged! 🎉

PR #424, opened 2022-11-03
Eliminate TTN Gas Cost

jakmeier:
Read the NEP rendered on GH: https://github.com/jakmeier/NEPs/blob/nep-424/neps/nep-0424.md

Governance forum discussion: https://gov.near.org/t/proposal-predictable-storage-gas-costs-removing-touched-trie-node-cost/30191


neps/nep-0424.md
```diff
@@ -0,0 +1,314 @@
...
+- Write per key byte costs need to be increased to at least 1 Ggas. This is
+  noticeable to users but should be acceptable.
+
+TODO: Include examples for how users are affected by the change.
+
+## Deploying the Feature
+
+Unfortunately, nearcore is not ready yet to support such low gas costs for
+writes. For now, it still has to do an expensive trie traversal for every write
+before it continues smart contract execution. This will be optimized in the
```
mm-near, line 230: I'm a little bit surprised by this statement - as in theory, this traversal is as expensive as the one during read, right?

jakmeier: Yes, traversal costs the same for read and write. This is based on the assumption that flat storage for read is complete and for write is is not.

We have some details left to do for reads as well but it will be ready for the same release that this proposal could possibly be included anyway. For writes, it will be significantly more work, so we want to keep doing the expensive traversal for writes.

jakmeier:
Yeah this is a draft and we are still figuring out crucial details before we can finalize the proposal.

I thought the moderators will ignore it until I set it as ready for review. It is pending details on flat storage implementation performance, which ultimately should guide specific gas numbers proposed here. These numbers are the most interesting part of it and while it remains unclear, I would recommend the WG to ignore this.

Regarding time line, I cannot guarantee it will be ready in January but it might be. February seems like a safer bet. I will ping as requested once it is ready.

ori-near:
Hi @jakmeier – we just wanted to revisit this NEP since we are nearing the end of February. Please let us know if you still plan to continue working on it and ping the @near/nep-moderators once you are ready for us to review it.  Thank you.

PR #399, opened 2022-09-29
Flat Storage

mzhangmzz:
Proposal of Flat Storage to improve state storage reads, which aims to make state storage more stable and make fees more predictable, and sketch the plan for future improvements.

PR #351, opened 2022-04-08
feat: add `standards` key to `ContractSourceMetadata`

esaminu:
This PR proposes the addition of an optional `standards` value to the `ContractSourceMetadata` type recently introduced in [NEP-330](https://github.com/near/NEPs/blob/master/neps/nep-0330.md) (#330). 

This will allow wallets and Dapps to be able to more easily identify what type of contract they are dealing with and what functionality to expect and render without having to test for those methods and events existing. 

One example use case of this is to introduce an extension that implements `buy` / `sell` for the [USN](https://github.com/binary-star-near/usn) contract and add it as `ContractSourceMetadata` so we can provide the functionality conditionally for fungible tokens in near/near-wallet#2568.

Additionally, this PR fixes missed updates of `ContractMetadata` to `ContractSourceMetadata` [here](https://github.com/near/NEPs/blob/master/neps/nep-0330.md?plain=1#L28) and [here](https://github.com/near/NEPs/blob/master/neps/nep-0330.md?plain=1#L84).

PR #428, opened 2022-11-14
Add support for signing in to more than one contract at once to Injected Wallet Standard

amirsaran3:
Added `signInMulti` method to solve the case where the dApp has more than one contract.

## NEP Status *(Updated by NEP moderators)*

SME reviews:

* [x] Wallet Standards SME: @joe-rlo https://github.com/near/NEPs/pull/428#issuecomment-1489552748
* [x] Wallet Standards SME: @andy-haynes  https://github.com/near/NEPs/pull/428#issuecomment-1490745864

Wallet Standards WG voting indications:

* [x] @Cameron-Banyan 
* [x] @MaximusHaximus 
* [x] @esaminu


ori-near:
Thank you @amirsaran3 for submitting this NEP.

As a moderator, I reviewed this NEP extension and it meets the proposed template guidelines. I am moving this NEP to the REVIEW stage and would like to ask the @near/wg-wallet-standards members to assign 2 Reviewers (Subject Matter Experts).

specs/Standards/Wallets/InjectedWallets.md
```diff
@@ -64,7 +64,7 @@ interface Network {
 }
 
 interface SignInParams {
-  permission: transactions.FunctionCallPermission;
+  permissions?: Array<transactions.FunctionCallPermission>;
```
MaximusHaximus, line 67: 1. What is the expected behaviour from a wallet when this method is called without a `permissions` property?

2. If a caller submits a payload matching the existing specification, with the non-pluralized shape `{ permission, accounts }`, what do we expect a wallet that conforms to this updated specification do?

3. This will be a breaking change for all wallets that have already implemented the specification with `permission` as the property name rather than `permissions`.  If a caller implements the spec as you are proposing in this PR, but they submit it to a wallet that has not yet updated their interface to be aware of the new pluralized property name, what will that wallet do?

MaximusHaximus (changes requested):
Just a few questions :)

specs/Standards/Wallets/InjectedWallets.md
amirsaran3: 1. The wallet will not create any function call access keys. Just like with the MyNearWallet `requestSignIn` method optionally takes in `contractId` as a parameter. It its not set no function call access keys will be created.
2. We can create a check for both specs for now.
3. I'm assuming that wallets are still in the progress of updating their interfaces to align with this NEP. The faster we update them about more changes it will better. This will also be a problem in for dApps after wallets implement the new interface. They will have to update their way of communicating with the wallet (unless they are using wallet-selector). It is recommended to use the wallet-selector for interacting with wallets so just updating wallet-selector interface will be enough. I'm not sure about this one though. Does this make sense?

MaximusHaximus:
@amirsaran3

1. Per our last meeting, please remove the portion of this PR that makes network change related modifications, especially given that the current communication flow is driven _from the dapp_ rather than _from the wallet_, and that account IDs cannot be the same on `testnet` and `mainnet` due to the fact that the top-level accounts are different and the same seed phrase used on testnet will result in a different implicit account than that same seed phrase used on mainnet (due to BIP39)

2. Unfortunately, this change as written will be a breaking change to a very low-level/fundamental API that all dapps use; we can't control how quickly dapps upgrade their wallet-selector implementation, nor can we control the speed at which wallets support changes to the API, so we need to make this a non-breaking change.
    - Instead of making a breaking change to the existing API, I suggest that we propose an entirely new method be added to the API -- my suggestion would be something like `signInMulti({ accounts: Array<string>, permissions: Array<transactions.FunctionCallPermission> })`. By making this a separate method, we gain two benefits:
      - We can easily detect support for the new API 
      - We can ensure that this functionality does not represent a breaking change for existing wallets

amirsaran3:
> @amirsaran3
> 
> 1. Per our last meeting, please remove the portion of this PR that makes network change related modifications, especially given that the current communication flow is driven _from the dapp_ rather than _from the wallet_, and that account IDs cannot be the same on `testnet` and `mainnet` due to the fact that the top-level accounts are different and the same seed phrase used on testnet will result in a different implicit account than that same seed phrase used on mainnet (due to BIP39)
> 2. Unfortunately, this change as written will be a breaking change to a very low-level/fundamental API that all dapps use; we can't control how quickly dapps upgrade their wallet-selector implementation, nor can we control the speed at which wallets support changes to the API, so we need to make this a non-breaking change.
>    
>    * Instead of making a breaking change to the existing API, I suggest that we propose an entirely new method be added to the API -- my suggestion would be something like `signInMulti({ accounts: Array<string>, permissions: Array<transactions.FunctionCallPermission> })`. By making this a separate method, we gain two benefits:
>      
>      * We can easily detect support for the new API
>      * We can ensure that this functionality does not represent a breaking change for existing wallets

I've made some updates.

The only change to the `signIn` method will be that the permission will be optional.

I've created a new `signInMulti` method that will have permissions as an argument and that will take in an array.

amirsaran3:
@MaximusHaximus I updated the permission parameter.

nearbuild:
Does this need the wallet working group to assign 2 SMEs?

frol:
@nearbuild Yes, I believe so.

As a moderator, I would like to ask the @near/wg-wallet-standards working group members to assign 2 Technical Reviewers to complete a technical review (see expectations). If you want to assign yourself, please mention that you are acting as the Technical Reviewers.

<details>

Technical Summary guidelines:

```

## Example

### Recommendation

Add recommendation

### Benefits

* Benefit

* Benefit

### Concerns

| - | - | - | - |   

```

</details>

esaminu:
As a working group member, I’d like to nominate @andy-haynes as a Subject Matter Expert to review this NEP.

pvolnov:
We already maintain previous multi-transactions standard in HERE Wallet
why we need new one?

```
{
    "transactions": [
        {
            "actions": {
                    "type": "function call",
                    "params": {
                        "gas": 30000000000000,
                        "method_name": "hi",
                        "args":"",
                        "deposit": 1
                    },
                },
            "receiverId": "bob.near",
        }
    ],
    "network": "mainnet",
}
```

https://docs.herewallet.app/tools-for-dapps/universal-sign-links

pvolnov:
Moreover, [this](https://docs.herewallet.app/tools-for-dapps/universal-sign-links)  standard does not require a `signer_id` and dApps will be able to get rid of mandatory authorization

Cameron-Banyan:
As a Working Group member, I would like to nominate Edward Chew (@edwardchew97) from Meteor Wallet to be a Subject Matter Expert to review this proposed NEP update.

andy-haynes:
> We already maintain previous multi-transactions standard in HERE Wallet
why we need new one?

@pvolnov is there an NEP for this?

The proposal in this PR seeks to extend an existing API in a specific use case, providing a straightforward way for dApps to continue using the `signIn`-based approach. If there are viable alternatives for dApps seeking a solution to this then those would ostensibly be options for the dApp developer as well.

edwardchew97:
### Technical Summary

This NEP update proposes adding a new `signInMulti` method that allows creating function call access keys for multiple smart contracts at once. Currently, we have a `signIn` method but it only allows single smart contract signIn. 

It is important to note that `signIn` and `connect` are different, with `signIn` generating a function call access key that may allow a DAPP to sign non-payable transactions on behalf of users.


### Recommendation
I endorse the approval of this update.


### Benefits
1. The `signInMulti` method will enable a more seamless user experience for dapps that require function call access keys for multiple smart contracts, which is likely to become more common with the introduction of BOS and gateways.
2. This update will provide a better developer experience by eliminating the need for developers to initiate multiple transactions and check their output one by one.


### Concerns
| - | - | - | - |   
| 1 | Backward compatibility. As @**[MaximusHaximus](https://github.com/MaximusHaximus)** suggested, the update should not break existing dapps.| The existing `signIn` method should not be updated| Resolved.|
| 2 | Near Wallet selector's `signIn` implementation takes only one account and current implementation allows an array of accounts. This create inconsistency between the standard and implementation.| I recommend that the `signInMulti` method take in only one account, as it is uncommon for users to want to create function call keys for multiple accounts at once. Creating multiple keys for multiple accounts is complex to display and cost a higher storage fees. | This is currently also an issue with `signIn` method and not being addressed yet but it might be better to address them in another proposal. |
|3|A malicious website could request `signIn` or `signInMulti` for user's fungible token smart contract and drain user's wallet by calling `ft_transfer`. This is more likely to happen if the wallet doesn't interpret and show the target smart contract properly.| Wallets should display contract ids properly during sign in and potentially detect if they are FT or NFT contract by checking if certain key methods are implemented.| This doesn't fall under the responsibility of this update but the update introduce a bigger risk. We might need to push this to be implemented on wallet selector side.

andy-haynes:
### Summary

The introduction of a signInMulti method to the injected wallet interface enables dApp developers to facilitate logins for multiple contracts in a single transaction, functionality not currently supported by the signIn method.


### Recommendation

I recommend the NEP to be approved.


### Benefits

DApp developers using a multi-contract architecture would be able to present a near-identical UX to wallets, providing a more simple and reliable login workflow.


### Concerns

I have nothing to new to add without restating the cogent points raised by @edwardchew97

2) I agree with the assessment that this is an awkward signature inherited by the existing `signIn` method and that it would make sense to address this in the context of operations taken against individual accounts as opposed to a set of accounts.

3) This is an interesting point on the FT contract scenario that I had not considered. Wallets having some awareness of the contracts deployed to accounts it manages certainly sounds like it would mitigate some risk around this, but I agree that it is part of a broader topic not directly relevant to this NEP.

ori-near:
As the moderator, I would like to thank @edwardchew97 and @andy-haynes for completing the technical review. Based on your comments above, it seems like this proposal is ready for the working group members for review. @near/wg-wallet-standards  – Can you please fully read this NEP and comment in the thread if you are leaning towards approving or rejecting it? Please make sure to include your rationale and any feedback that you have for the author. Once we get 2/3 voting indications, we will schedule a public call for the author to present the NEP and for the working group members to formalize the voting.

Cameron-Banyan:
As a Wallet working group member, I approve this NEP because it enables developers to specify which specific methods on smart contracts to permit and it will give users the option to opt-in to a single non-payable transaction to skip approval for different function cal access key. 

Based on previous discussions:
- The NEP addresses the concerns of backward compatibly with no breaking changes.

- As addressed by @edwardchew97 - although it is uncommon for users to want to create function calls keys for multiple accounts, I believe the option to do so can enable use cases for those who are interacting with multiple accounts at the same time for something like financial transactions. (ie: approving two accounts for HideYourCash + Rainbow Bridge).

- Although it will cost storage fees and be more complex to display, wallets will have **the option** to display this alongside contract_ids and the type of contract by detecting certain methods or through the passed NEP351 (https://github.com/near/NEPs/pull/351). I would appreciate if a reference implementation on displaying signInMulti with multiple function access keys on multiple accounts, contract detection, and gas estimation to show Wallet Builders to increase adoption of this NEP.

- If the design choice for just 1 account is favored, then I agree with @edwardchew97 that this be addressed in a separate proposal.

MaximusHaximus:
### Summary
The introduction of a signInMulti method to the injected wallet interface enables dApp developers to facilitate logins for multiple contracts in a single transaction, functionality not currently supported by the signIn method.

### Recommendation
Overall, recommend the NEP to be **approved**, but I do think that it should be modified slightly from its current proposal, in particular to make the signIn request specific to a _single_ account rather than multiple accounts.

### Benefits
DApp developers using a multi-contract architecture would be able to present a near-identical UX to wallets, providing a more simple and reliable login workflow.

### Concerns
#### 2. Near Wallet selector's signIn implementation takes only one account and the current implementation allows an array of accounts. This creates inconsistency between the standard and implementation.
I believe this is a valid concern -- the UI a wallet would need to display would be very complex and the flow for approving multiple transactions (1 per account) has a lot of possible failure modes compared to a single account.  Furthermore, it makes the concept of being signed in on the _dapp_ side far more complex, since you could be signed into more than 1 account with different contracts for each account, and when the limited-access-key allowance is exhausted for one of those contract <-> account pairings, the dapp and wallet UI experience for replenishing just that one key for one account will be made more complex also; atomicity in signing in per account makes all of these challenges easier to handle.

#### 3. A malicious website could request signIn or signInMulti for users' fungible token smart contract and drain their wallets by calling ft_transfer. This is more likely to happen if the wallet doesn't interpret and show the target smart contract properly.
I believe this isn't a big problem, since signing in doesn't create a _full access key_ (FAK), but instead a _limited access key_  (LAK) -- according to NEP141 an NEP171 specifications, you must attach a `deposit` of 1 yoctoNEAR in order to call `ft_transfer` or any other methods that can cause ownership of an asset to change, which _forces_ a FAK to be used to sign any transactions that are dangerous.  Because of this requirement, such transactions must be done via a wallet which should be programmed to show the user the action they are approving before signing it.


#### Barriers to implementation / changes in concerns for dapp developers
I'd like to bring attention to two knock-on effects that need to be considered:

A. Keystores - Currently, no keystores implemented by the NEAR JS ecosystem allow multiple keys to be stored for the same account ID at once; LocalStorageKeystore only contains a mapping of `accountId`->`Key`.  In order to allow dapps to actually seamlessly sign transactions for more than 1 key on the same account without hacking their own persistence and keystores, we will need a new keystore that can store keys in a mapping of `accountId + contractId` -> `Key`.

B. Limited Access Key Possible Bloat - Currently, when the LAK that a dapp has been using as a result of a previous sign in has run out of available allowance, the dapp developer must catch the `NotEnoughAllowance` error and prompt the user to sign in to their dapp again.  With the introduction of _multiple keys_, this could lead to 'key bloat' if the dapp developer always requests the user re-sign in for _all_ of the contracts that they had originally been issued LAKs for.  Since it is highly unlikely all of the keys will run out of allowance at once, signing in again for all contracts will probably leave keys behind with allowances on them that are not exhausted, which may confuse the user.

ori-near:
As the moderator, I would like to thank the @amirsaran3 for submitting this Wallet Standards NEP, and for the @near/wg-wallet-standards working group members for your review. Based on the voting indications above, it seems like this proposal is close to a decision. We are therefore scheduling a Wallet Standards Working group meeting this week, where this NEP can enter the final voting stage. Anyone can discuss the technical details by adding your comments to this discussion. And join the call to learn about the final decision and how to get more involved.

**Meeting Info**
📅 Date: Thursday, April 20, 5pm UTC
✏️ [Register](https://bit.ly/3UG9iSE)

@amirsaran3 – please let me know if you'll be able to make the call and the best way to contact you to share the detailed agenda.

esaminu:
As a wallet working group member, I lean towards approving this NEP.

### Benefits
This NEP will optimize UX for multi contract DApps and avoid multiple redirects. These are more and more common in the ecosystem and this NEP will benefit the UX for those DApps.

### Concerns

- The currently available keystores will have to catch up in order to support multiple keys per account
- We should add the new method to [the `Wallet` interface](https://github.com/amirsaran3/NEPs/blob/patch-1/specs/Standards/Wallets/InjectedWallets.md?plain=1#L104) for clarity in the NEP doc

esaminu:
Thanks for the submission @amirsaran3! 

Could you make the following modifications to this NEP?

1. Make `signInMulti` and `signIn` take a single `Account` instead of `Array<Account>`
2. Add `signInMulti` method to [the `Wallet` interface](https://github.com/amirsaran3/NEPs/blob/patch-1/specs/Standards/Wallets/InjectedWallets.md?plain=1#L104)
3. Embed the Benefits & Concerns section to the NEP document for future reference

Then we should be ready to land this!

ori-near:
The Wallet Standards Work Group met today to review this NEP extension and reached the following consensus:

**Status:** Approved

- 👍 @Cameron-Banyan: https://github.com/near/NEPs/pull/428#issuecomment-1507875733
- 👍 @MaximusHaximus: https://github.com/near/NEPs/pull/428#issuecomment-1508916315
- 👍 @esaminu: https://github.com/near/NEPs/pull/428#issuecomment-1512376919

Meeting Recording:
https://youtu.be/l-cUuFCylKY

Thank you @amirsaran3 for authoring this NEP! And for all the participants who were involved during the review process.

Next Steps:
-  @amirsaran3 Please complete the steps @esaminu outlined [above](https://github.com/near/NEPs/pull/428#issuecomment-1517063567)  
- [TBD] Add supporting keystore to NAJ  
- [TBD] Add wallet selector interface

amirsaran3:
> Thanks for the submission @amirsaran3!
> 
> Could you make the following modifications to this NEP?
> 
> 1. Make `signInMulti` and `signIn` take a single `Account` instead of `Array<Account>` in the argument
> 2. Add `signInMulti` method to [the `Wallet` interface](https://github.com/amirsaran3/NEPs/blob/patch-1/specs/Standards/Wallets/InjectedWallets.md?plain=1#L104)
> 3. Embed the Benefits & Concerns section to the NEP document for future reference
> 
> Then we should be ready to land this!

I updated the PR.

nearbuild:
Whats the status on this @frol thought it was voted on?

frol (approved):

kujtimprenkuSQA:
Hi, we have worked on adding the `signInMulti` to the API of the Wallet Selector, we would appreciate if the contributors on this NEP update could take a look/review the PR in Wallet Selector: https://github.com/near/wallet-selector/pull/811 so that we're all on the same page, Thanks.

PR #432, opened 2022-11-16
Update NEP-366

e-uleyskiy:
This update to NEP #366 contains the following changes
- Fixed broken links
- Added more error messages
- Added more details to the text

### DIFF
--- a/neps/nep-0366.md
+++ b/neps/nep-0366.md
@@ -115,13 +115,18 @@ The permissions are verified based on the variant of `public_key`:
 
 #### Errors
 
-**Validation Error**
+- If the Sender's account doesn't exist
+```rust
+/// Happens when TX receiver_id doesn't exist
+AccountDoesNotExist
+```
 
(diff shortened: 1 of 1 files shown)


frol:
@near/wg-protocol Given that NEP #366 is in the final comment period stage, please, review this PR and merge it (it is based on the PR #366, so it won't be merged to master, instead, it will be merged to the PR #366 branch)

jakmeier (commented):
@e-uleyskiy it looks like your branch contains already squashed and merged commits. Can you please rebase or cherry-pick the new commits on top of the meta-tx:HEAD? The diff is a bit hard to review like this. I looked through the new changes commit by commit to avoid it but it would be better to only have the new commits in.

@ori-near This looks all good and necessary to me. No significant changes in terms of the proposal, but added meaningful details. Just as a reminder, I'm not part of the WG but since you asked me earlier today, my recommend here is to accept and merge this PR.

PR #439, opened 2022-12-02
chore: reformat transaction validation section

pugachAG:
Bullet list layout indentation doesn't seem to look nice with multi-line code snippets, especially for the last section. In this PR I propose to use subtitles instead. This also includes a couple of minor fixes such as typos and rewording of signature validation description (to avoid potential confusion that transaction is actually signed by `public_key`).

### DIFF
--- a/specs/RuntimeSpec/Transactions.md
+++ b/specs/RuntimeSpec/Transactions.md
@@ -53,63 +53,91 @@ and potentially errors returned when they fail.
 
 ### Basic Validation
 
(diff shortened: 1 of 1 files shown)


ori-near (approved):
Hi @pugachAG – thank you for submitting this reformatting proposal. 

As the moderator, I noticed that @akhi3030, who is the Subject Matter Expert for the Protocol, approved it. Given that your proposal is only changing the styling but not the actual meaning of the specification, we are happy to merge it.

PR #442, opened 2022-12-13
fix: typos and confusing wording in meta transaction spec

austinabell:
Some things I noticed when looking over the meta transactions spec. I have no opinion if this comes in or not, but might make this more clear.


frol (approved):

frol:
@austinabell Thanks for proof-reading it and submitting the PR! 🚀

PR #444, opened 2022-12-21
Remove specs/Economics/README.md

mzhangmzz:
Remove the old specs/Economics/README.md file since it is migrated to specs/Economics/Economics.md. It is confusing to have both files. A few NEPs only made changes to one of them, this PR also brings Economics.md up to date


frol:
@mzhangmzz Thanks for aligning nomicon with the actual implementation! Can you recommend someone to review this change?

As a moderator, I approve this change, but I cannot attest if the information is accurate.

bowenwang1996:
@frol could you approve the PR so that we can merge it? Thanks!

frol (approved):

PR #447, opened 2023-01-10
doc: Fix gas price calculation formula and ADJ_FEE

akashin:
As according to the https://github.com/near/nearcore/blob/master/nearcore/src/config.rs#L139 and the code, the actual formula and value are different


frol (approved):

frol:
@akashin Thanks for aligning the nomicon with the actual implementation!

Issue #415, opened 2022-10-13
[Discussion] Removing storage staking for base account information


ilblackdragon:
TLDR: 
  - Allow to create accounts with 0 $NEAR balance. 
  - This would unblock Meta Tx NEP and wallets to create named accounts without exploitation

## Problem

Currently there is a blocker with #366 which spans from the problem that to initialize account that doesn't exist yet on the blockchain, one needs to deposit $NEAR into it prior to other actions happening.

Similar problem exist broadly for all account actions, where someone needs to deposit $NEAR into the account either via call to `.near` or via transfer to implicit address.

Historically we have been trying to allow people to create easily named accounts and were distributing some amount of $NEAR via wallet.near.org for newly created accounts. Due to ability for user to remove such account later, it led to a massive drain of the faucet by bots who would try to collect these free tokens.

New applications that are launching and want to initialize accounts for their users also face this issue. E.g. SweatCoin has been initializing accounts with $NEAR due to this and lack of meta transactions.

## Context

Storage staking was designed for NEAR to have a sustainable storage economics. Observing Ethereum we saw continues growth in storage usage which didn't account in the economics. 

Ethereum devs themself have been exploring other options of dealing with this, including storage rent. Early on we have implemented storage rent but it led to more issues due to all the tooling needing to understand continuously reducing amounts and impact on the user.

Storage staking was proposed as an alternative that also was designed to align economics of projects with broader NEAR economy. As projects grow and take more space on the blockchain they would need to lock more $NEAR tokens, reducing liquidity and amount available for staking in turn paying to validators.

Although it is still valid point, a deeper understanding of token economics lead to realization that growth of usage (and especially NEAR locked in various DeFi protocols or usage as a medium of exchange) benefits the protocol's economics way more than locking NEAR.

Observing the developers on NEAR, we also see two cases:
 - developers building correct storage staking model, which users don't appreciate and see as higher cost. E.g. Ref Finance requires users to deposit storage when they interact with application, but nobody ever released storage back, recovering the funds
 - developers forgo managing storage, having simplified codebase and just covering storage costs from their pocket (Aurora & Tonic), focusing on growing usage first.

## Idea

The proposal here is to forgo storage costs of the account information.

Specifically:
 - CreateAccount would cost more and burn tokens for storage via gas
 - Storage amount is 0 after account is created, not requiring to have NEAR balance
 - DeleteAccount can be considered for deprecation, as it's not useful anymore

[...]

PR #448, opened 2023-01-11
zero balance account NEP

bowenwang1996:
NEP for #415. Implementation can be found in https://github.com/near/nearcore/pull/8378

### DIFF
--- a/neps/nep-0448.md
+++ b/neps/nep-0448.md
@@ -64,7 +64,26 @@ Details of the changes described in the section above:
 },
 ```
 * Change the implementation of `get_insufficient_storage_stake` to take into account the number of keys an account has
-and whether it already has a contract deployed.
+and whether it already has a contract deployed. More specifically, the check looks roughly as follows:
+```rust
+/// Returns true if an account is a zero balance account
+fn check_for_zero_balance_account(account_id: &AccountId, state_update: &TrieUpdate) -> bool {
+    let account = state_update.get(&account_id);
+    // check whether the account has a contract deployed
+    if account.code_hash != CryptoHash::default() {
+        return false;
+    }
+    // check whether the account has contract data
+    if has_contract_data(account_id, state_update) {
+        return false;
+    }
+    let keys = get_all_access_keys(account_id, state_update);
+    if keys.len() > 3 {
+        return false;
+    }
+    true
+}
+```
 
 ## Drawbacks (Optional)
 
--- a/neps/nep-0448.md
+++ b/neps/nep-0448.md
@@ -13,7 +13,7 @@ Created: 10-Jan-2023
 A major blocker to a good new user onboarding experience is that users have to acquire NEAR tokens to pay for
 their account. With the implementation of [NEP-366](https://github.com/near/NEPs/pull/366), users don't necessarily have
 to first acquire NEAR tokens in order to pay transaction fees, but they still have to pay for the storage of their account.
-To address this problem, we propose allowing each account to have free storage for the account itself and up to three keys
+To address this problem, we propose allowing each account to have free storage for the account itself and up to four keys
 and account for the cost of storage in the gas cost of create account transaction. 
 
 ## Motivation
@@ -25,7 +25,8 @@ browser and directly starts using it without worrying about registration. Under 
 user and the application creates an account for the user and pays for transaction fees through meta transactions. Later on,
 the user may find other applications that they are also interested in and give them a try as well. At some point, the user
 graduates from the onboarding experience by acquiring NEAR tokens either through earning or because they like some experience
-so much that they would like to pay for it explicitly.
+so much that they would like to pay for it explicitly. Overall we want to have two full access keys for recovery purposes
+and two function call access keys so that users can use two apps before graduating from the onboarding experience.
 
 ## Rationale and alternatives
 
@@ -41,15 +42,22 @@ and more costly for users (around 0.006N per account creation).
 
 There are two main changes to the protocol:
 * Account creation cost needs to be increased. For every account, at creation time, 600 bytes of storage are reserved
-for the account itself + three keys. For function call access keys, the "free" ones cannot use `method_names` in order to
-reserve space. The cost of these bytes is paid through transaction fee. Note that there is already [discussion](https://github.com/near/NEPs/issues/415)
+for the account itself + two full access keys + two function call access keys. For function call access keys,
+the "free" ones cannot use `method_names` in order to minimize the storage requirement for an account.
+The number of bytes is calculated as follows:
+  - An account takes 100 bytes due to `storage_usage_config.num_bytes_account`
+  - A full access key takes 42 bytes and there is an additional 40 bytes required due to `storage_usage_config.num_extra_bytes_record`
+  - A function call access key takes 126 bytes and there is an additional 40 bytes required due to `storage_usage_config.num_extra_bytes_record`
+  - Therefore the total the number of bytes is `100 + (126 + 40) * 2 + (42 + 40) * 2 = 596`.
+
+  The cost of these bytes is paid through transaction fee. Note that there is already [discussion](https://github.com/near/NEPs/issues/415)
 around the storage cost of NEAR and whether it is reasonable. While this proposal does not attempt to change the entire
 storage staking mechanism, the cost of storage is reduced in 10x when converting to gas. A [discussion](https://gov.near.org/t/storage-staking-price/399)
 from a while ago mentioned this idea, and the concerns there were proven to be not real concerns. No one is deleting
 data from storage in practice and the storage staking mechanism does not really serve its purpose. That conversion means
 we increase the account creation cost to 6Tgas from 0.2Tgas
 * Storage staking check will not be applied if an account has <= 3 keys and does not have a contract deployed. If an
-account accrues more than 3 keys, however, it must pay for the storage of everything including those 3 keys. This makes
+account accrues more than 3 keys, however, it must pay for the storage of everything including those 4 keys. This makes
 the implementation simpler and less error-prone.
 
 ## Reference Implementation (Required for Protocol Working Group proposals, optional for other categories)
@@ -78,10 +86,7 @@ fn check_for_zero_balance_account(account_id: &AccountId, state_update: &TrieUpd
         return false;
     }
     let keys = get_all_access_keys(account_id, state_update);
-    if keys.len() > 3 {
-        return false;
-    }
-    true
+    num_full_access_key(keys) <= 2 && num_function_call_access_key <= 2
 }
 ```
 
--- a/neps/nep-0448.md
+++ b/neps/nep-0448.md
@@ -99,6 +99,29 @@ they pay today when their account is no longer a zero-balance account.
 - A more radical thought: we can separate out zero-balance accounts into its own trie and manage them separately. This
 may allow more customization on how we want zero-balance accounts to be treated.
 
+## Decision Context
+
+### 1.0.0 - Initial Version
+
+The initial version of NEP-448 was approved by Protocol Working Group members on February 9, 2023 ([meeting recording](https://youtu.be/ktgWXjNTU_A)).
+
+#### Benefits
+
+- Users can now onboard with having to acquire NEAR tokens
+- Together with [meta transactions](https://github.com/near/NEPs/pull/366), this allows a user to start interacting with an app on NEAR directly from their device without any additional steps
+- Solves the problem of meta transaction for implicit accounts
+
+#### Concerns
+
+| # | Concern | Resolution | Status |
+| - | - | - | - |
+| 1 | The number of full access keys allowed is too small | Could be done in a future iteration. | Resolved |
+| 2 | No incentive for people to remove zero balance account. | Very few people actually delete their account anyways. | Resolved |
+| 3 | UX of requiring balance after a user graduate from zero balance account to a regular account | The experience of graduating from zero balance account should be handled on the product side | Resolved |
+| 4 | Increase of account creation cost may break some existing contracts | A thorough investigation has been done and it turns out that we only need to change the contract that is deployed on `near` slightly | Resolved |
+| 5 | Account creation speed is slower due to increased cost | Unlikely to be a concern, especially given that the number of shards is expected to grow in the future | Resolved |
+| 6 | Cost of transfers to implicit account increases | Unlikely to break anything at the moment, and could be addressed in the future in a different NEP (see https://github.com/near/NEPs/issues/462 for more details) | Resolved |
+
 ## Copyright
 
 [copyright]: #copyright
--- a/neps/nep-0448.md
+++ b/neps/nep-0448.md
@@ -1,8 +1,8 @@
 ---
-NEP: 0
+NEP: 448
 Title: Zero-balance Accounts
 Author: Bowen Wang <bowen@near.org>
-DiscussionsTo: https://github.com/nearprotocol/neps/pull/0000
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/448
 Status: Draft
 Type: Protocol Track
 Created: 10-Jan-2023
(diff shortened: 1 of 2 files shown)


neps/nep-0448.md
```diff
@@ -0,0 +1,92 @@
...
+simpler. However, there may be a risk of spamming attack given that the cost of creating an account is around 0.2Tgas.
+In addition, with the current design, it is easy to further reduce the cost. Going the other way is more difficult.
+* Do not change how storage staking is calculated when converting to gas cost. This means that account creation cost would
+be around 60Tgas, which is both high in gas (meaning that the throughput is limited and more likely for some contract to break)
+and more costly for users (around 0.006N per account creation).
+
+## Specification
+
+There are two main changes to the protocol:
+* Account creation cost needs to be increased. For every account, at creation time, 600 bytes of storage are reserved
```
bowenwang1996, line 43: need a precise number here

neps/nep-0448.md
```diff
@@ -0,0 +1,92 @@
...
+for the account itself + three keys. For function call access keys, the "free" ones cannot use `method_names` in order to
+reserve space. The cost of these bytes is paid through transaction fee. Note that there is already [discussion](https://github.com/near/NEPs/issues/415)
+around the storage cost of NEAR and whether it is reasonable. While this proposal does not attempt to change the entire
+storage staking mechanism, the cost of storage is reduced in 10x when converting to gas. A [discussion](https://gov.near.org/t/storage-staking-price/399)
+from a while ago mentioned this idea, and the concerns there were proven to be not real concerns. No one is deleting
+data from storage in practice and the storage staking mechanism does not really serve its purpose. That conversion means
+we increase the account creation cost to 6Tgas from 0.2Tgas
+* Storage staking check will not be applied if an account has <= 3 keys and does not have a contract deployed. If an
+account accrues more than 3 keys, however, it must pay for the storage of everything including those 3 keys. This makes
+the implementation simpler and less error-prone.
```
mfornet, line 53: Some questions that have been answered before but should be answered here as well:

- What happens after a fourth key is added or a contract is deployed? Are all the previous keys counted toward the storage staking?
- What happens if an account has four keys and no contract, and suddenly a key is removed? Is it acceptable for it to have 0N for staking again? 
- Depending on the answer to the previous question, we should take care of the case where a contract modifies the state but is later removed without zeroing the state back.

bowenwang1996: Will update the proposal to be more explicit, but basically it is a binary thing: either an account has free storage for up to 3 keys or it has to pay for everything (including those keys that were "free" earlier).

bowenwang1996: @mfornet I added a more detailed description in the "reference implementation" section. PTAL

bowenwang1996:
@frol @ori-near what are the next steps for this NEP?

ori-near:
Thank you @bowenwang1996 for submitting this NEP.

As a moderator, I reviewed this NEP and it meets the proposed template guidelines. I am moving this NEP to the REVIEW stage and would like to ask the  @near/wg-protocol working group members to assign 2 Technical Reviewers to complete a technical review (see expectations). If you want to assign yourself, please mention that you are acting as the Technical Reviewers.

<details>

</details>

mfornet:
As a working group member, I'd like to nominate @mm-near and @DavidM-D (because his team is working on the onboarding experience) as Subject Matter Experts to review this NEP.

ori-near:
Hi @DavidM-D and @mm-near! Thank you for agreeing to review this NEP. Per the NEP process, the expectation is for you to complete the review within one week (Tuesday, January 31). Please find additional guidelines below.

<details>

* First, review the proposal. If you have any suggestions that could be fixed, leave them as comments to the author. It may take a couple of iterations to resolve any open comments.
  * A summary of benefits (example)
  * A summary of concerns and blockers that were identified on the way and their status (some will be resolved, others dismissed, etc) (example)
</details>

mm-near:
## Recommendation 
I recommend to approve this NEP - as it allows easier onboarding (especially in the case when onboarding is done via 3rd party using meta transactions)

## Benefits
Lack of the balance costs for 'small' accounts, removes incentive for people to abuse the account creation via meta transactions flows (which was a large blocker there).
It also means that users can easily add (a few) additional keys using meta-txn too.
  
## Concerns
I have two small concerns, but they don't change my decision to support this NEP.

The main concern I have is around proper messaging for users that would 'stop' belonging to zero balance accounts (for example when they try to add the 5th key etc) - as this would mean that they will 'suddenly' be cut from most of the meta-txn relayer flows and forced to graduate and get some NEAR from somewhere. This is something that relayers should be clearly aware of and handle/inform the users correctly.

Another small concern is, that with this change, people are no longer incentivised to remove the account (currently they could do it, to get the small deposit back). But I'm not sure how often they actually did that.

DavidM-D:
This NEP addresses a real issue that we can't create user accounts without users having access to NEAR. That being said I think this deals with a part of the problem rather than the root cause.

The root cause of the problem is that relayers only receive [gas refunds](https://near.github.io/nearcore/architecture/how/meta-tx.html#gas-refunds-in-meta-transactions) [not balance refunds](https://near.github.io/nearcore/architecture/how/meta-tx.html#balance-refunds-in-meta-transactions). Lifting this restriction was discussed as part of the meta transactions proposal but it required a new field on receipts.

My reading of this proposal is that is effectively allows for a relayer balance refund on the creation of a single type of account, with two full access keys and two function call access keys.

I wonder if we can generalize this "gas as a non refundable deposit" and use it in more situations where a storage deposit would be required, e.g. adding keys. I think this would lead to a more uniform programming model for applications that use relayers and those that don't.

Failing that I would like to propose allowing 4 full access keys as we require 1 full access key for every user device and one for the backup key on NEAR crowd/social.

Overall I approve of this proposal as a step forward, but I wonder if we can take it further.

ori-near:
As the moderator, I would like to thank @mm-near and @DavidM-D for completing the technical review. Based on your comments above, it seems like this proposal is ready for the working group members for review.

@near/wg-protocol – Can you please fully read this NEP and comment in the thread if you are leaning towards approving or rejecting it? Please make sure to include your rationale and any feedback that you have for the author.

mm-near:
I lean towards supporting this NEP (see my comments above for detailed technical info).

jakmeier:
I helped out with the analysis on the increased gas cost and want to share three of my conclusions here.

## Unresolved issue
There is one unresolved issue in the NEP.

> - We need to better understand the implication of increasing account creation cost. We should check whether any contract
> on mainnet would break because of this.

Based on my analysis in https://github.com/near/nearcore/issues/8427 I conclude there are no usage patterns of account creation that cannot be easily addressed. This should not be a blocker.

## Breaking a design invariant
Today, gas costs are used to pay for CPU time (1 Tgas = 1ms) and storage staking is used to pay for persistent storage. These two are completely orthogonal. The main consequences are:
- CPU cost is affected by dynamic gas pricing, storage is not.
- CPU workload is directly limited per chunk, storage  is not.

Mixing gas and storage costs as proposed breaks this design. 1 Tgas = 1 ms compute time no longer holds. 

That's not necessarily a reason to not make this change. I just want everyone in the WG to be aware that this is the first time we break the 1 Tgas = 1 ms rule. This is a fundamental change to how gas and storage costs interact and it will set a precedent. Presumable it will influence future decisions around how we treat storage costs.

My third point below describes the implications of this specific change, as account creation throughput would no longer be limited only by CPU time.

## Mass onboarding
Today it is possible to onboard ~2500 users per chunk and only using 500Tgas of its capacity. 1M users can easily be onboarded in under 10 minutes. With the proposed change, that reduces to ~167 users per chunk, or around 2 hours for 1M users. (This assumes 500 Tgas are actually available just for creating account, spikes from traffic of the new users making function calls could invalidate this assumption.)

Limiting onboarding to 1M users in 2 hours may or may not be problematic for future projects. It depends on how a big onboarding event is carried out.

The best example from the past is Sweatcoin, where all accounts were created slowly in the days before the event and the unlocking of the tokens happened all in short period of time (TGE on Sep 12, 2022). Such an event would not be affected, as they were already limited by the function call cost that is close to 5 TGas today and account creation itself happened earlier.

frol:
@mfornet @k3y0k3y - This NEP is currently blocked on getting voting indications from the Protocol Working Group members. Please, help us to get this NEP unblocked.

frol:
As the moderator, I would like to thank the author @bowenwang1996 for submitting this NEP extension, and @mm-near @DavidM-D @jakmeier and @near/wg-protocol work group members for your review. Based on the voting indications above, and support from @bowenwang1996 it seems like this proposal is close to a decision. We are therefore scheduling the third Protocol Work Group meeting next week, where this NEP can enter the final voting stage. Anyone can discuss the technical details by adding your comments to this discussion. And join the call to learn about the final decision and how to get more involved.

Meeting Info
📅 Date: Thursday, Feb 9, 4 PM UTC
✏️ [Register](https://bit.ly/3XZvTe4)

mfornet:

k3y0k3y:
I am in support of this proposal. Agree with the feedback that a UI design to indicate to a user when they are reaching key-limit and need NEAR to continue is needed. Also agree with the feedback that in documentation of using this service there should be clear indication to use it in limited use cases where meta-tx can cover the costs or show a clear on-ramp for earning NEAR.

bowenwang1996:
Responding to a few concerns mentioned by various people:
* The number of access keys allowed is too small. @DavidM-D suggested having four full access keys instead of 2. This is actually something we could consider changing, as adding one more access key only adds 82 bytes to storage. @jakmeier do you know how much more can we increase account creation cost before something else breaks?
* No incentive for people to delete zero balance accounts. In general we observed very little deletion of data, let alone accounts. It doesn't appear that people are *really* incentivized to delete their account even with storage staking.
* The experience of "graduating" from a zero balance account is suboptimal as a user suddenly have to pay for the storage of all their keys. I'd argue that here the impact on UX is quite minimal since in our onboarding experience, a user graduates from a zero balance account either because they've earned NEAR, or because they decide to fund their account. In either case, paying for an additional <1000 bytes worth of storage (0.01N) doesn't seem to be too much of a problem.
* Increase of account creation cost may break some contracts. Thanks to the awesome work done by @jakmeier, we now know that only the contract on `near` needs to be upgraded for this change.
* Account creation throughput is lower due to increased cost. I don't think the throughput here is a big concern. Today with 4 shards we can create ~668 accounts per block and we plan to have even more shards in the future.
* Gas now can be burnt for storage (vs. before it is just about compute). Yes this is a conscious change and it feels to me that this is in the right direction of how storage should work. As mentioned in https://github.com/near/NEPs/issues/415, storage staking has a number of issues including the difficulty of storage management and bad ux for end users. In addition, people do not really delete data from storage anyways. We should think in general how storage management works, as suggested in https://github.com/near/NEPs/discussions/185

jakmeier:
> This is actually something we could consider changing, as adding one more access key only adds 82 bytes to storage. @jakmeier do you know how much more can we increase account creation cost before something else breaks?

82 extra bytes, I assume the cost would still be < 7 Tgas, right? That seems completely fine.

If the total gas goes above 9Tgas, we have to look into `the-auction-factory.near`. But I think we could even go up to 20 Tgas and only need to quickly check 4 contracts that will probably turn out to be no problem anyway, as most of them allow to attach more gas by users to resolve the problem.

bowenwang1996:
> 82 extra bytes, I assume the cost would still be < 7 Tgas, right? That seems completely fine.

If we want to have 4 full access keys, which is 2 more than the current design, then we need to add 164 bytes, which brings the account creation cost to 7.6Tgas

jakmeier:
Oh, I see. 7.6 Tgas is also fine though, I just had to do some more checks.

## Details

Only `the-auction-factory.near` falls in the problematic range and it seems all offending calls are to `create_store`, when it has only 100 Tgas attached. It will forward too much to `on_create` and not have enough to cover the extra account creation cost.
Luckily, this is a direct call and users can just attach 110 Tgas instead and it will work again. 
Example: <a href="https://explorer.near.org/transactions/BD4d1QtryVp8R5LpngSr9M7JCG8YCffN6bzaqwsGYmBT#HtmRj7qbGofHmANWt9A6E2SpNjfzAWeQw6ZjxNJvSkmD" >HtmRj7qbGofHmANWt9A6E2SpNjfzAWeQw6ZjxNJvSkmD</a>

frol:
Thank you to everyone who attended the third NEAR Protocol Working Group meeting! The working group members reviewed the NEP and reached the following consensus:

**Status:** Approved

* https://github.com/near/NEPs/pull/448#issuecomment-1410748605
* https://github.com/near/NEPs/pull/448#issuecomment-1418590578
* https://github.com/near/NEPs/pull/448#issuecomment-1419218213

Meeting Recording: TBD next week

@bowenwang1996 Thank you for authoring this NEP!

Next Steps:

* [ ] @bowenwang1996 to embed the benefits & concerns into the NEP document ([see this example](https://github.com/near/NEPs/pull/418/commits/5c71fad77d7ddb57e0661174a365c5e96c8aa4ea))
* [ ] @near/nep-moderators to approve and merge the NEP
* [ ] @near/wg-protocol to finalize the account creation cost
* [ ] nearcore team to stabilize the NEP implementation https://github.com/near/nearcore/pull/8378

Implementation Status: [Implemented in nearcore, but not yet stabilized](https://github.com/near/nearcore/pull/8378)

bowenwang1996:
Update: after a discussion with @k3y0k3y we decided to change the number of full access keys allowed to 4. This increases the account creation cost to 7.6Tgas. As we see in the discussion above, it won't break more contracts.

jakmeier:
Today I learned: Every NEAR token transfer action to an implicit account costs transfer cost + account creation cost + add full access key cost. Even if the account already exists.

It makes sense in hindsight, we don't know if the account exists when we send the transfer. And for some reason NEAR uses transfer actions to create implicit account instead of create account actions, which I was long aware of. I just didn't connect the dots for what it means for normal transfers.

This has two consequences for the zero balance account that I haven't realized previously. Maybe the SMEs did think of it already, but to make sure we are not glancing over it, here are the implications:
- normal NEAR transfers to any implicit account address will now cost 0.23 Tgas + 0.2 Tgas + 7.6 Tgas = 8.03 Tgas
- If there are any contracts that use transfers to implicit accounts inside a indirect function call, it is possible they break with this change. My checks would not have covered these cases. I don't know an easy way to check for these either.
But tbh, I am not sure why this pattern would be used in the first place. It is not inherently clear to me if it's even worth worrying about such cases.

@bowenwang1996 @mm-near @DavidM-D 
Any thoughts on this? Do you think we can proceed as planned or does this substantially change the assumptions under which this has been accepted?

frol:
@jakmeier That is a really good call. Exchanges heavily rely on implicit accounts as they own implicit accounts for each user, and now users will pay higher fees to top-up their CEX account balances, and CEX might also use implicit accounts internally for their cold wallets.

bowenwang1996:
Updated the spec for zero balance account as per discussion [here](https://github.com/near/nearcore/pull/8566#issuecomment-1429985839). @jakmeier I created https://github.com/near/NEPs/issues/462 to document your concern, but as we discussed, it does not seem to be a blocker for this NEP. @firatNEAR we may want to check with partners who manage accounts on users' behalves and often do token transfers to implicit accounts.

bowenwang1996:
@frol summary of benefits and concerns:

#### Benefits
- Users can now onboard with having to acquire NEAR tokens
- Together with [meta transactions](https://github.com/near/NEPs/pull/366), this allows a user to start interacting with an app on NEAR directly from their device without any additional steps
- Solves the problem of meta transaction for implicit accounts

#### Concerns
| - | - | - | - |
| 1 | The number of full access keys allowed is too small | Could be done in a future iteration. | Resolved |
| 2 | No incentive for people to remove zero balance account. | Very few people actually delete their account anyways. | Resolved |
| 3 | UX of requiring balance after a user graduate from zero balance account to a regular account | The experience of graduating from zero balance account should be handled on the product side | Resolved |
| 4 | Increase of account creation cost may break some existing contracts | A thorough investigation has been done and it turns out that we only need to change the contract that is deployed on `near` slightly | Resolved |
| 5 | Account creation speed is slower due to increased cost | Unlikely to be a concern, especially given that the number of shards is expected to grow in the future | Resolved |
| 6 | Cost of transfers to implicit account increases | Unlikely to break anything at the moment, and could be addressed in the future in a different NEP (see https://github.com/near/NEPs/issues/462 for more details) | Resolved |

frol (approved):
As a NEP moderator, I am approving this PR following the Protocol Work Group approval and the decision context incorporated into the NEP document.

PR #455, opened 2023-01-30
NEP-455: Parameter Compute Costs

akashin:
NEP for Parameter Compute Costs (formerly known as Parameter Weights).

For the context, this NEP was previously named "Parameter Weights" and the idea was originally proposed in https://gov.near.org/t/proposal-gas-weights-to-fight-instability-to-due-to-undercharging/30919.
Now it's renamed into "Parameter Compute Cost" to more accurately represent the fact that it is about accounting how much compute time the parameters cost and using it to limit the chunk processing time.

### DIFF
--- a/neps/nep-0455.md
+++ b/neps/nep-0455.md
@@ -1,0 +1,219 @@
+---
+NEP: 455
+Title: Parameter Compute Costs
+Author: Andrei Kashin <andrei.kashin@near.org>, Jakob Meier <jakob@near.org>
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/455
+Status: Draft
+Type: Protocol Track
+Category: Runtime
+Created: 26-Jan-2023
+---
+
+## Summary
+
+Introduce compute costs decoupled from gas costs for individual parameters to safely limit the compute time it takes to process the chunk while avoiding adding breaking changes for contracts.
+
+## Motivation
+
+For NEAR blockchain stability, we need to ensure that blocks are produced regularly and in a timely manner.
+
+The chunk gas limit is used to ensure that the time it takes to validate a chunk is strictly bounded by limiting the total gas cost of operations included in the chunk.
+This process relies on accurate estimates of gas costs for individual operations.
+
+Underestimating these costs leads to *undercharging* which can increase the chunk validation time and slow down the chunk production.
+
+As a concrete example, in the past we undercharged contract deployment.
+The responsible team has implemented a number of optimizations but a gas increase was still necessary.
+[Meta-pool](https://github.com/Narwallets/meta-pool/issues/21) and [Sputnik-DAO](https://github.com/near-daos/sputnik-dao-contract/issues/135) were affected by this change, among others.
+Finding all affected parties and reaching out to them before implementing the change took a lot of effort, prolonging the period where the network was exposed to the attack.
+
+Another motivating example is the upcoming incremental deployment of Flat Storage, where during one of the intermediate stages we expect the storage operations to be undercharged.
+See the explanation in the next section for more details.
+
+## Rationale
+
+Separating compute costs from gas costs will allow us to safely limit the compute usage for processing the chunk while still keeping the gas prices the same and thus not breaking existing contracts.
+
+An important challenge with undercharging is that it is not possible to disclose them widely because it could be used to increase the chunk production time thereby impacting the stability of the network.
+Adjusting the compute cost for undercharged parameter eliminates the security concern and allows us to publicly discuss the ways to solve the undercharging (optimize implementation, smart contract or increasing the gas cost).
+
+This design is easy to implement and simple to reason about and provides a clear way to address existing undercharging issues.
+If we don't address the undercharging problems, we increase the risks that they will be exploited.
+
+Specifically for Flat Storage deployment, we [plan](https://github.com/near/nearcore/issues/8006) to stop charging TTN (touching trie node) gas costs, however the intermediate implementation (read-only Flat Storage) will still incur these costs during writes introducing undercharging.
+Setting temporary high compute costs for writes will ensure that this undercharging does not lead to long chunk processing times.
+
+## Alternatives
+
+### Increase the gas costs for undercharged operations
+We could increase the gas costs for the operations that are undercharged to match the computational time it takes to process them according to the rule 1ms = 1TGas.
+
+Pros:
+- Does not require any new code or design work (but still requires a protocol version bump)
+- Security implications are well-understood
+
+Cons:
+- Can break contracts that rely on current gas costs, in particular steeply increasing operating costs for the most active users of the blockchain (aurora and sweat)
+- Doing this safely and responsibly requires prior consent by the affected parties which is hard to do without disclosing security-sensitive information about undercharging in public
+
+In case of flat storage specifically, using this approach will result in a large increase in storage write costs (breaking many contracts) to enable safe deployment of read-only flat storage and later a correction of storage write costs when flat storage for writes is rolled out.
+With compute costs, we will be able to roll out the read-only flat storage with minimal impact on deployed contracts.
+
+### Adjust the gas chunk limit
+We could continuously measure the chunk production time in nearcore clients and compare it to the gas burnt.
+If the chunk producer observes undercharging, it decreases the limit.
+If there is overcharging, the limit can be increased up to a limit of at most 1000 Tgas.
+To make such adjustment more predictable under spiky load, we also [limit](https://nomicon.io/Economics/Economic#transaction-fees) the magnitude of change of gas limit by 0.1% per block.
+
+Pros:
+- Prevents moderate undercharging from stalling the network
+- No protocol change necessary (as this feature is already [a part of the protocol](https://nomicon.io/Economics/Economic#transaction-fees)), we could easily experiment and revert if it does not work well
+
+Cons:
+- Very broad granularity --- undercharging in one parameter affects all users, even those that never use the undercharged parts
+- Dependence on validator hardware --- someone running overspecced hardware will continuously want to increase the limit, others might run with underspecced hardware and continuously want to decrease the limit
+- Malicious undercharging attacks are unlikely to be prevented by this --- a single 10x undercharged receipt still needs to be processed using the old limit.
+Adjusting 0.1% per block means 100 chunks can only change by a maximum of 1.1x and 1000 chunks could change up to x2.7
+- Conflicts with transaction and receipt limit --- A transaction or receipt can (today) use up to 300Tgas.
+The effective limit per chunk is `gas_limit` + 300Tgas since receipts are added to a chunk until one exceeds the limit and the last receipt is not removed.
+Thus a gas limit of 0gas only reduces the effective limit from 1300Tgas to 300Tgas, which means a single 10x undercharged receipt can still result in a chunk with compute usage of 3 seconds (equivalent to 3000TGas)
+
+### Allow skipping chunks in the chain
+Slow chunk production in one shard can introduce additional user-visible latency in all shards as the nodes expect a regular and timely chunk production during normal operation.
+If processing the chunk takes much longer than 1.3s, it can cause the corresponding block and possibly more consecutive blocks to be skipped.
+
+We could extend the protocol to produce empty chunks for some of the shards within the block (effectively skipping them) when processing the chunk takes longer than expected.
+This way will still ensure a regular block production, at a cost of lower throughput of the network in that shard.
+The chunk should still be included in a later block to avoid stalling the affected shard.
+
+Pros:
+- Fast and automatic adaptation to the blockchain workload
+
+Cons:
+- Needs a voting mechanism to decide that processing of the chunk is slow (as opposed to the chunk producer being slow) and the chunk should be skipped within a block
+- Needs economic design for the rewards for skipped chunks
+
+## Specification
+
+- **Chunk Compute Usage** -- total compute time spent on processing the chunk
+
+- **Chunk Compute Limit** -- upper-bound for compute time spent on processing the chunk
+
+- **Parameter Compute Cost** -- the numeric value in seconds corresponding to compute time that it takes to include an operation into the chunk
+
+Today, gas has two somewhat orthogonal roles:
+
+1. Gas is money. It is used to avoid spam by charging users
+2. Gas is CPU time. It defines how many transactions fit in a chunk so that validators can apply it within a second
+
+The idea is to decouple these two by introducing parameter compute costs.
+Each gas parameter still has a gas cost that determines what users have to pay.
+But when filling a chunk with transactions, parameter compute cost is used to estimate CPU time.
+
+Ideally, all compute costs should match corresponding gas costs.
+But when we discover undercharging issues, we can set a higher compute cost (this would require a protocol upgrade).
+The stability concern is then resolved when the compute cost becomes active.
+
+The ratio between compute cost and gas cost can be thought of as an undercharging factor.
+If a gas cost is 2 times too low to guarantee stability, compute cost will be twice the gas cost.
+A chunk will be full 2 times faster when gas for this parameter is burned.
+This deterministically throttles the throughput to match what validators can actually handle.
+
+Compute costs influence the gas price adjustment logic described in https://nomicon.io/Economics/Economic#transaction-fees.
+Specifically, we're now using compute usage instead of gas usage in the formula to make sure that the gas price increases if chunk processing time is close to the limit.
+
+Compute costs **do not** count towards the transaction/receipt gas limit of 300TGas, as that might break existing contracts by pushing their method calls over this limit.
+
+Compute costs are static for each protocol version.
+
+### Using Compute Costs
+Compute costs different from gas costs are only a temporary solution.
+Whenever we introduce a compute cost, we as the community can discuss this publicly and find a solution to the specific problem together.
+
+In the best case, we find technical optimizations that allow us to decrease the compute cost to match the existing gas cost.
+
+In other cases, the only solution is to increase the gas cost.
+But the dApp developers who are affected by this change should have a chance to voice their opinion, suggest alternatives, and implement necessary changes before the gas cost is increased.
+
+## Reference Implementation
+
+The reference implementation can be found at https://github.com/near/nearcore/pull/8426.
+
+The compute cost is a numeric value represented as `u64` in time units.
+Value 1 corresponds to `10^-15` seconds or 1fs (femtosecond) to match the gas costs scale.
+
+By default, the parameter compute cost matches the corresponding gas cost.
+
+Compute costs should be applicable to all gas parameters, specifically including:
+- [`ExtCosts`](https://cs.github.com/near/nearcore/blob/6e08a41084c632010b1d4c42132ad58ecf1398a2/core/primitives-core/src/config.rs#L377)
+- [`ActionCosts`](https://cs.github.com/near/nearcore/blob/6e08a41084c632010b1d4c42132ad58ecf1398a2/core/primitives-core/src/config.rs#L456)
+
+Changes necessary to support `ExtCosts`:
+1. Track compute usage in [`GasCounter`](https://cs.github.com/near/nearcore/blob/51670e593a3741342a1abc40bb65e29ba0e1b026/runtime/near-vm-logic/src/gas_counter.rs#L47) struct
+2. Track compute usage in [`VMOutcome`](https://cs.github.com/near/nearcore/blob/056c62183e31e64cd6cacfc923a357775bc2b5c9/runtime/near-vm-logic/src/logic.rs#L2868) struct (alongside `burnt_gas` and `used_gas`)
+3. Store compute usage in [`ActionResult`](https://cs.github.com/near/nearcore/blob/6d2f3fcdd8512e0071847b9d2ca10fb0268f469e/runtime/runtime/src/lib.rs#L129) and aggregate it across multiple actions by modifying [`ActionResult::merge`](https://cs.github.com/near/nearcore/blob/6d2f3fcdd8512e0071847b9d2ca10fb0268f469e/runtime/runtime/src/lib.rs#L141)
+4. Store compute costs in [`ExecutionOutcome`](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/runtime/runtime/src/lib.rs#L266) and [aggregate them across all transactions](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/runtime/runtime/src/lib.rs#L1279)
+5. Enforce the chunk compute limit when the chunk is [applied](https://cs.github.com/near/nearcore/blob/6d2f3fcdd8512e0071847b9d2ca10fb0268f469e/runtime/runtime/src/lib.rs#L1325)
+
+Additional changes necessary to support `ActionCosts`:
+1. Return compute costs from [`total_send_fees`](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/runtime/runtime/src/config.rs#L71)
+2. Store aggregate compute cost in [`TransactionCost`](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/runtime/runtime/src/config.rs#L22) struct
+3. Propagate compute costs to [`VerificationResult`](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/runtime/runtime/src/verifier.rs#L330)
+
+Additionaly, the gas price computation will need to be adjusted in [`compute_new_gas_price`](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/core/primitives/src/block.rs#L328) to use compute cost instead of gas cost.
+
+## Security Implications
+
+Changes in compute costs will be publicly known and might reveal an undercharging that can be used as a target for the attack.
+In practice, it is not trivial to exploit the undercharging unless you know the exact shape of the workload that realizes it.
+Also, after the compute cost is deployed, the undercharging should no longer be a threat for the network stability.
+
+## Drawbacks
+
+- Changing compute costs requires a protocol version bump (and a new binary release), limiting their use to undercharging problems that we're aware of
+
+- Updating compute costs is a manual process and requires deliberately looking for potential underchargings
+
+- The compute cost would not have a full effect on the last receipt in the chunk, decreasing its effectiveness to deal with undercharging.
+This is because 1) a transaction or receipt today can use up to 300TGas and 2) receipts are added to a chunk until one exceeds the limit and the last receipt is not removed.
+Therefore, a single receipt with 300TGas filled with undercharged operations with a factor of K can lead to overshooting the chunk compute limit by (K - 1) * 300TGas
+
+- Underchargings can still be exploited to lower the throughput of the network at unfair price and increase the waiting times for other users.
+This is inevitable for any proposal that doesn't change the gas costs and must be resolved by improving the performance or increasing the gas costs
+
+- Even without malicious intent, the effective peak throughput of the network will decrease when the chunks include undercharged operations (as the stopping condition based on compute costs for filling the chunk becomes stricter).
+Most of the time, this is not the problem as the network is operating below the capacity.
+The effects will also be softened by the fact that undercharged operations comprise only a fraction of the workload.
+For example, the planned increase for TTN compute cost alongside the Flat Storage MVP is less critical because you cannot fill a receipt with only TTN costs, you will always have other storage costs and ~5Tgas overhead to even start a function call.
+So even with 10x difference between gas and compute costs, the DoS only becomes 5x cheaper instead of 10x
+
+## Unresolved Issues
+
+## Future possibilities
+We can also think about compute costs smaller than gas costs.
+For example, we charge gas instead of token balance for extra storage bytes in [NEP-448](https://github.com/near/NEPs/pull/448), it would make sense to set the compute cost to 0 for the part that covers on-chain storage if the throttling due to increased gas cost becomes problematic.
+Otherwise, the throughput would be throttled unnecessarily.
+
+A further option would be to change compute costs dynamically without a protocol upgrade when block production has become too slow.
+This would be a catch-all, self-healing solution that requires zero intervention from anyone.
+The network would simply throttle throughput when block time remains too high for long enough.
+Pursuing this approach would require additional design work:
+- On-chain voting to agree on new values of costs, given that inputs to the adjustment process are not deterministic (measurements of wall clock time it takes to process receipt on particular validator)
+- Ensuring that dynamic adjustment is done in a safe way that does not lead to erratic behavior of costs (and as a result unpredictable network throughput).
+Having some experience manually operating this mechanism would be valuable before introducing automation
+
+and addressing challenges described in https://github.com/near/nearcore/issues/8032#issuecomment-1362564330.
+
+The idea of introducing a chunk limit for compute resource usage naturally extends to other resource types, for example RAM usage, Disk IOPS, [Background CPU Usage](https://github.com/near/nearcore/issues/7625).
+This would allow us to align the pricing model with cloud offerings familiar to many users, while still using gas as a common denominator to simplify UX.
+
+## Copyright
+
+[copyright]: #copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+
+## References
+- https://gov.near.org/t/proposal-gas-weights-to-fight-instability-to-due-to-undercharging/30919
+- https://github.com/near/nearcore/issues/8032
--- a/neps/nep-0455.md
+++ b/neps/nep-0455.md
@@ -207,6 +207,25 @@ and addressing challenges described in https://github.com/near/nearcore/issues/8
 The idea of introducing a chunk limit for compute resource usage naturally extends to other resource types, for example RAM usage, Disk IOPS, [Background CPU Usage](https://github.com/near/nearcore/issues/7625).
 This would allow us to align the pricing model with cloud offerings familiar to many users, while still using gas as a common denominator to simplify UX.
 
+## Changelog
+
+### 1.0.0 - Initial Version
+
+This NEP was approved by Protocol Working Group members on March 16th, 2023 ([meeting recording](https://www.youtube.com/watch?v=4VxRoKwLXIs)):
+
+- [Bowen](https://github.com/near/NEPs/pull/455#issuecomment-1467023424)
+- [Marcelo](https://github.com/near/NEPs/pull/455#pullrequestreview-1340887413)
+- [Marcin](https://github.com/near/NEPs/pull/455#issuecomment-1471882639)
+
+#### Benefits
+
+- Among the alternatives, this is the easiest to implement.
+- It allows us to able to publicly discuss undercharging issues before they are fixed. 
+
+#### Concerns
+
+No concerns that need to be addressed. The drawbacks listed in this NEP are minor compared to the benefits that it will bring. And implementing this NEP is strictly better than what we have today.
+
 ## Copyright
 
 [copyright]: #copyright
--- a/neps/nep-0455.md
+++ b/neps/nep-0455.md
@@ -90,8 +90,7 @@ Pros:
 - Fast and automatic adaptation to the blockchain workload
 
 Cons:
-- Needs a voting mechanism to decide that processing of the chunk is slow (as opposed to the chunk producer being slow) and the chunk should be skipped within a block
-- Needs economic design for the rewards for skipped chunks
+- For the purpose of slashing, it is hard to distinguish situations when the honest block producer skips chunk due to slowness from the situations when the block producer is offline or is maliciously stalling the block production. We need some mechanism (e.g. on-chain voting) for nodes to agree that the chunk was skipped legitimately due to slowness as otherwise we introduce new attack vectors to stall the network
 
 ## Specification
 
--- a/neps/nep-0455.md
+++ b/neps/nep-0455.md
@@ -211,11 +211,11 @@ This would allow us to align the pricing model with cloud offerings familiar to 
 
 ### 1.0.0 - Initial Version
 
-This NEP was approved by Protocol Working Group members on March 16th, 2023 ([meeting recording](https://www.youtube.com/watch?v=4VxRoKwLXIs)):
+This NEP was approved by Protocol Working Group members on March 16, 2023 ([meeting recording](https://www.youtube.com/watch?v=4VxRoKwLXIs)):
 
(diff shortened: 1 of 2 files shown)


neps/nep-9999.md
```diff
@@ -0,0 +1,111 @@
...
+
+If we don't do this work, it would be harder to deploy Flat Storage safely without a mechanism to safely adjust gas costs.
+
+## Alternatives
+
+### Increase the gas cost for undercharged operations
+We could change the gas costs for the operations that are undercharged to match the time it takes to process them.
+
+However, this would break contracts that rely on current gas costs. It can also steeply increase operating costs for the most active users of the blockchain. I believe the protocol must serve its users and not the other way around. Therefore, increasing gas costs without prior consent by the affected parties can only be considered in absolute emergency scenarios.
+
```
jakmeier, line 38: We should also describe that:
- security related changes cannot be discussed in public and therefore we cannot always ask the smart contract devs for their opinion before we increase costs
- in case of the flat storage example, incrementing the cost to a safe value in the intermediate state would significantly increase the cost for many users, even though only very specific use cases are actually undercharged

akashin: Good point about need to disclose the security-sensitive information, added it.

For the flat storage example, I don't think this problem is specific for this alternative solution, it's just the fact that our gas parameters are not granular/nuanced enough to charge two workloads differently according to their real resource usage. Parameter Weights would suffer from similar problem of limiting chunk sizes with some workloads that are actually fine from the processing time perspective just for the sake of curbing some worst-case workloads that we have in mind.

akashin:
@jakmeier , I've drafted the rest of the sections, adding the main points that need to be covered there. Please take a look if there are any other topics that the NEP should highlight/clarify.

I also proposed a new terminology for the weights to make the naming a bit more future-looking. Specifically, I'm thinking of renaming "Parameter Weights" to "Parameter Compute Weights/Factors" to reflect the fact that this is a conversion from gas costs to compute costs. Similarly, I'm thinking of introducing a "Chunk Compute Limit" (specialized "Chunk Gas Limit") for the purpose of limiting the total compute usage in the chunk.
The idea is to more accurately reflect the fact that node hardware resources are multi-dimensional (CPU, RAM, Disk IOPS) and the limits during processing of the chunk should really be on each dimension separately as reaching limits on any of the dimensions would degrade chunk processing time significantly.
Gas still stays as a unit of account to simplify UX for users. This NEP in essense introduces a mechanism to convert compute resource usage to gas costs.

jakmeier (commented):
This is taking on a good shape already. It's a great basis to involve more protocol team members to discuss details.

neps/nep-9999.md
```diff
@@ -0,0 +1,184 @@
...
+resolved when the compute weight becomes active.
+
+The parameter compute weight can be thought of as an undercharging weight. If a gas cost is 1.5
+times too low to guarantee stability, the compute weight is 1.5. A chunk will be full 1.5 times
+faster when gas for this parameter is burned. This deterministically throttles the throughput to
+match what validators can actually handle.
+
+Parameter compute weights are static and are fixed at the Nearcore release time.
+
+They will not be served by a public RPC interface of the node.
```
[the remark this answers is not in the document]
akashin: Makes sense, dropped this.

neps/nep-9999.md
```diff
@@ -0,0 +1,184 @@
...
+
+Cons:
+- Very broad granularity. (Undercharging in one parameter affects all users, even those that never use the undercharged parts.)
+- Dependence on validator hardware. (Someone running overspecced hardware will continuously want to increase the limit, others might run with underspecced hardware and continuously want to decrease the limit.)
+- Malicious undercharging attacks are unlikely to be prevented by this. (A single 10x undercharged receipt still needs to be processed using the old limit. Adjusting 0.1% per block means 100 chunks can only change by a maximum of 1.1x and 1000 chunks could do a weight of 2.7).
+- Conflict with transaction & receipt limit. A transaction or receipt can (today) use up to 300Tgas. The effective limit per chunk is `gas_limit` + 300Tgas since receipts are added to a chunk until one exceeds the limit and the last receipt is not removed. Thus a gas limit of 0gas only reduces the effective limit from 1300Tgas to 300Tgas, which means undercharging of > 4.33x is not covered.
+
+### Allow skipping chunks in the chain
+- [ ] TODO: Is the reasoning why slow block production is undesired correct?
```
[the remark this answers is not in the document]
akashin: Good idea, I've posted the question on Zulip: https://near.zulipchat.com/#narrow/stream/295558-pagoda.2Fcore/topic/Q.3A.20Consensus.20under.20slower.20block.20production/near/326111705

akashin:
@jakmeier , in a process of polishing the Summary/Motivation/Spefication sections I came to conclusion that we need to do a bit more thinking around conceptual way we present this idea (as I've struggled to clearly write it down).

Here is my latest thinking trying to distill what we want to accomplish even further:

The root cause/tension behind underchargings is the fact that we want to use the average operation gas costs for the purpose of charging users (to avoid overcharging them and make their workload costs competetive with other chains), but we want to use **safe gas costs** for the purpose of limiting the chunk size (to limit the chunk production time in the worst case).

We mostly get away with making safe costs equal to average costs (as on average we don't hit the worst case), but in cases when the difference between the two is large (e.g. storage operations) this distinction becomes especially noticeable and calls for action.

So another conceptual way to think about this NEP is as an introduction of **safe gas costs** that approximate the worst case compute costs of operations.
As it is now, by default the safe gas cost matches the average gas cost.
As with weights, we can define the safe cost directly on per-parameter basis without the need for indirect weights (avoids working with rationals).

Then we will use safe gas costs for the purpose of counting towards chunk gas limit, but still use the existing gas cost for all other purposes (e.g. burning).

Overall, I think this is both easier to explain and also cuts to the root of the problem. What do you think?
As a next step, it would be good to check how this framing applies to existing undercharging situations as well as flat storage deployment.

neps/nep-9999.md
akashin: I've added a paragraph dedicated to Flat Storage deployment, please take a look whether this expresses what you had in mind.

neps/nep-0455.md
```diff
@@ -0,0 +1,188 @@
...
+In other cases, the only solution is to increase the gas cost.
+But the dApp developers who are affected by this change should have a chance to voice their opinion, suggest alternatives, and implement necessary changes before the gas cost is increased.
+
+## Reference Implementation
+
+The reference implementation can be found at https://github.com/near/nearcore/pull/8426.
+
+The Compute Cost is a numeric value represented in time units.
+
+- [ ] Which type will we use? Which units will we use?
```
jakmeier, line 130: Yeah we should decide and fill this in before marking this as ready for review.

We could use well-known types from the Rust ecosystem ([std](https://doc.rust-lang.org/std/index.html)::[time](https://doc.rust-lang.org/std/time/index.html)::[Duration](https://doc.rust-lang.org/std/time/struct.Duration.html#)) but probably just a `u64` (potentially with a strongly-typed wrapper) is better suited.

One reason not to use std Duration is it only has nanosecond granularity.
Current gas costs are in the granularity of millionth parts of a nanosecond, aka femtosecond (fs). I think the lowest value today is 98'562 fs. Therefore, ns granularity wouldn't be enough.

It feels like fs are a bit too granular and picoseconds (ps) (10^-12) would be granular enough. But to keep things in a simple relation to gas costs, I think we should just keep it as fs granularity (10^-15). A `u64` can then represent up to 18446 seconds ~ 5 hours. That seems reasonable to me.

jakmeier (commented):
@akashin I had a brief look over it. Left a few small grammar related comments and a thought or two on the content as well. But I think once these details are addressed, this looks ready to be submitted to the WG.

jakmeier:
Oh and we should update the title and description to reflect the new name. (Maybe even a quick explanation that it used to be called weights but isn't anymore could be useful in the PR comment.)

neps/nep-0455.md
akashin: I agree with matching compute costs with gas cost scales and using u64. Updated the section to describe this.

neps/nep-0455.md
```diff
@@ -154,28 +148,28 @@ Additionaly, the gas price computation will need to be adjusted in [`compute_new
...
-Thus compute weight would not have a full effect on the last receipt in the chunk
+Thus compute cost would not have a full effect on the last receipt in the chunk
 
 ## Unresolved Issues
 
 ## Future possibilities
-We can also think about compute weights smaller than 1.
-For example, if we want to charge gas instead of token balance for extra storage bytes, it would make sense to set the compute weight to 0 for the part that covers on-chain storage.
+We can also think about compute costs smaller than gas costs.
+For example, if we want to charge gas instead of token balance for extra storage bytes, it would make sense to set the compute cost to 0 for the part that covers on-chain storage.
```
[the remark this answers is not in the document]
akashin: Rewrote the sentence to mention this.

akashin:
@near/nep-moderators , this NEP is now ready for review.

For the context, this NEP was previous named "Parameter Weights" and the idea originally proposed in https://gov.near.org/t/proposal-gas-weights-to-fight-instability-to-due-to-undercharging/30919.
Now it's renamed into "Parameter Compute Cost" to more accurately represent the fact that it is about accounting how much compute time the parameters cost and using it to limit the chunk processing time.

frol:
Thank you @akashin for submitting this NEP.

As a moderator, I reviewed this NEP and it meets the proposed template guidelines. I am moving this NEP to the REVIEW stage and would like to ask the @near/wg-protocol working group members to assign 2 Technical Reviewers to complete a technical review. If you want to assign yourself, please mention that you are acting as the Technical Reviewers.

<details>


  * A summary of benefits

```
## Example

### Recommendation
Add recommendation

### Benefits

- Benefit
- Benefit

### Concerns

| - | - | - | - |   
```

</details>

bowenwang1996:
I nominate @jakmeier and @mzhangmzz as SME reviewers for this NEP.

akhi3030:
@blasrodri indicated on the Protocol Community Group that they will be interested in helping review this NEP.

jakmeier:
> I nominate @jakmeier and @mzhangmzz as SME reviewers for this NEP.

@near/nep-moderators 

As a listed co-author and one of the engineers originally proposing this idea, does it make sense for me to take on the role of an SME? Obviously, I can do it with little effort, given that I have already spent a lot of time discussing this topic with Andrei and others. But I wonder if this makes sense from a process perspective.

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+
+We could extend the protocol to produce empty chunks for some of the shards within the block (effectively skipping them) when processing the chunk takes longer than expected.
+This way will still ensure a regular block production, at a cost of lower throughput of the network in that shard.
+The chunk should still be included in a later block to avoid stalling the affected shard.
+
+Pros:
+- Fast and automatic adaptation to the blockchain workload
+
+Cons:
+- Needs a voting mechanism to decide that processing of the chunk is slow (as opposed to the chunk producer being slow) and the chunk should be skipped within a block
```
blasrodri, line 82: would this also have an impact on how the light client work, and the costs of state transition validations?

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+This process relies on accurate estimates of gas costs for individual operations.
+
+Underestimating these costs leads to *undercharging* which can increase the chunk validation time and slow down the chunk production.
+
+As a concrete example, during the incremental deployment of Flat Storage we will be going through intermediate stages where the storage operations will be undercharged.
+
+## Rationale
+
+Separating compute costs from gas costs will allow us to safely limit the compute usage for processing the chunk while still keeping the gas prices the same and thus not breaking existing contracts.
+After adding compute cost for target parameter we will be able to publicly discuss the ways to solve the undercharging (optimize implementation, smart contract or increasing the gas cost).
```
akhi3030, line 30:
```suggestion
After adjusting the compute cost for target parameter we will be able to publicly discuss the ways to solve the undercharging (optimize implementation, smart contract or increasing the gas cost).
```

akhi3030, line 30: I think we should add a sentence somewhere before here that explains that when an instance of undercharging is discovered, it is currently not possible to disclose the discovery widely because it could be used to increase the chunk production thereby impacting the stability of the network.

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+Thus compute cost would not have a full effect on the last receipt in the chunk
+
+## Unresolved Issues
+
+## Future possibilities
+We can also think about compute costs smaller than gas costs.
+For example, we charge gas instead of token balance for extra storage bytes in [NEP-448](https://github.com/near/NEPs/pull/448), it would make sense to set the compute cost to 0 for the part that covers on-chain storage if the throttling due to increased gas cost becomes problematic.
+Otherwise, the throughput would be throttled unnecessarily.
+
+A further option would be to change compute costs dynamically without a protocol upgrade when block production has become too slow.
```
blasrodri, line 175: what's the complexity of introducing this solution right away?

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+As a concrete example, during the incremental deployment of Flat Storage we will be going through intermediate stages where the storage operations will be undercharged.
+
+## Rationale
+
+Separating compute costs from gas costs will allow us to safely limit the compute usage for processing the chunk while still keeping the gas prices the same and thus not breaking existing contracts.
+After adding compute cost for target parameter we will be able to publicly discuss the ways to solve the undercharging (optimize implementation, smart contract or increasing the gas cost).
+
+This design is easy to implement and simple to reason about and provides a clear way to address existing undercharging issues.
+
+If we don't address the undercharging problems, we increase the risks that they will be exploited and also make it harder to deploy Flat Storage safely.
```
akhi3030, line 34: Not sure if we have sufficiently explained how this helps deploy flat storage safely yet.

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/455
+Status: Draft
+Type: Protocol Track
+Category: Runtime
+Created: 26-Jan-2023
+---
+
+## Summary
+
+Introduce compute costs decoupled from gas costs for individual parameters to allow to safely limit the compute time it takes to process the chunk.
```
blasrodri, line 14:
```suggestion
Introduce compute costs decoupled from gas costs for individual parameters to allow to safely limit the compute time it takes to process the chunk to ensure block production time is kept within bounds.
```

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+Cons:
+- Can break contracts that rely on current gas costs, in particular steeply increasing operating costs for the most active users of the blockchain (aurora and sweat)
+- Doing this safely and responsibly requires prior consent by the affected parties which is hard to do without disclosing the security-sensitive information about undercharging in public
+
+In case of flat storage specifically, using this approach will result in a large increase in storage write costs (breaking many contracts) to enable safe deployment of read-only flat storage and later a correction of storage write costs when flat storage for writes is rolled out.
+With compute costs, we will be able to roll out the read-only flat storage with minimal impact on deployed contracts.
+
+### Adjust the gas chunk limit
+We could continuously measure the chunk production time in nearcore clients and compare it to the gas burnt.
+If the chunk producer observes undercharging, it decreases the limit.
```
akhi3030, line 54: Might be worth adding one or more sentences to explain that we can only adjust a limit a bit at a time so in cases where a big adjustment is needed, it will take multiple blocks to get there.

neps/nep-0455.md
blasrodri, line 14:
```suggestion
Introduce compute costs decoupled from gas costs for individual parameters to allow to safely limit the compute time it takes to process the chunk while avoiding adding breaking changes for contracts.
```

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+
+Cons:
+- Very broad granularity --- undercharging in one parameter affects all users, even those that never use the undercharged parts
+- Dependence on validator hardware --- someone running overspecced hardware will continuously want to increase the limit, others might run with underspecced hardware and continuously want to decrease the limit
+- Malicious undercharging attacks are unlikely to be prevented by this --- a single 10x undercharged receipt still needs to be processed using the old limit.
+Adjusting 0.1% per block means 100 chunks can only change by a maximum of 1.1x and 1000 chunks could change up to x2.7
+- Conflicts with transaction and receipt limit --- A transaction or receipt can (today) use up to 300Tgas.
+The effective limit per chunk is `gas_limit` + 300Tgas since receipts are added to a chunk until one exceeds the limit and the last receipt is not removed.
+Thus a gas limit of 0gas only reduces the effective limit from 1300Tgas to 300Tgas, which means undercharging of > 4.33x is not covered.
```
akhi3030, line 68: can you briefly explain how 4.33x is computed?

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+This way will still ensure a regular block production, at a cost of lower throughput of the network in that shard.
+The chunk should still be included in a later block to avoid stalling the affected shard.
+
+Pros:
+- Fast and automatic adaptation to the blockchain workload
+
+Cons:
+- Needs a voting mechanism to decide that processing of the chunk is slow (as opposed to the chunk producer being slow) and the chunk should be skipped within a block
+- Needs economic design for the rewards for skipped chunks
+- No prior examples of this approach in other blockchains
```
akhi3030, line 84: Another con might be the noisy neighbour effect.  If I happen to be in the same shard as another contract that is exploiting some undercharging, the throughput of the shard is reduced and I am getting less work done.  I believe the proposal of this NEP will help in such scenarios.

neps/nep-0455.md
```diff
@@ -0,0 +1,191 @@
...
+3. Propagate compute costs to [`VerificationResult`](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/runtime/runtime/src/verifier.rs#L330)
+4. Store compute costs in [`ExecutionOutcome`](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/runtime/runtime/src/lib.rs#L266) and [aggregate them across all transactions](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/runtime/runtime/src/lib.rs#L1279)
+
+Additionaly, the gas price computation will need to be adjusted in [`compute_new_gas_price`](https://cs.github.com/near/nearcore/blob/578983c8df9cc36508da2fb4a205c852e92b211a/core/primitives/src/block.rs#L328) to use compute cost instead of gas cost.
+
+## Security Implications
+
+Changes in compute costs will be publicly known and might reveal an undercharging that can be used as a target for the attack.
+In practice, it is not trivial to exploit the undercharging unless you know the exact shape of the workload that realizes it.
+Also, after the compute cost is deployed, the undercharging should no longer be a threat for the network stability.
```
akhi3030, line 156: I think there is another security implication although it is probably a minor one.  

If I want to dos the system, I probably want to do it as cheaply as possible.  So I would use the publicly available compute costs information to make my attack as cheap as possible.  I don't think this is a huge problem yet as our network is generally operating way below capacity.

akashin: Yes, that is exactly what I was trying to convey with this paragraph.
Do you mean to add an additional consideration that the attacker might use the publicly known list of costs? I believe that is already the case given that all [gas parameters](https://github.com/near/nearcore/blob/master/core/primitives/res/runtime_configs/parameters.snap) are publicly known which is necessary for consensus, so I don't think we're introducing any new implications on that front.

neps/nep-0455.md
akashin: To be honest, I don't know because this idea is not fully thought though.

I would defer to @abacabadabacaba to answer this, given that this idea is my sketch of a sketch proposed by @abacabadabacaba during one of the discussions that I've included for completeness.

neps/nep-0455.md
akashin: @jakmeier did a great job explaining the additional complexity that this will introduce in https://github.com/near/nearcore/issues/8032#issuecomment-1362564330

TL;DR is:
- This would require some form of on-chain voting, given that inputs to the adjustment process are not deterministic (measurements of wall clock time it takes to process receipt on particular validator)
- Designing the dynamic adjustment in a safe way that does not lead to erratic behavior of costs (and as a result unpredictable network throughput) itself is not trivial. Having some experience manually operating this mechanism would be valuable before introducing automation

blasrodri: thanks

neps/nep-0455.md
akhi3030: Hmm... maybe I am missing some context here.  The way I understand it, let's say that we have a gas parameter X which is currently set to 10T Gas.  So if I want to dos the system by taking up all the gas in a single block, I need to produce transactions that issue X 130 times.  Now, let's say that the this parameter is being undercharged and we need to set the compute cost for this to 10.  As the compute cost knowledge is public, I can see it and I realise that in order to dos the system I only need to issue to issue X 13 times.  So in essence, the compute costs will help in figuring out where the most undercharging is happening so which parameters I choose to use up the available capacity the cheapest.

neps/nep-0455.md
akashin: I've added these considerations to the NEP.

neps/nep-0455.md
akashin: Good idea, I've added a sentence that describes this.

neps/nep-0455.md
akashin: Done.

neps/nep-0455.md
akashin: It rewrote this in a way that's easier to understand, PTAL

akhi3030: much better!

neps/nep-0455.md
akashin: Unfortunately, all of the proposals described in this NEP (except actually increasing gas costs) would allow the attacker to reduce throughput for all users of a shard where the undercharged operations are used at a lower cost. This is due to the fact that unless you correct the gas price to match the compute cost, there will always be an opportunity to exploit that difference by paying less gas for using more compute.

I've added this to Drawbacks section.

neps/nep-0455.md
akashin: Actually, you are right, this is a valid point and we should check how big is the margin right now between the gas usage and gas limit at peak times and how likely/severe will the throttling be when we enforce compute limits if we were to declare 10x undercharging on operations like storage. I'll investigate this and also add a note in the NEP.

Also, CC @jakmeier for any additional thoughts on this

neps/nep-0455.md
akhi3030: Yes, I suppose there is a noisy neighbour effect with all proposals.  Thanks for adding it to the drawbacks section.

neps/nep-0455.md
jakmeier: Yes, a DoS or even a legitimate user using up all capacity becomes cheaper if compute costs > gas cost.
At the gas price floor, 1 NEAR can buy 10 chunks of a shard. That's about 5 NEAR for a minute, or 300 NEAR for an hour. (The gas price inflation is supposed to kick in at make it 20 more expensive but on a single shard it wouldn't work.)

A compute cost with a 10x difference to the gas cost would in theory mean that the cost for blocking it for an hour goes from 300 NEAR down to 30 NEAR. While that is significantly cheaper, it doesn't really enable new types of attacks. My main concern would be that it becomes cheaper to try such things "for fun".

But I don't think it is an argument against compute costs. It's just a fact we need to consider whenever we set the compute cost higher than the gas cost.

In practice, it often isn't as bad. The planned difference for TTN alongside the flat storage MVP for example is less critical because you cannot fill a receipt with only TTN costs, you will always have properly charged other storage costs and ~5Tgas overhead to even start a function call. So let's say we put a 10x difference there, probably the DoS only becomes 5x cheaper instead of 10x.

akhi3030: Thanks for the explanation.  I think this has been discussed enough and for what its worth, I don't think we need to modify the NEP.

jakmeier: Hm, I think a quick note in the NEP wouldn't hurt. It is a relevant discussion point and we should expect WG members that will vote on it to only read the NEP itself, not all discussions.

akashin: I agree that in practice this shoudn't be a problem, but we need to keep this in mind when we introduce compute weights and look how the behavior of network will be affected at the moments when it operates at capacity.

I've added a note about this in Drawbacks section.

frol:
@akashin I see that you are fleshing out the NEP. Is it ready for review?

akashin:
> @akashin I see that you are fleshing out the NEP. Is it ready for review?

There are no substantial changes happening to the NEP itself, I'm mostly adding clarifications based on the feedback on the PR for completeness. Is there any specific process that I need to follow there to avoid the confusion?

I believe NEP is ready for review and it would be great to get feedback and ACK from SMEs within the next week to make progress towards voting as we would like this change to land before [Flat Storage MVP](https://github.com/near/NEPs/pull/399).

jakmeier:
@mzhangmzz and @blasrodri I believe you are SMEs for this, looking at previous comments. Could you please give your recommendations as described in the technical guidelines in the expansion of this comment: https://github.com/near/NEPs/pull/455#issuecomment-1428116231

Thank you :)

blasrodri:
> @mzhangmzz and @blasrodri I believe you are SMEs for this, looking at previous comments. Could you please give your recommendations as described in the technical guidelines in the expansion of this comment: [#455 (comment)](https://github.com/near/NEPs/pull/455#issuecomment-1428116231)
> 
> Thank you :)

Will go over it tomorrow again

neps/nep-0455.md
akhi3030: lgtm

neps/nep-0455.md
```diff
@@ -0,0 +1,214 @@
...
+- Malicious undercharging attacks are unlikely to be prevented by this --- a single 10x undercharged receipt still needs to be processed using the old limit.
+Adjusting 0.1% per block means 100 chunks can only change by a maximum of 1.1x and 1000 chunks could change up to x2.7
+- Conflicts with transaction and receipt limit --- A transaction or receipt can (today) use up to 300Tgas.
+The effective limit per chunk is `gas_limit` + 300Tgas since receipts are added to a chunk until one exceeds the limit and the last receipt is not removed.
+Thus a gas limit of 0gas only reduces the effective limit from 1300Tgas to 300Tgas, which means a single 10x undercharged receipt can still result in a chunk with compute usage of 3 seconds (equivalent to 3000TGas)
+
+### Allow skipping chunks in the chain
+Slow chunk production can pose a liveness risk as the nodes expect a regular chunk production during normal operation.
+If processing the chunk takes more than expected, the block at the corresponding height will not be constructed in time and other nodes will start skipping this height effectively stalling the block production.
```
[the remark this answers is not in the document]
akashin: Thank you for clarifying, I've updated this paragraph!

neps/nep-0455.md
```diff
@@ -0,0 +1,214 @@
...
+This way will still ensure a regular block production, at a cost of lower throughput of the network in that shard.
+The chunk should still be included in a later block to avoid stalling the affected shard.
+
+Pros:
+- Fast and automatic adaptation to the blockchain workload
+
+Cons:
+- Needs a voting mechanism to decide that processing of the chunk is slow (as opposed to the chunk producer being slow) and the chunk should be skipped within a block
+- Needs economic design for the rewards for skipped chunks
+- No prior examples of this approach in other blockchains
```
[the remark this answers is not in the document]
akashin: That's fair, remove it.

neps/nep-0455.md
```diff
@@ -0,0 +1,213 @@
...
+## Motivation
+
+For NEAR blockchain stability, we need to ensure that blocks are produced regularly and in a timely manner.
+
+The chunk gas limit is used to ensure that the time it takes to validate a chunk is strictly bounded by limiting the total gas cost of operations included in the chunk.
+This process relies on accurate estimates of gas costs for individual operations.
+
+Underestimating these costs leads to *undercharging* which can increase the chunk validation time and slow down the chunk production.
+
+As a concrete example, during the incremental deployment of Flat Storage we will be going through intermediate stages where the storage operations will be undercharged.
```
blasrodri, line 25: why are you sure that this will happen?

blasrodri:
I'm happy with this NEP. My only concern/unaddressed comment is [this](https://github.com/near/NEPs/pull/455/files#r1105804309)

neps/nep-0455.md
```diff
@@ -0,0 +1,213 @@
...
+In practice, it is not trivial to exploit the undercharging unless you know the exact shape of the workload that realizes it.
+Also, after the compute cost is deployed, the undercharging should no longer be a threat for the network stability.
+
+## Drawbacks
+
+- Changing compute costs requires a protocol version bump (and a new binary release), limiting their use to undercharging problems that we're aware of
+
+- Updating compute costs is a manual process
+
+- A transaction or receipt can (today) use up to 300Tgas.
```
mzhangmzz, line 171: These sentences can be reordered. I think it would be better to say what the drawback is first and then mention why. For example,

"It may be hard to use this solution to prevent undercharging issues more than 3X undercharging factor from affecting the blockchain. This is because 1) a transaction or receipt today can use up to 300 TGas and 2) receipts are added to a chunk until one exceeds the limit and the last receipt is not removed. Therefore, effectively, a single receipt with 300TGas filled with undercharged operations that are more than 3X undercharged can cause the compute cost to be above 900TGas"

neps/nep-0455.md
```diff
@@ -0,0 +1,213 @@
...
+The effective limit per chunk is `gas_limit` + 300Tgas since receipts are added to a chunk until one exceeds the limit and the last receipt is not removed.
+Thus compute cost would not have a full effect on the last receipt in the chunk
+
+- Underchargings can still be exploited to lower the throughput of the network at unfair price and increase the waiting times for other users.
+This is inevitable for any proposal that doesn't change the gas costs and must be resolved by improving the performance or increasing the gas costs
+
+- The throttling due to compute usage limit can potentially occur for chunks that take less than 1.3s to process, as the average performance for undercharged operations will often be better than the worst case.
```
mzhangmzz, line 178: I am a little confused about what this sentence means, it probably needs some rewording.

neps/nep-0455.md
akashin: Good call, I've added a paragraph describing past underchargings and added a pointer to Flat Storage undercharging explanation below.

neps/nep-0455.md
akashin: I've reworded the sentence as you've suggested

neps/nep-0455.md
akashin: This is just noting the fact that the effective peak throughput of the network will decrease with introduction of compute costs because now the condition that limits the chunk compute usage is more strict than it was before (as compute costs are expected to be higher than the gas costs).
But thinking more about this, we already have chunks that use 1000TGas but take only 100ms to process, so it's not like we're introducing any new major drawback here.

I've rephrased this removing the mention of 1.3s which is not relevant here and explaining how this relates to the previous drawback.

neps/nep-0455.md
```diff
@@ -0,0 +1,219 @@
...
+- Updating compute costs is a manual process and requires deliberately looking for potential underchargings
+
+- The compute cost would not have a full effect on the last receipt in the chunk, decreasing its effectiveness to deal with undercharging.
+This is because 1) a transaction or receipt today can use up to 300TGas and 2) receipts are added to a chunk until one exceeds the limit and the last receipt is not removed.
+Therefore, a single receipt with 300TGas filled with undercharged operations with a factor of K can lead to overshooting the chunk compute limit by (K - 1) * 300TGas
+
+- Underchargings can still be exploited to lower the throughput of the network at unfair price and increase the waiting times for other users.
+This is inevitable for any proposal that doesn't change the gas costs and must be resolved by improving the performance or increasing the gas costs
+
+- Even without malicious intent, the effective peak throughput of the network will decrease when the chunks include undercharged operations (as the stopping condition based on compute costs for filling the chunk becomes stricter).
```
mzhangmzz, line 184: Why is this the case? How do you define effective throughput? I think effective throughput will not decrease, since compute cost actually represents the real processing cost. Say that a transaction's compute cost is 100TGas and its gas cost is 50TGas. With the proposal, the network can only process 10 transactions per block, instead of 20. But I don't think that should count as throughput decreasing, because if the blockchain can't include more than 10 transactions without downgrading its performance (increasing block time)

mzhangmzz (commented):
### Recommendation
I recommend this proposal to be approved. 

### Benefits
- Among the alternatives, this is the easiest to implement.
- It allows us to able to publicly discuss undercharging issues before they are fixed. 

### Concerns
I don't have any concerns that need to be addressed. The drawbacks listed in this NEP are minor compared to the benefits that it will bring. And implementing this NEP is strictly better than what we have today.

blasrodri:
> ### Recommendation
> I recommend this proposal to be approved.
> 
> ### Benefits
> * Among the alternatives, this is the easiest to implement.
> * It allows us to able to publicly discuss undercharging issues before they are fixed.
> 
> ### Concerns
> I don't have any concerns that need to be addressed. The drawbacks listed in this NEP are minor compared to the benefits that it will bring. And implementing this NEP is strictly better than what we have today.

I agree with this assessment.

akashin:
@near/nep-moderators , given that this NEP have been approved by reviewers, can we please move it to the voting stage?

frol:
As the moderator, I would like to thank @mzhangmzz and @blasrodri for completing the technical review. Based on your comments above, it seems like this proposal is ready for the working group members for review.

@near/wg-protocol  – Can you please fully read this NEP and comment in the thread if you are leaning towards approving or rejecting it? Please make sure to include your rationale and any feedback that you have for the author.

neps/nep-0455.md
```diff
@@ -0,0 +1,219 @@
...
+
+Cons:
+- Very broad granularity --- undercharging in one parameter affects all users, even those that never use the undercharged parts
+- Dependence on validator hardware --- someone running overspecced hardware will continuously want to increase the limit, others might run with underspecced hardware and continuously want to decrease the limit
+- Malicious undercharging attacks are unlikely to be prevented by this --- a single 10x undercharged receipt still needs to be processed using the old limit.
+Adjusting 0.1% per block means 100 chunks can only change by a maximum of 1.1x and 1000 chunks could change up to x2.7
+- Conflicts with transaction and receipt limit --- A transaction or receipt can (today) use up to 300Tgas.
+The effective limit per chunk is `gas_limit` + 300Tgas since receipts are added to a chunk until one exceeds the limit and the last receipt is not removed.
+Thus a gas limit of 0gas only reduces the effective limit from 1300Tgas to 300Tgas, which means a single 10x undercharged receipt can still result in a chunk with compute usage of 3 seconds (equivalent to 3000TGas)
```
abacabadabacaba, line 79: This affects parameter compute costs as well.

neps/nep-0455.md
abacabadabacaba: This solution is transparent to light clients. In the simplest version, it doesn't even require a protocol change, because chunks can be skipped already, for example, if the chunk producer is offline. Without a protocol change, however, chunk producers will be penalised for skipping chunks, so it may be desirable to distinguish between slow chunks and slow chunk producers, which would require some sort of voting. Or just consider missing chunks to be normal and not penalise chunk producers as much.
@akashin I suggest to add an explanation why voting (and a protocol change) is necessary with this solution.

bowenwang1996:
As a working group member, I lean towards approving this NEP. This NEP provides a way to mitigate the risks of undercharging and allow protocol developers to openly discuss undercharging issues without creating security risks. It would be nice to have a way to adjust compute costs dynamically, but I think this NEP is a good first step.

neps/nep-0455.md
```diff
@@ -0,0 +1,219 @@
...
+
+Compute costs influence the gas price adjustment logic described in https://nomicon.io/Economics/Economic#transaction-fees.
+Specifically, we're now using compute usage instead of gas usage in the formula to make sure that the gas price increases if chunk processing time is close to the limit.
+
+Compute costs **do not** count towards the transaction/receipt gas limit of 300TGas, as that might break existing contracts by pushing their method calls over this limit.
+
+Compute costs are static for each protocol version.
+
+### Using Compute Costs
+Compute costs different from gas costs are only a temporary solution.
```
mfornet, line 130: I appreciate the intention of having different gas cost from compute cost to be temporary, but we should assume that the temporal period can be large, or even at some points permanent.

I recommend suggesting in this PR a workflow for developers that plan to change the compute cost, about how they plan it to be fixed (i.e for making gas cost match compute cost again). Maybe opening an issue in NEPs repository is ok, or adding entries on a new section of this NEP. So we can keep track off all current differences, and see what is the remediation plan.

What are the implications of having this difference permanently? 
- Metrics will be affected. (i.e gas cost != compute cost so blocks will be seen emptier than they really are)
- We are currently exposing profiling information of tx execution through the RPC. Should we expose both compute cost and gas cost? (Not sure about this one).

mfornet (approved):
The PR looks good to me. I left some comments that are not blocking its approval.

ori-near:
As the moderator, I would like to thank @akashin for submitting this NEP, and for the @near/wg-protocol working group members for your review. Based on the voting indications above, it seems like this proposal is close to a decision. We are therefore scheduling the fourth Protocol Work Group call, where this NEP can enter the final voting stage. Anyone can discuss the technical details by adding your comments to this discussion. And join the call to learn about the final decision and how to get more involved.

**Meeting Info**
📅 Date: Thursday, March 16, 4 PM UTC
✏️ [Register](https://bit.ly/3FqqVjg)

mm-near:
I lean towards approving this NEP.

As other folks have mentioned before, it allows us to 'stop the bleeding' (when we discover the undercharging) and buys us time to do the proper fix.

The main disadvantage I see, is that it still requires a new release + protocol change -- which might take couple weeks.

neps/nep-0455.md
akashin: I agree, I've mentioned this in more detail in "Drawbacks" section down below. I wasn't able to come up with a solution that mitigates this problem.
At least we have some good estimates on how bad the total processing time can get based on the undercharging factor and the receipt limit of 300TGas.

ori-near:
Thank you to everyone who attended the fourth Protocol Working Group meeting last Thursday! The working group members reviewed the NEP and reached the following consensus:

**Status:** Approved

- https://github.com/near/NEPs/pull/455#issuecomment-1467023424
- https://github.com/near/NEPs/pull/455#pullrequestreview-1340887413
- https://github.com/near/NEPs/pull/455#issuecomment-1471882639

Meeting Recording:
https://youtu.be/4VxRoKwLXIs

@akashin and @jakmeier  Thank you for authoring this NEP!

Next Steps:
 -  NEP moderator to embed the benefits & concerns into the NEP document

frol (approved):

frol:
@akashin Are we ready to merge this NEP? I just want to double-check that there are no outstanding last-minute fixes.

neps/nep-0455.md
akashin: Makes sense to me, I've rephrased the point about on-chain voting mentioning the actual problem of distinguishing slow production from offline/malicious nodes

akashin:
@frol , I did a few more cleanup steps, folding a large set of older commits into one and answering all pending questions and I think now the PR is ready to be merged.

PR #456, opened 2023-01-30
Fix variable names in `select_validators`

akashin:
The previous name `block_producers` was not defined and also not generic enough as the algorithm would be used for both block producer selection and chunk-only producer selection, so using `validators` instead.

Also, the function returns 2 outputs, so reflecting this in signature.


frol (approved):
The changes make sense to me, and as a moderator, I rely on your knowledge of nearcore implementation as one of the maintainers, so I approve this PR.

PR #457, opened 2023-02-01
Fix inaccuracies about chunk gas limit

akashin:
This PR fixes two inaccuracies:
- The fact that chunk includes one more receipt after reaching the gas limit
- The current gas limit being 1000 TGas


frol (approved):
The changes make sense to me, and as a moderator, I rely on your knowledge of nearcore implementation as one of the maintainers, so I approve this PR.

PR #459, opened 2023-02-03
adding security implications and example scenarios

kenobijon:


neps/nep-0452.md
```diff
@@ -155,6 +189,14 @@ Below are some references for linkdrop contracts. A specific implementation of t
 - [Keypom Contract](https://github.com/keypom/keypom)
 
 
+## Security Implications
+
+1. Linkdrop Creation
+Linkdrop creation involves creating keypairs associated with mapping to assets such as $NEAR, FTs, NFTs and access keys. There are security implications with the access keys associated with the keypairs. These access keys should be function call access keys and be limited to specific functionality. For example, keypairs can be geneerated with function call access keys which have permission to only call `claim` and `create_account_and_claim`. Another important security implication of linkdrop creation is to ensure that only one key is mapped to a set of assets at any given time. Linkdrop creation can happen in real-time or prior to being received. The mass creation of linkdrops involves creating multiple keypairs associated with mapping to assets and access keys. The security implications mentioned above are increased with the mass creation of linkdrops.
```
BenKurrek, line 195:
```suggestion
Linkdrop creation involves creating keypairs associated with mapping to assets such as $NEAR, FTs, NFTs and access keys. There are security implications with the access keys associated with the keypairs. These access keys should be function call access keys and be limited to specific functionality. For example, keypairs can be generated with function call access keys which have permission to only call `claim` and `create_account_and_claim`. Another important security implication of linkdrop creation is to ensure that only one key is mapped to a set of assets at any given time. Linkdrop creation can happen in real-time or prior to being received. The mass creation of linkdrops involves creating multiple keypairs associated with mapping to assets and access keys. The security implications mentioned above are increased with the mass creation of linkdrops.
```

BenKurrek (changes requested):

BenKurrek (approved):

PR #461, opened 2023-02-14
NEP-461: Prefix Tag for Signed Messages

gagdiez:
A proposal to include a standardized prefix in all signed messages that are not transactions. 

Particularly, we propose to prepend the prefix $2^{30}$ + NEP-number actionable messages (i.e. those that are processed on-chain), and the prefix $2^{31}$ + NEP-number to all non-actionable messages (i.e. those destined for off-chain use).

---

### SME reviews: 

- [ ] @DavidM-D Waiting for SME review ([Assigned on 2023-03-27](https://github.com/near/NEPs/pull/461#issuecomment-1486252759))
- [ ] @robin-near Waiting for SME review ([Assigned on 2023-03-27](https://github.com/near/NEPs/pull/461#issuecomment-1486252759))

### Voting indications:
Protocol Work Group voting has not started yet.


gagdiez:
@abacabadabacaba would you like to be a coauthor of this NEP, since it was basically your proposal? Please let me know.

gagdiez:
Please notice that right now we are proposing only to use the prefix in non-transactional, in this way, we don't need to rush a protocol change, only change meta-transactions NEP a little bit

morgsmccauley:
Sorry @gagdiez misclicked and can't move it back to draft..

ori-near:
Hi @gagdiez – thank you for starting this proposal. As the moderator, we labeled this PR as "Needs author revision" because we assume you are still working on it since you submitted it in "Draft" mode.

Please ping the @near/nep-moderators once you are ready for us to review it. We will review it again in early April, unless we hear from you sooner. We typically close NEPs that are inactive for more than two months, so please let us know if you need more time.

abacabadabacaba:
@gagdiez Sure, and we will also need to make sure that any existing proposals that involve using account keys adhere to this scheme.

neps/nep-0461.md
```diff
@@ -0,0 +1,72 @@
...
+All actionable messages, i.e. those that ARE validated on-chain, should have a structure which first parameter is a `prefix: u32` with fixed value $2^{30} + NEP-number$.
+
+For example, if we implement a new actionable message, destinated to be used on-chain, and the proposal has the number `NEP400`, then the structure must start with the prefix `1073742224`:
+
+```ts
+Structure{
+    prefix: u32 = 1073742224 // prefix fixed to 2**30 + 400 == 1073742224
+    // Rest of the attributes go here
+}
+```
```
jakmeier, line 48: Two technical details:
1. This reads as if the prefix must be included in the message itself. In the implementation, it feels more natural to not include it in the actual message, only in the input for the signature.
2. Do you want to sign the hash of the message or the message itself?

More specifically, NEP-366 (meta transaction) defines signatures like this:
```rust
// NEP-366
let delegate_action = DelegateAction {...};
let signature = sign(hash(delegate_action));
let signed_delegate_action = SignedDelegateAction {
    delegate_action
    signature,
}
```

There are several ways how the prefix can be added. 
What I think you want, based on this proposal, is this:

```rust
// NEP-366 with your proposal
let delegate_action = DelegateAction {
   prefix: [ 0xFF,0xFF,0xFF,0xFF]; // this would be the prefix
   ... // here go all the other values
};
let signature = sign(delegate_action); // note the this has no hash
let signed_delegate_action = SignedDelegateAction {
    delegate_action
    signature,
}
```

But what feels more natural, at least for meta transactions, is this:
```rust
// NEP-366 with prefix inside hash
let delegate_action = DelegateAction {...};
// prefix is only prefixed for the hash and signature, not in the actual message
let signature = sign(hash( [ 0xFF,0xFF,0xFF,0xFF] +  delegate_action));
let signed_delegate_action = SignedDelegateAction {
    delegate_action
    signature,
}
```
Including the prefix in the message itself seems like unnecessary overhead. A receiver knows from the context what the message is supposed to be. The 4 bytes are unnecessary overhead to send over the wire, store on chain, etc. Therefore I would suggest to write in the NEP that this prefix is added implicitly and doesn't have to be included in the actual struct.

Regarding hashing or not hashing, I am not sure if there is any gain from the extra hash, ed25519 hashes all the bytes anyway. But if we want this proposal to be compatible with meta transactions (or vice-versa) either we should define to use the hash or change the meta transaction implementation which is about to be stabilized any moment now.

neps/nep-0461.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Security Implications
+
+By implementing this NEP, we can be sure that all new messages do not collide with classic transactions.
+
+## Drawbacks
+
+A protocol change would be needed to ensure that `actionable messages` are treated as such, and not be confused with classic transactions.
```
jakmeier, line 68: I don't think that is true. The point of picking these high numbers is to avoid any potential conflict with existing transactions.

neps/nep-0461.md
gagdiez: Indeed, this is a great point, I have changed the prefix to be implicit

neps/nep-0461.md
gagdiez: removed the sentence

neps/nep-0461.md
gagdiez: I meant the hash to still be there, since this is how traditionally signing works in NEAR. I made this more explicit now. Thanks for the feedback!

Ekleog-NEAR:
A few questions (disclaimer: I came from reviewing https://github.com/near/nearcore/pull/8578):
- Why should the discriminant be explicit for classic transactions, but implicit for other messages?
- What if a NEP needs to define two message types? IMO we should just "be IANA" and have one number, not directly related to the NEP number, per message kind. We could then decide to reserve one range for "experimentation purposes" so it could be used during experiments and not-yet-approved NEPs
- Why do we need a different range for on-chain messages and off-chain messages? (It might be easier for a human to distinguish whether the message is on- or off-chain by peering at bytes, but we can just have our "be IANA" policy actually do that without enforcing that in the range)

I would also like to suggest an alternative design for the same problem, which should simplify implementation. We could for now simply say:
> if a message starts with "N461" (which should be way out of the range of valid transactions already), then parse it as `Nep461Message`, which is a borsh-serializable `enum` that can get new variants with any NEP

This would cost us 4 more bytes for the binary format of NEP 366, but I don’t think it’d actually be a problem? It would also help us not have to worry that much about having multiple types of messages, though we could still decide to introduce new numbers (with us as the IANA registry as per my questions above) if we ever hit a use case that would benefit from reducing the size by 4 bytes. (Long-term we could maybe envision transitioning the current `Transaction` to also be encoded as an N461 variant, which could help us simplify our code again, but this would be outside of the scope of this alternative proposal)

Basically, it would be like saying "we are IANA, currently messages starting with 0-64 are Transactions and messages starting with "N461" are Nep461Message", so it would be as extensible as the solution suggested by my questions above.

WDYT?

frol:
Thank you @gagdiez for submitting this NEP.

As a moderator, I reviewed this NEP and it meets the proposed template guidelines. I am moving this NEP to the REVIEW stage and would like to ask the @near/wg-protocol working group members to assign 2 Technical Reviewers to complete a technical review (see expectations). If you want to assign yourself, please mention that you are acting as the Technical Reviewers.

<details>


  * A summary of benefits

```
### Recommendation

Add recommendation

### Benefits

* Benefit
* Benefit

### Concerns

| - | - | - | - |   
```

</details>

jakmeier:
Great thought and valid question Léo! I am also interested in the answers.

Regarding your proposal

> Basically, it would be like saying "we are IANA, currently messages starting with 0-64 are Transactions and messages starting with "N461" are Nep461Message", so it would be as extensible as the solution suggested by my questions above.

I see one technical disadvantage. When a new NEP is in its prototyping, approval, or testing phase, we already need a number. The IANA approach would only assign the number once approved, I assume. Libraries would want to include a way to construct unapproved messages, to make prototyping possible. The "NEP-number = message tag" approach makes this straight forward.

Ekleog-NEAR:
I think in the "we are IANA" approach, we could deal with that by setting a range that would be for experimentation purposes, similar to what was done with eg. ipv4 class E. While developing the messages would then use magic numbers from this class using whatever scheme the devs want to use, and as soon as the NEP lands only the constant magic number would need to change to the allocated one :)

abacabadabacaba:
@Ekleog-NEAR @gagdiez We can reserve the range starting with $2^{31}+2^{30}$ for experimentation/private use. Basically, say that no standard is going to use such prefixes.

Regarding NEP number vs we are IANA approach: I think that it is desirable to have less bureocracy. If a NEP needs multiple prefixes, we can allocate some, but this should be rare.

Regarding using enum for different message types: this will also serialize to different prefixes. Moreover, it is unlikely that any code will need to deserialize a message as an enum, because each piece of code is going to deal only with one message type. In fact, a prefixed message shouldn't be stored anywhere, the prefix should only be added to create/verify a signature. So, this is different from how enums are normally used.

gagdiez:
Were are we standing with this, I assume it was already used for the early implementation of meta transactions? @firatNEAR

neps/nep-0461.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Security Implications
+
+By implementing this NEP, we can be sure that all new messages do not collide with classic transactions.
+
+## Drawbacks
+
+We will need to update [NEP366](https://github.com/near/NEPs/blob/master/neps/nep-0366.md) to follow this standard, and make sure all future NEPs follow it, which might be hard to track.
+
+If at any point there is an update on how transactions are encoded, this NEP could stop being valid.
```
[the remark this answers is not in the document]
mfornet: I propose introducing a "protocol update" that enforces the first four bytes of every "classic tx" to be in the proper range. This is the case today (implicitly), but we should add it to the spec, and an extra assertion to the client (no real protocol change required) so we don't break this in the future.

mfornet (commented):
### Recommendation

I recommend approving this NEP.

>  IMO we should just "be IANA"

This introduces some complexity which could be problematic (security-wise) for apps that tries to encode different types of messages in the first four bytes. Using the NEP number seems like a straightforward way to use this proposal. 

### Benefits

It addresses a potential vector of attack that could be created in the feature unintentionally. It brings awareness to NEP developers and helps align them on a proper solution that fits current and future use cases.

### Concerns

| - | - | - | - |   
| 1 | We need to make sure classic tx encoding is not changed in the future. | | |

Ekleog-NEAR:
> This introduces some complexity which could be problematic (security-wise) for apps that tries to encode different types of messages in the first four bytes. Using the NEP number seems like a straightforward way to use this proposal.

Uhhh I’m not sure I understand your concern? The reason why I said that we should "be IANA" is exactly because using the NEP number introduces quite a bit of code complexity in security-sensitive code. To be more precise, all the bit-twiddling in https://github.com/near/nearcore/pull/8578 would have been replaced with a simple `if discriminant == [the number we assign to the meta-tx NEP] { ... }`.

It does add some process complexity when creating a new NEP (we need to give it one ID), but I think this is much less harmful security-wise than having more code complexity in areas of the code dealing with signatures.

If we really want to use the NEP number, then at least we should remove the part about on-chain and off-chain messages being in two different ranges, that just adds useless complexity to our numbering scheme… unless there is a reason for this that I cannot think of?

bowenwang1996:
As a working group member, I nominate @DavidM-D and @robin-near as SMEs to review this NEP.

robin-near:
Re: IANA vs NEP, we could also take the approach often used by Google and Meta - reserve a global ID (maybe by committing to a file in the repo, maybe an enum definition) even for experimentation, and the production ID will be the same. Some IDs will end up being dropped if the NEP is not ultimately implemented, but it's fine to have gaps anyway. If we stick with using NEP, we could also add one more byte to the prefix to allow multiple messages.

Either way, overall it seems that the primary concern is with how to make sure everybody is aware of the same scheme and sticks to the same scheme. Could we tackle that with changing the code-level APIs so that it is very difficult to implement new features that involve signing messages not conforming to the scheme that we decide on? Not being able to compile one's code will force them to look for what they are supposed to do.

frol:
@gagdiez Hey, did you intent to close this NEP?

gagdiez:
@frol not at all :huge_face_palm: I'll check if there's a simple way to restore it, worst case scenario I can open a new PR and we reference the discussion

neps/nep-0461.md
```diff
@@ -0,0 +1,79 @@
...
+In practice, what users sign is the `sha256` hash of a [borsh-serialized](https://borsh.io) structure. This is:
+
+```ts
+sha256.hash(Borsh.serialize(MessageStructure))
+```
+
+Including a large `u32` prefix allows to easily distinguish transactions from messages, ensuring that they never overlap.
+
+```ts
+sha256.hash(Borsh.serialize<u32>(large-number) + Borsh.serialize(MessageStructure))
```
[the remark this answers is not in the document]
gagdiez: indeed, I actually feel that other types of operations will only add confusion. In the text however I have changed it

PR #470, opened 2023-03-22
Update status of NEP-455 to Final

akashin:
As it's been voted on and accepted


frol:
@akashin Thanks for catching it!

PR #471, opened 2023-03-24
docs(template): move alternatives down

robert-zaremba:
When reading a NEP, it's usually hard to start with _alternatives_ before we read the specification and the main design.
It would be more logical (and more clear) if we push the _alternatives_ section down.

Also, we don't need to call it "Rationale and Altenratives", because rationale should be clearly covered in the "Motivation" section.

### DIFF
--- a/nep-0000-template.md
+++ b/nep-0000-template.md
@@ -22,7 +22,7 @@ Title: NEP title
 
 Author: List of author name(s) and optional contact info. Examples FirstName LastName <satoshi@fakenews.org>, FirstName LastName (@GitHubUserName)>
 
-Status: The NEP status -- New | Approved | Deprecated. 
+Status: The NEP status -- New | Approved | Deprecated.
 
 DiscussionsTo (Optional): URL of current canonical discussion thread, e.g. GitHub Pull Request link.
 
@@ -42,8 +42,6 @@ LastUpdated: The Created header records the date that the NEP was assigned a num
 
 See example above -->
 
-
-
 ## Summary
 
 [Provide a short human-readable (~200 words) description of the proposal. A reader should get from this section a high-level understanding about the issue this NEP is addressing.]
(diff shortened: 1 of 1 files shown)


nep-0000-template.md
```diff
@@ -64,11 +58,11 @@ See example above -->
...
-* Its interaction with other features is clear.
-* Where possible, include a Minimum Viable Interface subsection expressing the required behavior and types in a target programming language. (ie. traits and structs for rust, interfaces and classes for javascript, function signatures and structs for c, etc.)
-* It is reasonably clear how the feature would be implemented.
-* Corner cases are dissected by example.
-* For protocol changes: A link to a draft PR on nearcore that shows how it can be integrated in the current code. It should at least solve the key technical challenges.
+- Its interaction with other features is clear.
+- Where possible, include a Minimum Viable Interface subsection expressing the required behavior and types in a target programming language. (ie. traits and structs for rust, interfaces and classes for javascript, function signatures and structs for c, etc.)
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.
+- For protocol changes: A link to a draft PR on nearcore that shows how it can be integrated in the current code. It should at least solve the key technical challenges.
```
robert-zaremba, line 65: auto formatting

nep-0000-template.md
```diff
@@ -42,8 +42,6 @@ LastUpdated: The Created header records the date that the NEP was assigned a num
 
 See example above -->
 
-
-
```
robert-zaremba, line 46: auto formatted

frol (approved):
Looks reasonable to me 👍 

@ori-near What do you think?

ori-near:
@frol I agree. Thank you @robert-zaremba for thinking about how to make the NEP process/documentation better. We appreciate it.

PR #472, opened 2023-03-24
github action: add Markdown linter

robert-zaremba:
Integrate two linter to be run on PRs:
* [write good](https://github.com/btford/write-good) -- to check english language
* [markdownlint](https://github.com/DavidAnson/markdownlint) for linting Markdown files

I disabled some of the default rules (eg the max line length).
The current setup is that it only checks the modified files. I run it with all files, and there are many errors. 

I propose to firstly merge this one, and in a separate PR fix existing files. Most of the issues are auto-fixable.


robert-zaremba:
I disabled some of the default rules (eg the max line length).

The current setup is that it only checks the modified files. I run it with all files, and there are many errors. 

I propose to firstly merge this one, and in a separate PR fix existing files. Most of the issues are auto-fixable.

robert-zaremba:
It seams that write_good is integration is not good enough to analyze the whole repo - it takes too much time. Updating to use it only with `neps`

robert-zaremba:
removing `write_good` all together... too much false positives.

PR #474, opened 2023-04-03
update NEP template

robert-zaremba:
* added Consequences and merged into it "considerations"
* moved Alternatives and Future Possibilities above the Consequences / Considerations
* Updated Changelog Concerns table
* Added example of NEP Status section for a Pull Request description in the NEP process (`nep-0001.md`)


nep-0000-template.md
```diff
@@ -70,51 +70,67 @@ The section should return to the examples given in the previous section, and exp
 
 [Explicitly outline any security concerns in relation to the NEP, and potential ways to resolve or mitigate them. At the very least, well-known relevant threats must be covered, e.g. person-in-the-middle, double-spend, XSS, CSRF, etc.]
 
-## Drawbacks (Optional)
```
robert-zaremba, line 73: drawbacks is merged into consequences.

nep-0000-template.md
```diff
@@ -70,51 +70,67 @@ The section should return to the examples given in the previous section, and exp
...
+> Template for Subject Matter Experts review for this version:
+> Status: New | Ongoing | Resolved
 
-## Copyright
+|   # | Concern | Resolution | Status |
+| --: | :------ | :--------- | -----: |
+|   1 |         |            |        |
+|   2 |         |            |        |
 
-[copyright]: #copyright
```
robert-zaremba, line 118: CI linter complains about this line, it thinks that it should be an image here. Probably CI is right with this complain, because Github doesn't render this line.

neps/nep-0001.md
```diff
@@ -68,13 +68,39 @@ The NEP process begins when an author submits a [NEP draft](#submit-a-nep-draft)
...
+- [ ] Role1: @github-handle
+- [ ] Role2: @github-handle
+
+Contract Standards WG voting indications (❔ | 👍 | :-1: ):
+
+- ❔ @frol
+- ❔ @mfornet
+- ❔ @abacabadabacaba
+- ❔ @fadeevab
+- ❔ @robert-zaremba
```
robert-zaremba, line 96: not sure if in a template we should list the names here. It makes it easier to copy paste.

nep-0000-template.md
```diff
@@ -70,51 +70,67 @@ The section should return to the examples given in the previous section, and exp
 
 [Explicitly outline any security concerns in relation to the NEP, and potential ways to resolve or mitigate them. At the very least, well-known relevant threats must be covered, e.g. person-in-the-middle, double-spend, XSS, CSRF, etc.]
 
-## Drawbacks (Optional)
+## Alternatives
```
robert-zaremba, line 73: alternatives should be before final _Consequences_

neps/nep-0001.md
frol: These handles only represent Contract Standards WG members, so the copy paste argument is not justified. Let's replace it with a placeholder as it would be cumbersome to update the template every time we add a new WG and invite/remove WG members.

nep-0000-template.md
```diff
@@ -70,51 +70,67 @@ The section should return to the examples given in the previous section, and exp
...
 
 #### Concerns
 
-[Template for Subject Matter Experts review for this version:]
-| # | Concern | Resolution | Status |
-| - | - | - | - |  
-| 1 | | | |
-| 2 | | | |
+> Template for Subject Matter Experts review for this version:
+> Status: New | Ongoing | Resolved
```
frol, line 127: I don't see a need to separate "new" and "ongoing". Historically, we used "resolved" and "unresolved". Do you have concerns about those?

Note, most of the time, SMEs do not have permission to edit the NEP document (submitted PR).

frol (commented):
Overall, this change does not flag any concerns from my side. There are a couple of tweaks here and there, so I lean towards approving it.

I would love to hear inputs from NEP authors and @ori-near before merging it.

neps/nep-0001.md
robert-zaremba: OK

robert-zaremba: removed names from the template

ori-near (approved):
Thank you for proposing these changes @robert-zaremba. The template flows better this way.

PR #479, opened 2023-04-27
fix: I believe that zero-balance accounts are now finalised

akhi3030:
- marks the NEP as finalised
- Adds it to the list of NEPs on the README.md


akhi3030:
CC: @bowenwang1996

frol (approved):

Issue #478, opened 2023-04-26
[DISCUSSION] Account extensions


firatNEAR:
The main idea for this proposal is to increase the usability of the network by introducing new features that decrease the barrier of entry for end-users. 

TLDR: allow accounts to have more than one contract on them that can communicate in sync, where users/developers can `subscribe` to contracts by deploying just the hash of a smart contract instead of full binary to different namespaces on the account.

### Motivation
Currently we are faced with multiple problems:

- Composing contracts is non-trivial.
- Even if we compose contracts, storage staking for the contracts is economically infeasible for certain users.
- Certain contracts are deployed over and over again, still costing the same for each user.
- There is no way of `acting behalf of` another account in the current design.
- There is no way for contracts to call each other synchronously. 

#### Composing contracts is non-trivial

As an end-user, who doesn’t know how to write smart-contracts, there is no way to compose multiple contracts and deploy them onto your account. Why is this relevant? Imagine a case where a user wants to deploy a multi-sig contract on their account but also wants to have the dead man switch at the same time on the account. Currently there is no way that they can do this without either learning intricacies of smart-contract development or asking the community to develop a smart contract.

The contract upgrade workflow would also be improved by the introduction of composable contracts: an upgrade component could control the upgrades of individual components instead of the entire contract (which risks bricking an account if a bad contract is deployed).

#### Even if we compose contracts, storage staking for the contracts is economically infeasible for certain users.

Let’s assume that the end-user found a way to compose these contracts manually. [The multi-sig core contract](https://github.com/near/core-contracts/blob/dad58eb5f968c25913e746028ad63980506f5890/multisig2/res/multisig2.wasm) currently is 333kB, which costs ~3N in storage staking. If we keep composing contracts their size will grow and the cost of storage staking for the contract will also keep increasing. For some users, it is not feasible to keep this amount of NEAR on their accounts just to achieve these kinds of basic functionalities. 

#### Certain contracts are deployed over and over again, still costing the same for each user

Certain dApps require a ‘proxy’ contract to fully function due to the fact that there is no way of ‘acting on behalf of` another account without actually having a contract deployed on the account. Thus, dApps, such as Keypom, deploy the same contract on each of their users’ accounts. This means a huge amount of cost to the developers/businesses or a huge amount of cost on the end-users based on how the developers decide to put the costs on.

#### There is no way for contracts to call each other synchronously

[...]

PR #480, opened 2023-05-12
NEP-480: Account Namespaces

encody:
https://github.com/near/NEPs/issues/478

### DIFF
--- a/neps/nep-0480.md
+++ b/neps/nep-0480.md
@@ -52,6 +52,10 @@ Once a namespace is created, this NEP does not provide a way to delete it. Howev
 
 The `FunctionCall` action is modified to include a `namespace` field. If the `namespace` field is not specified in an RPC call, the default namespace (the empty string) is used. The function call is then executed on the code deployed to the specified namespace.
 
+### `FunctionCallPermission`
+
+A `namespaces` field is added to `FunctionCallPermission`. This field contains a list of namespaces that the access key is allowed to call. If the `namespaces` field is not specified, the access key is allowed to call all namespaces.
+
 ### NEAR VM host functions
 
 The following host functions are added to the NEAR host environment:
--- a/neps/nep-0480.md
+++ b/neps/nep-0480.md
@@ -7,7 +7,7 @@ DiscussionsTo: https://github.com/near/NEPs/pull/480
 Type: Protocol
 Version: 1.0.0
 Created: 2023-05-12
-LastUpdated: 2023-05-12
+LastUpdated: 2023-06-20
 ---
 
 ## Summary
@@ -70,8 +70,8 @@ There is no need for a `signer_namespace` function, because a signer is never a 
 
 Some existing host functions will need to be updated to include namespaces:
 
-- `promise_batch_action_function_call`: Updated to include namespace.
-- `promise_batch_action_deploy_contract`: Updated to include namespace.
+- `promise_batch_action_function_call`: Calls default namespace. Introduced `promise_batch_action_function_call_ns` to call a specific namespace.
+- `promise_batch_action_deploy_contract`: Targets default namespace. Introduced `promise_batch_action_deploy_contract_ns` to target a specific namespace.
 - etc.
 
 Existing host functions will generally still refer to the account as a whole:
--- a/neps/nep-0480.md
+++ b/neps/nep-0480.md
@@ -64,7 +64,7 @@ The following host functions are added to the NEAR host environment:
 - `current_namespace() -> Namespace`: Returns the namespace of the current account, which may be the empty namespace.
 - `namespace_storage_usage(&Namespace) -> u64`: Returns the storage usage of the given namespace. This includes storage consumed by the contract and state, but not account data (e.g. balance, access keys). That is to say,
 
-  $$\texttt{storage\_usage()} \ge \sum_{n \in \N} \texttt{namespace\_storage\_usage} (n)$$
+  $$\verb|storage_usage()| \ge \sum_{n} \verb|namespace_storage_usage| (n)$$
 
 There is no need for a `signer_namespace` function, because a signer is never a smart contract, and can therefore never have a non-empty namespace.
 
--- a/neps/nep-0480.md
+++ b/neps/nep-0480.md
@@ -62,7 +62,11 @@ The following host functions are added to the NEAR host environment:
 
 - `predecessor_namespace() -> Namespace`: Returns the namespace of the predecessor account, which may be the empty namespace. In the case that the predecessor is not a smart contract, this is the empty namespace.
 - `current_namespace() -> Namespace`: Returns the namespace of the current account, which may be the empty namespace.
-- There is no need for a `signer_namespace` function, because a signer is never a smart contract, and can therefore never have a non-empty namespace.
+- `namespace_storage_usage(&Namespace) -> u64`: Returns the storage usage of the given namespace. This includes storage consumed by the contract and state, but not account data (e.g. balance, access keys). That is to say,
+
+  $$\texttt{storage\_usage()} \ge \sum_{n \in \N} \texttt{namespace\_storage\_usage} (n)$$
+
+There is no need for a `signer_namespace` function, because a signer is never a smart contract, and can therefore never have a non-empty namespace.
 
 Some existing host functions will need to be updated to include namespaces:
 
@@ -73,12 +77,13 @@ Some existing host functions will need to be updated to include namespaces:
 Existing host functions will generally still refer to the account as a whole:
 
 - `storage_usage() -> u64`: Returns the storage usage of the current account, including all namespaces.
+- `account_balance() -> u128`: Returns the entire balance of the current account. There is no namespace-specific accounting.
 
 ### RPC view calls
 
 #### `view_account`
 
-The `code_hash` field contains the hash of the code deployed to the default namespace. The `code_hashes` field contains a map of namespaces to code hashes.
+The `code_hash` field contains the hash of the code deployed to the default namespace. The `code_hashes` field contains a map of namespaces to code hashes, including the default namespace. (This means that the hash of the code deployed to the default namespace is returned twice. This is intentional, for backwards compatibility.)
 
 #### `view_state`
 
@@ -103,7 +108,8 @@ WIP
 ## Unresolved Issues (Optional)
 
 - How to delete a namespace?
-- Backwards compatibility issues could be resolved with a routing table.
+- How to enumerate namespaces from within a smart contract?
+- Backwards compatibility issues could be resolved with a per-account routing table that maps incoming method names to a [namespace, method name] pair.
 
 ## Alternatives
 
--- a/neps/nep-0480.md
+++ b/neps/nep-0480.md
@@ -1,7 +1,7 @@
 ---
 NEP: 480
 Title: Account namespaces
-Authors: Jacob Lindahl <@encody>
+Authors: Jacob Lindahl <@encody>, Firat Sertgoz <@firatNEAR>
 Status: Draft
 DiscussionsTo: https://github.com/near/NEPs/pull/480
 Type: Protocol
--- a/neps/nep-0480.md
+++ b/neps/nep-0480.md
@@ -95,7 +95,7 @@ The `view_code` call now accepts an optional `namespace` field. If it is not spe
 
 ## Reference Implementation
 
-WIP
+- https://github.com/near/nearcore/pull/8890
 
 ## Security Implications
 
--- a/neps/nep-0480.md
+++ b/neps/nep-0480.md
@@ -1,9 +1,9 @@
 ---
-NEP: 0
+NEP: 480
 Title: Contract namespaces
 Authors: Jacob Lindahl <@encody>
 Status: Draft
-DiscussionsTo: https://github.com/nearprotocol/neps/pull/0000
+DiscussionsTo: https://github.com/near/NEPs/pull/480
 Type: Protocol
 Version: 1.0.0
 Created: 2023-05-12
--- a/neps/nep-0480.md
+++ b/neps/nep-0480.md
@@ -1,6 +1,6 @@
 ---
 NEP: 480
-Title: Contract namespaces
+Title: Account namespaces
 Authors: Jacob Lindahl <@encody>
 Status: Draft
 DiscussionsTo: https://github.com/near/NEPs/pull/480
@@ -12,7 +12,7 @@ LastUpdated: 2023-05-12
 
 ## Summary
 
-This proposal introduces contract namespaces, which enable deploying multiple WASM binaries to a single account. A WASM binary is deployed to a unique, deployer-selected namespace. Subsequent deployments to the same namespace overwrite the code currently deployed to that namespace. The empty namespace is considered the default namespace.
+This proposal introduces account namespaces, which enable deploying multiple WASM binaries to a single account. A WASM binary is deployed to a unique, deployer-selected namespace. Subsequent deployments to the same namespace overwrite the code currently deployed to that namespace. The empty namespace is considered the default namespace.
 
 A WASM binary deployed to a non-default (non-empty) namespace can be invoked explicitly by specifying the namespace in the action. If no namespace is specified, the default namespace is used.
 
@@ -126,6 +126,8 @@ See [the Account Extensions upgrade master issue](https://github.com/near/NEPs/i
 
 ## Copyright
 
-[copyright]: #copyright
-
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+## References
+
+- https://gov.near.org/t/proposal-account-extensions-contract-namespaces/34227
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Changes to the protocol specification and standards are called NEAR Enhancement 
 | [0399](https://github.com/near/NEPs/blob/master/neps/nep-0399.md) | Flat Storage                             | @Longarithm @mzhangmzz                      | Review |
 | [0448](https://github.com/near/NEPs/blob/master/neps/nep-0448.md) | Zero-balance Accounts                    | @bowenwang1996                              | Final  |
 | [0455](https://github.com/near/NEPs/blob/master/neps/nep-0455.md) | Parameter Compute Costs                  | @akashin @jakmeier                          | Final  |
+| [0480](https://github.com/near/NEPs/blob/master/neps/nep-0480.md) | Account Namespaces                       | @encody                                     | Draft  |
 
 ## Specification
 
--- a/neps/nep-0000.md
+++ b/neps/nep-0000.md
@@ -1,0 +1,131 @@
+---
+NEP: 0
+Title: Contract namespaces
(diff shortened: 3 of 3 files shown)


neps/nep-0480.md
```diff
@@ -0,0 +1,137 @@
...
+
+Some existing host functions will need to be updated to include namespaces:
+
+- `promise_batch_action_function_call`: Updated to include namespace.
+- `promise_batch_action_deploy_contract`: Updated to include namespace.
+- etc.
+
+Existing host functions will generally still refer to the account as a whole:
+
+- `storage_usage() -> u64`: Returns the storage usage of the current account, including all namespaces.
```
akhi3030, line 75: do we need a way to get the storage used by the current namespace?

neps/nep-0480.md
```diff
@@ -0,0 +1,137 @@
...
+
+### RPC view calls
+
+#### `view_account`
+
+The `code_hash` field contains the hash of the code deployed to the default namespace. The `code_hashes` field contains a map of namespaces to code hashes.
```
akhi3030, line 81: Should this clarify if `code_hashes` map will also include the default namespace or not?

neps/nep-0480.md
```diff
@@ -0,0 +1,137 @@
...
+- Applications that detect smart contract updates by tracking the `code_hash` field from a `view_account` call will fail to account for updates to namespaces. (Note that the correct way to track code changes is by monitoring the blockchain for `DeployContract` actions targeting the account, not by tracking `code_hash`.)
+- As described by this NEP, namespaces all have full permission to act on behalf of the account, just as any smart contract.
+- If a namespaced contract interacts with a legacy contract (unaware of namespaces), it is possible that the legacy contract may save the account ID of the namespaced contract, but not the namespace. If the legacy contract subsequently attempts to interact with the namespaced contract, it will only be able to interact with the contract deployed to the default namespace instead. However, this is equivalent to the case in which an non-contract account signs the same set of actions to the legacy contract.
+
+## Drawbacks (Optional)
+
+## Unresolved Issues (Optional)
+
+- How to delete a namespace?
+- Backwards compatibility issues could be resolved with a routing table.
```
akhi3030, line 106: Do you plan on describing what a routing table is in this context?

akhi3030 (commented):
Should the NEP also specify how the token balance is shared or not between namespaces?

encody:
@near/nep-moderators This should be ready for another look!

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Security Implications
+
+- Applications that detect smart contract updates by tracking the `code_hash` field from a `view_account` call will fail to account for updates to namespaces. (Note that the correct way to track code changes is by monitoring the blockchain for `DeployContract` actions targeting the account, not by tracking `code_hash`.)
+- As described by this NEP, namespaces all have full permission to act on behalf of the account, just as any smart contract.
+- If a namespaced contract interacts with a legacy contract (unaware of namespaces), it is possible that the legacy contract may save the account ID of the namespaced contract, but not the namespace. If the legacy contract subsequently attempts to interact with the namespaced contract, it will only be able to interact with the contract deployed to the default namespace instead. However, this is equivalent to the case in which an non-contract account signs the same set of actions to the legacy contract.
+
+## Drawbacks (Optional)
+
+## Unresolved Issues (Optional)
```
frol, line 108:
```suggestion
## Drawbacks

## Unresolved Issues
```

bowenwang1996:
As a working group member, I'd like to nominate @akhi3030 and @nagisa as SMEs to review this NEP.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+- If a developer wants their smart contract to be upgradable, they have to manually write the upgrade functionality into their smart contract, and smart contract upgrades are complex, multi-step operations with many security considerations and opportunities to "brick" the account. Namespaces would allow a developer to deploy a semi-standardized upgrade module to a namespace, and have that manage the upgrade process for the other (default, etc.) namespaces.
+
+As this is a part of the [Account Extensions upgrade](https://github.com/near/NEPs/issues/478), some of the benefits of account namespaces are realized in conjunction with other proposals from the upgrade:
+
+- Certain contracts are frequently deployed on different accounts. "Codeless contracts" allow an account to deploy just the hash of a smart contract instead of the full binary. This, in conjunction with namespaces, would allow a user to pick and choose a set of account features to deploy with negligible gas and storage costs.
```
nagisa, line 31: This seems like an improvement that could be made and proposed independently of the account namespaces. Two reasons I would like this to happen:

1. I think combining the two features into one NEP makes this NEP substantially less self-contained and quite a bit harder to understand;
2. Even if the account namespaces’ proposal changes and evolves as part of coordination with the synchronous calls NEP, this seems like an improvement can largely stand on its own merit.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+We introduce a new primitive type: `Namespace`. A `Namespace` is a string that contains only lowercase letters, numbers, and underscores. The empty string is considered the default namespace.
+
+### `DeployContract` action
+
+The `DeployContract` action is modified to include a `namespace` field. If the `namespace` field is not specified in an RPC call, the default namespace (the empty string) is used. The attached code is then deployed to the specified namespace.
+
+If code has already been deployed to the specified namespace, it is replaced, leaving code deployed to other namespaces untouched.
+
+Once a namespace is created, this NEP does not provide a way to delete it. However, namespaces should not be considered permanent.
```
nagisa, line 49: Is there a reason why we’d need a concept of namespace creation? Would this creation step introduce semantics meaningfully different from the empty namespace existing for accounts that haven't deployed a contract yet?

If not, I think it would be quite a bit more straightforward to work with this feature if all namespaces were presumed to “exist” with a code_hash equal to `11111111111111111111111111111111` or whatever we use for invalid contract.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+### `FunctionCall` action
+
+The `FunctionCall` action is modified to include a `namespace` field. If the `namespace` field is not specified in an RPC call, the default namespace (the empty string) is used. The function call is then executed on the code deployed to the specified namespace.
+
+### `FunctionCallPermission`
+
+A `namespaces` field is added to `FunctionCallPermission`. This field contains a list of namespaces that the access key is allowed to call. If the `namespaces` field is not specified, the access key is allowed to call all namespaces.
```
nagisa, line 57: Is there a reason I’m not seeing why this deviates from the usual “no namespace ⇒ empty namespace” defaulting elsewhere in this NEP? I feel like it might be a notable implicit foot gun in the use of protocol. Consider for example:

1. Following a tutorial I deploy a contract to the empty (default) namespace;
2. Since the tutorial is namespace oblivious, they suggest adding a key without `namespace`;
3. Later on I deploy a different contract that requires strict access permissions to a namespace.

Forgetting I’ve done 2, I have accidentally allowed the important contract to be called arbitrarily by the old key(s).

The alternative is to equate “empty `namespaces` ⇒ default empty namespace” here as well, in which case the failure mode is that I attempt to call the new contract with an old key, notice it doesn't work and expand the permissions for the key.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+### NEAR VM host functions
+
+The following host functions are added to the NEAR host environment:
+
+- `predecessor_namespace() -> Namespace`: Returns the namespace of the predecessor account, which may be the empty namespace. In the case that the predecessor is not a smart contract, this is the empty namespace.
+- `current_namespace() -> Namespace`: Returns the namespace of the current account, which may be the empty namespace.
+- `namespace_storage_usage(&Namespace) -> u64`: Returns the storage usage of the given namespace. This includes storage consumed by the contract and state, but not account data (e.g. balance, access keys). That is to say,
```
nagisa, line 65: What would be the value of obtaining the storage usage of a different namespace than the one currently executing? If there isn’t one, might it make sense to require that the `Namespace` arugment is `current_namepsace()` otherwise a panic is raised for the time being?

Might be worthwhile to double check if an ability to do so does not preclude whatever other improvements we might want to make.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+  $$\verb|storage_usage()| \ge \sum_{n} \verb|namespace_storage_usage| (n)$$
+
+There is no need for a `signer_namespace` function, because a signer is never a smart contract, and can therefore never have a non-empty namespace.
+
+Some existing host functions will need to be updated to include namespaces:
```
nagisa, line 71: Unfortunately, I don’t think we can update functions easily as that would result in a ABI change which is relied upon by near-sdk et al. We’d need to introduce new ones, and possibly delegate to them from the old functions.

nagisa (commented):
This is an initial pass and largely just my brain dump.

neps/nep-0480.md
encody: Yes, that's what it says: this NEP _does not_ propose adding codeless contracts, but the combination of codeless contracts + namespaces is potentially really powerful.

neps/nep-0480.md
encody: Namespace creation is approximately equal to contract deployment in this NEP. If all namespaces are presumed to exist, how does an RPC account view call show namespaces?

neps/nep-0480.md
encody: Right now, a function call access key that does not specify a list of methods is allowed to call any method. Therefore, a function call access key that does not specify a list of namespaces is allowed to call any namespace.

neps/nep-0480.md
encody: Noted

neps/nep-0480.md
nagisa: Right. I feel like the stakes are likely somewhat different here in that deploying a new contract to an account in absence of namespaces overwrites the old one, and thus the user is conscious about both the previous and the new contract (and thus possibly is thinking of the access controls in context of both.) In the world of namespaces they no longer need to care about any previous state anymore.

But fair, I see how the proposed behaviour is consistent.

neps/nep-0480.md
nagisa: Hm, I'm not sure! I imagine it would list the non-`111...111` hashes, rather than the list of “namespaces”?

But counter-question to your question: what's the expected behaviour of calling a non-existent namespace? If my memory serves me right, in the namespace-less world of today, the behaviour is meaningfully different between calling a function in a non-existent account and calling a function in an existing account without a contract deployed.

akhi3030:
> As a working group member, I'd like to nominate @akhi3030 and @nagisa as SMEs to review this NEP.

@near/nep-moderators: looks like SMEs are assigned so maybe we need to update the labels on the NEP.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+- If a developer wants their smart contract to be upgradable, they have to manually write the upgrade functionality into their smart contract, and smart contract upgrades are complex, multi-step operations with many security considerations and opportunities to "brick" the account. Namespaces would allow a developer to deploy a semi-standardized upgrade module to a namespace, and have that manage the upgrade process for the other (default, etc.) namespaces.
+
+As this is a part of the [Account Extensions upgrade](https://github.com/near/NEPs/issues/478), some of the benefits of account namespaces are realized in conjunction with other proposals from the upgrade:
+
+- Certain contracts are frequently deployed on different accounts. "Codeless contracts" allow an account to deploy just the hash of a smart contract instead of the full binary. This, in conjunction with namespaces, would allow a user to pick and choose a set of account features to deploy with negligible gas and storage costs.
+
+- Permissions: It could be possible to set permissions on a particular namespace. For example, restricting the namespace `dao_multisig` to interactions with `dao.near` only.
```
akhi3030, line 33: Are these the only capabilities that you propose to enable?  What about arbitrary host functions, etc.?  Having read through the proposal in more detail, I think it is important to have fine grained capabilities for namespaces.  What you are proposing is to allow smart contract developers who are not proficient to deploy multiple wasm modules on their account.  These wasm modules could be buggy or malicious.  If this proposal is expected, I can imagine some users just having a single account with all the different contracts they want to try out / experiment with and without fine grained control over their capabilities, they could end up losing a lot of funds.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+Version: 1.0.0
+Created: 2023-05-12
+LastUpdated: 2023-05-12
+---
+
+## Summary
+
+This proposal introduces account namespaces, which enable deploying multiple WASM binaries to a single account. A WASM binary is deployed to a unique, deployer-selected namespace. Subsequent deployments to the same namespace overwrite the code currently deployed to that namespace. The empty namespace is considered the default namespace.
+
+A WASM binary deployed to a non-default (non-empty) namespace can be invoked explicitly by specifying the namespace in the action. If no namespace is specified, the default namespace is used.
```
akhi3030, line 17:
```suggestion
A WASM binary deployed to a non-default (non-empty) namespace can be invoked explicitly by specifying the namespace in the action. If no namespace or the empty string is specified, then the default namespace is used.
```

Based on the information below in the NEP.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+### `FunctionCallPermission`
+
+A `namespaces` field is added to `FunctionCallPermission`. This field contains a list of namespaces that the access key is allowed to call. If the `namespaces` field is not specified, the access key is allowed to call all namespaces.
+
+### NEAR VM host functions
+
+The following host functions are added to the NEAR host environment:
+
+- `predecessor_namespace() -> Namespace`: Returns the namespace of the predecessor account, which may be the empty namespace. In the case that the predecessor is not a smart contract, this is the empty namespace.
```
akhi3030, line 63: Hmm... so we are using the empty namespace to represent two different values: empty namespace in the predecessor or no smart contract.  Could this be a problem?  Is there a good way to distinguish between the two?

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+Some existing host functions will need to be updated to include namespaces:
+
+- `promise_batch_action_function_call`: Updated to include namespace.
+- `promise_batch_action_deploy_contract`: Updated to include namespace.
+- etc.
+
+Existing host functions will generally still refer to the account as a whole:
+
+- `storage_usage() -> u64`: Returns the storage usage of the current account, including all namespaces.
```
akhi3030, line 79: In the current proposal, then there are two ways of getting storage used.  `storage_usage()` and `namespace_storage_usage()`.  I can see some developers getting confused when `storage_usage()` does not add up to what they are seeing the current namespace is actually using (they might have forgotten that other namespaces exist).

In the ideal world, it would be great to have a single hostfunction with a signature like: `storage_usage(enum {EntireContract; Namespace(namespace)'})`.  Does something like this make sense?

If it does, I propose that we update the proposed `namespace_storage_usage()` to be something like the above and then mark `storage_usage()` as deprecated.  We probably never will be able to remove it but hopefully with sufficient documentation, we will minimise the confusion.

neps/nep-0480.md
encody: Maybe we need a new syntax for function call access keys, wherein they can be permitted to call "any method _on `xyz` namespace_"?

neps/nep-0480.md
encody: I'm not sure how all of the questions in this comment fit together.

Currently, I don't think that the typical non-tech-savvy user deploys any code to their accounts, but please correct me if I am wrong.

Specifically when speaking of non-tech-savvy users, the workflow we conceptualized is something like:
1. Alice wants to allow Bob to access her funds if she is inactive for 180 days.
2. Alice logs into her wallet.
3. Alice clicks on "Account Extensions."
4. Alice scrolls through the list of available extensions and selects "Dead Man Switch."
5. Alice inputs parameters:
    - Beneficiary ID: `bob.near`
    - Timeframe: 180 days
6. Alice clicks "Save."
7. Alice approves the contract deployment + initialization function call action:
    - Namespace: `dead-man-switch`
    - Code: `0x...`

neps/nep-0480.md
encody: That the empty namespace is the default namespace is specified elsewhere, but added for clarity.

neps/nep-0480.md
encody: I think we usually don't care too much whether a call comes from an EOA or contract, but if the distinction _is_ pertinent, one may use `env::signer_account_id()`.

neps/nep-0480.md
encody: I don't think I have enough information to have an informed opinion here. Making a decision one way or another has the potential to break contracts: if `storage_usage()` only references the current namespace, old contracts could incorrectly think the account is not using storage when it is, and if it references the whole account's usage (all namespaces), then old contracts could incorrectly think the current contract is using storage when it is not.

neps/nep-0480.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Motivation
+
+- Composing contracts is non-trivial. Currently, it is possible to deploy a single smart contract to a NEAR account using a `DEPLOY_CONTRACT` action. Subsequent `DEPLOY_CONTRACT` actions completely replace existing code with new code, while leaving all storage intact.
+
+- It is impossible for a non-developer user to compose multiple contract standards together into a single contract without actually performing some development work. Account namespaces would enable a user workflow wherein a user only needs to choose which contracts to deploy, and not worry about how to compose them.
+
+- If a developer wants their smart contract to be upgradable, they have to manually write the upgrade functionality into their smart contract, and smart contract upgrades are complex, multi-step operations with many security considerations and opportunities to "brick" the account. Namespaces would allow a developer to deploy a semi-standardized upgrade module to a namespace, and have that manage the upgrade process for the other (default, etc.) namespaces.
```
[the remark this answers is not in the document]
encody: Workflow example:
1. Alice creates `alice.near`.
2. Alice deploys main contract bytecode `v1` to the default namespace.
3. Alice deploys bytecode `upgrade-manager` to the `upgrade-manager` namespace. The `upgrade-manager` bytecode provides an `upgrade` method that performs a contract upgrade sequence (explained in later steps).
4. Alice deletes all full-access keys from `alice.near`.
5. Alice creates a new version of the main contract bytecode. She calls this version `v2`.
6. Alice calls `alice.near:upgrade-manager->upgrade(v2)`, passing in the `v2` bytecode.
7. The `upgrade-manager` code performs a contract upgrade sequence, deploying the `v2` bytecode to the default namespace of `alice.near`.
8. Turns out, the `v2` bytecode is invalid/unusable.
9. Alice creates another new version of the main contract bytecode, `v3`.
10. Alice calls the _still-functional_ `upgrade-manager` to perform the upgrade sequence with `v3`.
11. The account has a valid contract deployed to the default namespace, and continues to work as intended.

If namespaces did not exist, the `v2` deployment would have bricked the entire `alice.near` account.

walnut-the-cat:
Hello- NEP moderator here.

@encody , is it still ongoing project?

encody:
@walnut-the-cat This project is not currently undergoing active development.

PR #491, opened 2023-07-25
NEP: Non-refundable storage

jakmeier:
DRAFT ONLY

### DIFF
--- a/neps/nep-xxxx.md
+++ b/neps/nep-xxxx.md
@@ -1,0 +1,349 @@
+---
+NEP: TODO
+Title: Non-Refundable Storage Staking
+Authors: Jakob Meier <jakob@near.org>
+Status: Draft
+DiscussionsTo: https://gov.near.org/t/proposal-locking-account-storage-refunds-to-avoid-faucet-draining-attacks/34155
+Type: Protocol Track
+Version: 1.0.0
+Created: 2023-07-24
+LastUpdated: 2023-07-24
+---
+
+
+## Summary
+
+Non-refundable storage allows to create accounts with arbitrary state for users,
+without being susceptible to refund abuse.
+
+This is done by tracking non-refundable balance in a separate field of the
+account. This balance is only useful for storage staking and otherwise can be
+considered burned.
+
+
+## Motivation
+
+Creating new accounts on chain costs a gas fee and a storage staking fee. The
+more state is added to the account, the higher the storage staking fees. When
+deploying a contract on the account, it can quickly go above 1 NEAR per account.
+
+Some business models are okay with paying that fee for users upfront, just to
+get them onboard. However, it turns out if you do that today, the user can
+delete their new account and refund the storage staking fee to another account
+in their control. Thereby taking advantage of the business who sponsored their
+account.
+
+While spam avoidance is best handled at the application level, the protocol
+should at least allow to create accounts in a way that is not susceptible to
+such refund abuse. This would at least change the incentives such that creating
+fake users is no longer profitable.
+
+Non-refundable storage staking is a further improvement over
+[NEP-448](https://github.com/near/NEPs/pull/448) (Zero Balance Accounts) which
+addressed the same issue. But the limit on the state size of accounts makes it
+practically impossible to use sponsored accounts with contracts deployed on
+them.
+
+## Specification
+
+Users can opt-in to nonrefundable storage when creating new accounts. For that,
+we extend the `Transfer` action with a new field:
+
+```rust
+struct Transfer {
+    pub deposit: Balance,
+    /// If this flag is set, the balance will be added to the receiver's non-refundable balance.
+    pub nonrefundable: bool,
+}
+```
+
+To create a named account today, the typical pattern is a transaction with
+`CreateAccount`, `Transfer`, and `AddKey`. To make the funds nonrefundable, one
+can use the same pattern with the non-refundable flag set to true:
+```json
+"Actions": {
+  "CreateAccount": {},
+  "Transfer": { "deposit": "1000000000000000000000000", "nonrefundable": true },
+  "AddKey": { "public_key": ..., "access_key": ... }
+}
+```
+
+It is also possible it combine nonrefundable balance and normal (refundable)
+balance in the same account creation by adding a second `Transfer` action.
+```json
+"Actions": {
+  "CreateAccount": {},
+  "Transfer": { "deposit": "1000000000000000000000000", "nonrefundable": true },
+  "Transfer": { "deposit": "1000000000000000000000000", "nonrefundable": false },
+  "AddKey": { "public_key": ..., "access_key": ... }
+}
+```
+
+To create implicit accounts, the current protocol requires a single `Transfer`
+action without further actions in the same transaction. Setting it to
+non-refundable will make the full balance non-refundable. If users want to add
+refundable balance, it must done in a second transaction.
+
+If a non-refundable transfer arrives at an account that already exists, it will
+fail and the funds are returned to the predecessor.
+
+Finally, when querying an account for its balance, there will be an additional
+field `nonrefundable` in the output. Wallets will need to decide how they want
+to show it. They could, for example, add a new field called "non-refundable
+storage credits".
+```js
+// Account near
+{
+  "amount": "68844924385676812880674962949",
+  "block_hash": "3d6SisRc5SuwrkJnLwQb3W5pWitZKCjGhiKZuc6tPpao",
+  "block_height": 97314513,
+  "code_hash": "Dmi6UTRYTT3eNirp8ndgDNh8kYk2T9SZ6PJZDUXB1VR3",
+  "locked": "0",
+  "storage_paid_at": 0,
+  "storage_usage": 2511772,
+  "formattedAmount": "68,844.924385676812880674962949",
+  // this is new
+  "nonrefundable": "0"
+}
+```
+
+
+## Reference Implementation
+
+On the protocol side, the change to the `Transfer` struct requires changing the
+serialization format. To ensure backwards compatibility, we add a new variant to
+`enum Action`:
+
+```rust
+enum Action {
+  CreateAccount(CreateAccountAction), 
+  DeployContract(DeployContractAction),
+  FunctionCall(FunctionCallAction),
+  // kept for backwards compatibility but new clients should only use `TransferV2`
+  Transfer(TransferAction),
+  Stake(StakeAction),
+  AddKey(AddKeyAction),
+  DeleteKey(DeleteKeyAction),
+  DeleteAccount(DeleteAccountAction),
+  Delegate(super::delegate_action::SignedDelegateAction),
+  // this gets added in the end
+  TransferV2(TransferActionV2),
+}
+```
+
+Further, we have to update the account meta data representation in the state
+trie to track the non-refundable storage.
+
+```rust
+pub struct Account {
+    amount: Balance,
+    locked: Balance,
+    // this field is new
+    nonrefundable: Balance,
+    code_hash: CryptoHash,
+    storage_usage: StorageUsage,
+    // the account version will be increased from 1 to 2
+    version: AccountVersion,
+}
+```
+
+The field `nonrefundable` must be added to the normal `amount` and the `locked`
+balance calculate how much state the account is allowed to use. The new formula
+to check storage balance therefore becomes
+```rust
+amount + locked + nonrefundable >= storage_usage * storage_amount_per_byte
+```
+
+For old accounts that don't have the new field, the non-refundable balance is
+always zero.
+
+Conceptually, these are all changes on the protocol level. However,
+unfortunately, the account version field is not currently serialized, hence not
+included in the on-chain state.
+
+Therefore, as the last change necessary for this NEP, we also introduce a new
+serialization format new accounts.
+```rust
+    // new: prefix with a sentinel value to detect V1 accounts, they will have
+    //      a real balance here which is smaller than u128::MAX
+    writer.serialize(u128::MAX)?;
+    // new: include version number (u8) for accounts with version 2 or more
+    writer.serialize(version)?;
+    writer.serialize(amount)?;
+    writer.serialize(locked)?;
+    writer.serialize(code_hash)?;
+    writer.serialize(storage_usage)?;
+    // new: this is the field we added, the type is u128 like other balances
+    writer.serialize(nonrefundable)?;
+```
+
+Note that we are not migrating old accounts. Accounts created as version 1 will
+remain at version 1. This
+
+A proof of concept implementation for nearcore is available in this PR: [TODO]
+insert link to PoC PR
+
+## Security Implications
+
+There is a risk for users to accidentally create accounts with nonrefundable
+storage and lose the balance as a consequence. This is mitigated because the
+default behavior with `nonrefundable = false` will behave exactly like a normal
+transfer and not have this problem.
+
+Otherwise, we were not able to come up with security relevant implications.
+
+## Alternatives
+
+There are small variations in the implementation, and there completely different
+ways to look at the problem. Let's start with the small variations first.
+
+### Variation: Allow adding nonrefundable balance to existing accounts
+
+Instead of failing when a non-refundable transfer arrives at an existing
+account, we could add the balance to the existing non-refundable balance. This
+would be more user flexible to use, as a business can easily add more funds for
+storage even after account creation.
+
+The main problems with that are in implementation details. It would allow to add
+non-refundable storage to existing accounts, which would require some form of
+migration of the all accounts in the state trie.
+
+We could maybe do it lazily, i.e. read account version 1 and automatically
+convert it to version 2. However, that would break the assumption that every
+logical value in the merkle trie has a unique borsh representation, as there
+would be a account version 1 and a version 2 borsh serialization that both map
+to the same logical version 2 value.
+
+It is not 100% clear to me, the author, if this is a problem we could work
+around. However, the complications it would involve do not seem to be worth it,
+given that in the feature discussions nobody saw it as critical to add
+non-refundable balance to existing accounts.
+
+Another consideration is user friendliness. When someone wants to create a new
+implicit account but the account already exists, the current design will fail
+and return the tokens. In the alternative, this tokens would be added to the
+refundable storage and no longer be retrievable.
+
+### Variation: Allow refunds to original sponsor
+
+Instead of complete non-refundability, the tokens reserved for storage staking
+could be returned to the original account that created the account when an
+account is deleted.
+
+The community discussions ended with the conclusion that this feature would
+probably not be used and we should not implement it until there is real demand
+for it.
+
+### Alternative: Don't use smart contracts on user accounts
+
+Instead of deploying contracts on the user account, one could build a similar
+solution that uses zero balance accounts and a single master contract that
+performs all smart contract functionality required. This master contract can
+implement the [Storage Management]
+(https://nomicon.io/Standards/StorageManagement) standard to limit storage usage
+per user.
+
+This solution is not as flexible. The master account cannot make cross-contract
+function calls with the user id as the predecessor.
+
+### Alternative: Move away from storage staking
+
+We could also abandon the concept of storage staking entirely. However, coming
+up with a scalable, sustainable solution that does not suffer from the same
+refund problems is hard to come up with.
+
+One proposed design is a combination of sharing code between contracts and zero
+balance accounts. Basically, if somehow the deployed code is stored in a way
+that does not require storage staking by the user itself, maybe the per-user
+state is small enough to fit in the 770 bytes limit of zero balance accounts.
+(Questionable for most non-trivial use cases.)
+
+This alternative is much harder to design and implement. The proposal that has
+gotten the furthest so far is [Ephemeral
+Storage](https://github.com/near/NEPs/pull/485), which is pretty complicated and
+does not have community consensus yet. But even on that, nobody is currently
+working on. While we could wait for that to eventually make progress, the
+community is held back in their innovation because of the refund problem.
+
+## Future possibilities
+
+- We might want to add the possibility to make non-refundable balance transfers
+  from within a smart contract. This would require changes to the WASM smart
+  contract to host interface. Since removing anything from there is virtually
+  impossible, we shouldn't be too eager in adding it there but if there is
+  demand for it, we certainly can do it without much trouble.
+- We could later add the possibility to refund the non-refundable tokens to the
+  account who sent the tokens initially.
+- We could allow sending non-refundable balance to existing accounts.
+- If (cheap) code sharing between contracts is implemented in the future, this
+  proposal will most likely work well in combination with that. Per-user data
+  will still need to be paid for by the user, which could be sponsored as
+  non-refundable balance without running into refund abuse.
+
+
+## Consequences
+
+
+### Positive
+
+- Businesses can sponsor new user accounts without the user being able to steal
+  the sponsored tokens.
+
+### Neutral
+
+- Non-refundable tokens are removed from the circulating supply, i.e. burnt.
+
+### Negative
+
+- Understanding a user's balance become even more complicated than it already
+  is. (Instead of only `amount` amd `locked`, there will be a third component.)
+- There is no incentive anymore to delete an account and its state when the
+  backing tokens are not refundable.
+
+
+### Backwards Compatibility
+
+We believe this can be implemented with full backwards compatibility.
+
+## Unresolved Issues (Optional)
+
+All of these issues already have a proposed solution above. But nevertheless,
+these points are likely to be challenged / discussed:
+
+- Should we allow adding non-refundable balance to existing accounts? (proposal:
+  no)
+- Should we allow adding more non-refundable balance after account creation?
+  (proposal: no)
+- Should this NEP include a host function to send non-refundable balance from
+  smart contracts? (proposal: no)
+- How should a wallet display non-refundable balances? (proposal: up to wallet
+  providers, probably a new separate field)
+
+## Changelog
+
+### 1.0.0 - Initial Version
+
+> Placeholder for the context about when and who approved this NEP version.
+
+#### Benefits
+
+> List of benefits filled by the Subject Matter Experts while reviewing this
+> version:
+
+- Benefit 1
+- Benefit 2
+
+#### Concerns
+
+> Template for Subject Matter Experts review for this version: Status: New |
+> Ongoing | Resolved
+
+|   # | Concern | Resolution | Status |
+| --: | :------ | :--------- | -----: |
+|   1 |         |            |        |
+|   2 |         |            |        |
+
+## Copyright
+
+Copyright and related rights waived via
+[CC0](https://creativecommons.org/publicdomain/zero/1.0/).


jakmeier:
I believe this is ready to start the NEP process.
@near/nep-moderators can you please take a look if it meets the standards to be moved to REVIEW?

cc @DavidM-D @akhi3030

frol:
Thank you @jakmeier for submitting this NEP.

<details>

Technical Summary guidelines:

```

### Recommendation

Add recommendation

### Benefits

* Benefit

* Benefit

### Concerns

| - | - | - | - |

```

</details>

jakmeier:
@near/wg-protocol reminder, can you please assign SMEs as frol asked 2 weeks ago? This effort is entirely blocked on your input here. Thanks!

neps/nep-0491.md
```diff
@@ -0,0 +1,359 @@
...
+these points are likely to be challenged / discussed:
+
+- Should we allow adding non-refundable balance to existing accounts? (proposal:
+  no)
+- Should we allow adding more non-refundable balance after account creation?
+  (proposal: no)
+- Should this NEP include a host function to send non-refundable balance from
+  smart contracts? (proposal: no)
+- How should a wallet display non-refundable balances? (proposal: up to wallet
+  providers, probably a new separate field)
```
norwnd, line 330: Do wallets even need to display non-refundable balance to users (what for) ? My impression was "non-refundable balance" is working specifically towards hiding the complexity of storage-staking from users, is this even an achievable goal ? Blockchain explorers would probably need to display "non-refundable balance" though at least for info/debugging purposes.

On a similar topic, the name "non-refundable balance" might be unnecessarily vague given the intent of this proposal, wouldn't something like "dedicated storage balance" or "storage-tied balance" better describe the concept ? Since it's gonna show up in blockchain explorers this makes it a bit more self-explanatory, I think. Unless it does become a more generic concept eventually (in which case more generic name is warranted ofc).

norwnd (commented):
Hi, just some newbie comments here, feel free to ignore if I don't make much sense.

birchmd:
@bowenwang1996 I'm not sure who to nominate here as a technical reviewer on behalf of the protocol working group. Do you have someone in mind?

neps/nep-0491.md
jakmeier: Hi @norwnd , thanks for your input, it's very much welcome! :)

> Do wallets even need to display non-refundable balance to users (what for) ?

Not displaying it is a valid choice for a wallet IMO. But right now, I assume most wallets are doing some kind of computation to show how much of the balance is currently open for withdrawal, vs the amount that's reserved to cover storage staking. If they have this computation, they need to update how it works once this proposal gets through.

> My impression was "non-refundable balance" is working specifically towards hiding the complexity of storage-staking from users, is this even an achievable goal ?

Interesting question. The original motivation here isn't so much to simplify things. It's more about enabling a certain use case. But I can see how, if presented correctly, we can maybe make it simpler from a user point of view, as they don't even have to know there is storage staking for their account. I like that!

> wouldn't something like "dedicated storage balance" or "storage-tied balance"/"storage-restricted balance" better describe the concept ?

I'm open to name changes. I very much dislike "non-refundable balance". But I would like to have a clear distinction between "refundable balance that's currently locked due to storage usage" (which we already have today) and the new concept of "balance that can only ever be used for storage in this account and will be burnt when the account is deleted". In your suggestions, it seems like the line between the two is blurry.

I've considered "storage credits", but that sounds like you can trade them between accounts.

Another idea: Perhaps we should just "purchase" an amount of bytes, and add a field like `free_bytes`, as in, no storage staking required for these bytes. Sounds a bit crazy to me but maybe that would make the concept easier to grasp. WDYT?

akhi3030:
> @bowenwang1996 I'm not sure who to nominate here as a technical reviewer on behalf of the protocol working group. Do you have someone in mind?

@birchmd, @bowenwang1996: I am happy to volunteer.

neps/nep-0491.md
norwnd: > In your suggestions, it seems like the line between the two is blurry.

Yep, it was just to illustrate that point, your `free_bytes` concept feels much better (alternative names to consider: `purchased/prepayed/bought`+`_`+`bytes/storage` perhaps).

norwnd: > The original motivation here isn't so much to simplify things. It's more about enabling a certain use case. But I can see how, if presented correctly, we can maybe make it simpler from a user point of view, as they don't even have to know there is storage staking for their account.

Long term though, UX-wise both for **users** and **devs** the _"Alternative: Move away from storage staking"_ is still probably desirable, if achievable in sustainable manner.

neps/nep-0491.md
```diff
@@ -0,0 +1,359 @@
...
+to repeat this action for as long as the business has funds left in the faucet.
+
+The protocol should allow to create accounts in a way that is not susceptible to
+such refund abuse. This would at least change the incentives such that creating
+fake users is no longer profitable.
+
+Non-refundable storage staking is a further improvement over
+[NEP-448](https://github.com/near/NEPs/pull/448) (Zero Balance Accounts) which
+addressed the same issue but is limited to 770 bytes per account. By lifting the
+limit, sponsored accounts can be used in combination with smart contracts.
```
norwnd, line 43: Perhaps the following part could use a bit more details (expanded on, in some section below):

> By lifting the limit

with respect to how it interacts with "free 770 bytes per account" concept. Because strictly speaking, the way I understand this prop, it's not exactly a generalization of https://github.com/near/NEPs/pull/448 but rather a standalone feature that extends it conceptually.

It seems both features interplay nicely together, just want to make sure it doesn't create unnecessary burden for developers when they decide which one to go with.

bowenwang1996:
As a protocol working group member, I nominate @akhi3030 and @birchmd as SME reviewers for this NEP.

birchmd (commented):
Thanks for the proposal @jakmeier ! I am providing my SME review as requested by @bowenwang1996 

I appreciate your thoroughness in describing alternatives to this proposal. You anticipated many of the questions I would have asked!

My remaining concern is if the additional protocol complexity added by this change is justified by the claimed benefit of removing the incentive to spam and delete free accounts. It is true that it removes the "first-order" incentive for this behaviour, but there are "second-order" incentives a bad actor could have. For example, someone who holds many NEAR tokens may be incentivized to cause a large number of NEAR tokens to be taken out of circulation as a means of inflating the NEAR price. Or perhaps a competitor to the business offering the free accounts would be happy to drain the funds of their rival to gain market advantage.

My main point is that business that are offering pre-funded accounts to users will probably still need to implement some kind of spam protection themselves because there could still be bad actors that benefit from spamming. And if it is true that all businesses are implementing this spam protection themselves then do we also need this non-refundable protection in the protocol?

I understand this is a somewhat subjective question. If the community is asking for this feature then it probably provides value to them. But I just want to make sure we are recognizing that this is not a silver bullet to solve free account spam issues.

jakmeier:
@birchmd thanks for the review! You raise a very good question! I completely agree that we should carefully consider how important the feature is and weigh it against added complexity in the protocol.

@DavidM-D can you maybe make a stronger case for why we need a solution to the storage-refund problem? Specifically answering the comment by @birchmd .

wacban:
> someone who holds many NEAR tokens may be incentivized to cause a large number of NEAR tokens to be taken out of circulation as a means of inflating the NEAR price 

This is always going to be possible regardless of this proposal. Anyone holding near can just "not use it" or even go as far as "burn" it e.g. by sending it to an unusable account like implicit account with address 00000.... .

What I would worry about is the malicious case where an attacker drains the funds allocated for free accounts by spamming. For that reason I agree that we need spam protection. I don't think it's feasible as part of protocol as we cannot really tell apart spamming from increased traffic. I would rather see the business implement its own, app level, solution to this problem.

MaximusHaximus:
> My main point is that business that are offering pre-funded accounts to users will probably still need to implement some kind of spam protection themselves because there could still be bad actors that benefit from spamming. And if it is true that all businesses are implementing this spam protection themselves then do we also need this non-refundable protection in the protocol?
> 
> I understand this is a somewhat subjective question. If the community is asking for this feature then it probably provides value to them. But I just want to make sure we are recognizing that this is not a silver bullet to solve free account spam issues.


> @DavidM-D can you maybe make a stronger case for why we need a solution to the storage-refund problem? Specifically answering the comment by @birchmd .

Having been responsible for the services that provided a faucet based solution for sponsoring new account creation for quite some time in the NEAR wallet on wallet.near.org which was attacked by many bad actors who took advantage of this behaviour for their gain, I can add some color here.

While it's true that this change by itself won't resolve spamming sponsored new account creation as an attack vector entirely, it does eliminate the most _common_ incentive for exploiting sponsored account creation in the past, and it will change the _economy_ of exploiting the service, in that the attacker must spend their resources (time and money) to attack the service, but doing so will not provide them with NEAR as a direct reward for a successful exploit. This change will eliminate a huge cohort of possible bad actors, and even in the cases mentioned (e.g. competitive sabotage or purely malicious attackers), it will have a limiting effect on the scale of attacks.

Almost any form of anti-spam measures can be circumvented given enough effort by a determined attacker, _especially_ if you want people to be able to use the service without extremely invasive measures (e.g. passport / actual state issued ID / KYC uploads to ensure uniqueness of users).  The most important thing that dissuades mass exploitation is that it must cost attackers more (time or money) to circumvent the anti-spam protections that you put in place than makes it worthwhile for attackers to circumvent those protections. 

As an example -- in NEAR wallet, we implemented multiple anti-spam strategies including e-mail registration and multiple forms of reCaptcha (active and passive) -- all of those were eventually circumvented by attackers.  We even forced users to use a specific email provider that had their own strong anti-spam functionality so that emails could not be trivially created via automation.  I discovered that even the very small amount (**0.005N!**) of NEAR turned out to be valuable enough at the time that it was a very strong incentive in some countries.  I saw usage patterns that indicated that at least one person was _manually_ going through the steps of creating new email accounts and bypassing reCaptcha checks on a regular work schedule like an actual 9am-5pm job work day in their country.  Even though it took them several **minutes** to create each account, _and_ they were forced to spend money in order to distribute their requests across many IP addresses, and used their own time to solve reCaptcha prompts (or paid money to farm it out to a mechanical turk service) , it was still cost-effective for them to exploit the service.  I discovered that the value of the amount of NEAR that we needed to provide to facilitate a useful account was high enough that farming new accounts for that NEAR could make exploiters more money than a legitimate full-time job in some countries(!), even after their expenses were covered.  We made it as hard as we could reasonably do without extremely intrusive anti-spam measures, forcing them to do everything _manually_, with no real automation/scripting to their attack, and it was _still_ cost-effective for them to spend all day long bypassing our anti-spam mechanisms to create accounts.  Also, I learned that there are entire online communities of people who share exploitable systems they discover amongst each-other, so if you have a system that is cost-effective to exploit, you will end up with _many_  attackers trying exploit your service who are actually completely un-interested in your service or in harming you, but are there purely because they are interested in their own gains! 

Consider the _scale_ of the incentive that the current implementation can provide. Imagine for a moment that a developer wants to sponsor creation of new accounts and they want to deploy a smart contract to _each_ account when it is created, with some bespoke functionality that is a critical part of their app.  We could take the multi-sig contract as an example -- you need **3.5 NEAR** of balance to deploy that contract.  Even in today's market, that much NEAR is a non-trivial amount **each** time an attacker successfully exploits the service.  This gives attackers a large 'budget' to use to bypass any anti-spam measures put in place by the developer. Just imagine what kind of incentive it would give attackers in a _strong_ market if they could collect 3.5N every time they exploited a sponsored account service.  If you do the math with even _half_ of the value of NEAR's all-time-high you can see this is a really serious issue. If we eliminate the direct token incentive for exploiting, bad actors will be forced to invest their _own_ time and other resources into attacking the service; they won't be directly rewarded each time they do so.

Just to note that I too would prefer the "Allow refunds to original sponsor" variation, but any solution that eliminates the current incentive would be a huge improvement.

birchmd:
Thanks @MaximusHaximus for the detailed comment! Your first-hand experience with this is quite valuable in weighing the trade-offs of this proposal's utility vs protocol complexity.

@jakmeier how much additional complexity is there in the "refunds to original sponsor" variant? Maybe it would be worthwhile since it has been mentioned by both @wacban and @MaximusHaximus

jakmeier:
> @jakmeier how much additional complexity is there in the "refunds to original sponsor" variant? Maybe it would be worthwhile since it has been mentioned by both @wacban and @MaximusHaximus

It's not trivial but also not too much to ask for. I'm happy to include it if that's the decision.

In more details, we have to:
- Upon deletion, create two storage refund receipts instead of one.
  - I see no issue here.
- Store the sponsor id to the account meta data. Which sounds simple but has a few implications.
  - It will increase the account meta data by 69 bytes. Hence we should increase `storage_num_bytes_account` from 100 to 169 or something. But if we increase it, then it changes how many keys can be added to a zero balance account, unless we also increase the ZBA limit.
  - Old accounts might not have enough balance to cover the additional storage. We can probably get away with using the old limit for old accounts but it's a bit hacky. Or we could enforce the new limit on all accounts and let affected users figure out that they  have to add more balance before they can execute receipts again...


Finally, "Non-Refundable Storage" needs renaming even more if we implement it this way. :)

birchmd:
> It will increase the account meta data by 69 bytes. 

Could the field be optional so that old accounts need to only increase by 1 byte (a flag indicating the absence of a sponsor)?

Having a minimal impact on non-sponsored accounts is important in my opinion.

jakmeier:
> > It will increase the account meta data by 69 bytes.
> 
> Could the field be optional so that old accounts need to only increase by 1 byte (a flag indicating the absence of a sponsor)?
> 
> Having a minimal impact on non-sponsored accounts is important in my opinion.

The problem is that `storage_num_bytes_account` is a constant for all accounts.
But yeah, I guess we can change it and make the per-account meta data storage requirement a dynamically computed value. That might just be the easiest solution.

akhi3030:
I have some reservations with this proposal.  I have highlighted them in [here](https://github.com/near/NEPs/issues/505).  However, I don't think my reservations are crucial enough that they should prevent this work from moving forward.  So I am happy to support this work.  I will write a SME summary shortly.

DavidM-D:
Since there doesn't seem to be any substantial objection to this proposal can we start the process of merging this in?

akhi3030:
This is actually blocked on SME reviews and I am one of them.  Sorry for not getting it done till now.  I promise to get it done today.

akhi3030:
# Technical summary

## Background
In the NEAR protocol, smart contracts "pay" for the state that they are using by locking up tokens.  The tokens that are locked cannot be staked so the fee is the opportunity cost of being able to stake the tokens instead and earn some interest.  In order to create a smart contract with some state, someone has to deposit some tokens on that account.  

## Use case
There exist some business use cases where a third party is willing to fund the smart contract creation for many users.  Generalised a bit, party A is willing to create the smart contract for party B and also deposit the tokens needed to maintain the state in the smart contract.  This use case is currently not possible because once party A gives ownership of the smart contract to party B, B can delete the contract and get the tokens recovered back to them.  Party B can repeat this trick several times to drain a lot of funds from party A.

This proposal suggests a change to the protocol to enable the above use case.

## Actual proposal
The proposal allows the account creator to specify how much storage they are willing to pay for upfront and for which if the account is deleted, they do not want a refund back.  This proposal will eliminate the personal incentive that party B has (described above) to ask party A to create accounts for them and then to delete them.  Party B no longer stands to personally benefit from this.

## Complications
### Other malicious uses
Other forms of attacks remain such as party B can still go through the exercise of having party A create and then delete the accounts simply to hurt party A.  Addressing this problem is outside the scope of this NEP and if required can be built on top of this NEP.

### To refund or not to refund
It has been proposed that if and when party B does delete the account, instead of not producing any refund, the refund can be produced and sent back to A.  I personally do not see much value in adding this complication to this NEP at this moment in time because I do not see any clear incentives for party B to want to delete the account if they do not stand to gain anything from it.  And also if ever a usecase arises to support this, an additional NEP can be proposed.

# Recommendation

I do have some [reservations](https://github.com/near/NEPs/issues/505) with the overall approach we have for paying for storage in the NEAR protocol.  These concerns do not get in the way of approving this NEP.  For the current way of paying for storage, this NEP makes sense to me and I see this as a positive enhancement.

## Benefits

The main benefit of this NEP is to enable an important usecase that has the potential to onboard many users to NEAR and to support new usecases of NEAR (third party account creation) that were not possible earlier.

## Concerns

| - | - | - | - |   
| 1 | This makes account creation a bit more complicated.  Users need to understand details when creating accounts | The defaults remain the same, this is something that only a small set of power users need to use and be aware of. | Not an issue |
| 2 | As outlined in [this issue](https://github.com/near/NEPs/issues/505), chances of garbage state on the chain | This NEP enables an important use case.  If we do not add support for this use case, then people might try to create third party accounts and roundabout ways and still there will be garbage state on the network | This is already a problem and is not necessarily made worse by this proposal |

birchmd:
My initial SME review prompted some discussion and now I think it makes sense for me to post a final review.

I agree with the review by @akhi3030 where I think we should approve the proposal as-is. Perhaps in the future we can iterate on this to allow refunding to the sponsor, but for now we can leave it out because this feature adds significant additional complexity to the proposal and the current proposal already addresses concerns from the community (as seen in the comment from @MaximusHaximus ).

akhi3030:
@near/nep-moderators, @near/working-groups, @DavidM-D: the original author for this NEP (Jakob) is no longer employed at Pagoda and potentially will not be working on it actively anymore.  We may need someone to take ownership of this.  CC: @walnut-the-cat.

jakmeier:
As the original author, I'm happy to help if anything is unclear and needs my input. But yeah, I wrote this proposal for Pagoda's requirements and hence won't be driving this further at this point.

neps/nep-0491.md
```diff
@@ -0,0 +1,359 @@
...
+
+Some business models are okay with paying that fee for users upfront, just to
+get them onboarded. However, if a business does that today, their users can
+delete their new accounts and spend the tokens intended for storage staking in
+other ways. Since this is free for the user, they are financially incentivized
+to repeat this action for as long as the business has funds left in the faucet.
+
+The protocol should allow to create accounts in a way that is not susceptible to
+such refund abuse. This would at least change the incentives such that creating
+fake users is no longer profitable.
```
mfornet, line 38: An alternative solution to this problem that doesn't involve a change to the protocol.

The business creates an account with a deployed contract that will work in two modes. 
- Restricted mode: The user doesn't have a FullAccessKey, and can't delete the account
- Regular mode: The user has a FullAccessKey

On restricted mode, the user has a FunctionCallKey that allows calling into some methods of the contract that grants similar permissions to a FullAccessKey, but not all: [See all permissions here](https://docs.near.org/concepts/basics/accounts/access-keys#full-access-keys). In particular, it must not have permission to delete the account and get a refund.

To upgrade from restricted mode to regular mode, there is a method `upgrade` that will send the initial amount of NEAR deposited by the creator of the account (the business) and will attach a new FullAccessKey to the contract (the same as the initial FunctionCallKey). Anyone can call this method.

With this proposal, there will be no incentives to steal money from the business (by creating and deleting accounts), since the user can only delete the account in normal mode, after depositing the initial amount.

The business can upgrade accounts for those users that they already trust to not be acting maliciously. (i.e their goal is not to recover the money, but to avoid incentives for malicious users to deplete their funds).

## Drawback

The accounts in restricted mode don't behave in the same way as owning a regular account, since all actions need to be done through a special method (i.e. to do any Transfer or FunctionCall to another contract). 

This might be acceptable if the contract is a multi-sig (or similar interface) where the user is not expected to do Transfers or FunctionCalls directly anyway; or if the account is expected to be used on a specific application only, that is aware of this mode. 

To support using restricted accounts for arbitrary (allowed) use cases, wallets should be aware of it, which requires creating a standard about it. An alternative (that requires a protocol update), is creating a new type of key, in addition to FullAccessKey and FunctionCallKey, we can have GranularAccessKey, which works similarly to a FullAccessKey but doesn't have permission to do some actions (in this case it won't have permission to Delete the contract).

walnut-the-cat: cc. @jbajic

DavidM-D:
@firatNEAR Just a heads up that this is a blocker for multichain relayers, which we're hoping to get out by EoY

neps/nep-0491.md
jakmeier: > How does a user interact with a fungible token contract without a Full Access Key?

@DavidM-D  As I understand the idea, it goes through a contract deployed on the account that forwards arbitrary calls.

> On restricted mode, the user has a FunctionCallKey that allows calling into some methods of the contract that grants similar permissions to a FullAccessKey, but not all

So for example, you call `user.near::proxy([{action: "Function Call", "receiver": "ft.near", "method": "ft_transfer", "arg": "100", "balance": "1"}])` which will be unpacked by the contract and create the actual request as `ft.near::ft_transfer("100")` to the FT contract as a cross-contract call.

I think we discussed this as the "proxy account model" before, in the variation where it's deployed on every account. The increased storage and gas costs were one of the problems we saw in this idea. Also the latency is increased by at least one block time.

But more importantly, I think we assumed we want to give full access to users, which leads again to the storage refund problem. Now, the proxy contract gets around this problem by stripping the rights to delete the account or redeploy the contract code from the user. That's fine and I think we considered this already but assumed for the upgrade path we need to hold a full access key on the sponsor, which we didn't like for potential abuse issues. However, this seems to solve it:

> To upgrade from restricted mode to regular mode, there is a method upgrade that will send the initial amount of NEAR deposited by the creator of the account (the business) and will attach a new FullAccessKey to the contract (the same as the initial FunctionCallKey). Anyone can call this method.

So the upgrade path is also part of the contract. Nobody has full access for the account, except the contract code. I really like this idea by @mfornet 

Of course it's possible I missed something. WDYT think @DavidM-D, could it work for your use case?

jbajic: So to support the alternative solution, we would need to add another type of key (beforementioned **GranularAccessKey**) for this specific use case in which we don't want users to delete their account. I am a bit confused with this delegation of actions through FunctionCall, why would the restricted account have to use FunctionCall?

I am also fine with the existing solution, I feel like the alternative is slightly better but I may lack understanding to see if it is much better to pursue it instead of what we already have.

walnut-the-cat: @mfornet , @DavidM-D , @firatNEAR , shall we set up a meeting sometime early next week to sort it out so we can get to the alignment asap? EoY is right around the corner and we need to move fast

DavidM-D: I'm keen for that.

One question I'd ask is, ignoring the contract cost, could we have implemented Zero Balance Accounts entirely on the application level using a KeyPomesque solution?

mfornet: @walnut-the-cat I can join a call next week as well. Use the email: mfornet94 at gmail

> One question I'd ask is, ignoring the contract cost, could we have implemented Zero Balance Accounts entirely on the application level using a KeyPomesque solution?

@DavidM-D Zero Balance Accounts no, but maybe you could have achieved your goal anyway (which is not to avoid paying but avoid getting scammed by the users as I understand). 

How are the users expected to use those accounts, or interact with the contract you are deploying? (i.e what is the goal of the contract). 

If the contract is somehow "similar" in spirit to a multi-sig, where the user needs to go through a contract method to perform any action, then KeyPom contract will fit you. Otherwise **drawbacks** from the original comment in this thread applies.

walnut-the-cat: @mfornet , if you can share some of your availability (e.g. timezone), I will send out an invitation! I can check schedule for @DavidM-D and @firatNEAR from my end

mfornet: @walnut-the-cat I can do any time between (6:00 - 14:00) UTC, I could do later, but not every day. I recommend you create an event, and we follow up in email.

neps/nep-0491.md
```diff
@@ -0,0 +1,359 @@
...
+  "AddKey": { "public_key": "...", "access_key": "..." }
+}
+```
+
+To create implicit accounts, the current protocol requires a single `Transfer`
+action without further actions in the same transaction. With this proposal,
+setting it to non-refundable will make the full balance non-refundable. If users
+want to add refundable balance on top, it requires a second transaction.
+
+If a non-refundable transfer arrives at an account that already exists, it will
```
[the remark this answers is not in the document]
jbajic: Because the non-refundable balance exists only for storage staking and it would be confusing if somebody would top up non-refundable balance and then it all gets burned when accounts gets deleted.

neps/nep-0491.md
```diff
@@ -0,0 +1,359 @@
...
+
+Therefore, as the last change necessary for this NEP, we also introduce a new
+serialization format for new accounts.
+
+```rust
+// new serialization format for `struct Account`
+
+// new: prefix with a sentinel value to detect V1 accounts, they will have
+//      a real balance here which is smaller than u128::MAX
+writer.serialize(u128::MAX)?;
```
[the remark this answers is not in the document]
jbajic: Markdown does not support links to paragraphs from code sections. I think the explanation is not that far away from this part, I could add it as a comment in code if you think it is valuable?

neps/nep-0491.md
```diff
@@ -0,0 +1,359 @@
...
+### Neutral
+
+- Non-refundable tokens are removed from the circulating supply, i.e. burnt.
+
+### Negative
+
+- Understanding a user's balance become even more complicated than it already
+  is. (Instead of only `amount` and `locked`, there will be a third component.)
+- There is no incentive anymore to delete an account and its state when the
+  backing tokens are not refundable.
```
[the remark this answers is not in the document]
jbajic: I think the negative aspects of:

> - Understanding a user's balance become even more complicated than it already
>   is. (Instead of only `amount` and `locked`, there will be a third component.)

is the users experience, since the user must now know the nature of another balance in their account and have and understanding of what is possible with it and what is not.
and for the second statement:

> - There is no incentive anymore to delete an account and its state when the
>   backing tokens are not refundable.
the negative side is that we can have a lot of accounts with their state and no reason for their owners to delete them, since they get nothing out of it, which leaves "dead" accounts in circulation.

neps/nep-0491.md
```diff
@@ -0,0 +1,359 @@
...
+to iterate over all existing accounts and re-merklize. That's infeasible within
+a single block time and stopping the chain would be disruptive.
+
+We could maybe migrate lazily, i.e. read account version 1 and automatically
+convert it to version 2. However, that would break the assumption that every
+logical value in the merkle trie has a unique borsh representation, as there
+would be a account version 1 and a version 2 borsh serialization that both map
+to the same logical version 2 value. This could lead to different
+representations of the same chunk in memory, which might be used in attacks to
+force a double-sign by innocent validators.
```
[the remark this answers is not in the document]
jbajic: The issues that arise with converting accounts from v1 to v2 are due to the Borsch serialization, currently we have a guarantee that there is no two binary representations that deserialize into the same object of same type, but with different types we can have two same binary representation (or as you refer to it as physical representation)

neps/nep-0491.md
```diff
@@ -0,0 +1,359 @@
...
+
+- Non-refundable tokens are removed from the circulating supply, i.e. burnt.
+
+### Negative
+
+- Understanding a user's balance become even more complicated than it already
+  is. (Instead of only `amount` and `locked`, there will be a third component.)
+- There is no incentive anymore to delete an account and its state when the
+  backing tokens are not refundable.
+
```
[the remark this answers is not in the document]
jbajic: Isn't that the same as the last one?

walnut-the-cat:
@frol and @ori-near , The NEP is ready for another SME review and potentially Working group meeting for final approval. I believe @mfornet has been involved as a SME and is onboard with the current proposal mentioned in the NEP. Could you guide how we can move forward from here?

frol:
> While the new proposal by @mfornet is valid and can solve the target issues, as development for zero balance account has been ongoing for multiple quarters and launch deadline is around the corner, cost to make it happen properly may be too high at this moment.

@jakmeier, @DavidM-D, @firatNEAR, @akhi3030, Were the costs of adoption and fixing it in the future considered?

Help me to understand the motivation for this NEP.

> Non-refundable storage staking is a further improvement over https://github.com/near/NEPs/pull/448 (Zero Balance Accounts) which addressed the same issue but is limited to 770 bytes per account. By lifting the limit, sponsored accounts can be used in combination with smart contracts.

1. If business sponsoring a new account does not need to deploy a smart contract they should be fine with Meta-transactions to further compensate the costs, and users cannot spend balance as it is zero
2. If business sponsoring a new account needs to deploy a smart contract on the sponsored account (what is the use-case, by the way?), let it operate like a proxy as @mfornet suggested and have a central contract doing whatever OR creating one zero-balance account (e.g. `frol.near`) and a subaccount with the smart contract deployed to it (`contract.frol.near`)

This NEP does not talk about transferability of the non-refundable tokens; I assume, non-refundable tokens cannot be transferred, but there is nothing in this NEP. It will be quite confusing for users to see 5 NEAR balance (500kb contracts are common to see) which they cannot use for anything but storing data.

walnut-the-cat:
> Were the costs of adoption and fixing it in the future considered

I don't think 'fixing' is the right term here as we are not building anything that will break. Having said that, we discussed future impact of each option and the conclusion was that the difference is not that much. If we want to change things based on all the hindsight we have learned, regardless of the solution we take here, such change will be significant enough to make difference in two solutions trivial.

frol:
If you will decide to still pursuit this path, I would like to at least ask you to consider a better API:

Instead of having `Transfer` and `TransferV2`, where the first one is soft-deprecated, and the second one gets a pretty ugly `nonrefundable` boolean flag, I would suggest to leave `Transfer` as is, and don't deprecate it, and create a new action (naming is hard: `ReserveStorage` / `DepositStorage` / `CoverStorageCost` / `BurnForStorage` / `TransferForStorage` / `StorageTransfer` / `TransferAndLockForStorage` - the goal is to communicate that the tokens can be used only for storage and cannot be transferred).

wacban:
Either way TransferV2 or the new name you may also consider adding versioning to it, in order to make it future proof.

DavidM-D:
@frol 
> If business sponsoring a new account does not need to deploy a smart contract they should be fine with Meta-transactions to further compensate the costs, and users cannot spend balance as it is zero

I'm not sure I fully understand this point. Meta transactions allow only for gas payments on behalf of the user, not NEAR, FT or NFT transactions. This also prevents sponsoring of storage deposit.

Even without deployed contracts, this proposal is useful. Currently ZBA can only have a fixed number of keys (6). Since we need 1 key for recovery, 1 for each device and 1 for each LAK on an application we can very quickly run through that.

> If business sponsoring a new account needs to deploy a smart contract on the sponsored account (what is the use-case, by the way?), let it operate like a proxy as @mfornet suggested and have a central contract doing whatever OR creating one zero-balance account (e.g. frol.near) and a subaccount with the smart contract deployed to it (contract.frol.near)

In terms of use cases, they're identical to the current use cases for smart contracts on accounts except we don't require users to purchase NEAR to use them. Some examples are multisig, NETH and account recovery using an Oauth token (FastAuth).

Proxy contracts are the approach we're taking, but as I understand it there still needs to be a contract deployed on the users account.

Could you explain the benefit of deploying a contract to a subaccount? Does it prevent the user from getting a refund on the storage somehow?

jakmeier:
@frol I'm a bit out of the loop and can't say with confidence why we still need this proposal. But my understanding is that all use cases, including contract-deployed-on-each-user-account, can already be implemented without protocol changes. It seems to be more an issue of facilitating  project timelines based on [this comment](https://github.com/near/NEPs/pull/491#issuecomment-1790846143):

> While the new proposal by @mfornet is valid and can solve the target issues, as development for zero balance account has been ongoing for multiple quarters and launch deadline is around the corner, cost to make it happen properly may be too high at this moment.

I have to admit, it's not obvious to me why implementing NEP-491 is expected to be quicker than an implementation that doesn't change the protocol but I trust the group of people who discussed it and reached this conclusion according to the comment linked above.

Regarding API:

> If you will decide to still pursuit this path, I would like to at least ask you to consider a better API:
>
> Instead of having Transfer and TransferV2, where the first one is soft-deprecated, and the second one gets a pretty ugly nonrefundable boolean flag, I would suggest to leave Transfer as is, and don't deprecate it, and create a new action (naming is hard: ReserveStorage / DepositStorage / CoverStorageCost / BurnForStorage / TransferForStorage / StorageTransfer / TransferAndLockForStorage - the goal is to communicate that the tokens can be used only for storage and cannot be transferred).

Yeah absolutely, this seems like a better API naming. The original naming API comes from me and it is bad. At a first glance, I really like `ReserveStorage` or `BurnForStorage`. Maybe the field `nonrefundable_balance` could become `reserved_storage` and that would also make it clearer how to show it in a wallet to the user. But I think @jbajic you should probably be the engineer to work out a better API in details to improve on my initial take.

jakmeier:
@wacban 
> Either way TransferV2 or the new name you may also consider adding versioning to it, in order to make it future proof.

Yes! Currently action receipts are independent of a protocol version and hence we cannot make breaking changes to action serialization formats. Versioning could solve that.

But I think perhaps versioning should be on a per-receipt level rather than on a per-action level. Why? (1) because of the overhead of a version field per actions and (2) because I don't want to deal with e.g. mixed Transfer V1 and Transfer V2 in the same receipt.

This versioning per-receipt could be as simple as including the intended protocol version in the action receipt. (Instead of introducing a new action-version counter that runs parallel to protocol versions.)

But because I think this would be the better way to implement versioning, I also think it is off-topic here.

walnut-the-cat:
Hello- NEP moderator here.

@akhi3030 and @birchmd , could you double confirm whether you are approving this NEP as SMEs?

mfornet:
> I have to admit, it's not obvious to me why implementing NEP-491 is expected to be quicker than an implementation that doesn't change the protocol but I trust the group of people who discussed it and reached this conclusion according to the comment linked above.

During the discussion, I understood the implementation of this NEP is almost ready to land, and will provide a simpler solution to their use case.

> While the original proposal by NEP-491 adds more complexity to Protocol, which we all want to avoid if possible, the expected outcome of NEP-491 is essentially subset of what already exists and does not make the situation worse.

To elaborate on this. This NEP is promoting the usage of non-refundable storage, (i.e, storage that will be in the state forever). This is however not a strong argument against the NEP, since in practice almost no one is cleaning unused storage. The state growth needs to be solved apart.

jakmeier:
> During the discussion, I understood the implementation of this NEP is almost ready to land, and will provide a simpler solution to their use case.

Yes, when I handed it over, the work on the protocol team's side was mostly ready, with the main leftover work being to agree on a better naming, finishing open discussions, and getting final approval.
But the introduction of a new action type has consequences for tooling across the stack outside the protocol team's realm. This, plus the natural cycle through testnet is bound to delay the feature.

As I see it, the counter proposal seems to be more engineering work but has no external dependencies and could thus be implemented on a tighter timeline if prioritized.

birchmd:
@walnut-the-cat with @frol 's good suggestion to make non-refundable storage transfer a separate action as opposed to a modification of the existing transfer action I think the NEP needs updating to specify this in detail.

In general I am still supportive of this NEP if the community thinks it is valuable (which it seems they do given the amount of activity on this NEP). @mfornet 's contract-level solution also solves the problems at hand, but at the cost of more complexity in tooling (packaging actions as function calls to a controller/proxy contract instead of sending them directly as transactions). This tooling complexity seems greater than the protocol-level complexity of allowing non-refundable transfers, which I think is why people like @DavidM-D continue to push for this NEP.

akhi3030:
As an SME, my opinions are aligned with birchmd's [message](https://github.com/near/NEPs/pull/491#issuecomment-1821603288) above.  If we are introducing a separate action, it will be good to see the NEP updated to specify that.

jbajic:
> As an SME, my opinions are aligned with birchmd's [message](https://github.com/near/NEPs/pull/491#issuecomment-1821603288) above. If we are introducing a separate action, it will be good to see the NEP updated to specify that.

Thanks for the comment @frol @birchmd @akhi3030 , I will be working on updating the NEP then, and let you know when it is ready for another review

walnut-the-cat:
@akhi3030 and @birchmd , could you have another look at the updated NEP and confirm your approval as SMEs? Thank you!

neps/nep-0491.md
```diff
@@ -0,0 +1,386 @@
...
+To create implicit accounts, the current protocol requires a single `Transfer`
+action without further actions in the same transaction and this has not changed
+with this proposal:
+
+```json
+"Actions": {
+  "CreateAccount": {},
+  "ReserveStorage": { "deposit": "1000000000000000000000000" },
+  "Transfer": { },
```
birchmd, line 91:
```suggestion
  "Transfer": { "deposit": "0" },
```

neps/nep-0491.md
```diff
@@ -0,0 +1,386 @@
...
+## Reference Implementation
+
+On the protocol side, we need to add new action:
+
+```rust
+enum Action {
+  CreateAccount(CreateAccountAction), 
+  DeployContract(DeployContractAction),
+  FunctionCall(FunctionCallAction),
+  // kept for backwards compatibility but new clients should only use `TransferV2`
```
birchmd, line 129: Outdated comment

neps/nep-0491.md
```diff
@@ -0,0 +1,386 @@
...
+
+```rust
+amount + locked + nonrefundable >= storage_usage * storage_amount_per_byte
+```
+
+For old accounts that don't have the new field, the non-refundable balance is
+always zero. Adding non-refundable balance later is not allowed. If a transfer
+is made to an account that already existed before the receipt's actions are
+applied, execution must fail with
+`ActionErrorKind::OnlyImplicitAccountCreationAllowed{ account_id: AccountId }`.
```
birchmd, line 171: Should this be a new error kind instead? Something like `ActionErrorKind::OnlyReserveStorageOnAccountCreation`

neps/nep-0491.md
```diff
@@ -0,0 +1,386 @@
...
+A proof of concept implementation for nearcore is available in this PR: 
+https://github.com/near/nearcore/pull/9346
+
+
+## Security Implications
+
+There is a risk for users to accidentally create accounts with nonrefundable
+storage and lose the balance as a consequence. This is mitigated because the
+default behavior with `nonrefundable = false` will behave exactly like a normal
+transfer and not have this problem.
```
birchmd, line 208: Needs updating to say something like "This is mitigated because non-refundable storage is a new action type which will not be the default for tooling used in creating new accounts (e.g. near-cli, Near Wallet)"

neps/nep-0491.md
```diff
@@ -0,0 +1,386 @@
...
+user to call methods of the contract that controls the `FullAccessKey` and 
+allows user some functionality but not all, e.g. not allowing account deletion.
+The user in restricted mode could also upgrade an account by sending the 
+initial amount of NEAR deposited by the account creator and will attach a new
+`FullAccessKey`.
+
+The downside is that accounts in restricted mode do not behave as normal 
+accounts and need to do all actions through `FunctionCall`. To support
+the restricted accounts protocol upgrade is needed, but that can be mitigated by
+adding an alternative key `GranularAccessKey` similar to `FullAccessKey` but
```
birchmd, line 302: This idea of a `GranularAccessKey` sounds like a separate alternative from the idea of a proxy account. The proxy account idea does not require a protocol change (the user is only given a FunctionCall access key, which already exists). The complexity with the proxy contract idea is on the tooling side because actions like adding access keys to the account need to be converted to function calls instead of being direct actions.

Another complexity is in developing the contract a business would use. Since there is no protocol-level way to compose functionality from multiple Wasm modules (Account Extensions is on hold), the business would need to include the proxy logic with their business logic in the same contract. Including both functionalities together increases the development complexity of such a contract.

neps/nep-0491.md
jbajic: Ah you are right, it should be a new error since it can be triggered independently of `ActionErrorKind::OnlyImplicitAccountCreationAllowed`.

neps/nep-0491.md
jbajic: I would just remove the issue because it cannot happen accidentally since it requires an explicit new action

neps/nep-0491.md
bowenwang1996: What is the conclusion of this discussion? @mfornet @DavidM-D @jakmeier

neps/nep-0491.md
```diff
@@ -0,0 +1,381 @@
...
+  "CreateAccount": {},
+  "ReserveStorage": { "deposit": "1000000000000000000000000" },
+  "Transfer": { "deposit": "100" },
+  "AddKey": { "public_key": "...", "access_key": "..." }
+}
+```
+
+To create implicit accounts, the current protocol requires a single `Transfer`
+action without further actions in the same transaction and this has not changed
+with this proposal:
```
bowenwang1996, line 85: I am confused here. It seems like it does change with this proposal?

neps/nep-0491.md
```diff
@@ -0,0 +1,381 @@
...
+user in restricted mode could also upgrade an account by sending the initial
+amount of NEAR deposited by the account creator and will attach a new
+`FullAccessKey`.
+
+The downside of this idea is additional complexity on the tooling side because
+actions like adding access keys to the account need to be converted to function
+calls instead of being direct actions. And the complexity on the business side
+is that it needs to include the proxy logic with their business logic in the
+same contract, increasing the complexity of development.
+
```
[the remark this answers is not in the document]
bowenwang1996: Agree with your assessment on the complexity of the `GranularAccessKey` change. I think Marcelo's proposal of proxy smart contracts is definitely worth considering though

Sarah-NEAR:
Hi, strangers! 👋 As the moderator (on behalf of @ori-near), I would like to thank @jakmeier, @walnut-the-cat, @akhi3030, and @birchmd for completing the technical review. Based on your comments above, it seems like this proposal is ready for the working group members for review.
@bowenwang1996, @birchmd, @mfornet – Can you please fully read this NEP and comment in the thread if you are leaning towards approving or rejecting it? Please make sure to include your rationale and any feedback that you have for the author.

mfornet:

neps/nep-0491.md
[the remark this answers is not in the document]
staffik: See https://github.com/jakmeier/NEPs/blob/non_refundable_storage/neps/nep-0491.md#variation-allow-adding-nonrefundable-balance-to-existing-accounts

neps/nep-0491.md
jbajic: The creation of implicit accounts without reserved storage does not change, the request stays the same, I will remove the `ReserveStorage` action below

ori-near:
As the moderator, I would like to thank the @jakmeier for submitting this Protocol NEP, and for the @near/wg-protocol work group members for your review. Based on the voting indications above, it seems like this proposal is close to a decision. We are therefore scheduling the sixth Protocol Work group call this week, where this NEP can enter the final voting stage. Anyone can discuss the technical details by adding your comments to this discussion. And join the call to learn about the final decision and how to get more involved.

**Meeting Info**
📅 Date: Friday, Dec 15, 4:30 PM UTC
✨ [Add to calendar](https://bit.ly/Protocol-Call-Dec15)✨

bowenwang1996:
As a working group member, I lean towards rejecting this proposal for the following reasons:
* What Marcelo proposed [in this comment](https://github.com/near/NEPs/pull/491#discussion_r1349496234) seems like a sensible solution that does not require changes to the protocol. It does create a bit more work on the application level, but we already did similar work in the past to support multisig. 
* It seems to me that the main issue this proposal tries to address is the cost of deploying a smart contract on user accounts, which is a nontrivial amount that users can take back by deleting their account. I definitely agree that this is a problem but I'd rather that we tackle this specific problem through approaches like global smart contracts, which would also be needed down the line for [web3 wallet support](https://github.com/near/NEPs/issues/518). As a result, I am not sure how much value this proposal would provide.
* This proposal introduces an additional 16 bytes of storage per account, which is roughly a 20% overhead. I think we should always be very careful with putting more data in the state to avoid unnecessarily growth in state size.

mm-near:
As a working group member, I lean towards rejecting this proposal:

* Let's start by rolling out the proxy account feature and then see what the community thinks. If there are many scenarios where it doesn't quite fit, we can always circle back to discuss this NEP again.
* Introducing a new Action within the system carries considerable implications across various components, including wallets and other related systems. Such an addition should be considered a measure of last resort especially for actions which involve the movement of tokens. It presents a potential opportunity for a grief attack, where an attacker could deceive a user into creating an account and 'ReserveStorage' for a disproportionately large amount of tokens.

bowenwang1996:
Posting on behalf on the protocol working group:
- We have some concerns regarding this protocol change, including not being the best option to address the problem and introducing a new action that needs support by all tooling
- At the same time, the option of not making the protocol change poses its own challenges and could be more complex to adopt since it requires support from different wallets.

Considering that storage refund is a major blocker for a few important applications, notably fast auth account recovery, the protocol working group agreed to approve this NEP to unblock them.

walnut-the-cat:
During Protocol working group meeting, @mm-near suggested updating the action name to clarify that token will be burnt when reserving storage. 

The currently proposed name is **ReserveStorage** and I would like to collect feedback and get alignment on what the new name should be.

To keep the discussion moving, here is my proposal on naming principles
* *Purpose* first then *consequence*. Action name should emphasize what the action is for first before listing side effects.
* Avoid confusing words. *Stake* or *reserve* shall be avoid as they implies non-permanent, reversible behavior. 
* Stay simple. There may be multiple outcomes and side effects of the action, but focus on the most critical ones.

When it comes to the actual naming, my brain doesn't go further than something like **GrantStorageByBurningToken**. If anyone has a suggestion/thought, happy to hear.

cc. @staffik

Sarah-NEAR:
Thank you to everyone who attended the sixth Protocol Work Group call today! The work group members reviewed the NEP and reached the following consensus:

**Status:** Approved

 👍@birchmd:[ https://github.com/near/NEPs/pull/491#issuecomment-1850078450](https://github.com/near/NEPs/pull/491#issuecomment-1850078450)
 👍 @mfornet: https://github.com/near/NEPs/pull/491#issuecomment-1851941757
 👎 **@mmnear**: https://github.com/near/NEPs/pull/491#issuecomment-1854851236
 👎@bowenwang1996: https://github.com/near/NEPs/pull/491#issuecomment-1854363076

After further discussion, Bowen broke the tie and the working group approved the NEP.

**Meeting Recording:**
https://bit.ly/483Wo6X

@jakmeier Thank you for authoring this NEP!

**Next Steps:**
 @walnut-the-cat finalize decision on renaming ReserveStorage

walnut-the-cat:
> During Protocol working group meeting, @mm-near suggested updating the action name to clarify that token will be burnt when reserving storage.
> 
> The currently proposed name is **ReserveStorage** and I would like to collect feedback and get alignment on what the new name should be.
> 
> To keep the discussion moving, here is my proposal on naming principles
> 
>     * _Purpose_ first then _consequence_. Action name should emphasize what the action is for first before listing side effects.
> 
>     * Avoid confusing words. _Stake_ or _reserve_ shall be avoid as they imply non-permanent, reversible behavior.
> 
>     * Stay simple. There may be multiple outcomes and side effects of the action, but focus on the most critical ones.
> 
> 
> When it comes to the actual naming, my brain doesn't go further than something like **GrantStorageByBurningToken**. If anyone has a suggestion/thought, happy to hear.
> 
> cc. @staffik

@jakmeier , @birchmd , @mfornet , @DavidM-D , please share your thoughts on the naming :)

jakmeier:
`WithNonrefundableStorage` is neat. I like how it makes it obvious it is not a general purpose action.

Some people didn't like "nonrefundable" to begin with. But I just can't think of something that sounds nicer without becoming more vague. The alternatives I found in the discussion are "dedicated storage balance", "storage-tied balance", "storage-restricted balance" by @norwnd in [their comment in August](https://github.com/near/NEPs/pull/491/files#r1290469418). To me, these sound more vague than nonrefundable storage but I think it's worth brining them up again here.
But I acknowledge that it can be confusing for users to see a balance called "nonrefundable storage" listed in their wallet.

`StoragePurchase` would also work. But I imagine that can also lead to confusion. Some people might see this action in the docs and conclude they have to purchase storage explicitly every time, when in reality this is only needed in very specific circumstances. The same goes for an  `BurnForStorage` which was [proposed a month ago](https://github.com/near/NEPs/pull/491#issuecomment-1813352619) by @frol among other suggestions.

In conclusion, I personally think `WithNonrefundableStorage` is the most purpose driven and the least confusing of all the names proposed so far. It's not the simplest name but it feels like a trade-off worth making. And for UX, I would push the problem to wallet implementations. They can always present a term that's easier to understand to their users if they don't want to use "nonrefundable storage" in their UI.

frol:
> `WithNonrefundableStorage` is neat. I like how it makes it obvious it is not a general purpose action.

This semantically limits this action to account creation context (at least in my head), and I don't see a reason to have this limitation. I don't like the idea of introducing "WithNonrefundableStorage" as an **Action**.

One suggestion that was not considered here is to introduce a more generic extendible action (draft naming `ConfigureAccount`):

```js
{
  "actions": {
    "CreateAccount": {},
    "Transfer": {
      "deposit": "500100",
    },
    "AddKey": ...,
    "ConfigureAccount": [
      {
        "property": "NONREFUNDABLE_STORAGE",
        "value": "500000"
      }
    ]
  }
}
```

The benefits I see here:

* It is more future-proof and potentially other properties could be set this way
* The `Transfer` action still includes full transferred amount, so existing tooling won't need to have a special treatment (we already need to handle FunctionCall.deposit and Transfer.deposit, so it is not a huge problem, but may smooth out the adoption)

birchmd:
A general `ConfigureAccount` action is a nice idea. I agree it makes the change more extensible in the future. But it feels like a bit of scope creep in this NEP so I wonder if it is too late to include this modification to the proposal.

> This semantically limits this action to account creation context

Yes, this was my intention. The current NEP says

```
If a non-refundable transfer arrives at an account that already exists, it will fail and the funds are returned to the predecessor.
```

so I wanted the name of the action to reflect this.

frol:
> Yeah the proposal as it stands actually only works for account creation.
> But it feels like a bit of scope creep in this NEP so I wonder if it is too late to include this modification to the proposal.

I don't want to drop bombs here, so I will leave it up to the WG. Another naming idea is `CreateAccountWithNonrefundableStorage`, so at least it is aligned with the other actions being verbs.

staffik:
> > Yeah the proposal as it stands actually only works for account creation.
> > But it feels like a bit of scope creep in this NEP so I wonder if it is too late to include this modification to the proposal.
> 
> I don't want to drop bombs here, so I will leave it up to the WG. Another naming idea is `CreateAccountWithNonrefundableStorage`, so at least it is aligned with the other actions being verbs.

For a named account creation the transaction would be: 
```
CreateAccount
CreateAccountWithNonrefundableStorage
...
```
that looks a bit weird to me.

The original (temporary) name for the action was `TransferV2` and the existing implementation allows to create implicit accounts with non-refundable storage (using `TransferV2` as the only action in a transaction). We can disallow it but I don't see a reason to add this limitation. However, `WithNonrefundableStorage` as the only action in a transaction would look weird too.

staffik:
What about `NonrefundableStorageTransfer` ?
It conveys 3 points:
- it is a transfer
- it is for storage
- it is non-refundable

It is quite close to `WithNonrefundableStorage` yet it is a verb and it makes sense both in
```
CreateAccount
NonrefundableStorageTransfer
...
```
and as a standalone action for implicit account creation.

It is quite long but I could not find a better alternative to the `Nonrefundable` part. I considered `LockedStorageTransfer` but it might suggest that it can be unlocked.

staffik:
For anyone interested, the PR is ready for review https://github.com/near/nearcore/pull/9600.
I've used `NonrefundableStorageTransfer` action name because lack of a feedback, but it can be changed easily during the review process.

bowenwang1996:
@staffik @jakmeier question: it looks like the nonrefundable balance is essentially burnt but the burning is not recorded on the protocol level. What was the rationale behind this decision vs explicitly burning the balance and record `permanent_storage_bytes` on the account (instead of `nonrefundable`) and modify the invariant to be
```rust
amount + locked >= max(0, storage_usage - permanent_storage_bytes) * storage_amount_per_byte
```
This way the burning of tokens is reflected on the protocol level (total supply is decreased). At the same time, it is less confusing for users and integration partners who need to understand what `nonrefundable` means

staffik:
> @staffik @jakmeier question: it looks like the nonrefundable balance is essentially burnt but the burning is not recorded on the protocol level. What was the rationale behind this decision vs explicitly burning the balance and record `permanent_storage_bytes` on the account (instead of `nonrefundable`) and modify the invariant to be
> 
> ```rust
> amount + locked >= max(0, storage_usage - permanent_storage_bytes) * storage_amount_per_byte
> ```
> 
> This way the burning of tokens is reflected on the protocol level (total supply is decreased). At the same time, it is less confusing for users and integration partners who need to understand what `nonrefundable` means

I think one rationale was to make it easier if we decide to implement returning the nonrefundable balance to the sponsor. But otherwise, `permanent_storage_bytes` looks much better to me.

bowenwang1996:
@staffik @jakmeier thanks for the context. Given the discussion throughout this NEP, it doesn't seem like we will end up implementing return of balance to the sponsor given its complexity. So should we just change the implementation to properly reflect that the nonrefundable balance is burnt?

staffik:
> @staffik @jakmeier thanks for the context. Given the discussion throughout this NEP, it doesn't seem like we will end up implementing return of balance to the sponsor given its complexity. So should we just change the implementation to properly reflect that the nonrefundable balance is burnt?

If no one disagrees, I think we should change the implementation. I will just confirm with Coinbase if they are ok with this change.

walnut-the-cat (approved):

PR #492, opened 2023-07-27
proposal to restrict ethereum address on NEAR

bowenwang1996:
Proposal to restrict the creation of Ethereum addresses on NEAR to prevent loss of funds and enable future possibilities for Ethereum wallets to support NEAR transactions.

Implementation: https://github.com/near/nearcore/pull/9365


pvolnov:
What restrictions are there on creating an address through the registrar account? Will it require a signature with the Ethereum private key to prevent spam?

abacabadabacaba:
Seems hacky. Why not restrict creation of top-level accounts entirely? Is there a use case for user-created top-level accounts (other than implicit accounts)?

Making a wallet that automatically switches between NEAR and Ethereum based on address doesn't sound good, because this approach cannot be extended to support other chains (many of which use the same address format as Ethereum). Similarly, this proposal may prevent users from accidentally withdrawing their tokens to NEAR, but not to any of the other chains, so this solution doesn't scale.

By the way, do exchanges even prevent withdrawals to a non-existent address? In most blockchains, transferring tokens to a non-existent address is not a problem, so many exchanges may not have any code to detect that.

A much better approach would be to standardize a set of prefixes denoting the chain, e.g. prepend `near:` to every NEAR address, `eth:` to every Ethereum address, and so on. If the wallets refuse to send tokens to an address without a correct prefix, the possibility of making a mistake would be greatly reduced, and this works with all chains.

mfornet:
This seems to be a problem to be solved in the wallet / web interface, rather than in the protocol. Agree with @abacabadabacaba that this proposal feels hacky. What if someone pastes an ethereum address without `0x` prefix, that could be a source of confusion as well. We can restrict altogether the creation of TLD without going through registrar, and have this logic on the registrar itself, rather than in the protocol.

neps/nep-0492.md
```diff
@@ -0,0 +1,74 @@
...
+
+Ethereum wallets such as Metamask could potentially support NEAR transactions through meta transactions.
+
+## Consequences
+
+In the short term, no Ethereum addresses would be allowed to be created, but this change would not create any problem for users.
+
+### Backwards Compatibility
+
+There is no backwards compatibility concern.
```
mfornet, line 70: To some extent there is given there are ~5k such addresses already.

frol:
Thank you @bowenwang1996 for submitting this NEP.

<details>

Technical Summary guidelines:

```

### Recommendation

Add recommendation

### Benefits

* Benefit

* Benefit

### Concerns

| - | - | - | - |

```

</details>

bowenwang1996:
yeah I think restricting all TLDs (other than implicit accounts) make sense.

bowenwang1996:
@abacabadabacaba @mfornet the proposal is updated to restrict the creation of all top level accounts (other than implicit accounts)

mfornet (approved):

neps/nep-0492.md
```diff
@@ -0,0 +1,74 @@
...
+anyone can create such an account. This has unfortunately caused a few incidents where users lose their funds due to either a scam or careless behaviors.
+For example, when a user withdraw USDT from an exchange to their NEAR account, it is possible that they think they withdraw to Ethereum and therefore enter their Eth address.
+If this address exists on NEAR, then the user would lose their fund. A malicious actor could exploit this can create known Eth smart contract addresses on NEAR to trick users to send tokens to those addresses. With the proliferation of BOS gateways, including Ethereum ones, such exploits may become more common as users switch between NEAR wallets and Ethereum wallets (mainly metamask).
+
+In addition to prevent loss of funds for users, this change allows the possibility of Ethereum wallets supporting NEAR transactions, which could enable much more adoption of NEAR. The exact details of how that would be done is outside the scope of this proposal.
+
+There are currently ~5000 Ethereum addresses already created on NEAR. It is also outside the scope of this proposal to discuss what to do with them.
```
birchmd, line 27: What about other non-Ethereum-like top level accounts whos creation would have been restricted by this proposal?

bowenwang1996:
@frol given that both Marcelo and Michael have approved the proposal, looks like we have a majority of support from the protocol working group. What else do we need to merge this proposal?

neps/nep-0492.md
bowenwang1996: There are a total of 11423 accounts on mainnet that are top-level accounts longer than 32 characters and that are not implicit accounts. I don't think there is a clear pattern that people use top-level accounts longer than 32 characters for.

frol:
@bowenwang1996 Thank you for reaching out! I moved this NEP to final comment period and will ask @near/nep-moderators to schedule a Protocol WG call to present this NEP to community and formally vote on it.

birchmd (approved):
As a member of the protocol working group, I approve this NEP.

Some notes from the discussion we had at the last open working group meeting:

- The title should be changed to reflect the increased scope (all top level accounts disallowed, not just Ethereum-like)
- Concerns around censorship should be addressed in follow-up discussions (i.e. actually spec out the top-level account name auction system; and in the meantime make the current registrar account more permissive).

mfornet (approved):
As a member of the Protocol WG I approve this NEP

ori-near:
Thank you to everyone who attended the fifth Protocol Work Group call today! The work group members reviewed the NEP and reached the following consensus:

**Status:** Approved

 - * 👍 @birchmd:https://github.com/near/NEPs/pull/492#pullrequestreview-1607239172
 --* 👍 @mfornet: https://github.com/near/NEPs/pull/492#pullrequestreview-1607311478

Meeting Recording:
https://youtu.be/oi0ooi9QBuk

@bowenwang1996  Thank you for authoring this NEP!

Next Steps:
 - @bowenwang1996 Finish and merge implementation https://github.com/near/nearcore/pull/9365
- @bowenwang1996 Stabilize the feature

neps/nep-0492.md
```diff
@@ -0,0 +1,74 @@
+---
+NEP: 492
+Title: Restrict creation of Ethereum Addresses
+Authors: Bowen Wang<bowen@near.org>
+Status: Draft
```
frol, line 5:
```suggestion
Authors: Bowen Wang <bowen@near.org>
Status: Final
```

frol (approved):

PR #393, opened 2022-09-14
feat: adding soulbound token

robert-zaremba:
Soulbound Tokens (SBT) are non transferrable NFTs. Even though tranferability is not available, we define a recoverability mechanism. 

SBTs are well suited of carrying proof-of-attendence NFTs, proof-of-unique-human "stamps" and other similar credibility-carriers.

---

## NEP Status *(Updated by NEP moderators)*

SME reviews:

* [x] @KazanderDad (NDC GWG): https://github.com/near/NEPs/pull/393#pullrequestreview-1380240545
* [x] @alexastrum (Tenamint): https://github.com/near/NEPs/pull/393/#issuecomment-1541004902

Contract Standards WG voting indications (❔ | 👍 | :-1: ):

* 👍 @frol: https://github.com/near/NEPs/pull/393#pullrequestreview-1499578010
* 👍 @fadeevab: https://github.com/near/NEPs/pull/393#pullrequestreview-1506223006
* -- @robert-zaremba (can't vote myself)

Wallet Standards WG voting indications:

* ❔ @Cameron-Banyan 
* ❔ @MaximusHaximus 
* ❔ @esaminu  

## Concerns

| #   | Concern | Resolution                                                                                                                                                                               | Status |
| --- | --- | -- | -- |
| 1   | [Rober] Should we Emit NEP-171 Mint and NEP-171 Burn by the SBT contract (in addition to SBT native events emitted by the registry)? If the events will be emitted by registry, then we need new events to include the contract address. | [Robert] Don't emit NFT events. SBT is not NFT. Support: @alexastrum | open |
| 2   | [Robert] remove `memo` in events. The `memo` is already part of the transaction, and should not be needed to identify transactions. Processes looking for events, can easily track transaction through event and recover `memo` if needed. | currently removed, consequently also removed from registry transactions . Support: @alexastrum |  open |
| 3   | [Token Spam](https://github.com/near/NEPs/pull/393/#discussion_r1163938750) | [Robert]: we have a `Burn` event. Added example `sbt_burn` function, but keeping it not as a part of required interface. Event should be enough. | open |
| 4   | [Multiple registries](https://github.com/near/NEPs/pull/393/#discussion_r1163951624). Registry source of truth [comment](https://github.com/near/NEPs/pull/393/#issuecomment-1531766643) | Robert: this is a part of the design: permissionless approach. [Justification for registry](https://github.com/near/NEPs/pull/393/#issuecomment-1540621077) | open |
[...]

PR #504, opened 2023-09-06
NEP: SBT Metadata Standard

robert-zaremba:
A standard interface for [SBT](https://github.com/near/NEPs/pull/393) Metadata: tokens, classes and issuers.
This allows an SBT registry and issuers to be interrogated for its name and for details an SBT represent.

The proposal addresses potential challenges related to metadata reconstruction, off-chain storage security, owner/operator dependence, and data privacy.

Related update to NEP-393: https://github.com/near/NEPs/pull/393/commits/ad56e5f4d95100a7d051a2f6c80952f960e1ce1d

---

## NEP Status *(Updated by NEP moderators)*

SME reviews:

* [ ] sme1
* [ ] sme2 

Contract Standards WG voting indications (❔ | 👍 | :-1: ):

* ❔ @frol: https://github.com/near/NEPs/pull/393#pullrequestreview-1499578010
* ❔ @fadeevab: https://github.com/near/NEPs/pull/393#pullrequestreview-1506223006
* -- @robert-zaremba (can't vote myself)

Wallet Standards WG voting indications:

* ❔ @Cameron-Banyan 
* ❔ @MaximusHaximus 
* ❔ @esaminu  

## Concerns


codingshot:
- Adding .com to label for social media seems like a waste of space
- event emitted should match contract_metadata_update https://github.com/near/NEPs/pull/423/commits or it should be clear whether which individual SBT is updated vs the contract metadata as a whole
- interesting choice for contact to be a list
- soundcloud is uncessary and general website should be replaced with this as an example

I think there is limitations with just png and svg, in future object files and gif could be including. Was recently talking to a team from Solart about bringing a metadata specific chain to handle the robustness of document types. and information about the underlying file.

neps/nep-504.md
```diff
@@ -0,0 +1,242 @@
...
+
+```rust
+pub struct ClassMetadata {
+    /// Issuer class name. Required.
+    pub name: String,
+    /// If defined, should be used instead of `contract_metadata.symbol`.
+    pub symbol: Option<String>,
+    /// Icon content (SVG) or a link to an Icon. If it doesn't start with a scheme (eg: https://)
+    /// then `contract_metadata.base_uri` should be prepended.
+    pub icon: Option<String>,
```
fadeevab, line 47: Limiting the `icon` format to the "Data Url" format makes sense from the perspective of protecting fellow developers from unintentionally triggering any SSRF vulnerabilities with URL parsers.

neps/nep-504.md
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// An event emitted by an SBT Issuer when Class Metadata are updated.
+type Class {
+  classes: ClassId[];  // list of class IDs whose metadata has been updated
+}
+
+/// An event emitted by an SBT Issuer when Contract Metadata is updated.
+/// Note: it's an empty object
+type Class { }
```
fadeevab, line 162: Did you mean `Contract`?
I would disambiguate event types: maybe `ClassData` and `ContractData` would sound better (?)

neps/nep-504.md
```diff
@@ -0,0 +1,242 @@
...
+- When possible, events should contain aggregated data, with respect to the standard function related to the event.
+
+The Contract and Class Metadata are stored and managed by the Issuer contract. This NEP covers only the Issuer contract. `TokenMetadata` is stored in the SBT registry, consequently Token Metadata updates and events are defined in NEP-393.
+
+```typescript
+type Nep504Event {
+  standard: "nep504";
+  version: "1.0.0";
+  event: "class" | "contract";
+  data: Class | Contract ;
```
fadeevab, line 152:
```suggestion
  data: Class | Contract;
```

neps/nep-504.md
```diff
@@ -0,0 +1,240 @@
...
+  },
+  // Sub object defining contact resources. Each value must be a list.
+  "contact": {
+    "email": ["contact@example.com", "contact2@example.com"], // must be a list
+    "telegram": ["telegram_handle_1", "telegram_handle2"], // must be a list
+  },
+  // Default token icon
+  "icon": {
+    "png": "https://genadrop.mypinata.cloud/ipfs/...",
+    "svg": "https://genadrop.mypinata.cloud/ipfs/..."
```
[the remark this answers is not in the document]
fadeevab: @robert-zaremba ...or it may be just a part of the URL: "https://image.com/?w=1024&h=748"

neps/nep-504.md
```diff
@@ -0,0 +1,242 @@
...
+    social_media: Map<String, String>,
+    // keys are contact types
+    contact: Map<String, Vec<String>>,
+    // keys are image format
+    icon: Map<String, Vec<String>>,
+    // keys are video format
+    video: Map<String, Vec<String>>,
+    // keys are audio format
+    audio: Map<String, Vec<String>>,
+}
```
fadeevab, line 85: comments on every field....

neps/nep-504.md
```diff
@@ -0,0 +1,242 @@
...
+NEAR doesn't have a well structured Metadata standard. The [NFT Metadata](https://github.com/near/NEPs/blob/master/neps/nep-0177.md) doesn't clearly define the `reference` extension. In fact, the NEP-393 Metadata is defined based on NFT Metadata. The schema defined here can be used by other token variations as well.
+
+Issuers, contract implementers and users demand a guideline and a minimum schema to make better integration with tools and projects. This document defines a minimum common schema for reference fields of `ContractMetadata` and `TokenMetadata`. Moreover, while working with various issuers, we found that issuers and minters often duplicate lot of data in the `TokenMetadata`. We found that it is unnecessary and it would be useful to create a new type: `ClassMetadata`, that defines common data for all tokens of a given (issuer, class) pair.
+
+## Specification
+
+NEP-393 `TokenMetadata` is minimalistic. It only requires data which a contract can reason about, and all user facing details (such as title, description) are meant to be part of the reference.
+Similarly, `ContractMetadata` only defines issuer basic info.
+
+### Class Metadata
```
fadeevab, line 34: The big question is: where the class metadata is actually used? Any API (view/call methods)?

neps/nep-504.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Reference Field
+
+The `reference` is attribute is part of all Metadata structures (Token, Class and Contract). It must be a valid JSON string or an URL link to a valid JSON file. In the latter case we recommend to set `reference_hash` to facilitate secure off-chain metadata storage and a mechanism for consistency checks.
+It's up to the issuer if he wants to store `reference` fully on chain (as a JSON string) or off-chain. The latter case won't guarantee data availability, but may provide faster data access.
+
+We define a `ReferenceBase` type. All entries are optional, and don't have to be set. If an entry is set, then it must have type compatible with `ReferenceBase`. The Reference object can extend the `ReferenceBase` and provide more fields.
+
+```rust
+struct ReferenceBase {
```
fadeevab, line 71: The second big question is: should the reference field be standardized at all? The structure fields look like taken from a concrete/specialized project, and all of the fields are... optional.

frol: I also don't understand why these fields that are proposed to be standardized are part of `reference`, which was designed to be a link to off-chain additional information.

> It must be a valid JSON string or ...

This is an abuse of the `reference` field. Just add more fields where you need them. That is, I suggest putting these fields on the `*Metadata` level as individual fields instead of creating unnecessary nesting. Metadata structs can have any number of additional fields on them, and that does not contradict the minimum metadata structure as at the API level it is communicated through JSONs and additional fields can be easily ignored.

frol (changes requested):

neps/nep-504.md
fadeevab: @robert-zaremba I think I still don't get whether it's returned via API, or is posted as a log string in the L1 block (?).

neps/nep-504.md
```diff
@@ -0,0 +1,249 @@
...
+    /// An URL to an Icon. To protect fellow developers from unintentionally triggering any
+    /// SSRF vulnerabilities with URL parsers, we don't allow to set an image bytes here.
+    /// If it doesn't start with a scheme (eg: https://)
+    /// then `contract_metadata.base_uri` should be prepended.
+    pub icon: Option<String>,
+    /// JSON or an URL to a JSON file with more info. If it doesn't start with a scheme
+    /// (eg: https://) then base_uri should be prepended.
+    pub reference: Option<String>,
+    /// Base64-encoded sha256 hash of JSON from reference field. Required if `reference` is included.
+    pub reference_hash: Option<Base64VecU8>,
```
frol, line 54: @robert-zaremba Let me try to deliver my point once again:

https://github.com/near/NEPs/blob/6eee9dc14ff64dd2eac24e4ae704e0b8fbb18ba9/neps/nep-0177.md#L45-L46

`reference` field is designed to be a **link** and nothing else.

If you want to have more fields stored as part of `ClassMetadata`, just add those fields on the same level as `name`, `symbol`, ... fields - this is not a breaking change and won't affect consumers of the API as they will just ignore unsupported fields. There is no need in string-encoding JSON data into the `reference` field.

frol (changes requested):

neps/nep-504.md
frol: > My reasoning is to have a strict minimum amount of keys in ClassMetadata, and add more structure for other common requirements, so tools and websites can display more information about tokens.

There is no reason to overcomplicate things. If you want to standardize fields, add them to ClassMetadata optional or not. `reference` is needed when the metadata is too big to be stored inline. There is no reason to store JSON as a string inside JSON value.

> In such case, we would need to extend the ClassMetadata as more requirements will emerge from the ecosystem. So far, for the SBTs we were working with, we need different pictures (size / format), website address, social_media links ...

If you will want to standardize on those, you will still need to extend the NEP and some structure, be it `ClassMetadata` or some other struct, it does not really change things. Also, you don't always need to standardize every single field as that may limit the usability of the whole spec if those fields will be used sparsely.

> I would like to hear more opinions from the tool designers and other projects dealing with tokens.

Sure, make a call in community channels.

robert-zaremba: > you will still need to extend the NEP and some structure

This is what an appendix is for.

We need to have a structure to make sure tools will have a common way to handle the data we describe in this PR. Can be in reference (as originally proposed), or by adding more optional attributes to the top level Metadata. Either way we need to have that structure.

frol: > We need to have a structure to make sure tools will have a common way to handle the data we describe in this PR. Can be in reference (as originally proposed), or by adding more optional attributes to the top level Metadata. Either way we need to have that structure.

I don't argue against having the defined structure, I just suggest moving it to where it belongs (new optional fields on `ClassMetadata`), instead of abusing `reference` field.

robert-zaremba: Sure, I'm OK with that, see the second part of the [post](https://github.com/near/NEPs/pull/504#discussion_r1376331389).

frol: > ClassMetadata is not open ended.

It is, actually, open ended by JSON spec, and that should be encouraged to be used whenever it makes sense.

> Sure, I'm OK with that, see the second part of the https://github.com/near/NEPs/pull/504#discussion_r1376331389.

Alright, so you are waiting for inputs from someone else besides me, right?

robert-zaremba:
Status update: this proposal, while still valid, needs market review and verification. Tools or related projects didn't provide such review. Until this will be done, the PR will stay on hold.

garikbesson:

PR #510, opened 2023-09-25
nep-393: add bool return to sbt_revoke_by_owner

robert-zaremba:
In the [reference](https://github.com/near-ndc/i-am-human/blob/cfc92b1/contracts/registry/src/registry.rs#L362) implementation we return `bool` in the `registry.sbt_revoke_by_owner` to signal if the operation should continue or it's finished.

We forgot to push this update to the standard.

This update also makes the API consistent with `sbt_recover` and `sbt_soul_transfer`

### DIFF
--- a/neps/nep-0393.md
+++ b/neps/nep-0393.md
@@ -405,10 +405,8 @@ trait SBTRegistry {
     fn sbt_revoke(&mut self, tokens: Vec<TokenId>, burn: bool);
 
     /// Similar to `sbt_revoke`. Allows SBT issuer to revoke all tokens by holder either by
-    /// burning or updating their expire time. The function will try to revoke
-    /// at most `MAX_REVOKE_PER_CALL` tokens (to fit into the tx gas limit),
-    /// so when an owner has many tokens from the issuer, the issuer may need to
-    /// call this function multiple times, until all tokens are revoked.
+    /// burning or updating their expire time. When an owner has many tokens from the issuer,
+    /// the issuer may need to call this function multiple times, until all tokens are revoked.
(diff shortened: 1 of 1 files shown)


sczembor:
This change enhances the method to maintain consistency with other methods that may run a risk of running out of gas and require repeated execution. Consequently, it eliminates the necessity for an extra view call to verify the completion status of the revocation process.

KazanderDad (approved):
This change adds a boolean to signal if the operation is finished, which ensures consistency with other methods that may run a risk of running out of gas and require repeated execution. If the result is false then the method should be called again until true is returned. This also eliminates the necessity for an extra view call to verify the completion status of the revocation process.

neps/nep-0393.md
```diff
@@ -404,8 +404,17 @@ trait SBTRegistry {
     /// Must also emit `Burn` event if the SBT tokens are burned (removed).
     fn sbt_revoke(&mut self, tokens: Vec<TokenId>, burn: bool);
 
-    /// Similar to `sbt_revoke`, but revokes all `owner`s tokens issued by the caller.
-    fn sbt_revoke_by_owner(&mut self, owner: AccountId, burn: bool);
+    /// Similar to `sbt_revoke`. Allows SBT issuer to revoke all tokens by holder either by
+    /// burning or updating their expire time. The function will try to revoke
+    /// at most `MAX_REVOKE_PER_CALL` tokens (to fit into the tx gas limit),
```
frol, line 409: What is `MAX_REVOKE_PER_CALL`? It is the only reference here in this document. I believe it should be removed here.

frol (changes requested):

frol (approved):

Issue #246, opened 2021-07-28
Multi Token Contract Standard 


zcstarr:
### Summary
A standard interface for  a multi token standard that supports fungible, semi-fungible, and tokens of any type, allowing for ownership, transfer, and batch transfer of tokens generally regardless of specific type. This standard aims to make it possible to represent these multi tokens in a way that allows these tokens to be represented across chains.

### Motivation
Having a single contract represent both NFTs and FTs can greatly improve efficiency as demonstrated by Enjin Coin. The ability to make batch requests with multiple asset classes can reduce a many transactions, transaction to a single transaction to trade around both NFTs and FTs that are a part of same token contract. Moreover, this flexibility has led to building proper infrastructure for ecosystems, that often have to issue many tokens under management such as games.

Finally on other chains, such as Ethereum it would be nice to be able to bridge a token contract from near to ETH or to other chains. Having a multi token standard that can represent many different classes of tokens is highly desirable

### Background
- Current NEP PR:
  - PR #245
- Prior art:
  - EIP-1155 : https://github.com/ethereum/EIPs/issues/1155

  - This NEP derives some examples and format from: https://github.com/near/NEPs/pull/21

  - Example of an NFT Series: https://github.com/near-apps/gnr8

  - Things from the NFT Discussions: https://github.com/near/NEPs/discussions/171

  - Things from the NFT Discussions: https://gov.near.org/t/nft-standard-discussion/853

- Discussions out of band:
   - https://gov.near.org/t/multi-token-standard-discussion/2917
   - https://github.com/shipsgold/multi-token-standard-impl/tree/main/meetings 
- WIP/impl:
  -  https://github.com/shipsgold/multi-token-standard-impl/tree/feat/initial-token
  - https://github.com/shipsgold/multi-token-standard-impl/pull/1

### Core Trait
```
pub trait MultiTokenCore {
    /// Basic token transfer. Transfer a token or tokens given a token_id. The token id can correspond to  
    /// either a NonFungibleToken or Fungible Token this is differeniated by the implementation.
    ///
    /// Requirements
    /// * Caller of the method must attach a deposit of 1 yoctoⓃ for security purposes
    /// * Contract MUST panic if called by someone other than token owner or,
    /// * If using Approval Management, contract MUST nullify approved accounts on
    ///   successful transfer.
    /// * TODO: needed? Both accounts must be registered with the contract for transfer to
    ///   succeed. See see https://nomicon.io/Standards/StorageManagement.html
    ///
[...]

PR #511, opened 2023-10-11
Fix nep245

marco-sundsk:
Now, we have a revision of this standard. This time, we focus on the combination of FT and NFT in those core functions cause we noticed some misunderstandings on FT and NFT workflow on near network in those interfaces. Beside that, we also make a supplement to the nep-0245.md to include all the extensions while leaving the one in specs untouched as a comparision base point, so that you could have a clear view about what have been changed in this revision.  

The plan is to have an agreement on nep-0245.md first, then the corresponding fix would be applied to those in specs. And the standard code implementation would be the last.  

The detailed explaination of this revision could be found in the [discussion](https://github.com/near/NEPs/issues/246).


neps/nep-0245.md
```diff
@@ -326,14 +309,11 @@ The following behavior is required, but contract authors may name this function
 //   `approvals` is an array of expected `approval_list` per `token_ids`. 
 //   If a `token_id` does not have a corresponding `approvals_list` then the entry in the 
 //   array must be marked null.
-//   `approvals_list` is an array of triplets of [`owner_id`,`approval_id`,`amount`]. 
-//   `owner_id` is the valid Near account that owns the tokens.   
+//   `approvals_list` is an array of triplets of [`approved_account_id`,`approval_id`].
```
ruseinov, line 312:
```suggestion
//   `approvals_list` is an array of tuples of [`approved_account_id`,`approval_id`]. 
```

neps/nep-0245.md
```diff
@@ -382,24 +362,379 @@ Contracts which want to make use of `mt_transfer_call` and `mt_batch_transfer_ca
...
+  updated_at: string|null, // When token was last updated, Unix epoch in milliseconds
+  extra: string|null, // Anything extra the MT wants to store on-chain. Can be stringified JSON.
+  reference: string|null, // URL to an off-chain JSON file with more info.
+  reference_hash: string|null // Base64-encoded sha256 hash of JSON from reference field. Required if `reference` is included.
+}
+
+// For base info, either base_id or base would be included, depends on implementation.
+// Suggest to return base to save client from extral query on base info.
+type MTTokenMetadataAll = {
+  base_id: string | null, // id in type MTBaseTokenMetadata
```
ruseinov, line 434: This seems confusing, should this be instead restricted to one or the other (base_id vs including the metadata itself)?

neps/nep-0245.md
```diff
@@ -62,12 +62,12 @@ To recap, we choose to create this standard, to improve interoperability, develo
...
 ```ts
-// The base structure that will be returned for a token. If contract is using
-// extensions such as Approval Management, Enumeration, Metadata, or other
-// attributes may be included in this structure.
+// The base structure that will be returned for a token.
 type Token = {
   token_id: string,
-  owner_id: string | null
+  owner_id: string | null, // not null if the token is a NFT
+  metadata: MTTokenMetadataAll | null // not null if Metadata extension is enabled
```
ruseinov, line 69: How/when does this field get assigned to a token?

ruseinov: Curious if we need to deal with Metadata at all at this stage, wouldn't it be nicer to go with iterative approach and get the barebone tokens right up to the contract implementation and then deal with extensions?

neps/nep-0245.md
```diff
@@ -92,21 +92,17 @@ type Token = {
...
 //    providing information for a transfer
-
-
 function mt_transfer(
   receiver_id: string,
   token_id: string,
-  amount: string,
-  approval: [owner_id: string, approval_id: number]|null,
+  amount: string, 
+  approval_id: number|null,
```
ruseinov, line 105: Same goes for approvals, do we really want this to be a part of the interface.
Why don't we just do the barebones MT standard without the extensions and then figure out how those will play together with what we have?

neps/nep-0245.md
```diff
@@ -382,24 +362,379 @@ Contracts which want to make use of `mt_transfer_call` and `mt_batch_transfer_ca
...
+The primary value of tokens comes from their metadata. While the [core standard](Core.md) provides the minimum interface that can be considered a multi token, most artists, developers, and dApps will want to associate more data with each token, and will want a predictable way to interact with any MT's metadata.
+
+NEAR's unique [storage staking](https://docs.near.org/concepts/storage/storage-staking) approach makes it feasible to store more data on-chain than other blockchains. This standard leverages this strength for common metadata attributes, and provides a standard way to link to additional offchain data to support rapid community experimentation.
+
+This standard also provides a `spec` version. This makes it easy for consumers of Multi Tokens, such as marketplaces, to know if they support all the features of a given token.
+
+Prior art:
+
+- NEAR's [Fungible Token Metadata Standard](../FungibleToken/Metadata.md)
+- NEAR's [Non-Fungible Token Metadata Standard](../NonFungibleToken/Metadata.md)
```
ruseinov, line 387: Those links seem broken.

frol:
Hi @marco-sundsk @ruseinov, thank you for pushing this NEP forward! I see that this is still an on-going effort, so as a NEP moderator, I marked this PR as draft and marked that it needs reference implementation (as per our discussion in Telegram, we don't need a full SDK-grade library implementation, we just need a standalone contract that implements all these methods and some client that demonstrates that the API surface is not lacking and ergonimic)

Please ping the @near/nep-moderators once you are ready for us to review it.

victorchimakanu:
Hi @marco-sundsk @ruseinov, i'd like to understand the current status of this NEP.

ruseinov:
> Hi @marco-sundsk @ruseinov, i'd like to understand the current status of this NEP.

The status is that we have a small MVP in another repo that's supposed to drive this effort. I'm currently busy with other things, but planning to eventually come back to it.

ruseinov:
> Hi @ruseinov @marco-sundsk Do you still want to proceed with this NEP-fix?

At some point. Currently swamped, but hoping to come back to it sometime in the future.

PR #519, opened 2023-11-17
WIP: yield execution

akhi3030:


neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,164 @@
...
+```rust
+/// Instructs the protocol that the smart contract is not ready yet to respond
+/// to its caller yet.  The smart contract promises to call `yield_resume()`
+/// within `yield_num_blocks` blocks.  When `yield_resume()` is called, the
+/// protocol will call the method on the smart contract that is identified by
+/// `method_name_len` and `method_name_ptr` and this method will be expected to
+/// either respond to the caller or create another promise.
+///
+/// If the contract fails to call `yield_resume()` within `yield_num_blocks`,
+/// then the protocol will call the method on the smart contract that is
```
DavidM-D, line 48:
```suggestion
/// then the protocol may call the method on the smart contract that is
```
I don't think this needs to be required in the protocol, rather it's an option the protocol has.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,164 @@
...
+
+```rust
+/// Instructs the protocol that the smart contract is not ready yet to respond
+/// to its caller yet.  The smart contract promises to call `yield_resume()`
+/// within `yield_num_blocks` blocks.  When `yield_resume()` is called, the
+/// protocol will call the method on the smart contract that is identified by
+/// `method_name_len` and `method_name_ptr` and this method will be expected to
+/// either respond to the caller or create another promise.
+///
+/// If the contract fails to call `yield_resume()` within `yield_num_blocks`,
```
DavidM-D, line 47:
```suggestion
/// If the contract fails to, in a `yield_resume()`, call `value_return`, `panic` or `abort` within `yield_num_blocks`,
```
This makes the API much more flexible, but does it make it harder to implement?

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,164 @@
...
+The proposal is to add the following host functions to the NEAR protocol:
+
+
+```rust
+/// Instructs the protocol that the smart contract is not ready yet to respond
+/// to its caller yet.  The smart contract promises to call `yield_resume()`
+/// within `yield_num_blocks` blocks.  When `yield_resume()` is called, the
+/// protocol will call the method on the smart contract that is identified by
+/// `method_name_len` and `method_name_ptr` and this method will be expected to
+/// either respond to the caller or create another promise.
```
DavidM-D, line 45:
```suggestion
/// `method_name_len` and `method_name_ptr` and this method may respond to the caller.
/// Once the method has responded, `yield_resume` can no longer be called on this promise.
```
The difference here is that the promise index remains the same with multiple calls to resume. The current design means that you need to have at most one resume call in flight at a time or the second one will fail.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,164 @@
...
+/// improves the devX by allowing the developer to specify how to divide up the
+/// remaining gas.
+///
+/// Return value: u64: Similar to the
+/// [promise_create](https://github.com/near/nearcore/blob/a908de36ab6f75eb130447a5788007e26d05f93e/runtime/near-vm-runner/src/logic/logic.rs#L1281)
+/// host function, this function also create a promise and returns an index to
+/// the promise.
+pub fn yield_create(
+    method_name_len: u64,
+    method_name_ptr: u64,
```
DavidM-D, line 65: Why have the method name specified here rather than in `yield_resume`? Since we already have to validate the input of `yield_resume` this change would appear to make this strictly more powerful.

Happy to keep it this way if it's easier to implement though.

DavidM-D (commented):
Looks like a really good start and I'm broadly happy with it in it's current form. The biggest thing I'd change is the changing promise ID each time resume is called, everything else is nits.

neps/nep-519-yield-execution.md
akhi3030: Hmm... I would've thought that this must happen at some point otherwise the smart contract might leak memory.  If the contract created some state and then called `yield_create` which it then wanted to free in the callback, that might never happen.  

Also I see this as analogous to the case where contract A calls B and B fails to respond to A.  Then the protocol generates a response for A.

neps/nep-519-yield-execution.md
akhi3030: I'm happy to make this change.  I don't think it will make the implementation trickier.  I don't even remember why I chose it this way now.

neps/nep-519-yield-execution.md
akhi3030: If I understood the idea properly, then I think we need to think about the semantics some more.  Let's say the following has happened:

- contract A called the signer to get a signature and the signer has called yield create.
- contract B called the signer to get a signature and the signer has called yield create.
- now some other contract C calls the signer and the signer `panic`s, which one of the above two outstanding executions should we terminate?  I don't think we can really pick between them so the only sensible choice seems to be to terminate both.  It doesn't seem like that that is a good choice either.

neps/nep-519-yield-execution.md
akhi3030: Ah interesting, yes I think we should support this.  I will add this suggestion.  I don't have a good intuition for whether or not it will make the implementation trickier.  Let's find out.

akhi3030 (commented):
Thanks for the feedback @DavidM-D.  I've added some comments.  I am going to make some changes to the NEP to remove the concept of `yield_num_blocks`.  Unfortunately, given the crummy github UI, it might become challenging to continue the comment threads after the changes. We will have to stumble along somehow.

neps/nep-519-yield-execution.md
saketh-are: I had written it this way because it made sense to me for the prepaid gas amount to be specified in the same place as the method which is being prepaid for. If we separate them the smart contract author will have to think about whether the gas prepaid in one place (`yield_create`) will be sufficient for the method call specified in a different place (`yield_resume`). 

That said, overall I do like the added flexibility and am happy with the change.

neps/nep-519-yield-execution.md
DavidM-D: So if contract C calls the signer and that call panics before it calls `yield_create` then the call from contract C should panic and none of the others.

This refers to a situation where contract D calls `yield_create`, the signer calls `yield_resume` and then the resumed call panics, ending contract D's suspended call. It shouldn't effect any other outstanding calls.

neps/nep-519-yield-execution.md
DavidM-D: OK I agree with this, it makes life easier for contract developers if there is always a response eventually.

neps/nep-519-yield-execution.md
saketh-are: This might not work for a different reason; we need to know the signature of the callback method to support composability with other promises. For example if we wish to compile something like `promise_create(...).and_then(...)`, we should know the type returned by the promise callback.

bowenwang1996:
At the protocol working group meeting yesterday, different members expressed concerns on these new runtime APIs. I'll attempt the summarize them below:
* There is nothing that stops someone from emulating the API in the sdk by saving the promise in the state and triggering it again once it is ready to be resumed. The call stack is discontiguous but that could potentially be fixed on the tooling level and it seems that Aurora already does this.
* Whether or not the delayed promise has an expiration time, the introduction of this API could lead to a new type of failure that does not previously exist where the contract execution fails with no way to clean up the state. It could be fine for the chain signature use case, but if some (potentially malicious) developer uses this in a defi contract for oracles (for example), then it is possible that the contract call stalls/fails in the middle, which results in users losing access to their fund.
* The API does not specify who can call "resume". This means that a malicious attacker can just call resume with invalid data to make a transaction fail. While the attacker does need to spend gas themselves, it is quite cheap. This allows for a relatively easy DoS attack, especially given that the gas that a user needs to spend may be larger than what an attacker needs to spend to cancel the transaction. In addition, failure in a chain of calls A -> B -> C may also lead to inconsistent state in one of the smart contract involved.

cc @mm-near @birchmd @mfornet

akhi3030:
>     * There is nothing that stops someone from emulating the API in the sdk by saving the promise in the state and triggering it again once it is ready to be resumed. The call stack is discontiguous but that could potentially be fixed on the tooling level and it seems that Aurora already does this.

This main motivation for this work is the MPC project.  If we can satisfy the requirements without having to change the protocol, that much better.  

Can you provide some links to how aurora does this?  Maybe then @DavidM-D and @itegulov can review them.  

> 
>     * Whether or not the delayed promise has an expiration time, the introduction of this API could lead to a new type of failure that does not previously exist where the contract execution fails with no way to clean up the state. It could be fine for the chain signature use case, but if some (potentially malicious) developer uses this in a defi contract for oracles (for example), then it is possible that the contract call stalls/fails in the middle, which results in users losing access to their fund.

I am not sure I completely understand this point.  Especially about not having a way to clean up the state.  If things expire, the protocol will call the contract.  So is the contract may not handle that case properly?  Isn't that similar to how the current Promise API works?

> 
>     * The API does not specify who can call "resume". This means that a malicious attacker can just call resume with invalid data to make a transaction fail. While the attacker does need to spend gas themselves, it is quite cheap. This allows for a relatively easy DoS attack, especially given that the gas that a user needs to spend may be larger than what an attacker needs to spend to cancel the transaction. 
If contract A called `yield_create`, only contract A can call `resume` on that promise.  No other contract can.  

> In addition, failure in a chain of calls A -> B -> C may also lead to inconsistent state in one of the smart contract involved.

Can you please expand on this?  If A is calling B, then B can fail for any number of reasons today, (it can trap, run out of gas, have invalid state).  Is the problem that we are introducing a new failure mode now that B has to handle?

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,164 @@
...
+2. When called, the contract updates some contract state which is being monitored by external indexers to indicate that a new signing request has been received.  It also defers replying to the caller.
+3. The indexers observe the new signing request, they compute a signature and call another function `fn signature_available(Signature, ...)` on the signer contract.
+4. The signer contract validates the signature and if validate, replies to the original caller.
+
+Today, the NEAR protocol has no sensible way to defer replying to the caller in step 2 above.  This proposal proposes adding two following new host functions to the NEAR protocol:
+
+- `yield`: this can be called by a contract to indicate to the protocol that it is not ready yet to reply to its caller.
+- `resume`: a contract can use this mechanism to indicate to a protocol that it is now ready to reply to a caller that it had deferred earlier.
+
+If these two host functions were available, then `yield` would be used in step 2 above and `resume` would be used in step 4 above.
```
mfornet, line 32: Can you expand on the motivation section why is the new interface of `yield` / `resume` essential for this workflow. What you describe can be done with two transactions today. 

1. Signer contract provides a function `fn sign_payload(Payload, ...)` **[SAME]**
2. When called, the contract updates some contract state which is being monitored by external indexers to indicate that a new signing request has been received. It also defers replying to the caller **on a future transaction**.
3. The indexers observe the new signing request, they compute a signature and call another function **in a new transaction** `fn signature_available(Signature, ...)` on the signer contract. 
4. The signer contract validates the signature and if validate, replies to the original caller. **[SAME]**

This workflow requires two transactions (on step 1 by the user, and on step 3 by the MPC network).

akhi3030: >  It also defers replying to the caller on a future transaction.

How can a contract do that today?  My understanding is that if contract A calls B today, one of the following things must happen:

1. B must provide a reply
2. B calls another contract (potentially itself)
3. If B does not provide a reply or traps or runs out of gas, etc. then the protocol generates a response on its behalf.

Are you proposing option 2 above?

birchmd: I think Marcelo is proposing option 1 and there is a contract-level protocol for these kinds of off-chain computation. Consider the following sketch implementation:

```rust
struct FunctionCall {
    contract: AccountId,
    method_name: String,
    gas: u64,
    deposit: u128,
}

struct Request<T> {
    input: T,
    callback: FunctionCall,
}

enum Response<T> {
    // `id` computed as the hash of the `Request` concatenated with any relevant environment data (e.g. timestamp).
    // We will assume that requests with identical `id` values will also have identical results when the computation completes.
    Pending { id: CryptoHash },
    Complete { value: T },
    // An error case to allow the contract to reject computations it doesn't want to do.
    InvalidRequest { message: String },
}

fn off_chain_computation(&mut self, request: Request) -> Response {
    let id = hash_with_environment(&request);
    // Caching could be optional
    if let Some(value) = self.cache.get(id) {
        return Response::Complete { value };
    }
    // Verify that env::deposit() >= request.callback.deposit  + storage_cost(request),
    // along with any other business logic that might determine an invalid request.
    if let Err(message) = self.validate_request(&request) {
        return Response::InvalidRequest { message};
    }
    self.requests.insert(id, request);
    // Maybe also emit an event here to make indexer's lives easier
    Response::Pending
}

fn complete_computation(&mut self, id: CryptoHash, value: T) {
    // TODO: access control (what accounts can submit a response?)

    // Optional caching. Could make this some kind of LRU cache to bound storage usage.
    self.cache.insert(id, value);

    if let Some(request) = self.requests.remove(id) {
        // Maybe also emit an event here to make indexer's lives easier
        let cb = request.callback;
        // TODO: could consider refunding the storage cost of the request also since it has been removed.
        env::promise_return(cb.contract.call(cb.method_name).args(id, value).gas(cb.gas).deposit(cb.deposit));
    }
}
```

With the above scheme any contract calling `off_chain_computation` would need to be able to properly handle the `Response::Pending` case. Usually this will mean saving the `id` into its state so that when the callback is invoked later the contract will be able to resume whatever computation it wanted to do.

This scheme does not have an explicit timeout, it would be up to the caller to enforce that. For example it could save a timestamp with the request id and in the callback function check if there was a timeout. In the case the off-chain computation never completes then it would also be up to the caller to trigger their callback method themselves to do the on-chain timeout handling logic.

An important note with this scheme is that whatever service submits the result of the off-chain computation also pays for the gas needed to execute the callback. A service may limit the amount it is willing to pay by adding a request validation rule limiting the amount of gas attached to the callback. If the service severely limits the amount of gas it is willing to spend then it could be that the callback could only persist the response and the caller would need to submit another transaction that does further computation.

In my opinion, it is that "who pays" complexity which makes a scheme like this not a universal solution to this problem and does make a protocol change worthwhile.

akhi3030: To make sure I understand this correctly, I think the flow is as following:

- Contract B provides some service that requires offchain computation.
- Contract A calls B to use the service
- B updates its state and then replies to A that the computation is in progress
- Some time later, the offchain computation finishes and B is called by the offchain service
- B sends another message to A providing the results of the offchain computation.

Is that correct?

birchmd: Option 2 you provided @akhi3030 is potentially interesting. The contract would call itself in a loop waiting to either run out of gas or see the response in its storage and send it back to the caller. The issue I see with that kind of scheme is how expensive it would be in terms of gas usage. I think it takes around 2 Tgas to make a cross-contract call (even a self-call), so with a 300 Tgas limit there would only be 150 blocks of wait time. And of course you could not dedicate the whole transaction to waiting for the response if you wanted to do any computation before or after, so maybe in practice you could only wait for at most 50 blocks (100 Tgas) to get a response and I am not sure if that is long enough.

birchmd: > To make sure I understand this correctly, I think the flow is as following:
> 
> * Contract B provides some service that requires offchain computation.
> * Contract A calls B to use the service
> * B updates its state and then replies to A that the computation is in progress
> * Some time later, the offchain computation finishes and B is called by the offchain service
> * B sends another message to A providing the results of the offchain computation.
> 
> Is that correct?

Yes, that is correct

akhi3030: I think @DavidM-D 's team is currently using option 2.  They view it as a hacky workaround till a more proper solution is available.

Option 1 above may not be generally possible as it does not follow the usual request - response procedure.  As contract A (in the above example) is not always under our control.  

The way I see this, we in protocol need requirements for what kind of usecase needs to be supported.  If the current protocol is good enough, that is great and nothing to be done 😃

DavidM-D: Yup, we're not really happy with the call loop approach, it's tremendously brittle and will fail if you do simple things, like call for multiple signatures.

To some extent you're correct that we can pick any arbitrary API so we can use submit and poll systems. The problem is that every contract that calls this endpoint will also have to have application level submit and poll. It also means that we need to design a version of the indexer that is call stack aware. This would be a tremendous amount of work and would lead to function colouring for anything that has a signature anywhere in it's call tree.

bowenwang1996:
> I am not sure I completely understand this point. Especially about not having a way to clean up the state. If things expire, the protocol will call the contract. So is the contract may not handle that case properly? Isn't that similar to how the current Promise API works?

I misread the proposal and therefore thought incorrectly that the execution would just fail at that point.

> If contract A called yield_create, only contract A can call resume on that promise. No other contract can.

Yes, I guess the concern here is that a contract developer can shoot themselves in the foot if they are not careful with how they valid the computation result that is passed to `yield_resume` because while only the contract itself can resume the computation. Anyone can call the method (`signature_available` in this case) that provides the computation result. If there is no proper validation of the result, then a malicious attacker can easily submit something that causes the original transaction to fail. I don't think this is a concern for MPC signature per se since that contract will be carefully written and audited, but still worth noting for the general use case.

akhi3030:
> Yes, I guess the concern here is that a contract developer can shoot themselves in the foot if they are not careful with how they valid the computation result that is passed to `yield_resume` because while only the contract itself can resume the computation. Anyone can call the method (`signature_available` in this case) that provides the computation result. If there is no proper validation of the result, then a malicious attacker can easily submit something that causes the original transaction to fail. I don't think this is a concern for MPC signature per se since that contract will be carefully written and audited, but still worth noting for the general use case.

Yes, this is always possible.  Luckily, we have some nice tools (https://github.com/aurora-is-near/near-plugins/blob/master/near-plugins/src/access_controllable.rs) available that could be used.

neps/nep-519-yield-execution.md
mfornet: > Yup, we're not really happy with the call loop approach, it's tremendously brittle and will fail if you do simple things, like call for multiple signatures.

Isn't the idea elaborated by Michael (https://github.com/near/NEPs/pull/519#discussion_r1444876559) a better and more robust approach than a loop? 
- Is it a concern who pays for the gas in the second transaction (maybe it could be solved with delegated transactions).
- Or is it a concern about the engineering complexity of this approach? I found it feasible but complex.

Btw what Michel described is how Aurora works internally, so there is some experience in the ecosystem with those patterns, and we know they work.

> we can pick any arbitrary API so we can use submit and poll systems. The problem is... 

You described a lot of complexity on the application layer without this feature, but notice that the alternative that is being proposed here is to move this complexity into the protocol, which IMO is a bad idea. The protocol should be as simple as possible (it is not simple by any means today, but we should try to avoid further complexity if possible).

This change is quite invasive, and it is hard to foresee unintended consequences. 
- This requires updating the SDK
- Documentation to ensure developers know how to use this feature properly and don't shoot themselves in the foot.
- Increase the bar to hire auditors, given they need to be aware of this quite specific feature in NEAR-Protocol
- Hard to maintain at the protocol level
- Any future change to the protocol, including gas/balance accounting, must make sure yield/resume keeps working (backward compatibility)
- This feature is quite broad in what it enables, but it is being designed with one application in mind.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,192 @@
...
+/// use this function to indicate that it may now be ready to reply to it.  When
+/// this is called, then the protocol will call the method that the smart
+/// contract referred to in the earlier `promise_yield_create()` call.
+///
+/// `data_id_len` and `data_it_ptr`: This value was returned in an earlier call
+/// to `promise_yield_create` and uniquely identifies which yielded execution
+/// should be resumed.
+///
+/// `payload_len` and `payload_ptr`: the smart contract can provide an optional
+/// list of arguments that should be passed to the method that will be resumed.
```
saketh-are, line 92: nit: This is not really a list of arguments from the yield/resume perspective, just an arbitrary byte sequence. The rust sdk makes it easy to jsonify a tuple of objects and parse them back out, but that is not relevant at the level of the host function.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,192 @@
...
+DiscussionsTo: https://github.com/near/NEPs/pull/519
+Type: Protocol
+Version: 0.0.0
+Created: 2023-11-17
+LastUpdated: 2023-11-20
+---
+
+## Summary
+
+Today, when a smart contract is called by a user or another contract, it has no sensible way to delay responding to the caller.  There exist use cases where contracts would benefit from being able to delay responding till some arbitrary time in the future.  This proposal introduces such possibility into the NEAR protocol.
```
saketh-are, line 15: > till some arbitrary time

This feels potentially misleading to me given that we enforce a timeout. Could it be clearer to say something like "await a future transaction" rather than "delay responding"?

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Motivation
+
+There exist some situations where when a smart contract on NEAR is called, it will only be able to provide an answer at some arbitrary time in the future.  So the callee needs a way to defer replying to the caller till this time in future.
+
+Examples include when a smart contract (`S`) provides MPC signing capabilities parties external to the NEAR protocol are computing the signature.  The rough steps are:
+1. Signer contract provides a function `fn sign_payload(Payload, ...)`.
+2. When called, the contract updates some contract state which is being monitored by external indexers to indicate that a new signing request has been received.  It also defers replying to the caller.
```
saketh-are, line 23: In practice I think the signers will directly index the transactions on the account and observe the receipts generated by yield (no need for any storage in contract state). But perhaps it is easier to explain things with this example.

neps/nep-519-yield-execution.md
akhi3030: Hopefully improved now.

neps/nep-519-yield-execution.md
akhi3030: Hopefully improved now.  I'll add the example in a subsequent commit.

neps/nep-519-yield-execution.md
akhi3030: Based on our last discussion, I believe it is fine to resolve this now.

neps/nep-519-yield-execution.md
akhi3030: Hopefully improved now.

neps/nep-519-yield-execution.md
akhi3030: Believe this is resolved now.

akhi3030:
@near/nep-moderators: I believe this PR should be ready for review.

bowenwang1996:
As a working group member, I'd like to nominate @birchmd and @nagisa as subject matter experts to review the NEP.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+/// this is called, then the protocol will call the method that the smart
+/// contract referred to in the earlier `promise_yield_create()` call.
+///
+/// `data_id_len` and `data_it_ptr`: This value was returned in an earlier call
+/// to `promise_yield_create` and uniquely identifies which yielded execution
+/// should be resumed.
+///
+/// `payload_len` and `payload_ptr`: the smart contract can provide an
+/// additional optional blob of arguments that should be passed to the method
+/// that will be resumed. These will be appended to the list that was provided
```
birchmd, line 94: This looks inaccurate. In the reference implementation, the data is not appended to the input provided by `promise_yield_create` (which will be available from the `input` host function). Instead the data passed by `promise_yield_resume` is made available via the `promise_result` host function. It is important for the documentation to be technically accurate for contract developers that are trying to use this feature.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+/// `promise_yield_resume()` within X (TBD) blocks.  When
+/// `promise_yield_resume()` is called, the protocol will call the method on the
+/// smart contract that is identified by `method_name_len` and `method_name_ptr`
+/// and this method may or may not respond to the caller.
+///
+/// `arguments_len` and `arguments_ptr` provide an initial blob of arguments
+/// that will be passed to the method.
+///
+/// If the contract fails to call `promise_yield_resume()` within X blocks, then
+/// the protocol will call the method with a timeout error.
```
birchmd, line 52: In the reference implementation the method is called without any promise results (i.e. the `promise_result_count` host function will return 0). It will be up to the contract to interpret this missing data as a timeout error (probably there will be convenience macros/functions in `near-sdk` for this). Again I think it is important to be clear in the documentation about how developers will interact with this feature.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Specification
+
+The proposal is to add the following host functions to the NEAR protocol:
+
+
+```rust
+/// Instructs the protocol that the smart contract is not yet ready to respond
+/// to its caller yet.  The smart contract promises to call
+/// `promise_yield_resume()` within X (TBD) blocks.  When
```
birchmd, line 43: The reference implementation has X = 200 blocks.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+
+There are some problems with this approach though.  After the change of flow of calls; `B` is now going to be paying for gas for various executions that `A` should have been paying for.  Due to bugs or malicious intent, `B` could forget to call `A` with the signature.  If `A` is calling `B` deep in a call tree and `B` replies to it without actually providing an answer, then `A` would need a mechanism to keep the call tree alive while it waits for `B` to call it with the signature in effect running into the same problem that this NEP is attempting to solve.
+
+## Future possibilities
+
+[Describe any natural extensions and evolutions to the NEP proposal, and how they would impact the project. Use this section as a tool to help fully consider all possible interactions with the project in your proposal. This is also a good place to "dump ideas"; if they are out of scope for the NEP but otherwise related. Note that having something written down in the future-possibilities section is not a reason to accept the current or a future NEP. Such notes should be in the section on motivation or rationale in this or subsequent NEPs. The section merely provides additional information.]
```
birchmd, line 203: Please complete or remove the remaining sections as needed. We cannot merge this with the template text.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+Examples include when a smart contract (`S`) provides MPC signing capabilities parties external to the NEAR protocol are computing the signature.  The rough steps are:
+
+1. Signer contract provides a function `fn sign_payload(Payload, ...)`.
+2. When called, the contract defers replying to the caller.
+3. External indexers are monitoring the transactions on the contract; they observe the new signing request, compute a signature, and call another function `fn signature_available(Signature, ...)` on the signer contract.
+4. The signer contract validates the signature and if validate, replies to the original caller.
+
+Today, the NEAR protocol has no sensible way to defer replying to the caller in step 2 above.  This proposal proposes adding two following new host functions to the NEAR protocol:
+
+- `yield`: this can be called by a contract to indicate to the protocol that it is not ready yet to reply to its caller.
```
birchmd, line 30:
```suggestion
- `promise_yield_create`: this can be called by a contract to indicate to the protocol that it is not ready yet to reply to its caller.
```

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+- `resume`: a contract can use this mechanism to indicate to a protocol that it is now ready to reply to a caller that it had deferred earlier.
```
birchmd, line 31:
```suggestion
- `promise_yield_resume`: a contract can use this mechanism to indicate to a protocol that it is now ready to reply to a caller that it had deferred earlier.
```

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+
+If these two host functions were available, then `yield` would be used in step 2 above and `resume` would be used in step 4 above.
```
birchmd, line 33:
```suggestion
If these two host functions were available, then `promise_yield_create` would be used in step 2 above and `promise_yield_resume` would be used in step 4 above.
```

birchmd (changes requested):
As a SME reviewer, I am excited about this proposal and the new use cases it will enable for Near smart contracts. I like that the proposed implementation is minimally invasive in the sense that it mostly uses the existing promises mechanism without change (the only new addition is the ability to directly create data receipts via the new `promise_yield_resume` host function). This is an elegant solution in my opinion.

My comments on the NEP are only about ensuring the documentation is clear so that contract developers are able to use this feature as soon as it is available. It might also be nice to include an example contract for this reason.

neps/nep-519-yield-execution.md
saketh-are: In the reference implementation in the case of a timeout `promise_result_count()` will still be 1  and `promise_result(0)` will return [PromiseResult::Failed](https://github.com/near/near-sdk-rs/blob/9bb16c2d08d522baf52ddd126e3fe0f01386f2ca/near-sdk/src/types/vm_types.rs#L20). I would consider adjusting the documentation to reflect that there is no specific "timeout error," just a generic promise failure.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+## Motivation
+
+There exist some situations where when a smart contract on NEAR is called, it will only be able to provide an answer at some arbitrary time in the future.  So the callee needs a way to defer replying to the caller till this time in future.
+
+Examples include when a smart contract (`S`) provides MPC signing capabilities parties external to the NEAR protocol are computing the signature.  The rough steps are:
+
+1. Signer contract provides a function `fn sign_payload(Payload, ...)`.
+2. When called, the contract defers replying to the caller.
+3. External indexers are monitoring the transactions on the contract; they observe the new signing request, compute a signature, and call another function `fn signature_available(Signature, ...)` on the signer contract.
+4. The signer contract validates the signature and if validate, replies to the original caller.
```
nagisa, line 26:
```suggestion
4. The signer contract validates the signature and if valid, replies to the original caller.
```

or perhaps even

```suggestion
4. The signer contract validates the signature and replies to the original caller.
```

neps/nep-519-yield-execution.md
nagisa, line 31: `promise_yield_resume` by itself does not imply a response to the caller, and now that I think neither does `promise_yield_create` require that the promise is passed along to `promise_return` – if these two are not combined in some way the intended behaviour of delaying a reply is not achieved.

So perhaps:

```suggestion
- `promise_yield_create`: allows setting up a continuation function that should only be executed after `promise_yield_resume` is invoked. Together with `promise_return` this allows delaying the reply to the caller;
- `promise_yield_resume`: indicates to the protocol that the continuation to the yield may now be executed.
```

neps/nep-519-yield-execution.md
nagisa, line 33:
```suggestion
If these two host functions were available, then `yield` would be used to implement step 2 and `resume` would be used for step 3 of the motivating example above.
```

Nit: It has been a little while since the motivating example has been presented, so better refer more specifically to what steps we're referring to.

(Also, the yields are resumed by the indexers etc. to allow the contract to continue execution; that's step 3.)

neps/nep-519-yield-execution.md
birchmd: Thanks for the correction @saketh-are !

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+If these two host functions were available, then `yield` would be used in step 2 above and `resume` would be used in step 4 above.
+
+## Specification
+
+The proposal is to add the following host functions to the NEAR protocol:
+
+
+```rust
+/// Instructs the protocol that the smart contract is not yet ready to respond
+/// to its caller yet.  The smart contract promises to call
```
nagisa, line 42: Same note as earlier, `promise_yield_create` by itself does not imply this behaviour. It only occurs in combination with `promise_return`. Here I think the `promise_yield_create` should be documented about its own behaviour and only that, since it is a specification of these functions.

There is also no real punishment to anybody if they don't call a corresponding `promise_yield_resume`, so I would rephrase the 2nd sentence in terms of the number of blocks the protocol is willing to keep track of the receipt before timing it out and invoking the resumption function with no data.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+///
+/// Similar to the `gas` parameter in
+/// [promise_create](https://github.com/near/nearcore/blob/a908de36ab6f75eb130447a5788007e26d05f93e/runtime/near-vm-runner/src/logic/logic.rs#L1281),
+/// the `gas` parameter is a prepayment for the gas that would be used to
+/// execute the method.
+///
+/// `gas_weight`: as specified in
+/// [this](https://github.com/near/NEPs/blob/master/neps/nep-0264.md) NEP, this
+/// improves the devX of specifying a portion of the remaining gas for executing
+/// the method instead of specifying a precise amount.
```
nagisa, line 62: We don't need to reiterate the motivation for that NEP. Better refer to `promise_create` the same way like it was done for `gas`.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+///
+/// `register_id`: is used to identify the register that will be used by the
+/// protocol to return unique token referring to this yielded execution to the
+/// contract.  The contract will have to pass this value when it calls
+/// `promise_yield_resume`.
```
nagisa, line 67: The protocol isn't really involved in handling of registers. Maybe:

```suggestion
/// `register_id`: is used to identify the register that will be filled 
/// with a unique resumption token. This token is used with
/// `promise_yield_resume` to resolve the continuation receipt set up by this
/// function.
```

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+    method_name_len: u64,
+    method_name_ptr: u64,
+    arguments_len: u64,
+    arguments_ptr: u64,
+    gas: u64,
+    gas_weight: u64,
+    register_id: u64,
+) -> u64;
+
+/// When a smart contract has postponed replying to its caller earlier, it can
```
nagisa, line 83: Similar comment about "reply postponement" – these functions don't inherently do that, so don't make it part of their specification description.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+/// use this function to indicate that it may now be ready to reply to it.  When
+/// this is called, then the protocol will call the method that the smart
+/// contract referred to in the earlier `promise_yield_create()` call.
+///
+/// `data_id_len` and `data_it_ptr`: This value was returned in an earlier call
+/// to `promise_yield_create` and uniquely identifies which yielded execution
+/// should be resumed.
```
nagisa, line 90: It should be noted that the `promise_yield_resume` may only resume delayed receipts that were created by the same contract.

(This can also be implicit by defining somewhere that these functions operate in sender-is-receiver mode only.)

nagisa (commented):
Mostly wording and text nits.

One thing I have a hard time wrapping my head around it the necessity for the "Reference Implementation" section. The pseudocode provided there seemed to me to be both borderline useless by the lack of detail necessary to reproduce the functional and protocol-affecting behaviour (e.g. the resumption token can only resume yields produced by the same contract) while at the same time promising details that we probably should refrain from promising (such as that the resumption token exactly matches the data input IDs.)

If it were up to me, I would link the reference implementation PR only, but if we need to keep it, I would like to comb through it with a fine comb to make sure it covers all the facets and aspects of the specification (even if it is pseudo-code.)

neps/nep-519-yield-execution.md
akhi3030: I think all the unnecessary sections should be removed now.

akhi3030:
> One thing I have a hard time wrapping my head around it the necessity for the "Reference Implementation" section

I agree, I also found it a bit difficult to figure out how much details to present here.  Inspired by [this NEP](https://github.com/near/NEPs/blob/27e975ae12b56c8879c24d430153a08fb7588a9e/neps/nep-0514.md) I think it is fine to just provide a link to the PR.  I will make the change to this PR now.  And let the SMEs decide if they are happy with the change or not.

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+    let num_bytes = method_name.len() + arguments.len();
+    self.gas_counter.pay_per(yield_create_bytes, num_bytes);
+
+    // Create a receipt with a single data dependency which will be resolved by
+    // either by the resume call or the timeout.
+    let data_id = self.generate_data_id();
+    let receipt = ReceiptMetadata {
+        output_data_receivers: vec![],
+        input_data_ids: vec![data_id],
+        actions: vec![],
```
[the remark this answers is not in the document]
akhi3030: This section is removed now.

neps/nep-519-yield-execution.md
akhi3030: Should be fixed now.

neps/nep-519-yield-execution.md
akhi3030: @saketh-are: is your recommendation then to change the sentence to something like the following:

"If the contract fails to call `promise_yield_resume()` within X blocks, then the protocol will call the method with a promise failure."

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Summary
+
+Today, when a smart contract is called by a user or another contract, it has no sensible way to delay responding to the caller till it has observed another future transaction.  This proposal introduces this possibility into the NEAR protocol.
+
+## Motivation
+
+There exist some situations where when a smart contract on NEAR is called, it will only be able to provide an answer at some arbitrary time in the future.  So the callee needs a way to defer replying to the caller till this time in future.
+
+Examples include when a smart contract (`S`) provides MPC signing capabilities parties external to the NEAR protocol are computing the signature.  The rough steps are:
```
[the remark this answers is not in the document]
akhi3030: should be improved now.

neps/nep-519-yield-execution.md
akhi3030: Much better wording.  Looks like I cannot commit suggestion so adding manually as a commit.

neps/nep-519-yield-execution.md
akhi3030: Added.

neps/nep-519-yield-execution.md
akhi3030: Done.

akhi3030:
@birchmd, @nagisa: thank you very much for the review.  I believe I have addressed all the feedback that you provided.  There is an ongoing discussion on the edge case around when a timeout has happened but the callback has not executed and the contract calls `resume`.  Once we resolve the discussion, I will update the spec.  But this should be a minor change.  So I believe that this should be ready for you to look once again.

neps/nep-519-yield-execution.md
saketh-are: Really we guarantee the following:
- If `promise_yield_resume` is called before the yield expires, the callback will be executed with `PromiseResult::Successful` containing the payload passed to yield_resume. 
- If resume is not called by the time the yield expires, the callback will be executed with a `PromiseResult::Failed`.
- The yield will not expire within X blocks, but it is at the discretion of the protocol exactly when it will expire after X blocks have passed.

victorchimakanu:

SME reviews:

* Protocol SME @nagisa: [Support for yielded execution #519]((https://github.com/near/NEPs/pull/519#pullrequestreview-1904175947))
* Protocol SME @birchmd  [Support for yielded execution #519]((https://github.com/near/NEPs/pull/519#pullrequestreview-1903751019))

* ❔ @bowenwang1996 : [Link to voting indication]
* ❔ @birchmd : [Link to voting indication]
* ❔ @bowenwang1996 : [Link to voting indication]

birchmd (approved):
Thank for updating the NEP @akhi3030 ! It looks good to me now.

As both an SME and working group member I approve this NEP and look forward to seeing the feature go live on testnet in the future.

victorchimakanu:
As the moderator, I would like to thank the author @akhi3030  for submitting this NEP extension, and the work group members for your review. Based on the voting indications above, it seems like this proposal is close to a decision. We are therefore scheduling the 8th Protocol Work group meeting next week.

Anyone can discuss the technical details by adding your comments to this discussion. And join the call to learn about the final decision and how to get more involved.

### Meeting Info:
📅 Date: Friday, March 8th, 4 PM UTC
✏️ Register [here](https://bit.ly/protocol-work-group-call-8)

neps/nep-519-yield-execution.md
```diff
@@ -0,0 +1,183 @@
...
+/// `payload_len` and `payload_ptr`: the smart contract can provide an
+/// additional optional blob of arguments that should be passed to the callback
+/// that will be resumed.  These are available via the `promise_result` host
+/// function.
+pub fn promise_yield_resume(
+    data_id_len: u64,
+    data_id_ptr: u64,
+    payload_len: u64,
+    payload_ptr: u64,
+) -> ();
```
nagisa, line 106: One more small change that has happened to the implementation since the last review:

```suggestion
/// See `promise_yield_create()` for more details.  This host function can be
/// used to resolve the continuation that was set up by
/// `promise_yield_create()`.  The contract calling this function must be the
/// same contract that called `promise_yield_create()` earlier.  This host
/// function cannot be called for the same resumption token twice or if the
/// callback specified in `promise_yield_create()` has already executed.
///
/// `data_id_len` and `data_it_ptr`: Used to pass the unique resumption token
/// that was returned to the smart contract in the `promise_yield_create()`
/// function.
///
/// `payload_len` and `payload_ptr`: the smart contract can provide an
/// additional optional blob of arguments that should be passed to the callback
/// that will be resumed.  These are available via the `promise_result` host
/// function.
///
/// The function returns `1` if submitting the payload was successful. This 
/// guarantees that the yielded callback will be executes with one of the
/// successfully submitted payloads as the input. Otherwise (e.g. if the yield 
/// receipt has already timed out) `0` will be returned, indicating that this
/// payload could not be submitted successfully.
pub fn promise_yield_resume(
    data_id_len: u64,
    data_id_ptr: u64,
    payload_len: u64,
    payload_ptr: u64,
) -> u32;
```

> This host function cannot be called for the same resumption token twice or if the callback specified in `promise_yield_create()` has already executed.

We may also want to relax this requirement. I.e. allow returning `true` more than exactly once. In that scenario, the callback is allowed to execute with any of the payloads that have been supplied and for which `true` was returned.

cc @saketh-are for double-check.

saketh-are: +1, if we can relax this it will make things simpler for us. The guarantees we can more easily provide are:
- `promise_yield_resume` will return true iff the callback has not yet been executed.
- If the user gets back true from `promise_yield_resume`, the callback will execute with a user-provided payload rather than a timeout signal.
- If the user calls resume multiple times and gets back multiple trues, the callback will be executed with one of the given payloads and the others will be safely ignored.

Hopefully we can come up with a simple and clear way to condense all of this into the docs.

bowenwang1996:
As a working group member, I lean towards accepting this proposal. It is a great step towards enabling smart contracts to interact with offchain computation while still preserving the call stack.

mfornet:
As a working group member, I lean toward approving this NEP.

This NEP solves a problem that can't be solved today, which is having simultaneously:
- Delayed responses
- Guaranteed execution of callback
- Abstraction from the caller about the delayed response.

Delayed responses are particularly useful to seamlessly use off-chain computation on-chain. This could involve for example oracle queries (eg `fetch_tweet(tweet_id)`) or privacy protocol execution such as signing some data offchain.

mm-near:
As a working group member, I lean toward approving this NEP.

* one small comment: in comments you say:
```
/// `data_id_len` and `data_it_ptr`: Used to pass the unique resumption token
/// that was returned to the smart contract in the `promise_yield_create()`
/// function.
```
But `promise_yield_create` seems to be returning u64

akhi3030:
>     * one small comment: in comments you say:
> 
> 
> ```
> /// `data_id_len` and `data_it_ptr`: Used to pass the unique resumption token
> /// that was returned to the smart contract in the `promise_yield_create()`
> /// function.
> ```
> 
> But `promise_yield_create` seems to be returning u64

Yeah, the comment is referring to the resumption token returned in the register.  I'll update the comment to address this.

neps/nep-519-yield-execution.md
```diff
@@ -92,18 +92,20 @@ pub fn promise_yield_create(
...
 /// additional optional blob of arguments that should be passed to the callback
 /// that will be resumed.  These are available via the `promise_result` host
 /// function.
 ///
-/// The function returns `1` if submitting the payload was successful. This 
-/// guarantees that the yielded callback will be executes with one of the
-/// successfully submitted payloads as the input. Otherwise (e.g. if the yield 
-/// receipt has already timed out) `0` will be returned, indicating that this
-/// payload could not be submitted successfully.
+/// This function can be called multiple times.  If it is called successfully
```
saketh-are, line 102: nit: "can be called multiple times with the same data id."

akhi3030: fixed.

walnut-the-cat:
can we merge this into master?

walnut-the-cat (approved):

PR #527, opened 2024-01-11
Revise the summary and motivation, add high level flow.

robin-near:


neps/nep-0509.md
```diff
@@ -12,58 +12,107 @@ LastUpdated: 2023-09-19
...
-* Then, validators will not store state locally and be randomly assigned to a shard at a given block height. Once a validator receives a chunk, along with its state witnesses, it verifies the state transition of the chunk, signs an approval, and sends it to the next block producer, similar to how it works today.
+* Under stateless validation, the responsibility of a chunk producer extends to packaging transactions and receipts and annotating them with state witnesses. This extended role will be called "chunk proposers".
+* The state witness of a chunk is defined to be a subset of the trie state, alongside its proof of inclusion in the trie, that is needed to execute a chunk. A state witness allows anyone to execute the chunk without having the state of its shard locally. 
+* Then, at each block height, validators will be randomly assigned to a shard, to validate the state witness for that shard. Once a validator receives a chunk along with its state witness, it verifies the state transition of the chunk, signs a chunk endorsement and sends it to the block producer. This is similar to, but separate from, block approvals and consensus.
```
Longarithm, line 20:
```suggestion
* Then, at each block height, validators will be randomly assigned to a shard, to validate the state witness for that shard. Once a validator receives both a chunk and its state witness, it verifies the state transition of the chunk, signs a chunk endorsement and sends it to the block producer. This is similar to, but separate from, block approvals and consensus.
```
Technically this is the case currently

neps/nep-0509.md
```diff
@@ -12,58 +12,107 @@ LastUpdated: 2023-09-19
...
+* The block producer waits for sufficient chunk endorsements before including a chunk into the block it produces.
```
Longarithm, line 21:
```suggestion
* The block producer waits for sufficient chunk endorsements before including a chunk into the block it produces and doesn't include a chunk if it is not endorsed enough.
```
nit

neps/nep-0509.md
```diff
@@ -12,58 +12,107 @@ LastUpdated: 2023-09-19
...
-  * Validator assignment for both chunk validation and block validation should not create any security vulnerability.
-* Validators should be rewarded with the same amount as they are now.
 * No validator needs to track all shards.
-* Only chunk producers need to maintain state locally.
-* State witness size should be small enough to delivered over network.
-* Majority of undercharging issues should be mitigated.
-* Validator roles are updated accordingly to reflect the necessary changes with stateless validation.
-* [TBD] Resharding should work as expected after stateless validation in place.
+* Security of protocol must not degrade.
+  * Validator assignment for both chunk validation and block validation should not create any security vulnerabilities.
```
Longarithm, line 42: I think the old point "Moreover, majority of undercharging issues should be mitigated." deserves a point here. Because it is not tracking all shards what allows in-memory trie to be used

Longarithm (approved):

neps/nep-0509.md
walnut-the-cat: > at each block height, validators will be randomly assigned to a shard

Is this true? I understand this is possible, but is this what we want to do? Wonder if it's better to leave the option open (e.g. mentioning possibility of assigning validators at every N height)

neps/nep-0509.md
walnut-the-cat: Should we mention what 'sufficient' really means? 2/3?

neps/nep-0509.md
walnut-the-cat: Any reason we are removing the following statement?

> Only chunk producers need to maintain state locally.

neps/nep-0509.md
```diff
@@ -74,7 +123,7 @@ TBD. [Explain the proposal as if you were teaching it to another developer. This
 
 The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.]
 
-## Validatior role change
+## Validatior Role Change
```
staffik, line 126: Validator

walnut-the-cat (approved):

walnut-the-cat:
@robin-near, let's try to merge this in by Wednesday.

neps/nep-0509.md
robin-near: These are mostly technicalities :)

I removed the undercharging point because it's not a "high level requirement". It's more of a good-to-have side effect. It's not a requirement for stateless validation to solve "a majority of undercharging issues" (which isn't even well defined).

For "only chunk producers need to maintain state locally", I also don't think this is a requirement. Even if everybody is required to track some state (and it's not all the state because there is already a point saying no validators needs to track all shards), stateless validation is still successful. If we're talking about the desire that chunk validators who don't also act as chunk producers do not need to track any shard at all, I think the motivation there is that they don't need to have beefy machines to track any state, but that is also covered by the point "Any additional load on network and compute should not negatively affect existing functionalities of *any* node in the blockchain", and that the cost of these should be acceptable. Also, "only chunk producers need to maintain state locally" seems to be a particular solution rather than a requirement, as we're also changing the definition of chunk producers.

neps/nep-0509.md
```diff
@@ -12,58 +12,107 @@ LastUpdated: 2023-09-19
...
+We propose a change to the following parts of the chunk and block production flow:
+
+* When a chunk producer produces a chunk, in addition to collecting transactions and receipts for the chunk, it will also produce a `ChunkStateWitness`.
+  * The `ChunkStateWitness` contains whatever data necessary to prove that this chunk's header should indeed be what is being produced:
+    * As it is today, all fields of the `ShardChunkHeaderInnerV2`, except `tx_root`, are uniquely determined by the blockchain's history based on where the chunk is located (i.e. its parent block and shard ID).
+    * The `tx_root` is based on the list of transactions proposed, which is at the discretion of the chunk producer. However, these transactions must be valid (i.e. the sender accounts have enough balance and the correct nonce, etc.).
```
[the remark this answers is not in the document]
robin-near: The next point after this says that the tx_root corresponds to a valid set of transactions, so I suppose implicitly that means the validation of ChunkStateWitness would validate every claim that it makes (including the tx_root corresponding to a valid set of transactions).

neps/nep-0509.md
```diff
@@ -12,58 +12,107 @@ LastUpdated: 2023-09-19
...
+* A chunk validator, upon receiving a `ChunkStateWitness`, validates the state witness and determines if the chunk header is indeed correctly produced. If so, it sends a `ChunkEndorsement` to the current block producer.
+  * A `ChunkEndorsement` contains the chunk hash along with a signature proving the endorsement by the chunk validator. It implicitly carries a weight equal to the amount of the chunk validator's stake that is assigned to this shard for this block. (See Validator Shuffling).
+* As the existing logic is today, the block producer for this block waits until either all chunks are ready, or a timeout occurs, and then proposes a block containing whatever chunks are ready. Now, the notion of readiness here is expanded to also having more than 2/3 of chunk endorsements by weight.
+  * This means that if a chunk does not receive enough chunk endorsements by the timeout, it will not be included in the block. In other words, the block only contains chunks for which there is already a consensus of validity. This is the key reason why we will no longer need challenges.
```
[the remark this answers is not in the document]
robin-near: Sure.

neps/nep-0509.md
```diff
@@ -12,58 +12,107 @@ LastUpdated: 2023-09-19
...
+  * The 2/3 fraction has the denominator being the total stake assigned to validate this shard, *not* the total stake of all validators.
```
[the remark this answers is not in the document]
robin-near: It's necessary, I was going to cover that in the validator selection algorithm section but that isn't complete. I added a reference to that section here.

neps/nep-0509.md
robin-near: The NEP is titled Stateless Validation Stage 0, and we don't currently have a plan to do every N (not clear there's a real benefit yet), so I think we may as well just be concrete about what exactly we're doing?

neps/nep-0509.md
robin-near: Yeah that's in the specification section. I wanted to keep the summary section high-level.

walnut-the-cat (approved):

PR #528, opened 2024-01-25
nep-393: add class metadata interface

robert-zaremba:
In https://github.com/near/NEPs/pull/393 we defined Issuer (an entity authorized to mint SBTs in the registry) and SBT Class. We also defined Issuer Metadata and Token Metadata, but we didn't provide interface for class metadata. This was implemented in the reference implementation (in one of the subsequent revisions), but was not backported to the NEP. In this PR

* we fix the name of the issuer interface from `SBTContract` to `SBTIssuer`. The original name is wrong and we oversight it in reviews. We talk everywhere about the issuer entity and issuer contract (even the header is _SBT Issuer interface_).
* Renames `ContractMetadata` to `IssuerMetadata`.
* added `ClassMetadata` struct and `sbt_class_metadata` function to the `SBTIssuer`.

### DIFF
--- a/neps/nep-0393.md
+++ b/neps/nep-0393.md
@@ -294,14 +294,14 @@ pub struct IssuerMetadata {
 /// ClassMetadata defines SBT class wide attributes, which are shared and default to all SBTs of
 /// the given class. Must be provided by the Issuer contract. See the `SBTIssuer` trait.
 pub struct ClassMetadata {
-    /// Issuer class name. Required.
+    /// Issuer class name. Required to be not empty.
     pub name: String,
-    /// If defined, should be used instead of `contract_metadata.symbol`.
+    /// If defined, should be used instead of `IssuerMetadata::symbol`.
     pub symbol: Option<String>,
     /// An URL to an Icon. To protect fellow developers from unintentionally triggering any
     /// SSRF vulnerabilities with URL parsers, we don't allow to set an image bytes here.
     /// If it doesn't start with a scheme (eg: https://)
-    /// then `contract_metadata.base_uri` should be prepended.
+    /// then `IssuerMetadata::base_uri` should be prepended.
     pub icon: Option<String>,
     /// JSON or an URL to a JSON file with more info. If it doesn't start with a scheme
     /// (eg: https://) then base_uri should be prepended.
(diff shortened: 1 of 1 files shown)


frol:
Thank you @robert-zaremba for submitting this NEP.

As a moderator, I reviewed this NEP and it meets [the proposed template guidelines](https://github.com/near/NEPs/blob/master/nep-0000-template.md). I am moving this NEP to [the REVIEW stage](https://github.com/near/NEPs/blob/master/neps/nep-0001.md#nep-lifecycle) and would like to ask the @near/wg-contract-standards working group members to assign 2 Technical Reviewers to complete a technical review (see expectations below).

<details>

Technical Summary guidelines:

```

### Recommendation

Add recommendation

### Benefits

* Benefit

* Benefit

### Concerns

| - | - | - | - |

```

</details>

fadeevab (changes requested):
### Recommendation

Fix the broken references in the documentation, typos, and wording.

### Benefits

Overall, the patch is not complex, everything looks fine.

### Concerns

| - | - | - | - |
| 1 | Broken references after renaming the struct | Fixes suggested | Done |
| 2 | Wording and typos | Fixes suggested | Done |

fadeevab (approved):
All suggestions are implemented, table updated: https://github.com/near/NEPs/pull/528#pullrequestreview-1848562574

KazanderDad:
As a SME I've reviewed this NEP and find that it fills a useful purpose. It clarifies some of the potential points of confusion for anyone wishing to implement SBTs, including some that I ran into myself when deploying a new issuer recently. 

I might suggest rethinking the hashing requirement of Base64 encoded SHA256. While we can recognize that this - if done right - is a little more compact with 44 characters; it is not a commonly available method and the risk for implementing wrong is high. Casual users who try to find online SHA256 hashing tools will almost exclusively find tools that instead output hexadecimal characters, resulting in a 64-character string. A very likely risk is that they will attempt to Base64-encode this string (instead of the binary), resulting in a much longer output than intended, and resulting in a lot of confusion. A more usable requirement might be "a SHA256 hash expressed as a hexadecimal string"

frol (commented):
As a Contract Standards Work Group member, I reviewed and lean to approving this NEP. That said, I believe we reached the quorum here to approve it, and since it is just an extension to the existing NEP, I suggest we go ahead and merge it.

@robert-zaremba Please, update the CHANGELOG section with 1.1.0 version summarizing the change and the decision.

frol (approved):

victorchimakanu:

**Status: Voting** 

SME reviews:

* Contract Standards SME @KazanderDad: https://github.com/near/NEPs/pull/528#issuecomment-1916048845
* Contract Standards SME @fadeevab: https://github.com/near/NEPs/pull/528#pullrequestreview-1848562574

Contract Standards Work Group voting indications (❔ | 👍 | 👎 ):

* 👍 @frol https://github.com/near/NEPs/pull/528#pullrequestreview-1865638134
* ? @robert-zaremba
* 👍 @fadeevab https://github.com/near/NEPs/pull/528#pullrequestreview-1848562574

PR #533, opened 2024-02-19
NEP-330 extension: Build details extension

Canvinus:
### Overview
This update introduces build details to the contract metadata, containing necessary information about how the contract was built. This makes it possible for others to reproduce the same WASM of this contract. The idea first appeared in the [cargo-near SourceScan integration thread](https://github.com/near/cargo-near/issues/131).

### Benefits

This NEP extension gives developers the capability to save all the required build details, making it possible to reproduce the same WASM code in the future. This ensures greater consistency in contracts and the ability to verify source code. With the assistance of tools like SourceScan and cargo-near, the development process on NEAR becomes significantly easier.

### DIFF
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -78,6 +78,7 @@ When using a Docker image as a reference, it's important to specify the digest o
 ### Paths Inside Docker Image
 
 During the build, paths from the source of the build as well as the location of the cargo registry could be saved into WASM, which affects reproducibility. Therefore, we need to ensure that everyone uses the same paths inside the Docker image. We propose using the following paths:
+
 - `/home/near/code` - Mounting volume from the host system containing the source code.
 - `/home/near/.cargo` - Cargo registry.
 
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -58,8 +58,8 @@ type Standard {
 type BuildInfo {
     build_environment_ref: string, // reference to a reproducible build environment docker image, e.g. "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment.
     source_code_snapshot: string, // reference to the source code snapshot that was used to build the contract, e.g. "git+https://github.com/near-DevHub/neardevhub-contract.git#335e89edec95d56a4744e7160c3fd590e41ec38e" or "ipfs://<ipfs-hash>"
-	  contract_path: string|null, // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
-	  build_command: string[], // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
+    contract_path: string|null, // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
+    build_command: string[], // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
 }
 ```
 
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -47,7 +47,7 @@ type ContractSourceMetadata = {
   version: string|null, // optional, commit hash being used for the currently deployed wasm. If the contract is not open-sourced, this could also be a numbering system for internal organization / tracking such as "1.0.0" and "2.1.0".
   link: string|null, // optional, link to open source code such as a Github repository or a CID to somewhere on IPFS.
   standards: Standard[]|null, // optional, standards and extensions implemented in the currently deployed wasm e.g. [{standard: "nep330", version: "1.1.0"},{standard: "nep141", version: "1.0.0"}].
-  build_details: BuildDetails|null, // optional, details that are required for contract WASM reproducibility.
+  build_info: BuildInfo|null, // optional, details that are required for contract WASM reproducibility.
 }
 
 type Standard {
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -31,22 +31,35 @@ There is a lot of information that can be held about a contract. Ultimately, we 
 
 Successful implementations of this standard will introduce a new  (`ContractSourceMetadata`) struct that will hold all the necessary information to be queried for. This struct will be kept on the contract level.
 
-The metadata will include three optional fields:
+The metadata will include optional fields:
 
-- `version`: a string that references the specific commit hash or version of the code that is currently deployed on-chain. This can be included regardless of whether or not the contract is open-sourced and can also be used for organizational purposes.
+- `version`: a string that references the specific commit hash or version of the code currently deployed on-chain. If the link is to a GitHub repo URL, this is crucial, as it is needed to make the contract reproducible.
 - `link`: a string that references the link to the open-source code. This can be anything such as Github or a CID to somewhere on IPFS.
 - `standards`: a list of objects (see type definition below) that enumerates the NEPs supported by the contract. If this extension is supported, it is advised to also include NEP-330 version 1.1.0 in the list (`{standard: "nep330", version: "1.1.0"}`).
+- `build_details`: a build details object (see type definition below) that contains all the necessary information about how the contract was built, making it possible for others to reproduce the same WASM of this contract.
 
 ```ts
 type ContractSourceMetadata = {
   version: string|null, // optional, commit hash being used for the currently deployed wasm. If the contract is not open-sourced, this could also be a numbering system for internal organization / tracking such as "1.0.0" and "2.1.0".
   link: string|null, // optional, link to open source code such as a Github repository or a CID to somewhere on IPFS.
   standards: Standard[]|null, // optional, standards and extensions implemented in the currently deployed wasm e.g. [{standard: "nep330", version: "1.1.0"},{standard: "nep141", version: "1.0.0"}].
+  build_details: BuildDetails|null, // optional, details that are required for contract WASM reproducibility.
 }
 
 type Standard {
     standard: string, // standard name e.g. "nep141"
     version: string, // semantic version number of the Standard e.g. "1.0.0"
+}
+
+type BuildDetails {
+	env_image: EnvImage, // reference to a reproducible build environment image, e.g. "{tag: "sourcescan/cargo-near:0.6.0", digest: "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"}"
+	contract_path: string|null // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
+	build_command: string[] // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
+}
+
+type EnvImage {
+	tag: string // tag of the image e.g. "sourcescan/cargo-near:0.6.0"
+	digest: string // a sha256 hash unique, immutable identifier for a container image, used to verify that the image under the same tag has not changed e.g. "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"
 }
 ```
 
@@ -58,30 +71,39 @@ function contract_source_metadata(): ContractSourceMetadata {}
 
 ## Reference Implementation
 
-As an example, say there was an NFT contract deployed on-chain which was currently using the commit hash `39f2d2646f2f60e18ab53337501370dc02a5661c` and had its open source code located at `https://github.com/near-examples/nft-tutorial`. This contract would then include a struct which has the following fields:
+As an example, consider an NFT contract located at the relative path `nft-contract`, which was deployed using the `cargo near deploy --no-abi` environment image `sourcescan/cargo-near:0.6.0`. Its latest commit hash is `39f2d2646f2f60e18ab53337501370dc02a5661c`, and its open-source code can be found at `https://github.com/near-examples/nft-tutorial`. This contract would then include a struct with the following fields:
 
 ```ts
 type ContractSourceMetadata = {
-  version: "39f2d2646f2f60e18ab53337501370dc02a5661c"
+  version: "39f2d2646f2f60e18ab53337501370dc02a5661c",
   link: "https://github.com/near-examples/nft-tutorial",
   standards: [
     {
-        standard: "nep330", 
-        version: "1.1.0"
+      standard: "nep330",
+      version: "1.1.0"
     },
     {
-        standard: "nep171", 
-        version: "1.0.0"
+      standard: "nep171",
+      version: "1.0.0"
     },
     {
-        standard: "nep177", 
-        version: "2.0.0"
+      standard: "nep177",
+      version: "2.0.0"
     }
-  ]
+  ],
+  build_details: {
+    env_image: {
+      tag: "sourcescan/cargo-near:0.6.0",
+      digest: "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"
+    },
+    contract_path: "nft-contract",
+    build_command: ["cargo", "near", "deploy", "--no-abi"]
+  }
 }
+
 ```
 
-If someone were to call the view function `contract_metadata`, the contract would return:
+Calling the view function `contract_metadata`, the contract would return:
 
 ```bash
 {
@@ -100,9 +122,19 @@ If someone were to call the view function `contract_metadata`, the contract woul
             standard: "nep177", 
             version: "2.0.0"
         }
-    ]
+    ],
+    build_details: {
+	    env_image: {
+	      tag: "sourcescan/cargo-near:0.6.0",
+	      digest: "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"
+	    },
+	    contract_path: "nft-contract",
+	    build_command: ["cargo", "near", "deploy", "--no-abi"]
+	}
 }
 ```
+
+This could be used by SourceScan to reproduce the same WASM using the build details and to verify the on-chain WASM code with the reproduced one.
 
 An example implementation can be seen below.
 
@@ -113,17 +145,31 @@ pub struct Contract {
     pub contract_metadata: ContractSourceMetadata
 }
 
-// Standard structure
+/// Standard structure
 pub struct Standard {
     pub standard: String,
     pub version: String
 }
 
+/// EnvImage structure
+pub struct EnvImage { 
+	pub tag: String, 
+	pub digest: String, 
+}
+
+/// BuildDetails structure
+pub struct BuildDetails { 
+	pub env_image: EnvImage, 
+	pub contract_path: Option<String>, 
+	pub build_command: Vec<String>, 
+}
+
 /// Contract metadata structure
 pub struct ContractSourceMetadata {
-    pub version: String,
-    pub link: String,
-    pub standards: Vec<Standard>
+    pub version: Option<String>,
+    pub link: Option<String>,
+    pub standards: Option<Vec<Standard>>,
+    pub build_details: Option<BuildDetails>,
 }
 
 /// Minimum Viable Interface
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -58,7 +58,7 @@ type Standard {
 type BuildInfo {
     build_environment: string, // reference to a reproducible build environment docker image, e.g. "docker.io/sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment. When using a Docker image as a reference, it's important to specify the digest of the image to ensure reproducibility, since a tag could be reassigned to a different image.
     source_code_snapshot: string, // reference to the source code snapshot that was used to build the contract, e.g. "git+https://github.com/near-DevHub/neardevhub-contract.git#335e89edec95d56a4744e7160c3fd590e41ec38e" or "ipfs://<ipfs-hash>"
-    contract_path: string|null, // relative path to contract crate within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
+    contract_path: string|null, // relative path to contract crate within the source code e.g. "contracts/contract-one". Often, it is the root of the repository, so can be omitted. 
     build_command: string[], // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
 }
 ```
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -56,8 +56,8 @@ type Standard {
 }
 
 type BuildInfo {
-    build_environment: string, // reference to a reproducible build environment docker image, e.g., "docker.io/sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment. When using a Docker image as a reference, it's important to specify the digest of the image to ensure reproducibility, since a tag could be reassigned to a different image.
-    source_code_snapshot: string, // reference to the source code snapshot that was used to build the contract, e.g., "git+https://github.com/near/cargo-near-new-project-template.git#9c16aaff3c0fe5bda4d8ffb418c4bb2b535eb420" or "ipfs://<ipfs-hash>". It is important to have Cargo.lock inside the source code snapshot to ensure reproducibility.
+    build_environment: string, // reference to a reproducible build environment docker image, e.g., "docker.io/sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment.
+    source_code_snapshot: string, // reference to the source code snapshot that was used to build the contract, e.g., "git+https://github.com/near/cargo-near-new-project-template.git#9c16aaff3c0fe5bda4d8ffb418c4bb2b535eb420" or "ipfs://<ipfs-hash>".
     contract_path: string|null, // relative path to contract crate within the source code, e.g., "contracts/contract-one". Often, it is the root of the repository, so can be omitted. 
     build_command: string[], // the exact command that was used to build the contract, with all the flags, e.g., ["cargo", "near", "build", "--no-abi"].
 }
@@ -68,6 +68,16 @@ In order to view this information, contracts must include a getter which will re
 ```ts
 function contract_source_metadata(): ContractSourceMetadata {}
 ```
+
+### Ensuring WASM Reproducibility
+
+#### Build Environment Docker Image
+
+When using a Docker image as a reference, it's important to specify the digest of the image to ensure reproducibility, since a tag could be reassigned to a different image.
+
+#### Cargo.lock
+
+It is important to have `Cargo.lock` inside the source code snapshot to ensure reproducibility. Example: https://github.com/near/core-contracts.
 
 ## Reference Implementation
 
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -55,7 +55,7 @@ type Standard {
     version: string, // semantic version number of the Standard e.g. "1.0.0"
 }
 
-type BuildDetails {
+type BuildInfo {
 	env_image: EnvImage, // reference to a reproducible build environment image, e.g. "{tag: "sourcescan/cargo-near:0.6.0", digest: "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"}"
 	contract_path: string|null // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
 	build_command: string[] // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -1,14 +1,14 @@
 ---
 NEP: 330
 Title: Source Metadata
-Author: Ben Kurrek <ben.kurrek@near.org>, Osman Abdelnasir <osman@near.org>
+Author: Ben Kurrek <ben.kurrek@near.org>, Osman Abdelnasir <osman@near.org>, Andrey Gruzdev <@canvi>
 DiscussionsTo: https://github.com/near/NEPs/discussions/329
 Status: Approved
 Type: Standards Track
 Category: Contract
-Version: 1.1.0
+Version: 1.2.0
 Created: 27-Feb-2022
-Updated: 13-Jan-2023
+Updated: 19-Feb-2023
 ---
 
 ## Summary
@@ -214,6 +214,13 @@ The extension NEP-351 that added Contract Metadata to this NEP-330 was approved 
 | 2 | NEP-330 and NEP-351 should be included in the list of the supported NEPs | There seems to be a general agreement that it is a good default, so NEP was updated | Resolved |
 | 3 | JSON Event could be beneficial, so tooling can react to the changes in the supported standards | It is outside the scope of this NEP. Also, list of supported standards only changes with contract re-deployment, so tooling can track DEPLOY_CODE events and check the list of supported standards when new code is deployed | Won’t fix |
 
+### 1.2.0 - Build Details Extension
+
+The NEP extension adds build details to the contract metadata, containing necessary information about how the contract was built. This makes it possible for others to reproduce the same WASM of this contract. The idea first appeared in the [cargo-near SourceScan integration thread](https://github.com/near/cargo-near/issues/131).
+
+#### Benefits
+
+- This NEP extension gives developers the capability to save all the required build details, making it possible to reproduce the same WASM code in the future. This ensures greater consistency in contracts and the ability to verify source code. With the assistance of tools like SourceScan and cargo-near, the development process on NEAR becomes significantly easier
 
 ## Copyright
 
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -107,7 +107,7 @@ type ContractSourceMetadata = {
 
 ```
 
-Calling the view function `contract_metadata`, the contract would return:
+Calling the view function `contract_source_metadata`, the contract would return:
 
 ```bash
 {
--- a/neps/nep-0330.md
+++ b/neps/nep-0330.md
@@ -56,7 +56,7 @@ type Standard {
 }
 
 type BuildInfo {
-    build_environment: string, // reference to a reproducible build environment docker image, e.g. "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment.
+    build_environment: string, // reference to a reproducible build environment docker image, e.g. "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment. When using a Docker image as a reference, it's important to specify the digest of the image to ensure reproducibility, since a tag could be reassigned to a different image.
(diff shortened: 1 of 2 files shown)


frol:
Thank you @Canvinus for submitting this NEP.

As a moderator, I reviewed this NEP and it meets [the proposed template guidelines](https://github.com/near/NEPs/blob/master/nep-0000-template.md). I am moving this NEP to [the REVIEW stage](https://github.com/near/NEPs/blob/master/neps/nep-0001.md#nep-lifecycle) and would like to ask the @near/wg-contract-standards working group members to assign 2 Technical Reviewers to complete a technical review (see expectations below).

<details>

Technical Summary guidelines:

```

### Recommendation

Add recommendation

### Benefits

* Benefit

* Benefit

### Concerns

| - | - | - | - |

```

</details>

lachlanglen:
This looks useful. Any thoughts on adding `commit_hash` while you're at it to remove ambiguity of `version` field?

> `version`: a string that references the specific commit hash or version of the code currently deployed on-chain.

Having a field that might be one thing **or** something else is not super clear IMO. I personally have added `commit_hash` in my own implementation of this standard for my contracts, as I find it useful to have both a version (e.g. `v1.0.0`) **and** a commit hash.

Canvinus:
I totally agree with that point, @lachlanglen. I've just tried to retain as much as possible from previous contributors and ensure compatibility with already deployed contracts that have the metadata.

frol (commented):
I was originally suggesting this NEP extension, so I, obviously support it. I would love to hear if there are other opinions from other WG members cc @fadeevab @robert-zaremba

neps/nep-0330.md
```diff
@@ -31,23 +35,36 @@ There is a lot of information that can be held about a contract. Ultimately, we
 
 Successful implementations of this standard will introduce a new  (`ContractSourceMetadata`) struct that will hold all the necessary information to be queried for. This struct will be kept on the contract level.
 
-The metadata will include three optional fields:
+The metadata will include optional fields:
 
-- `version`: a string that references the specific commit hash or version of the code that is currently deployed on-chain. This can be included regardless of whether or not the contract is open-sourced and can also be used for organizational purposes.
+- `version`: a string that references the specific commit hash or version of the code currently deployed on-chain. If the link is to a GitHub repo URL, this is crucial, as it is needed to make the contract reproducible.
 - `link`: a string that references the link to the open-source code. This can be anything such as Github or a CID to somewhere on IPFS.
```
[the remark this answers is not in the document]
frol: @lachlanglen Would this approach of a "frozen" link work for you? I believe it should be a good replacement for a `commit_hash` field which assumes that the code is stored on a public git provider.

neps/nep-0330.md
```diff
@@ -31,23 +35,36 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
+
+type BuildDetails {
+	env_image: EnvImage, // reference to a reproducible build environment image, e.g. "{tag: "sourcescan/cargo-near:0.6.0", digest: "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"}"
+	contract_path: string|null // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
+	build_command: string[] // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
+}
+
+type EnvImage {
+	tag: string // tag of the image e.g. "sourcescan/cargo-near:0.6.0"
+	digest: string // a sha256 hash unique, immutable identifier for a container image, used to verify that the image under the same tag has not changed e.g. "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"
```
fadeevab, line 66: What's the function of the `digest` field? Is it just a long unique number, or is it supposed to be a hash of something specific?

frol:
> One more: ☝ `build_command` cannot guarantee obtaining exactly the same image

@fadeevab That is why `build_details` is a combination of 4: "snapshot" reference to the source code, work dir where the build has happened withing the code snapshot, build command, and the docker image reference (name [tag] + unique identifier [digest]). You may want to review [this comment](https://github.com/near/cargo-near/issues/131#issuecomment-1937707851).

@Canvinus I am not completely sold on your naming, but I would love to hear other opinions since I am obviously biased. I tried to keep the naming generic enough to allow some other solutions but Docker for env, and other solutions besides GitHub or even git (ipfs, tarball on s3, etc).

neps/nep-0330.md
fadeevab: @frol So, the `digest` is a hash of a _Docker_ container image on the hub.docker.com, correct?

The word "docker" should be mentioned at least once in the NEP, because it's a bit confusing.

fadeevab:
@frol 

> > One more: ☝ `build_command` cannot guarantee obtaining exactly the same image
> 
> @fadeevab ... and the docker image reference (name [tag] + unique identifier [digest]). You may want to review [this comment](https://github.com/near/cargo-near/issues/131#issuecomment-1937707851).

I didn't realize that:
1. The "image" means a "docker image", not a "contract binary" in a raw form.
2. The `tag` is a hub.docker.com image tag, not a `git` tag.
3. The `digest` is the docker image digest.

(And right, `tag`/`digest` is just a Docker Hub way to encode an image).

Suggestions:
- `EnvImage` could be renamed to `DockerImage`.
- `BuildDetails` should be clarified better in the NEP.
- I still think that `contract_path` is unnecessary (although it's optional of course).
- A developer should be advised to commit `Cargo.lock`.

☝I see that the NEP is biased toward defining build environments via public docker images.
🤔However, I still can imagine not using `BuildDetails` and having just a `link` to a CI/CD public release to fit this NEP to my needs.
🙂Overall, I faced some confusion after jumping here from my bubble.

> > `version` could be mandatory, e.g. version is mandatory for public crates or NPM packages.
> If someone does not want to expose the version, it is up to them.
> That would be a breaking change for NEP-330

[Here](https://github.com/near/NEPs/#neps) I saw that the NEP is on the review and my thoughts were that the NEP is not already adopted. Need to update the README's table.

> It is really designed to be flexible. It is not a requirement to have this metadata to deploy your contract on the chain, right?

There is a case when the whole `ContractSourceMetadata` is empty.
I was also thinking about making a `version` mandatory to make at least 1 field mandatory. I would also require developers to define 

----

The bottom line is that the NEP can be improved with more clarification and examples.

----

P.S.: On a side note, I once had in my experience that a compiler attached a timestamp into a binary, so the digest never was the same.  I'm not sure it's true nowadays for `cargo`/`rustc`. Nevertheless, I'm biased toward the GitHub Workflow+Release model which has proven to be simple and effective.

neps/nep-0330.md
Canvinus: > The word "docker" should be mentioned at least once in the NEP, because it's a bit confusing.

Good point. I will clarify this to prevent confusion. Additionally, it's worth noting that the digest might differ across various registries. Therefore, the structure should also be modified as follows:
```ts
type DockerImage {
  registry: string // Docker image registry e.g. "ghcr.io", "hub.docker.com"
  tag: string // tag of the Docker image e.g. "sourcescan/cargo-near:0.6.0"
  digest: string // a sha256 hash unique, immutable identifier for a Docker image, used to verify that the image under the same tag has not changed e.g. "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"
}
```

neps/nep-0330.md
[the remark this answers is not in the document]
Canvinus: > Thus, I would move this build-related source code link to the `build_details` and don't touch the top-level `link` or `version` at all.

I agree. I appreciate the approach of using a "frozen" link, as it appears to resolve all the issues related to the "link" and "version" fields. The reason I didn't add it initially was to avoid confusion for anyone reviewing the metadata. Therefore, the name of the new field should be distinctly different from "link". In my opinion, "source_snapshot" fits well.

Canvinus:
> @Canvinus I am not completely sold on your naming, but I would love to hear other opinions since I am obviously biased. I tried to keep the naming generic enough to allow some other solutions but Docker for env, and other solutions besides GitHub or even git (ipfs, tarball on s3, etc).

I like this way of naming images because it's more flexible. But, I think we should also check if a Docker image is the right one by using its digest. So, we could suggest using the image's "digest" instead of its "tag" by default. For example, `[REPOSITORY]@[DIGEST]` or `ghcr.io/[OWNER]/[IMAGE]@[DIGEST]`. This method is simpler because we just need one string, and it could work for other approaches, not just Docker.

```rust
  /// Reference to a reproducible build environment, e.g. Docker image reference:
  /// "docker.io/sourcescan/cargo-near:0.6.0"
  build_environment_ref: Option<String>,
```

Canvinus:
@fadeevab, would having a "frozen" link as the source for the CI/CD release, along with a field like "build_environment_ref" as a string that could point to the commit of the CI/CD that published the release, meet your requirements?

fadeevab:
@Canvinus I think I could use the `link` field and put the commit hash in the `version` field, so no additional fields is required. Also, commit can be found/recorded in the public release builds,  e.g. https://github.com/DecentralBankDAO/usn/releases/tag/v2.3.4

robert-zaremba:
> EnvImage could be renamed to DockerImage.

Let's use `BuildImage`. There can be other container images than hub.docker.com

robert-zaremba:
> I think I could use the link field and put the commit hash in the version field, so no additional fields is required. 

@fadeevab I think it's good to keep both `version` and `field`. If a code is uploaded without a version control (which should be a very rare case), then version would just be human friendly semantic number (eg `v1.0.1`) and must be the same as package / crate version.

neps/nep-0330.md
```diff
@@ -31,23 +35,36 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
     version: string, // semantic version number of the Standard e.g. "1.0.0"
 }
+
+type BuildDetails {
+	env_image: EnvImage, // reference to a reproducible build environment image, e.g. "{tag: "sourcescan/cargo-near:0.6.0", digest: "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"}"
+	contract_path: string|null // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
+	build_command: string[] // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
+}
+
+type EnvImage {
```
robert-zaremba, line 64:
```suggestion
type Container {
```

neps/nep-0330.md
```diff
@@ -31,23 +35,36 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
   standards: Standard[]|null, // optional, standards and extensions implemented in the currently deployed wasm e.g. [{standard: "nep330", version: "1.1.0"},{standard: "nep141", version: "1.0.0"}].
+  build_details: BuildDetails|null, // optional, details that are required for contract WASM reproducibility.
 }
 
 type Standard {
     standard: string, // standard name e.g. "nep141"
     version: string, // semantic version number of the Standard e.g. "1.0.0"
 }
+
+type BuildDetails {
```
robert-zaremba, line 58:
```suggestion
type BuildInfo {
```

neps/nep-0330.md
```diff
@@ -31,23 +35,36 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
 - `standards`: a list of objects (see type definition below) that enumerates the NEPs supported by the contract. If this extension is supported, it is advised to also include NEP-330 version 1.1.0 in the list (`{standard: "nep330", version: "1.1.0"}`).
+- `build_details`: a build details object (see type definition below) that contains all the necessary information about how the contract was built, making it possible for others to reproduce the same WASM of this contract.
 
 ```ts
 type ContractSourceMetadata = {
   version: string|null, // optional, commit hash being used for the currently deployed wasm. If the contract is not open-sourced, this could also be a numbering system for internal organization / tracking such as "1.0.0" and "2.1.0".
   link: string|null, // optional, link to open source code such as a Github repository or a CID to somewhere on IPFS.
   standards: Standard[]|null, // optional, standards and extensions implemented in the currently deployed wasm e.g. [{standard: "nep330", version: "1.1.0"},{standard: "nep141", version: "1.0.0"}].
+  build_details: BuildDetails|null, // optional, details that are required for contract WASM reproducibility.
```
robert-zaremba, line 50:
```suggestion
  build_info: BuildInfo|null, // optional, details that are required for contract WASM reproducibility.
```

robert-zaremba: I think _details_ sounds more like the output details from the build, not _how to_ build.

neps/nep-0330.md
```diff
@@ -31,23 +35,36 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
 }
 
 type Standard {
     standard: string, // standard name e.g. "nep141"
     version: string, // semantic version number of the Standard e.g. "1.0.0"
 }
+
+type BuildDetails {
+	env_image: EnvImage, // reference to a reproducible build environment image, e.g. "{tag: "sourcescan/cargo-near:0.6.0", digest: "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"}"
+	contract_path: string|null // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted.
```
robert-zaremba, line 60:
```suggestion
	contract_path: string|null // relative path to contract crate within the source code e.g. "contracts/contract1". Often, it is the root of the repository, so can be omitted. 
```

neps/nep-0330.md
```diff
@@ -31,23 +35,36 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
 
 type Standard {
     standard: string, // standard name e.g. "nep141"
     version: string, // semantic version number of the Standard e.g. "1.0.0"
 }
+
+type BuildDetails {
+	env_image: EnvImage, // reference to a reproducible build environment image, e.g. "{tag: "sourcescan/cargo-near:0.6.0", digest: "sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471"}"
+	contract_path: string|null // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
+	build_command: string[] // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
```
robert-zaremba, line 61: instead of list of strings, should we just use a string, with proper flags?

robert-zaremba (commented):
### Recommendation

This pull request is an improvement to the NEP-330 v1.1.0 and I lean to approve it once the small concerns listed below are resolved.

### Concerns

- let's be more specific about optional and required fields. For example, if using a version control system (git...)  then `version` must be required.
- be more specific about version and link - I suggest to add more examples (made few suggestions).
- solve naming. Suggestions: env_image -> container, build_details -> build_info

neps/nep-0330.md
fadeevab: @robert-zaremba A container is a running instance of the image, so I believe it's still an image.

neps/nep-0330.md
frol: @robert-zaremba The reasoning behind the list of strings is that you can just pass those directly to `exec`. Having a string there would require shell parsing and may introduce unnecessary side-effects, e.g. if you would need to escape whitespace in the command, etc.

neps/nep-0330.md
frol: Thanks for providing your reasoning. I can see your point, I am fine with "info".

neps/nep-0330.md
frol: Originally, it was named `build_environment_ref`, and it was expected to be either `docker.io/frol/near-builder@hash-of-the-image` or something less conventional (torrent link, IPFS cid, etc)

neps/nep-0330.md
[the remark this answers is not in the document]
Canvinus: Fixed in f47d5ede90e0e542b99ac9f81b257789f21d7b4c

neps/nep-0330.md
Canvinus:
```ts
type BuildInfo {
    build_environment_ref: string, // reference to a reproducible build environment image, e.g. "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment.
```

Now using `build_environment_ref` in 18968ea09c965845b92ab7fdb39a8a1df319b7cd

neps/nep-0330.md
```diff
@@ -58,30 +75,39 @@ function contract_source_metadata(): ContractSourceMetadata {}
...
 ## Reference Implementation
 
-As an example, say there was an NFT contract deployed on-chain which was currently using the commit hash `39f2d2646f2f60e18ab53337501370dc02a5661c` and had its open source code located at `https://github.com/near-examples/nft-tutorial`. This contract would then include a struct which has the following fields:
+As an example, consider an NFT contract located at the relative path `nft-contract`, which was deployed using the `cargo near deploy --no-abi` environment image `sourcescan/cargo-near:0.6.0`. Its latest commit hash is `39f2d2646f2f60e18ab53337501370dc02a5661c`, and its open-source code can be found at `https://github.com/near-examples/nft-tutorial`. This contract would then include a struct with the following fields:
 
 ```ts
 type ContractSourceMetadata = {
-  version: "39f2d2646f2f60e18ab53337501370dc02a5661c"
+  version: "39f2d2646f2f60e18ab53337501370dc02a5661c",
   link: "https://github.com/near-examples/nft-tutorial",
```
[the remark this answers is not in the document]
Canvinus: Updated here 04c54a5040d4f73e76b15087022c652d82823693

neps/nep-0330.md
Canvinus: 35bbcded50a0419a5a24090bd2b0e27b8bcafed5

neps/nep-0330.md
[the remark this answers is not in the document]
Canvinus:
```ts
type BuildInfo {
    build_environment_ref: string, // reference to a reproducible build environment docker image, e.g. "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment.
    source_code_snapshot: string, // reference to the source code snapshot that was used to build the contract, e.g. "git+https://github.com/near-DevHub/neardevhub-contract.git#335e89edec95d56a4744e7160c3fd590e41ec38e" or "ipfs://<ipfs-hash>"
    contract_path: string|null, // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
    build_command: string[], // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
}
```

Now using `source_code_snapshot` in 18968ea09c965845b92ab7fdb39a8a1df319b7cd

neps/nep-0330.md
robert-zaremba: honestly, `Env` in `EnvImage` doesn't sound right. `build_environment_ref` is even more cryptic. Why not `BuilderImage` . We can also say `DockerImage` if we don't want to use _Container_

Canvinus: I like the name 'builder_image', but I think the original idea was to make it more abstract. This way, the reference could be something other than just a Docker image, for example, a link to a CD pipeline that creates the release.

neps/nep-0330.md
robert-zaremba: ok, fair enough.

neps/nep-0330.md
```diff
@@ -31,23 +35,32 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
 type Standard {
     standard: string, // standard name e.g. "nep141"
     version: string, // semantic version number of the Standard e.g. "1.0.0"
 }
+
+type BuildInfo {
+    build_environment_ref: string, // reference to a reproducible build environment docker image, e.g. "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment.
+    source_code_snapshot: string, // reference to the source code snapshot that was used to build the contract, e.g. "git+https://github.com/near-DevHub/neardevhub-contract.git#335e89edec95d56a4744e7160c3fd590e41ec38e" or "ipfs://<ipfs-hash>"
+	  contract_path: string|null, // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
+	  build_command: string[], // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
```
fadeevab, line 62: Fix the indentation.
```suggestion
    contract_path: string|null, // relative path to contract folder within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted. 
    build_command: string[], // the exact command that was used to build the contract, with all the flags e.g. ["cargo", "near", "build", "no-abi"].
```

neps/nep-0330.md
```diff
@@ -100,10 +120,18 @@ If someone were to call the view function `contract_metadata`, the contract woul
...
             version: "2.0.0"
         }
-    ]
+    ],
+    build_info: {
+	      build_environment_ref: "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471",
+        source_code_snapshot: "git+https://github.com/near-examples/nft-tutorial.git#39f2d2646f2f60e18ab53337501370dc02a5661c",
+	      contract_path: "nft-contract",
+	      build_command: ["cargo", "near", "deploy", "--no-abi"]
+	  }
```
fadeevab, line 129: Fix the indentation 
```suggestion
    build_info: {
        build_environment_ref: "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471",
        source_code_snapshot: "git+https://github.com/near-examples/nft-tutorial.git#39f2d2646f2f60e18ab53337501370dc02a5661c",
        contract_path: "nft-contract",
        build_command: ["cargo", "near", "deploy", "--no-abi"]
    }
```

fadeevab: I like flatting the structure! 👌
The `_ref` suffix is a bit out of the overall naming convention, could it be like a full word `_reference`, or maybe drop it, e.g. `build_environment`? Or `build_environment_link` (cause we already use the `link` word in a different place)?

robert-zaremba: let's drop it.

Canvinus: dropped `_ref` in aa84739feadf6b04bbfd7fdd06102b6d189367c8

Canvinus:
@frol, @fadeevab, @lachlanglen, @robert-zaremba, thank you everyone for your insightful reviews. It looks like I have applied all the suggestions that were presented here. The only thing left to do is to refine the naming so that it satisfies everyone and aligns with their vision.

neps/nep-0330.md
Canvinus: Changed here 66db5e8f2de6c32eaa15d05c01d811b2f4ee0456.
About the path example, I am referencing to the template here: https://github.com/near/cargo-near-new-project-template.

neps/nep-0330.md
```diff
@@ -31,23 +35,32 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
+    build_info: BuildInfo|null, // optional, details that are required for contract WASM reproducibility.
 }
 
 type Standard {
     standard: string, // standard name e.g. "nep141"
     version: string, // semantic version number of the Standard e.g. "1.0.0"
 }
+
+type BuildInfo {
+    build_environment: string, // reference to a reproducible build environment docker image, e.g. "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment. When using a Docker image as a reference, it's important to specify the digest of the image to ensure reproducibility, since a tag could be reassigned to a different image.
```
frol, line 59: I suggest we recommend to use `docker.io/` prefix for Docker Hub images, so when anyone introduces torrent link or something, it is easier to tell them apart.

```suggestion
    build_environment: string, // reference to a reproducible build environment docker image, e.g. "docker.io/sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471" or something else pointing to the build environment. When using a Docker image as a reference, it's important to specify the digest of the image to ensure reproducibility, since a tag could be reassigned to a different image.
```

neps/nep-0330.md
```diff
@@ -31,23 +35,32 @@ There is a lot of information that can be held about a contract. Ultimately, we
...
+    source_code_snapshot: string, // reference to the source code snapshot that was used to build the contract, e.g. "git+https://github.com/near-DevHub/neardevhub-contract.git#335e89edec95d56a4744e7160c3fd590e41ec38e" or "ipfs://<ipfs-hash>"
+    contract_path: string|null, // relative path to contract crate within the source code e.g. "src/contract". Often, it is the root of the repository, so can be omitted.
```
frol, line 61: `src/` might be misleading and readers may think they should somehow reference a file within `src/` folder (standard in Rust projects), so I suggest to use more neutral example:

```suggestion
    contract_path: string|null, // relative path to contract crate within the source code e.g. "contracts/contract-one". Often, it is the root of the repository, so can be omitted. 
```

frol:
### Recommendation

As a subject matter expert, I recommend @near/wg-contract-standards to approve this NEP.

### Benefits

* This is a non-breaking extension to NEP-330
* This extension brings a lot of potential for permissionless reproducible builds (I can image anyone would be able to just do `cargo near verify devhub.near` (just an idea) which will fetch the metadata, and build the contract locally based on the `build_info` and verify that the deployed contract Wasm file matches the source code - this will improve discoverability and confidence
* SourceScan and Explorers will be able to provide more details to the users and bring more visibility and potential collaboration in the space of cross-contract interactions

### Concerns

| - | - | - | - |
| 1 | Are these details enough to have a reproducible build | Consensus is that the combination of build environment snapshot, source code snapshot, build path, and build command is enough | Resolved |
| 2 | Various naming considerations | The agreement is reached: `{ build_info: { build_environment, source_code_snapshot, contract_path, build_command } }` | Resolved |

neps/nep-0330.md
```diff
@@ -58,35 +71,42 @@ function contract_source_metadata(): ContractSourceMetadata {}
...
+            standard: "nep171",
+            version: "1.0.0"
+        },
+        {
+            standard: "nep177",
+            version: "2.0.0"
+        }
+    ],
+    build_info: {
+        build_environment: "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471",
```
fadeevab, line 95: @Canvinus  Also, you should add `docker.io` here:
```suggestion
        build_environment: "docker.io/sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471",
```

neps/nep-0330.md
```diff
@@ -100,10 +120,18 @@ If someone were to call the view function `contract_metadata`, the contract woul
             standard: "nep177", 
             version: "2.0.0"
         }
-    ]
+    ],
+    build_info: {
+        build_environment: "sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471",
```
fadeevab, line 125: @Canvinus  And here:
```suggestion
        build_environment: "docker.io/sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471",
```

fadeevab (changes requested):

neps/nep-0330.md
Canvinus: Yes, thanks

neps/nep-0330.md
```diff
@@ -58,35 +71,42 @@ function contract_source_metadata(): ContractSourceMetadata {}
 
 ## Reference Implementation
 
-As an example, say there was an NFT contract deployed on-chain which was currently using the commit hash `39f2d2646f2f60e18ab53337501370dc02a5661c` and had its open source code located at `https://github.com/near-examples/nft-tutorial`. This contract would then include a struct which has the following fields:
+As an example, consider an NFT contract located at the relative path `nft-contract`, which was deployed using the `cargo near deploy --no-abi` environment docker image `sourcescan/cargo-near@sha256:bf488476d9c4e49e36862bbdef2c595f88d34a295fd551cc65dc291553849471`. Its latest commit hash is `39f2d2646f2f60e18ab53337501370dc02a5661c`, and its open-source code can be found at `https://github.com/near-examples/nft-tutorial`. This contract would then include a struct with the following fields:
```
fadeevab, line 74: @Canvinus @frol @robert-zaremba  Could we suggest having a `Cargo.lock` in the contract folder?

I think this way: if the contract was hacked because of a security issue in a 3rd-party dependency, then such an issue may not be reproducible after the dependency upgrade (in the least "patch" version triplet).

Also, I checked the `nft-tutorial/nft-contract`, it has `[patch.crates-io]` which leads to a custom crate, and it may be hard to reproduce the contract with such a dependency.

fadeevab:
### Recommendation

@near/wg-contract-standards  As the Contract Standards WG member, I lean towards approving this NEP extension _after resolving the concerns in the table below_.

### Benefits

The standard extension effectively adds a `build_info` to the contract metadata in a **backward-compatible manner**.

```diff
type ContractSourceMetadata = {
    version: string|null,
    link: string|null,
    standards: Standard[]|null,
+    build_info: BuildInfo|null, // optional, details required for contract WASM reproducibility.
}
```

The  NEP provides examples of how the fields of `BuildInfo` can be customized and how build reproducibility can be achieved.

### Concerns

| - | - | - | - |
| 1 | Considerations about `BuildInfo` structure complexity | Strusture was flattened and simplified | Resolved |
| 2 | How to properly use the NEP  | More details and examples added | Resolved |
| 3 | Build reproducibility via Docker images | Details about how to use `build_environment` and `build_command` | Resolved |
| 4 | Concerns about 3rd party dependencies which affect reproducibility | Suggest committing `Cargo.lock` | In Progress |
| 5 | Using public CI/CD releases instead of Docker environment | Using `build_environment` link to a custom env | To be done |

neps/nep-0330.md
robert-zaremba: Yes, good idea!

robert-zaremba:
### Recommendation

@near/wg-contract-standards As the Contract Standards WG member, I lean towards approving this NEP extension _after resolving the Cargo.lock suggestion  form @fadeevab_


### Concerns

| - | - | - | - |
| 1 | be more specific about optional and required fields | spec updated | Resolved |
| 2 | be more specific about version and `link` field  | `link` is an URL to version of repository when available | Resolved |
| 3 | consider changing field names | discussed and renamed | Resolved |

neps/nep-0330.md
Canvinus: Added note about Cargo.lock in 222d2602059765f1a127223840a72f2de64bb219
And changed the example repo to https://github.com/near/cargo-near-new-project-template in 35bed0fbc75783132b7ca95a11c9c51974844824

fadeevab: @Canvinus Thanks but Сargo.lock is still necessary because dependencies of dependencies may change.

Canvinus: @fadeevab, I have info about this in the comment of source_code_snapshot, where do you think it is also necessary to include information about Cargo.lock

fadeevab: @Canvinus
Thank you for your clarification questions!

1. I see that https://github.com/near/core-contracts have `Cargo.lock`s, it could be used as an example.
2. I realize we already use  https://github.com/near/cargo-near-new-project-template: is it possible to commit `Cargo.lock` there?
3. I would suggest moving that sentence from the code comment to the specification (somewhere before the reference implementation).

```
### Cargo.lock

It is important to have `Cargo.lock` inside the source code snapshot to ensure reproducibility. Example: https://github.com/near/core-contracts.
```
👇
image

Canvinus: Great that most of this project is still compiling, thanks to Cargo.lock: https://github.com/near/cargo-near-new-project-template. But I thought, isn't that a problem that they are using an outdated SDK that is not even capable of having contract source metadata? Wouldn't this be confusing to use as a reference?

Canvinus: Added new section `Ensuring WASM Reproducibility` c06b054f4ec19e77ce92df2fe30a4bab5b8e0ea7 and moved info out of comments into it

fadeevab: > Added new section `Ensuring WASM Reproducibility` [c06b054](https://github.com/near/NEPs/commit/c06b054f4ec19e77ce92df2fe30a4bab5b8e0ea7) and moved info out of comments into it

Awesome!

>  Wouldn't this be confusing to use as a reference?

An alternative is to add `Cargo.lock` to https://github.com/near/cargo-near-new-project-template and use it as a reference. It would be the best option probably...

fadeevab:
@victorchimakanu we still have unfinished Cargo.lock issue https://github.com/near/NEPs/pull/533#discussion_r1523848009 @Canvinus what's the status?..

Canvinus:
@fadeevab, I talked with @frol about it, and they will update cargo-near-new-project-template to include Cargo.lock. As soon as that happens, I will include the latest commit of cargo-near-new-project-template as an example.

fadeevab:
@Canvinus @frol Agree! I created a tracking issue https://github.com/near/cargo-near-new-project-template/issues/9 (don't forget it!), let's move forward with this NEP

fadeevab (approved):

robert-zaremba (approved):

victorchimakanu:
As the moderator, I would like to thank the author @Canvinus and @alexzinin for submitting this NEP, and the NEAR contracts standards work group members for their reviews. Based on the voting indications above, it seems like this proposal is close to a decision. We are therefore scheduling the **Fifth Contract standards Work group meeting** next week, where this NEP can enter the final voting stage. Anyone can discuss the technical details by adding your comments to this discussion. And join the call to learn about the final decision and how to get more involved.”

#### Meeting Info

NEP Discussion: NEP-330
📅 Date: Thursday April 18th 2024 5PM UTC
✏️ [Add to Calendar](https://calendar.google.com/calendar/event?action=TEMPLATE&tmeid=N3YyZXJpZ2xwaGRicmRydTFhc2JjMDR1NjYgY19mNTRlZDM3ZmQ5MjMyN2FjZGM3ZTQzNDNmZTQwNzIyYWU1Nzk3YjZjODI5MjliYTkzZTlmM2E4OWM2OTY1N2FiQGc&tmsrc=c_f54ed37fd92327acdc7e4343fe40722ae5797b6c82929ba93e9f3a89c69657ab%40group.calendar.google.com)

Canvinus:
> @Canvinus It makes sense to fix a linter error Wrong url: https://rust-random.github.io/rand/rand/distributions/weighted/alias_method/struct.WeightedIndex.html Good url: https://rust-random.github.io/rand/rand/distributions/struct.WeightedIndex.html

https://github.com/near/NEPs/pull/533/commits/ccf93305be4f729ea6637721a4ce730d7d10869d
Looks like it still has bunch of linter errors)

fadeevab:
> > @Canvinus It makes sense to fix a linter error Wrong url: https://rust-random.github.io/rand/rand/distributions/weighted/alias_method/struct.WeightedIndex.html Good url: https://rust-random.github.io/rand/rand/distributions/struct.WeightedIndex.html
> 
> [Looks like it still has bunch of linter errors)](https://github.com/near/NEPs/pull/533/commits/ccf93305be4f729ea6637721a4ce730d7d10869d)

Oh, right :) Usually, a markdown formatter in VS Code fixes 99% of those issues for me, e.g. placing spaces around headers. I guess, you should try.

frol:
Thank you to everyone who attended the Contract Standards Working Group meeting today! The working group members reviewed the NEP and reached the following consensus:

**Status: Approved**

* 👍 @frol https://github.com/near/NEPs/pull/533#issuecomment-1983429172
* 👍 @robert-zaremba https://github.com/near/NEPs/pull/533#issuecomment-1983820442
* 👍 @fadeevab https://github.com/near/NEPs/pull/533#issuecomment-1983820442
Meeting Recording: https://youtu.be/JQu8tObzhlM

@Canvinus Thank you for championing the extension to NEP-330

@fadeevab @robert-zaremba Thank you for all the review!

Next steps:

* @frol should merge the NEP PR
* Adopt the implementations on the [cargo-near](https://github.com/near/cargo-near/pull/134), SourceScan, and beyond (e.g. NearBlocks)

PR #536, opened 2024-03-13
NEP-536: Reduce the number of refund receipts

bowenwang1996:
This proposal aims to reduce the number of refund receipts and optimize the performance of the protocol. An implementation can be found [here](https://github.com/near/nearcore/compare/drop-gas-price-difference-refunds?expand=1)


neps/nep-0536.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Motivation
+
+Refund receipts create nontrivial overhead: they need to be merklized and sent across shards. In addition, the processing of refund receipts requires additional storage reads and writes, which is not optimal for the performance of the protocol. In addition, when there is congestion, refund receipts may be delayed during execution. Whenever this happens, it requires two additional storage writes to store a gas refund receipt and two additional reads and writes when they are later processed, which incurs a significant overhead. To optimize the performance of the protocol under congestion, it is imperative that we reduce the number of refund receipts.
+
+## Specification
+
+Pessimistic gas pricing is removed as a part of this change. This means that transactions that do not involve function calls will not generate gas refund receipts as a result. For function calls, this proposal changes the gas refund to be
+```
+refund = max(0, gas_refund - 5Tgas);
```
mm-near, line 25: please mention how 5T was selected. 

And how will this impact the final pricing of a simple transfer and of a "simple" function call.

jakmeier:
Looks like a positive change to me and has my support in principle.

Although it has a different goal, the proposal aligns nicely with my proposal to simplify gas pricing (https://github.com/near/NEPs/discussions/500) where option 1 is just removing pessimistic gas pricing.

As I read it, the only difference is that on top of removing pessimistic gas pricing, you also reduce refunds by a fixed amount (5 Tgas). I'm not so sure about this change. If developers care about the lost gas value, they will start to estimate exactly how much gas a call needs, which can be difficult and may lead to spurious failures in their apps. And because there are no ways to query gas parameters on chain, this is bound to lead developers to use even more hard coded gas numbers in their smart contracts and relying tightly on them. Which makes it even harder for protocol changes to change gas costs.
So yeah, I don't like it because I believe it makes life of contract developers and protocol developers more difficult in the long term.

neps/nep-0536.md
jakmeier: Yeah 5 Tgas is curious to me. Especially since a refund is just a transfer and the cost of a transfer is only 230 Ggas.

5 Tgas seems like overcharging a lot to incentives developers to fiddle with their gas numbers. If this means it's considered essential for contract developers to do exact gas estimates from now on, that's a massive decline in Devx if you ask me and I am not sure if this is really worth it.

neps/nep-0536.md
```diff
@@ -0,0 +1,97 @@
...
+DiscussionsTo: https://github.com/near/NEPs/pull/536
+Type: Protocol
+Version: 1.0.0
+Created: 2024-03-12
+LastUpdated: 2023-03-12
+---
+
+## Summary
+
+[Gas refund](https://docs.near.org/concepts/basics/transactions/gas#attach-extra-gas-get-refunded) is a mechanism that allows users to get refunded for gas that is not used during the execution of a smart contract. Due to [pessimistic gas pricing](https://docs.near.org/concepts/basics/transactions/gas-advanced#pessimistic-gas-price-inflation), however, even transactions that do not involve function calls generate refunds because users need to pay at a high price and get refunded the difference. Gas refunds lead to nontrivial overhead in runtime and other places, which hurts the performance of the protocol. This proposal aims to reduce the number of gas refunds and prepare for future changes that completely remove gas refunds.
```
jakmeier, line 15: > Gas refunds lead to nontrivial overhead in runtime and other places, which hurts the performance of the protocol.

I'm curious how much this really is, do we have numbers on it? I would expect there is about 1 refund (0.23 Tgas) for 1 function call (at least 4.8 Tgas). So I would think it's less than 5% of total workload in terms of gas.
Also, the size of a transfer is tiny and shouldn't be a problem for state witness sizes.

Am I missing something?

neps/nep-0536.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Reference Implementation
+
+The protocol changes are as follows:
+* When a transaction is converted to a receipt, there is no longer a `pessmistic_gas_price` multiplier when the signer balance is deducted. Instead, the signer is charged `transaction_gas_cost * gas_price`. There is no gas refund for actions other than function calls.
+* For function calls, if X gas is attached during the execution of a receipt and Y gas is burnt, then max(0, X-Y-5Tgas) is refunded at the original gas price.
+
+## Security Implications
+
+This change may lead to less correct charging for transactions when there is congestion. However, the entire gas price mechanism needs to be rethought any ways and when only one or two shards are congested, the gas price wouldn't change so there is no difference.
```
mfornet, line 37: This brings back the known attack of a malicious user clogging the network for a low gas price.

The attack:
The malicious user sends multiple transactions at a low gas price enough to fill a block. Each transaction calls the same contract in a loop until it runs out of gas, spanning a significant period of time.

The cost of the attack with this change will be smaller since the gas is paid at the cost of the gas in the beginning.

This attack can be done repeatedly by waiting until the gas price decreases, keeping the network from the POV of genuine users as either congested or with high gas prices.

neps/nep-0536.md
evgenykuzyakov: Note, that the gas refunds also touch the access key from which they were issued to refund the gas. And the transfer is about 460 (since it's send + exec)

The penalty should be large enough to also give developers estimation range. For example, if a developer estimates their call to be from 7 to 9Tgas. They can attach 10Tgas and no refunds will be generated. This will decrease the load on the shard by removing extra refunds.

neps/nep-0536.md
evgenykuzyakov: Valid concern which should be addressed with receipt execution order based either on gas price or priority fee, but it's a separate NEP. Even right now the attack above is not mitigated with the pessimistic gas because you can spawn multiple transactions in a short period of time with low attached gas. It only makes it marginally more expensive which is not worth sacrificing for better UX.

evgenykuzyakov:
We've had a discussion with @bowenwang1996 and came with slightly different mechanism.

We should introduce a gas refund receipt cost (about 650Ggas, see below) and "Gas over-attaching penalty". 
The penalty will be up to 5% of the refund gas. The goal is to discourage developers from over attaching too much gas. Examples:
- If the contract wants to refund 280Tgas, burning 5% of it would about 14Tgas which is significant cost and developers would be encouraged to optimize it on the front-end.
- If refund is 100Tgas, then 5% is 5Tgas - still a lot and achieves the goal 
- If the refund is <10Tgas (very common case for cross-contract call self-callbacks), the penalty should be just 500Ggas, which is less than the gas refund cost. So only the fixed refund cost will be charged from gas to spawn the gas refund receipt. No UX will be broken for legacy cross-contract call contracts, so long as frontend correctly estimates the required gas in worst case scenario.

The 650Ggas comes from the following (based on https://github.com/near/nearcore/blob/master/core/parameters/res/runtime_configs/parameters.snap#L13)
- action_receipt_creation (send_not_sir + execution): 216119000000
- action_transfer (send_not_sir + execution): 230246125000
- action_add_function_call_key (send_not_sir + execution): 204435250000

We need `action_add_function_call_key` because gas refund might refund the allowance for a given limited signer access key, so should expect it to be included.

Note, for implicit accounts, there is no extra cost, because it `system` can't create a new implicit account during a refund.

birchmd:
Thanks for continuing the discussion here @evgenykuzyakov . I am not sure I fully understand your proposal. I'll rephrase it below and please correct me if I get parts of it wrong.

1. We keep @bowenwang1996 's current proposal of removing pessimistic gas cost so that only FunctionCall actions can have refunds (all other actions will be charged exactly the right amount of gas at execution so no refunds are needed).
2. Any FunctionCall action which would have a refund of less than 650 Ggas will no longer receive any refund. This contributes to the goal of reducing the number of refund receipts.
3. Any FunctionCall action which would have a refund of more than 650 Ggas is now charged the additional 650 Ggas (the cost of running the refund) as well as a 5% "over-estimation fee". The purpose of the additional fee is to encourage front-end developers and end-users interacting via cli to submit accurate gas estimates which do not trigger refunds.

If I understand this correctly then I have a question about the 650 Ggas refund cost. Does the current gas estimation logic for receipts and transfers already include the cost of running the refund? I know these costs are estimated by executing the runtime in some kind of metered environment, but I'm not sure what exactly is included. If the refund cost is already rolled in then I don't think we want to double-charge for it and perhaps other costs could go down. Of course if we are currently not charging for refunds at all then this is an extra important change!

> No UX will be broken for legacy cross-contract call contracts, so long as frontend correctly estimates the required gas in worst case scenario.

If my understanding of the proposal as I have written it above is correct then I don't think we need to worry about backwards compatibility because the only change will be to the amount of refunded gas (and this cannot have any impact on the amount of gas attached to promises).

bowenwang1996:
> Does the current gas estimation logic for receipts and transfers already include the cost of running the refund? I know these costs are estimated by executing the runtime in some kind of metered environment, but I'm not sure what exactly is included. If the refund cost is already rolled in then I don't think we want to double-charge for it and perhaps other costs could go down.

It does. So as part of this change, we will rerun the estimator to reduce the action gas costs. This has the benefit that for non-function-call actions, because refunds are removed, they will be cheaper.

bowenwang1996:
I updated the proposal to reflect https://github.com/near/NEPs/pull/536#issuecomment-1998159850. @mfornet @mm-near @birchmd please take a look again

neps/nep-0536.md
```diff
@@ -0,0 +1,104 @@
...
+* If refund is 100Tgas, then 5% is 5Tgas, which is still significant and discourages developers from doing so.
+* If the refund is <10Tgas (very common case for cross-contract call self-callbacks), the penalty should be just 500Ggas, which is less than the gas refund cost. So only the fixed refund cost will be charged from gas to spawn the gas refund receipt. No UX will be broken for legacy cross-contract call contracts, so long as frontend correctly estimates the required gas in worst case scenario.
+
+
+## Reference Implementation
+
+The protocol changes are as follows:
+* When a transaction is converted to a receipt, there is no longer a `pessmistic_gas_price` multiplier when the signer balance is deducted. Instead, the signer is charged `transaction_gas_cost * gas_price`. There is no gas refund for actions other than function calls.
+* For function calls, if X gas is attached during the execution of a receipt and Y gas is burnt, then `max(0, X-Y-refund_cost)` is refunded at the original gas price where `refund_cost = max(REFUND_FIXED_COST, 0.05 * X-Y)`.
```
birchmd, line 39:
```suggestion
* For function calls, if X gas is attached during the execution of a receipt and Y gas is used+burnt, then `max(0, X-Y-refund_cost)` is refunded at the original gas price where `refund_cost = max(REFUND_FIXED_COST, 0.05 * X-Y)`. In the case the refund is 0 then no refund receipt is generated.
```

I think it is important to say "used" here in addition to burnt because gas can be attached to other promises as well as burnt in Wasm computation.

neps/nep-0536.md
```diff
@@ -0,0 +1,104 @@
...
+## Motivation
+
+Refund receipts create nontrivial overhead: they need to be merklized and sent across shards. In addition, the processing of refund receipts requires additional storage reads and writes, which is not optimal for the performance of the protocol. In addition, when there is congestion, refund receipts may be delayed during execution. Whenever this happens, it requires two additional storage writes to store a gas refund receipt and two additional reads and writes when they are later processed, which incurs a significant overhead. To optimize the performance of the protocol under congestion, it is imperative that we reduce the number of refund receipts.
+
+## Specification
+
+Pessimistic gas pricing is removed as a part of this change. This means that transactions that do not involve function calls will not generate gas refund receipts as a result. For function calls, this proposal introduces cost of refund to be
+```
+REFUND_FIXED_COST = action_receipt_creation + action_transfer + action_add_function_call_key
+refund_cost = max(REFUND_FIXED_COST, 0.05 * gas_refund);
```
birchmd, line 26: I think it would be nice to justify this 5% number. How is it chosen instead of 6% or 4%? What network data is used to make this decision?

birchmd (commented):
I like the changes that have been made to the proposal. I think it would still be good to justify the 5% part of the new receipt cost.

neps/nep-0536.md
bowenwang1996: It is an arbitrary number. The goal here is to start penalizing users for attaching too much gas. The percentage should be large enough so that someone attaching 100Tgas or more is penalized for doing so but also small enough so that a callback with a fixed amount of gas reserved are not heavily impacted because the contract is potentially already locked. I think a smaller percentage would work well too.

neps/nep-0536.md
```diff
@@ -0,0 +1,104 @@
...
+and creating unnecessary refunds. Some examples:
+* If the contract wants to refund 280Tgas, burning 5% of it would be about 14Tgas, which is a significant cost and developers would be encouraged to optimize it on the frontend.
+* If refund is 100Tgas, then 5% is 5Tgas, which is still significant and discourages developers from doing so.
+* If the refund is <10Tgas (very common case for cross-contract call self-callbacks), the penalty should be just 500Ggas, which is less than the gas refund cost. So only the fixed refund cost will be charged from gas to spawn the gas refund receipt. No UX will be broken for legacy cross-contract call contracts, so long as frontend correctly estimates the required gas in worst case scenario.
+
+
+## Reference Implementation
+
+The protocol changes are as follows:
+* When a transaction is converted to a receipt, there is no longer a `pessmistic_gas_price` multiplier when the signer balance is deducted. Instead, the signer is charged `transaction_gas_cost * gas_price`. There is no gas refund for actions other than function calls.
```
evgenykuzyakov, line 38: There might be a refund. Since we generate a refund per receipt, not per action. In case an action fails, the gas for all actions after it will be refunded. We'll use the same formula as for unused gas after a function call

neps/nep-0536.md
birchmd: > small enough so that a callback with a fixed amount of gas reserved are not heavily impacted

Shouldn't fixed gas callbacks be unaffected by this change? This change effectively makes refunds smaller, but there is no refund for gas reserved for promises until after those receipts have already executed.

bowenwang1996: The callbacks themselves are unaffected, but the amount of gas refund would be

victorchimakanu:

**Status: Review** 

SME reviews:

* Protocol SME @jakmeier : 
* Protocol SME @birchmd: 
* Protocol SME @evgenykuzyakov : 
* Protocol SME @mm-near:

* ? @bowenwang1996 
* ? @birchmd 
* ? @mfornet 
*  ? @mm-near

mm-near:

One important outcome of this proposal, is that "attaching MAX possible gas is now penalised" (as AFAIK in the past it was not). This is a large discrepancy from many of the other blockchains (for example Ethereum), so it should be clearly documented.

I'd also like to make sure that this new fee is clearly shown in the explorer -- so that users can see, the consequences of their gas overestimation - and exert some pressure on the app developers/wallets to improve the gas estimations (otherwise, if this is displayed together with all the other fees, users will not notice, and will just consider NEAR "expensive").

neps/nep-0536.md
```diff
@@ -0,0 +1,104 @@
...
+* For function calls, if X gas is attached during the execution of a receipt and Y gas is used+burnt, then `max(0, X-Y-refund_cost)` is refunded at the original gas price where `refund_cost = max(REFUND_FIXED_COST, 0.05 * X-Y)`. In the case the refund is 0 then no refund receipt is generated.
+* Tokens burnt on refund cost is counted towards tx_balance_burnt and the part over `REFUND_FIXED_COST` is not counted towards gas limit to avoid artificially limiting throughput.
+* Because refund cost is now separate, action costs do not need to account for refund and therefore should be recalculated and reduced.
+
+## Security Implications
+
+This change may lead to less correct charging for transactions when there is congestion. However, the entire gas price mechanism needs to be rethought any ways and when only one or two shards are congested, the gas price wouldn't change so there is no difference.
+
```
robert-zaremba, line 46: Let's consider the following scenario:
* tx with 300TG is included block b1, and converted into receipts (containing function calls). 
* Eventually, only 50TG was used to process all receipts.

How does this impact the block limit?

bowenwang1996: It doesn't affect how gas used is computed, so there is no impact on block limit

flmel (approved):

jakmeier:
I'm currently implementing this NEP. See https://github.com/near/nearcore/issues/13397 for the progress.

I have one question regarding the specification to the authors (@bowenwang1996 @evgenykuzyakov) and wg.members (@birchmd, @mfornet, @mm-near).  (also cc @graphite)

What should happen when the actual gas price during execution is different than the purchased gas price?

- Option A: Charge the block gas price. If that exceeds the balanced paid upfront, charge as much possible and write the rest of as gas deficit.
- Option B: Always charge the gas price noted on the receipt. The difference to the block gas price is written off as gas deficit in the balance checker.

In  both cases, we refund the difference between Ⓝ paid for gas and the Ⓝ value of the burnt gas. It's just that in option A, we use two different gas prices to calculate the Ⓝ value, whereas in option B we use the same price twice.

The current behaviour is option A. It's in line with the idea that pessimistic gas pricing tries to predict price movements during execution.

Without pessimistic gas pricing, option A starts looking odd IMO.

Example: A user attaches 100 Tgas and they burn through 50 Tgas. Usually, they get a refund of 50 Tgas. But if the gas price went up by 20%, with Option A, we only refund NEAR tokens equivalent to 40Tgas at the purchase price. Or if the price goes down 20%, we end up refunding tokens equivalent to 60 Tgas. With option B, we always refund tokens equivalent to 50 Tgas in value.

If the same user instead attached exactly 50Tgas, an increased price leads to a deficit of (gas_delta * 50 Tgas). Both options write it of in the balance checker that the system has to absorb. But if the price decreases, there is a surplus of (gas_delta * 50 Tgas) instead. Option A refunds the surplus to the user, whereas option B absorbs the surplus in the system.

See the two tables below to clearly see what happens:

  | Attached gas | Value attached gas at 0.1 N / 1000 Tgas | Burnt gas | Value burnt gas | Option A | Option B
-- | -- | -- | -- | -- | -- | --
Burn @ 0.08 N / 1000 Tgas | 100 Tgas | 0.01 Ⓝ | 50 Tgas | 0.004 Ⓝ | Refund 0.006Ⓝ (60Tgas)| Refund 0.005Ⓝ (50Tgas) 
Burn @ 0.10 N / 1000 Tgas | 100 Tgas | 0.01 Ⓝ | 50 Tgas | 0.005 Ⓝ | Refund 0.005Ⓝ (50Tgas) | Refund 0.005Ⓝ (50Tgas) 
Burn @ 0.12 N / 1000 Tgas | 100 Tgas | 0.01 Ⓝ | 50 Tgas | 0.006 Ⓝ | Refund 0.004Ⓝ (40Tgas) | Refund 0.005Ⓝ (50Tgas) 


However, if he user attaches exactly 50 Tgas, things start looking odd:

  | Attached gas | Value attached gas at 0.1 N / 1000 Tgas | Burnt gas | Value burnt gas | Option A | Option B
-- | -- | -- | -- | -- | -- | --
Burn @ 0.08 N / 1000 Tgas | 50 Tgas | 0.005 Ⓝ | 50 Tgas | 0.004 Ⓝ | **Refund 0.001Ⓝ (10 Tags)** | No refund, surplus of 0.001 Ⓝ
Burn @ 0.10 N / 1000 Tgas | 50 Tgas | 0.005 Ⓝ | 50 Tgas | 0.005 Ⓝ | No refund | No refund 
Burn @ 0.12 N / 1000 Tgas | 50 Tgas | 0.005 Ⓝ | 50 Tgas | 0.006 Ⓝ | No refund, deficit of 0.001 Ⓝ | No refund, deficit of 0.001 Ⓝ

The case in bold (option B, first row) has an extra refund even though the receipt burnt exactly as much gas as was attached. This will be visible in the explorer and may lead to confusion.

The surplus / deficit, on the other hand, is not visible on the explorer. It's just for blockchain internal bookkeeping, ensuring a clean balance sheet on every block.

Therefore, if we want things to be easier to understand for users, I suggest we use option B. Just ignore the gas delta.

On the other hand, if we want to keep things closer to how they are today, we should use option A, since this is how refunds were designed initially.

jakmeier:
Also, any objections to use `REFUND_FIXED_COST = 1 Tgas` in the implementation?

The original proposal in this NEP was 5 Tgas and was later changed because people (including me) asked where it came from. (see https://github.com/near/NEPs/pull/536#discussion_r1522597108)

The latest NEP version describes it as:

`REFUND_FIXED_COST = action_receipt_creation + action_transfer + action_add_function_call_key`

which I calculated to be

`2 * 108_059_500_000 + 2 * 115_123_062_500 + 2 * 102_217_625_000 = 650_800_375_000 = 0.650_800_375 Tgas`


I don't think the intention was to add complicated interactions between parameters, nor to make the penalty fee calculation overly complicated. Further, part of the explanation was that we want to penalize enough to actually reduce the number of refunds. Thus, I suggest we round up to 1 Tgas, rather than use 0.650800375 Tgas.

In practice, this means gas refunds cost at least 1Tgas. Simple enough for any developers to grasp in one sentence.

graphite:
> suggest we use option B. Just ignore the gas delta.

@jakmeier I agree with this one.

Jaynbala1:
b4181162af5f78a7fc5ca5587b4e627c3310f881

PR #538, opened 2024-03-18
Fix GitHub links in nep-0455.md

akashin:
The links to "cs.github.com" are now broken as it was removed. So we replace them with the usual GitHub links.


akashin:
@near/nep-moderators , friendly ping :)

robert-zaremba (approved):
Thanks for the fixes!

frol (approved):
@akashin Thank you for catching and fixing it!

PR #539, opened 2024-03-22
Cross-Shard Congestion Control

jakmeier:
[Latest rendered view](https://github.com/jakmeier/NEPs/blob/congestion-control/neps/nep-0539.md).

**Status**: Approved

**Meeting Recording:**
https://www.youtube.com/watch?v=O1MOBmxKqhI

* 👍 @birchmd:
https://github.com/near/NEPs/pull/539#pullrequestreview-2068533114
* 👍 @mfornet:
https://github.com/near/NEPs/pull/539#pullrequestreview-2070744236
* 👍 @mm-near:
https://github.com/near/NEPs/pull/539#issuecomment-2129717307
* 👍 @bowenwang1996:
https://github.com/near/NEPs/pull/539#issuecomment-2123522826

---------


jakmeier:
And a first draft of "the story behind" is also available: https://github.com/jakmeier/nearcore/blob/congestion-docs/docs/architecture/how/receipt-congestion.md

While the NEP focusses on specifying the proposed changes, the story behind explains our thought process why these changes lead to the desired consequences.

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+```python
+def filter_tx(tx):
+  INCOMING_CONGESTION_THRESHOLD = 0.5
+  if incoming_congestion(tx.receiver_shard_id) > INCOMING_CONGESTION_THRESHOLD
+    tx.reject()
+  else
+    tx.accept()
+```
+
+*TODO: fine-tune `INCOMING_CONGESTION_THRESHOLD`*
```
bowenwang1996, line 133: We likely need a lower one if maximum is 100Pgas. From a user experience point of view, I think over 10Pgas of delay is already quite suboptimal and should trigger the congestion control mechanism

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+# old
+tx_limit = 500 Tgas if len(delayed_receipts) < 20_000 else 0
+```
+
+We replace the formula for the transaction limit to depend on the
+`general_congestion` computed in the previous shard variable (between 0 and 1):
+
+```python
+# new
+MIN_GAS = 5 Tgas
```
staffik, line 105: Should it be 5 Tgas or 0 Tgas?

jakmeier: I think 5 Tgas is better but it's up for debate.

The idea here is that we allow a small amount of gas for accepting new transactions to other shard, if we know those other shards have `incoming_congestion = 0`. (see condition further below)

The thing is, if we allow it to go to zero, we allow complete starvation of a shard by other shards.
In practical terms, if my accounts happens to be on the same shard as HOT and everyone uses HOT, I can no longer do anything on any shard with my account, because I need to start at my own shard. Even if I only spend 2 Tgas on function call conversion on the congested shard and then 302 Tgas on a different shard which is currently idle.

Also, the harm of including 5 Tgas seems small since we still have 995 Tgas left for working on receipts to reduce the congestion.

staffik: I see, makes sense.

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+In the pseudo code above, we borrow the [`mix`](https://docs.gl/sl4/mix)
+function from GLSL for linear interpolation.
+
+> `mix(x, y, a)`
+> 
+> `mix` performs a linear interpolation between x and y using a to weight between
+> them. The return value is computed as $x \times (1 - a) + y \times a$. 
+
+
+Plus, we add the additional rule to reject all transactions to shard with an
```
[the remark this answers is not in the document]
jakmeier: Kind of yes, in the proposal we are only rejecting it at the chunk producer level. Everything else is not a protocol change and will not make it into this NEP. So we don't really define what should be treated as accepted or not accepcted.

Your described ideal world can certainly be implemented with the proposed spec here. If we take the vocabulary from [issue #6837](https://github.com/near/nearcore/issues/6837), anything except wait_until=None should not return success before a transaction is actually in a block. And the rejection filter happens before inclusion in a block.

In the most naive solution, we wait until the chunk producer rejects it and don't tell the RPC node waiting for it. This would be bad and the transaction just keeps waiting until some timeout.

Middle-ground: We signal back to the RPC node that the tx was rejected. This can be tricky since we forward it to several chunk producers and maybe some of them accept while others reject, if congestion is just building up or just being resolved. But we can just wait for all of them to answer before telling the user that it was rejected.

Dream scenario: RPC nodes run the rejection filter before even forwarding and reject it instantly. This also reduced load on the network to chunk producers.

jakmeier:
> A summary of my understanding is that each shard is going to advertise how much queue space it has available and other shards will take that into account when constructing their chunks and accepting new transactions. Is that a fair summery?

Yes, that sounds exactly right.

>     1. Shard A is congested and shard B and C both have a ton of receipts for it.  Assuming all shards are created equal, how do we make sure that the remaining queue space is shared fairly between B and C?  Is it by relying on the linear interpolation?

We don't give any guarantees about fairness. We hope that measures are reducing incoming transactions sharp enough that congestion resolves quickly and everyone can send again. But yes, linear interpolation of how much bandwidth (measured in gas) each shard can send per chunk should help in most practical scenarios, as the newly available space in the incoming queue of the congested shard is shared evenly across all sending shards.

>     2. Shard A is congested and shard B has a ton of receipts for it and shard C has no receipts for it.  How do we make sure that we are able to provide all the queue space to B and do not reserve any for C?

There is only one big incoming queue, without accounting per shard. So in this example, shard B can fill it up entirely. Shard C will be sad when it wants to send a single receipt and sees the queue full. But I personally think it's a good trade-off to make.

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+more and more transactions that produce receipts for the popular contract.
+Therefore the number of delayed receipts keeps growing indefinitely.
+
+Cross-shard congestion control addresses this issue by stopping new transactions
+at the source and delaying receipt forwarding when the receiving shard has
+reached its memory limit.
+
+## Specification
+
+The proposal adds fields to chunks headers, it adds a trie columns, it changes
```
wacban, line 64:
```suggestion
The proposal adds fields to chunks headers, it adds a new trie column, it changes
```

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+  incoming_congestion: u16
+  general_congestion: u16
+}
+```
+
+This adds 4 bytes to the chunk header, increasing it from 374 bytes to 378
+bytes in borsh representation. (Assuming no validator proposals included.)
+This in turn increases the block size by 4 bytes per shard.
+
+While the numbers are stored as `u16`,for the remainder of the NEP
```
wacban, line 85:
```suggestion
While the numbers are stored as `u16`, for the remainder of the NEP
```

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+
+Cross-shard congestion control addresses this issue by stopping new transactions
+at the source and delaying receipt forwarding when the receiving shard has
+reached its memory limit.
+
+## Specification
+
+The proposal adds fields to chunks headers, it adds a trie columns, it changes
+the rules to select transactions, and it changes the chunk execution flow. The
+next four sections specifies each of these changes in more details.
```
wacban, line 66:
```suggestion
next four sections specify each of these changes in more details.
```

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+    - (new) Non-local receipts above `outgoing_gas_limit[receipt.receiver]` for
+      the respective shard go to the outgoing receipts buffer.
+4. Execute receipts in the order of `local`, `delayed`, `incoming`.
+    - Don't stop before all receipts are executed or more than 1000 Tgas have
+      been burnt. Burnt gas includes the burnt gas from step 3.
+    - Outgoing receipts up to what is left in
+      `outgoing_gas_limit[receipt.receiver]` per shard (after step 3) go to the
+      outgoing receipts list of the chunk.
+    - (new) Above `outgoing_gas_limit[receipt.receiver]` receipts go to the
+      outgoing buffer.
```
wacban, line 173:
```suggestion
    - (new) Outgoing receipts above `outgoing_gas_limit[receipt.receiver]`
      go to the outgoing receipts buffer.
```

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+    memory_consumption += sum([receipt.size() for receipt in postponed_receipts_queue])
+    memory_consumption += sum([receipt.size() for receipt in outgoing_receipts_buffer])
+  
+    general_congestion = memory_consumption / MAX_CONGESTION_MEMORY_CONSUMPTION
+    general_congestion = min(1.0, general_congestion)
+    ```
+
+*TODO: fine-tune `MAX_CONGESTION_INCOMING_GAS` and `MAX_CONGESTION_MEMORY_CONSUMPTION`*
+
+In the formula above, we receipt gas and size are defined as:
```
wacban, line 195:
```suggestion
In the formula above, the receipt gas and the receipt size are defined as:
```

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+
+def size(receipt):
+  return len(borsh(receipt))
+```
+
+### Change to Trie
+
+We store the outgoing buffered receipts in the trie, similar to delayed receipts
+but in their own separate column. Therefore we add a trie column
+`DELAYED_RECEIPT_OR_INDICES: u8 = 13;`. Then we read and write analogue to tje
```
wacban, line 209:
```suggestion
`BUFFERED_RECEIPT_OR_INDICES: u8 = 13;`. Then we read and write analogue to the
```

Is this the name you wanted to go for? Anyway it shouldn't be the same as for delayed receipts :)

wacban (commented):
a few nits, typos and such

neps/nep-0539.md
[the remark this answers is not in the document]
akhi3030: sounds good.  Agreed that such discussions are outside the scope of the NEP and should definitely be possible with this NEP.

akhi3030:
> > A summary of my understanding is that each shard is going to advertise how much queue space it has available and other shards will take that into account when constructing their chunks and accepting new transactions. Is that a fair summery?
> 
> Yes, that sounds exactly right.
> 
> > ```
> > 1. Shard A is congested and shard B and C both have a ton of receipts for it.  Assuming all shards are created equal, how do we make sure that the remaining queue space is shared fairly between B and C?  Is it by relying on the linear interpolation?
> > ```
> 
> We don't give any guarantees about fairness. We hope that backpressure measures are reducing incoming transactions sharp enough that congestion resolves quickly and everyone can send again. But yes, linear interpolation of how much bandwidth (measured in gas) each shard can send per chunk should help in most practical scenarios, as the newly available space in the incoming queue of the congested shard is shared evenly across all sending shards.
> 
> > ```
> > 2. Shard A is congested and shard B has a ton of receipts for it and shard C has no receipts for it.  How do we make sure that we are able to provide all the queue space to B and do not reserve any for C?
> > ```
> 
> There is only one big incoming queue, without accounting per shard. So in this example, shard B can fill it up entirely. Shard C will be sad when it wants to send a single receipt and sees the queue full. But I personally think it's a good trade-off to make.

Generally happy with your responses here.  One other approach I have seen (and implemented in the past) to guarantee fairness is some sort of credit based queuing.  This lets a receiving entity decide in fine grain how much of its queue it wants to dedicate to each sender.  It is natural to use this mechanics to implement fair sharing or to arbitrary types of prioritisation as well (e.g. one shard is able to send 2x more than another).  The drawback of course is more state tracking and complex implementation.  So I'm happy with the proposed approach.

neps/nep-0539.md
```diff
@@ -0,0 +1,307 @@
...
+
+def size(receipt):
+  return len(borsh(receipt))
+```
+
+### Change to Trie
+
+We store the outgoing buffered receipts in the trie, similar to delayed receipts
+but in their own separate column. Therefore we add a trie column
+`BUFFERED_RECEIPT_OR_INDICES: u8 = 13;`. Then we read and write analogue to the
```
saketh-are, line 209: Minor detail: I know we use this pattern for `DELAYED_RECEIPT_OR_INDICES`, but it seems to be that way for historical reasons ([see commit message here](https://github.com/near/nearcore/commit/7365af44942f984a23ae7c1956d9b96802e52c9b)).

For this new queue it would be clearer to have separate `BUFFERED_RECEIPT` and `BUFFERED_RECEIPT_INDICES` columns.

akhi3030:
Another question popped into my head earlier.  AFAIU, creating a promise in NEAR is infallible i.e. contract A on shard 1 can always create a receipt for contract B on shard 2.  Further, it is the case that without actually executing the receipt against contract A, we cannot know for sure whether or not it will call contract B.  In the worst case, many different contracts on many different shards can all target the same contract (or a set of contracts on a shard).  

Does the proposed solution handle such scenarios?  Is the filter operation defined going to apply to the receipts created above?

jakmeier:
> Another question popped into my head earlier. AFAIU, creating a promise in NEAR is infallible i.e. contract A on shard 1 can always create a receipt for contract B on shard 2. Further, it is the case that without actually executing the receipt against contract A, we cannot know for sure whether or not it will call contract B. In the worst case, many different contracts on many different shards can all target the same contract (or a set of contracts on a shard).
> 
> Does the proposed solution handle such scenarios? Is the filter operation defined going to apply to the receipts created above?

The filter operation is only applicable to transactions, not to receipts. Once receipts are created, we commit to execute them.

The described situation is indeed problematic. Of course, that's exactly what backpressure is for.

If shard 3 becomes congested, shard 1 and 2 can still create receipts for shard 3 but they are forced to keep them in their outgoing buffer before forwarding. This way, shard 3 is protected from additional inflow. Eventually, shards 1 and 2 may also become congested and the backpressure spreads further out to all shards trying to send something to them. Eventually all shards are congested and no more new transactions anywhere are accepted.

Unfortunately, it is still not handled perfectly. We only apply backpressure based on incoming congestion, to avoid deadlocks. But if we are able to handle incoming receipts quickly, it is possible shard 1 keeps filling its outgoing buffer for shard 2, growing it faster than it can forward receipts in it. But because the incoming queue is always empty, it does not apply backpressure. (cc @wacban we should probably simulate with the latest changes that decouple incoming and outgoing congested to see how bad this can become.)

akhi3030:
I think I understand the high level explanation.  The drawback is that in the worst case, due to one shard not keeping up, it is possible that the entire network has to stop accepting new transactions.  I am still happy with this solution and see this as a very good next step to build.  Once built, I can imagine further refinements where we can address such cases as well.

wacban:
@akhi3030 

> This lets a receiving entity decide in fine grain how much of its queue it wants to dedicate to each sender.

If I understand correctly this could be implemented by splitting the delayed receipts queue into one queue per sending shard and then implementing some fair way to pull receipts from this set of queues. This makes sense but I would rather keep this NEP in the current simpler form and work on top of it in follow ups. The good news is that as far as I can tell the current proposal should be easily extendable to what you're suggesting. 

> A summary of my understanding is that each shard is going to advertise how much queue space it has available and other shards will take that into account when constructing their chunks and accepting new transactions. Is that a fair summery?

That is correct, just to add a detail to it, each shard will advertise two numbers, one representing the fullness of the "outgoing queues" and one representing the fullness of the "incoming queue". Those two types of congestion are treated differently which allows us to better adapt the measures to the specific workload that the network is under.

akhi3030:
@wacban: perfect, sounds like a solid plan to me.  I am always happy to build incrementally.

wacban:
I implemented the model of the strategy proposed in the NEP. I am now analysing different workloads to make sure that the strategy can handle them well. I will be sharing results and suggestions here as I progress. 

### AllToOne workload. 

In this workload all shards send direct transactions to a single shard that becomes congested. 

The strategy does a rather bad job at dealing with this workload as the outgoing buffers grow in gas without a reasonable limit. The memory limit is never exceeded because the receipts are small but the number and gas of receipts grows beyond acceptable values. 

The reason is that the current proposal does not take the gas accumulated in outgoing buffers into account. 

My suggestion would be to replace memory congestion with `general_congestion` as following:

```rust
ShardChunkHeaderInnerV3 {
  // as is
  incoming_congestion: u16,
  // memory -> general
  general_congestion: u16,
}
```

```
	// Same as in NEP
    MAX_CONGESTION_MEMORY_CONSUMPTION = 500 MB
    memory_consumption = 0
    memory_consumption += sum([receipt.size() for receipt in delayed_receipts_queue])
    memory_consumption += sum([receipt.size() for receipt in postponed_receipts_queue])
    memory_consumption += sum([receipt.size() for receipt in outgoing_receipts_buffer])

    memory_congestion = memory_consumption / MAX_CONGESTION_MEMORY_CONSUMPTION
    memory_congestion = min(1.0, memory_congestion)
    
    // New
    // Similar to memory but summing up gas instead of size
    MAX_CONGESTION_GAS_BACKLOG = 100 PG
    gas_backlog = 0
    gas_backlog += sum([receipt.gas() for receipt in delayed_receipts_queue])
    gas_backlog += sum([receipt.gas() for receipt in postponed_receipts_queue])
    gas_backlog += sum([receipt.gas() for receipt in outgoing_receipts_buffer])

    gas_congestion = gas_backlog / MAX_CONGESTION_GAS_BACKLOG
    gas_congestion = min(1.0, memory_congestion)
    
    // New
    general_congestion = max(memory_congestion, gas_congestion)
```

I implemented the suggestion in the model and the results are quite good - both the incoming queue and outgoing buffers display bounded, periodic behaviour. 

In the picture below, each period is characterized by four phases:
* phase 1 - rapid growth
  * incoming gas grows to 150PG 
  * outgoing gas grows to 100PG 
  * shards send plenty of load to the loaded shard (0)
* phase 2 - incoming decline
  * incoming gas drops to 100PG 
  * outgoing gas stays at to 100PG 
* phase 3 - outgoing decline
  * incoming gas stays at 100PG
  * outgoing gase drops to 0PG
* phase 4 - incoming decline
  * incoming gas drops to 50PG - the threshold for accepting transactions
  * outgoing gas stays at 0PG

We can probably smooth it out further by replacing the hard incoming congestion threshold with linear interpolation. It's not a priority right now so I'll leave it as is. 

Screenshot 2024-04-05 at 13 54 21

jakmeier:
I believe this is now ready to move forward.

@walnut-the-cat can you help us getting SME reviews?

@bowenwang1996 According to our conversation, I included the changes to save all information in the chunk header, rather than only the chunk extra, to allow stateless validation to work without downloading the previous chunk. Please also check the "Validation Changes" section for a better understanding of potential requirements on stateless validation.

@wacban I allowed myself to add two TODOs in the document regarding resharding considerations. I don't know if we need to mention anything at all, as I understand the current way of resharding will not be used again? Your expertise would be most appreciated.

@robin-near and @akashin, both of you have contributed in large parts to earlier discussions. Wacban and I wouldn't have gotten here without your valuable contributions and all the ideas we discussed of the year+. If you have a moment, please take a look at the alternative sections and let me know if I missed something or perhaps misrepresent something. Also, if you have resources that we should link to, please make sure they are **publicly accessible** and send them to me, so I can add them.

bowenwang1996:
As a working group member, I nominate @robin-near and @akashin as subject matter experts to review this NEP.

neps/nep-0539.md
```diff
@@ -0,0 +1,661 @@
...
+    - Non-local receipts up to `outgoing_gas_limit[receipt.receiver]` for the
+      respective shard go to the outgoing receipts list of the chunk.
+    - (new) Non-local receipts above `outgoing_gas_limit[receipt.receiver]` for
+      the respective shard go to the outgoing receipts buffer.
+    - (new) For each receipt added to the outgoing buffer, add `receipt.gas()`
+      to `outgoing_congestion` and `receipt.size()` to `receipt_bytes` for
+      the local shard.
+5. Execute receipts in the order of `local`, `delayed`, `incoming`, `yield-resume time-out`.
+    - Don't stop before all receipts are executed or more than 1000 Tgas have
+      been burnt. Burnt gas includes the burnt gas from step 3.
```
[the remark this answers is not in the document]
jakmeier: This should be step 4, thanks for pointing it out. Forgot to change the reference when I added step 1.

walnut-the-cat:
@akashin and @robin-near , friendly reminder to share your review on this NEP

victorchimakanu:

**Status: Review** 

SME reviews:

* Protocol SME @robin-near 
* Protocol SME @akashin

* ❔ @bowenwang1996 
* ❔ @birchmd 
* ❔ @mfornet 
* ❔@mm-near

walnut-the-cat:
@bowenwang1996 , it seems @akashin is OOO this week, should we nominate someone else to expedite?

akashin:
I skimmed the proposal and overall approach looks good to me, great work!
I will aim to finish a more thorough review of each section by tomorrow.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+For more details, a spreadsheet with the full analysis can be found here:
+https://docs.google.com/spreadsheets/d/1Vt_-sgMdX1ncYleikYY8uFID_aG9RaqJOqaVMLQ37tQ/edit#gid=0
+
+#### Queue sizes
+
+The parameters are chosen to strike a balance between guarantees for short
+queues and utilization. 20 PGas delayed receipts means that incoming receipts
+have to wait at most 20 chunks to be applied. And it can guarantee 100%
+utilization as long as the ratio between burnt and attached gas in receipts is
+above 1 to 20.
```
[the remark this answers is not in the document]
jakmeier: No, we don't enforce it in our proposal.

It is an interesting idea to investigate. But note that some patterns inevitably burn less than 1/20 in the first step. A delegate action, for example, burns very little gas on the relayer's shard. Any enforcement measures needs to consider what happens to such use cases.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+### 
+
+### Positive
+
+- Accepted transaction have lower latencies compared to today under congestion.
+- Storage and memory requirements on validator for storing receipts are bounded.
+
+### Neutral
+
+- More transactions are rejected at the chunk producer level.
```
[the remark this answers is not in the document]
jakmeier: This is an important topic and a great point to be raised. Let me describe my thinking, which also reflects what we discussed in the last weekly congestion control meeting.

From a protocol point of view, we add a condition to reject transactions, meaning it would be invalid to add them to a chunk. What happens in the transaction pool isn't really relevant on that level, hence unspecified, or simply implementation choices.

The question regarding transaction pool queueing depends on at least two implementation choices.

- Are transactions removed from the transaction pool if they are rejected once due to congestion?
- Are RPC nodes filtering transaction according to current congestion before forwarding them to chunk producers?

Personally, I hope we can go for "Yes and Yes". RPC nodes should only accept a transaction if it has a high probability of being accepted. In case of rejection, users get immediate feedback. It's up to wallets to resend as often as appropriate. And chunk-producers should aggressively remove transaction that can't be included in chunks at the moment. For good UX, this rejection should also propagate back to the user through the RPC nodes. But if it's rare enough, we might get away without this in the initial deployment.

If we implement is as "No and No," then you are exactly right, we will generally queue more in the transaction pool until its size limit is reached.

"Yes and No" is also possible, which keeps the transaction pool short. But here it is even more important to propagate the rejection information back to the user, or else we have terrible UX since RPC nodes keep waiting for inclusion of the tx in a chunk when it was already dropped. But this back propagation is non-trivial to implement, given that multiple chunk producers are involved, so we might have to ship it later than the base congestion control mechanism.

"No and Yes" is also worth considering. Probably the easiest implementation without breaking UX. It's just odd because RPC nodes essentially control how large the transaction pool grows, rather than the chunk-producers that actually store it. But it could be a stepping stone towards "Yes and Yes".

Which ever way we go, I believe we don't need to change anything on the protocol level. And rejected transaction handling is fundamentally a client-side question that I believe should not be defined on the protocol level. In this discussion here, we only need to ensure the specification allows for decent client implementations.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+account is on a shard with a full delayed receipts queue.
+
+Forwarding of receipts that are not directly produced by a transaction, namely
+cross-contract calls and delegated receipts, is limited by the receivers overall
+congestion level. This includes the gas of delayed receipts and the gas of
+receipts that have not been forwarded, yet, due to congestion control
+restrictions. Additionally, the memory consumption of both receipt types can
+also cause congestion.
+
+This proposal replaces the local congestion control rules already in place.
```
akashin, line 30: Nit: I think it supercedes/extends the existing rules rather than fully replacing them (we still have transaction pool limit and delayed receipt queue limit).

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+Today, transactions are taken from the chunk producer's pool until `tx_limit` is
+reached, where `tx_limit` is computed as follow.
+
+```python
+# old
+tx_limit = 500 Tgas if len(delayed_receipts) < 20_000 else 0
+```
+
+We replace the formula for the transaction limit to depend on the
+`incoming_congestion` computed in the previous shard variable (between 0 and 1):
```
akashin, line 110: Do you mean computed in the previous chunk for this shard?

neps/nep-0539.md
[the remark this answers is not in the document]
akashin: Got it.

True, this would be quite tricky to get right with multi-hop transactions.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+
+## Consequences
+
+
+### 
+
+### Positive
+
+- Accepted transaction have lower latencies compared to today under congestion.
+- Storage and memory requirements on validator for storing receipts are bounded.
```
akashin, line 679: I don't understand why outgoing receipts buffers are bounded in size. Or at least I only understand why this is a case under the assumption that `filter_tx` above is working effectively. Otherwise, I can imagine a case when we keep accepting and processing transactions to shard X (due to a low value of `incoming_congestion`) that generate an outgoing receipt (as a cross-contract call) to a congested shard Y which can't all be processed timely and accumulate in unbounded fashion.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+
+For example, estimations on how much gas is burnt on each shard could help with
+better load distribution in some cases.
+
+We also forsee that the round-robin scheduling of shard allowed to forward even
+under full congestion is not perfect. It is a key feature to make deadlocks
+provably impossible, since every shard is guaranteed to make a minimum progress
+after N rounds. But it could be beneficial to allocate more bandwidth to shards
+that actually have something to forward, or perhaps it would be better to stop
+forwarding anything for a while. The current proposal allows chunk producers to
```
akashin, line 663: > The current proposal allows chunk producers to experiment with this without a protocol version change.

True, but practically speaking, chunk producers will use the default implementation in nearcore. So we need to make sure that the signals that we provide (e.g. `allowed_shard`) are enough to iterate to a reasonable policy that deals with the congestion pattern seen on mainnet today.

In my understanding, round-robin on average introduces N/2 = 3 blocks of latency (where N=6 is the number of shards) to cross-contract calls. This could be noticeable for some large contracts like `sweat` and `sweat.claim` that have a lot of communication and are located on different shards that are often congested.

I wonder if there is some way to make the behavior of `allowed_shard` smoother (e.g. allowing multiple senders in each round).

akashin: Actually, maybe under congestion, the 3 blocks delay is not that big of a deal on top of 20 blocks delay from the delayed receipts queue. Do simulations that you ran provide any insights here?

neps/nep-0539.md
jakmeier: Absolutely right, outgoing buffers are not hard-bounded to the memory size parameter we give it. We should probably clarify what we mean by a bound.

In simulations, we tried to push it past the soft-limit as much as possible. One example is in the spreadsheet. You can see 1.3 GB memory usage for the "Big Linear Imbalance" workload, with the limit at 1GB. This workload has 500 B / 60 Tgas receipts that indirectly produce 2 MB / 10 Tgas receipts.

But we tried with many other workloads, too. Every workload will fill up the system with some amount of receipts before it locks down new traffic from entering. And max memory usage then essentially depends on what memory the receipts in the system use to finish enough (child) receipts to allow more traffic to enter again.

This arguing only works if the system actually locks down every workload eventually. This is hard, if not impossible, to prove for general dynamic workloads. That's why we decided to simulate as much as we could.

In practice, we can still cause the memory requirements to 3-4x the limit. But all simulated workloads stabilized around that or lower. In comparison, today's "strategy" would lead to >500GB memory usage within 2000 blocks even for the aforementioned "Big Linear Imbalance" workload.

All that said, we are aware that for malicious workloads, the max in bytes could still be larger than we can handle. There are a couple malicious ideas we never tested. But from the beginning, we put malicious workloads as second priority behind getting a basic congestion control for regular users in place.

neps/nep-0539.md
jakmeier: Yes, we do think round-robin is the best choice in the short-term. We just also want to give the protocol team the ability to experiment with it and change it, once we see how congestion control manifests itself on mainnet.

There are main ideas why we think round-robin is best.

1. This is a last-resort measure. The linear reduction of forwarding limits is designed to prevent it from happening in the first place. If we hit it, we are badly congested.
2. If we hit this round-robin because of incoming congestion, we don't introduce extra latency. Anything forwarded is waiting behind a long delayed queue anyway, it might as well wait in outgoing buffers.
3. If we hit it because of outgoing or memory congestion, we are in trouble and are much more concerned about deadlock prevention than about minimizing latencies. The only alternative solution we know of that does not deadlock is a global transaction stop. But global-transaction stop leads to terrible fairness problems, where a single heavily used shard affects all shards and can even cause them to become congested as well due to reduced capacity on them.
4. Allowing multiple shards to send at once quickly leads to conflicts with the 300 Tgas max gas value of a receipt. We went with the simplest possible solution: a single allowed shard.

jakmeier: > Actually, maybe under congestion, the 3 blocks delay is not that big of a deal on top of 20 blocks delay from the delayed receipts queue. Do simulations that you ran provide any insights here?

As far as simulations go, it's kind of hard to get into this phase and stay in it. Once in round-robin, it often goes in-and-out of it every few blocks.

"Relayed hot" in the spreadsheet is probably the best example, which has some extra graphs at the bottom of the document. Using the limits we have in the proposal, the median transaction delay is at 19 blocks, the 90th percentile at 24 blocks. This includes 2 cross-shard hops but I'm pretty sure only one is operating under round-robin from time to time. So yeah, 3 blocks isn't a huge deal.

akashin:
### Technical summary
Over the last few months, NEAR mainnet has been experiencing a considerable increase in usage, which led to regular congestion on shards 2, 3 and 5.

Usually, transactions on NEAR execute within a few blocks and the latency for the end users is on the order of a few seconds. However, during these periods of congestion, users regularly had to wait more than 15 minutes for their transaction to be processed and sometimes the wait time was over 1 hour. This is a considerable degradation in the user experience and it needs to be addressed to make the users of NEAR happy.

This NEP introduces a number of mechanisms to control the order in which transactions/receipts are admitted, exchanged between shards and processed as well as a concrete policy that if carefully tuned to balance end-user latency with overall system throughput and peak memory overhead.

### Recommendation
I recommend approving this NEP.

### Benefits
- Establishes much tighter latency guarantees for admitted transactions (20 blocks compared to 1000s of blocks in the past)
- Actually deals with cross-shard congestion which lead to unbounded latencies in the past
- Thorough evaluation of the proposed solution on a range of relevant workloads

### Concerns
| - | - | - | - |   
| 1 | There is no hard-limit on the size of the buffers for receipts | The soft-limit has been shown to work effectively on simulated workloads. Moreover, this NEP strictly improves the limit that we have today | Resolved  |
| 2 | User transactions will fail due to being rejected | Users already have to deal with this today and it needs to be addressed regardless of this NEP | Resolved |

robin-near (commented):
Added some comments on clarifying the content - I'll take another look once that is addressed. Also https://github.com/jakmeier/NEPs/pull/3 for the typos and grammar nits that would introduce too much clutter if I left them as comments.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+  // ...all fields from ShardChunkHeaderInnerV2
+
+  congestion_info: CongestionInfo
+}
+```
+
+Where alternatives for congestion info are decribed in more detail below in the 
+`Efficiently computing the congestion information` section. 
+
+Depending on the alternative solution this adds between 5 and 48 bytes to the
```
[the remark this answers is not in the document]
jakmeier: Good idea. I moved it to the alternative sections, making it concrete in the specification part of the NEP.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+
+Chunk validators must validate that the two rules above are respected in a
+produced chunk.
+
+### Changes to chunk execution
+
+We change chunk execution to hold back receipt forwarding to congested shards.
+This has two effects. 
+
+1. It spreads the memory consumption from the congested shard to the shards
```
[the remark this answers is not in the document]
jakmeier: Yeah I like that wording much better :) Done as suggested.

neps/nep-0539.md
jakmeier: True, I've added a few words to describe what we keep and what we replace.

neps/nep-0539.md
jakmeier: Yes. The sentence should be fixed now to make more sense.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+2. When user demand is consistently higher than the system can handle, this
+   mechanism lets backpressure propagate shard-by-shard until the shards
+   responsible for accepting too many receipts are reached and transaction
+   filtering kicks in.
+
+To accomplish this, we add 3 new steps to chunk execution (enumerate as 1, 2, 6
+below) and modify how outgoing receipts are treated in the transaction
+conversion step (3) and in the receipts execution step (4).
+
+The new chunk execution then follows this order.
```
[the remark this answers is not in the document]
jakmeier: Sorry, there were too many rounds of renaming things and in the end it was rather inconsistent. I completely missed that, thanks for pointing it out. It should be fixed now.

neps/nep-0539.md
```diff
@@ -0,0 +1,741 @@
...
+```
+
+We replace the formula for the transaction limit to depend on the
+`incoming_congestion` computed in the previous shard variable (between 0 and 1):
+
+```python
+# new
+MIN_TX_GAS = 20 Tgas
+MAX_TX_GAS = 500 Tgas
+tx_limit = mix(MAX_TX_GAS, MIN_TX_GAS, incoming_congestion)
```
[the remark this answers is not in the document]
jakmeier: Yes that makes a lot of sense. I've done so now, please take another look.

jakmeier:
Thanks a lot to @akashin and @robin-near for taking the time to read through our proposal and giving valuable feedback! I really appreciate your expertise to ensure we end up with the best possible solution to move congestion control one step forward.

Sorry about the subpart quality in the grammar, and just in general. I thought we had the NEP cleaned up much better, otherwise I wouldn't have asked for SME reviews. I think we rushed a bit too much then, as we wanted to get the NEP processed started as soon as possible.

I have tried my best to fix it up now and added a new section about important concepts. Please, @robin-near, can you take another look? Let me know if something is still not well defined or not written clearly.

jakmeier:
Oh and in the time since the last changes, we added "missed chunks congestion" as an additional indicator. I have added it to the concepts section and to the "Changes to chunk execution" section.

It's a bit of a last minute change, not something we initially wanted to address. But for stateless validation, Near Protocol needs a way to limit incoming receipts even when chunks are missed. This NEP introduces all the required tools to solve that problem, so it seemed worth it to include. But if preferred by the working group, we could also separate it out as its own NEP that builds on top of congestion control.

@wacban, since you spear-headed and implemented this, can you please double-check that I got the details around missed chunk congestion right?

birchmd (commented):
As a working group member, I lean towards approving this proposal. While Near aims to scale such that it can handle the load users place on the network, it is still critical that Near remains usable under all loads. This congestion handling protocol accomplishes this goal while leaving room for transaction prioritization in the future.

One note I would like to make is that front-ends may need to update their retry logic to specifically handle the "transaction rejected due to congestion" error. This should be communicated clearly along with the protocol change which includes congestion control.

bowenwang1996:
As a working group member, I lean towards approving this NEP. It is a major step towards addressing congestion related stability issues and improving the user experience of NEAR.

neps/nep-0539.md
```diff
@@ -0,0 +1,788 @@
...
+
+Forwarding of receipts that are not directly produced by a transaction, namely
+cross-contract calls and delegated receipts, is limited by the receiver's overall
+congestion level. This includes the gas of delayed receipts and the gas of
+receipts that have not been forwarded, yet, due to congestion control
+restrictions. Additionally, the memory consumption of both receipt types can
+also cause congestion.
+
+This proposal extends the local congestion control rules already in place. It
+keeps the transaction pool size limit as it but replaces the old delayed receipt
```
mfornet, line 31:
```suggestion
keeps the transaction pool size limit as is but replaces the old delayed receipt
```

neps/nep-0539.md
```diff
@@ -0,0 +1,788 @@
...
+**Receipt limits** are measured in gas and size. Gas in this context refers to
+the maximum amount of gas that could be burn when applying the receipt. The
+receipt size is how much space it takes in memory, measured in bytes.
+
+The **congestion level** is an indicator between 0 and 1 that determines how
+strong congestion prevention measures should be. The maximum congestion measured
+is reached at congestion level 1. This value is defined separately for each
+shard and computed as the maximum value of the following congestion types.
+
+- **Incoming congestion** increases linearly with the amount of gas is in
```
mfornet, line 102:
```suggestion
- **Incoming congestion** increases linearly with the amount of gas in
```

mfornet (commented):
As a working group member I lean toward approving this proposal.

One observation. Say that an account on shard A wants to interact with a contract on shard B. Shard B is congested so that the transaction will be rejected. A (not so simple) alternative for the user is to route their transaction through a collocated contract on shard C (user -> A -> C -> B). The receipt between C and B will be delayed, but the transaction got in any way (is there any advantage to this?). Eventually, if everyone keeps doing this, C will get congested, and given there is nothing special about C, if these routing contracts are collocated on every shard and users do this, all shards will eventually get congested due to one app on shard B.

The situation described above exists beyond this proposal. I'm highlighting it since it will continue to exist.

neps/nep-0539.md
```diff
@@ -0,0 +1,788 @@
...
+    - Non-local receipts up to `outgoing_gas_limit[receipt.receiver]` for the
+      respective shard go to the outgoing receipts list of the chunk.
+    - (new) Non-local receipts above `outgoing_gas_limit[receipt.receiver]` for
+      the respective shard go to the outgoing receipts buffer.
+    - (new) For each receipt added to the outgoing buffer, add `receipt.gas()`
+      to `outgoing_congestion` and `receipt.size()` to `receipt_bytes` for
+      the local shard.
+5. Execute receipts in the order of `local`, `delayed`, `incoming`, `yield-resume time-out`.
+    - Don't stop before all receipts are executed or more than 1000 Tgas have
+      been burnt. Burnt gas includes the burnt gas from step 4.
```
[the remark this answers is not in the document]
jakmeier: Step 4 converts transaction to receipts. The send gas of the created receipts and all inner actions are burnt in that step.

jakmeier:
@robin-near You wrote that you want to take another look. Note that a WG meeting and the voting on the NEP is scheduled for this Friday. If you have any concerns about the proposal, please raise them as early as possible so they can be incorporated in the decision.

mm-near:
As a working group member I lean toward approving this proposal.

I have two meta comments:
* similar to @mfornet comment above - with this change, it will suddenly matter on which shard your account is located (if you happen to be collocated with some popular contract, more of your transactions will fail etc). It should be clearly stated in documentation.

* I'd suggest that the shard `congestion_level` info is clearly visible in the explorers. So that regular users can quickly see what's going on (and that not all of the system is under load).

flmel (approved):

PR #541, opened 2024-04-03
NEP-541: Transaction priority fee

bowenwang1996:
Proposal to add transaction priority fee to the protocol


neps/nep-0541.md
```diff
@@ -0,0 +1,196 @@
...
+## Alternatives
+
+One clear alternative is to not have FIFO processing at all and make all receipt processing purely based on priority. This, however, would create the problem that some receipts without priority can be indefinitely delayed when receipts with higher priority keeps coming in.
+Then there is the question of whether those receipts should pay for its storage in the state since there is no bound on when they may get processed.
+Furthermore, since there is no way to cancel a receipt after it is created, even if a user notices that their receipt is stuck in the queue for a long time and would like to cancel,
+they won't be able to do it.
+
+## Future possibilities
+
+This NEP should be combined with [NEP-513](https://github.com/near/NEPs/pull/539). They together redefine how congestion is handled and how users can still send transactions during congestion. It is possible to explore more complex mechanism on priority fees when there is congestion. For example, the protocol could require that transactions to a congested shard must attach a priority fee and even place a minimum on the priority fee based on previous chunk's priority fees.
```
jakmeier, line 139:
```suggestion
This NEP should be combined with [NEP-539](https://github.com/near/NEPs/pull/539). They together redefine how congestion is handled and how users can still send transactions during congestion. It is possible to explore more complex mechanism on priority fees when there is congestion. For example, the protocol could require that transactions to a congested shard must attach a priority fee and even place a minimum on the priority fee based on previous chunk's priority fees.
```

neps/nep-0541.md
```diff
@@ -0,0 +1,196 @@
...
+# gas limit refers to the gas limit assigned to process receipts with priority
+def process_priority_receipts(delayed_receipts, incoming_receipts, gas_limit):
+    if delayed_receipts.empty() and incoming_receipts.empty():
+        return
+    gas_used = 0
+    while gas_used < gas_limit:
+        delayed_receipt_head = if delayed_receipts.empty() {-Inf} else {delayed_receipts.top() }
+        incoming_receipt_head = if incoming_receipts.empty() {-Inf} else {incoming_receipts.top() }
+        receipt = None
+        if delayed_receipt_head.priority > incoming_receipts_head.priority:
```
jakmeier, line 92: Are prioritized delayed receipts executed before local receipts? What about prioritized local receipts?
Today's order among receipts is 
- local receipts
- delayed receipts
- new incoming receipts

Note that if we don't execute local receipts in the first chunk, they end up in the delayed receipts queue one chunk later.

neps/nep-0541.md
```diff
@@ -0,0 +1,196 @@
...
+A max heap is created in the state for storing priorities of delayed receipts. More specifically, we add a trie column `RECEIPTS_PRIORITY_QUEUE: u8 = 13`. Each value in this heap is of the form
+```rust
+struct ReceiptPriority {
+    /// index of the corresponding receipt in the delayed receipts queue. Use u32 here to save some space in the state
+    receipt_index: u32
+    /// priority of the receipt
+    priority: u64
+}
+```
+Since each value is quite small and we don't expect the delayed receipt queue to be extremely long (even if there are 100k delayed receipts, it is only 1.2MB in size), and it only needs to be read and write once per chunk, this heap can be stored as a single value in the state.
```
jakmeier, line 77: In principle I agree. But one question for stateless validation, do these trie changes end up in the state witness?
Because 1.2MB every chunk does not sound completely trivial.

neps/nep-0541.md
jakmeier, line 139: Oh and how are we dealing with priority in the outgoing buffers that NEP-539 currently proposes?

I would probably suggest that, at least initially, draining the outgoing buffers should follow a strict FIFO order. The priority already helped to get inside faster, so maybe the "cross-shard delay" can be fair.

Alternatively, we can add more priority queues (one extra for each receiving shard, per shard) and give truly fast speed even to cross-contract calls during congestion. It would certainly be a better user experience, since otherwise one can still wait a long time for a cross-contract call during congestion, no matter how much one pays.

I just feel perhaps an iterative approach would be better than changing so much all at once, without the experience of how congestion control works in practice.

jakmeier (commented):
Looks pretty solid already. I left some thought in the comments as  was reading through.

walnut-the-cat:
how do we determine 'priority fee'? Is the amount preset by protocol? Does a user have to guess and attach arbitrary amount, hoping it would be high enough?

neps/nep-0541.md
bowenwang1996: yes they are executed before local receipts. The way to think of it is as follows: priority execution always happens first and regular execution happens. During regular execution, we preserve the order of execution we do today (local receipts, delayed receipts, and incoming receipts). We can also change the order within regular execution, but that is orthogonal to this proposal.

neps/nep-0541.md
```diff
@@ -0,0 +1,196 @@
...
+
+A max heap is created in the state for storing priorities of delayed receipts. More specifically, we add a trie column `RECEIPTS_PRIORITY_QUEUE: u8 = 13`. Each value in this heap is of the form
+```rust
+struct ReceiptPriority {
+    /// index of the corresponding receipt in the delayed receipts queue. Use u32 here to save some space in the state
+    receipt_index: u32
+    /// priority of the receipt
+    priority: u64
+}
+```
```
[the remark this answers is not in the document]
bowenwang1996: yes

neps/nep-0541.md
bowenwang1996: yeah that is a very good point. On the other hand, it does feel to me quite wasteful to store many such small values in the state, which would also result in many state reads.

neps/nep-0541.md
bowenwang1996: yeah I think that makes sense

neps/nep-0541.md
bowenwang1996: @jakmeier I thought about it a bit more and came up with a different idea. Instead of trying to maintain a heap within the trie, we can use the priority of receipts as trie keys under a new trie column. Since the key is a u64, it means there are at most 16 trie nodes for each value. It makes proof quite easy and works nicely with stateless validation. In addition, it is quite difficult to attack because one needs to attach priority fees to get a new value created, so in practice it would most likely not be a concern. After we switch to stateless validation, this will all be memory accesses so it's definitely not going to be a problem.

jakmeier: Yeah I like that idea. We can just use the merkle trie key order to sort things, no need for a priority heap. :)

Perhaps we could even round priorities to `u16` if we want 4 instead of 16 trie nodes per receipts. (~16k levels of priority seem enough to me, but I could be wrong)
That would also make attacks more expensive, as the priority increasing steps would need to be higher.

But one question. I'm reading your proposal as the key being `TrieKey::PriorityReceipts{priority: u64}`. But what is the value? I guess `{ receipts: Vec<CryptoHash> }` would make sense, storing receipt ids to lookup the value in the delayed queue. Is that what you have in mind?

walnut-the-cat:
> > how do we determine 'priority fee'? Is the amount preset by protocol? Does a user have to guess and attach arbitrary amount, hoping it would be high enough?
> 
> It is an arbitrary amount decided by a user

It's a good starting point, but it may not be intuitive enough. How can a user ensure that their txn will be 'prioritized' with the minimum amount of premium? Or is that out of our scope?

birchmd:
The protocol working group met today and had a lively discussion about this NEP along with #539 

The primary concern raised with this proposal is if there are no protocol limits on priority fees then validators can collude to create an effective minimum fee (by censoring transactions that are below their desired fee). But we think this can be addressed by combining the idea of a priority fee with the congestion metrics present in #539 .

It would work something like this. Below a certain threshold the system is considered "not congested" and receipt priority is ignored. Beyond this threshold, the priority is used as the back-pressure mechanism by making the minimum required priority higher as the queue fills up. If a receipt does not have a high enough priority to be added to the target shard's incoming queue, then it remains in the outgoing queue of its source shard. If a new transaction does not have a high enough priority fee so that the priority of its initial receipt is high enough to be added to the shard's queue then it is rejected. To prevent receipts from being stuck forever, the priority of receipts in outgoing queues can increase over time so that eventually either the congestion will alleviate or the priority will be high enough that the target shard will accept it.

As part of this proposal, it was suggested that 100% of the priority fee be burned so that there is not an incentive for validators to artificially keep the system in a congested state.

jakmeier:
@birchmd I am not quite sure I understand the concerns you a trying to address.

> validators can collude to create an effective minimum fee

If validators collude, the entire system is compromised, isn't it? Maybe I am missing what attacker model did you assumed for your discussion.

> As part of this proposal, it was suggested that 100% of the priority fee be burned so that there is not an incentive for validators to artificially keep the system in a congested state.

But then chunk producer have no incentive to actually include higher priority fee transactions over normal fee transactions they simply "like" more for one reason for another.

Keep in mind, chunk producer ultimately hold the power over which transactions are included on chain at all. If we don't give any of the priority fees to them, they can extract that value in other ways. For example, they could offer subscriptions to prioritize transactions from certain accounts and make extra profit this way. With the right pricing, this would also be cheaper for users, so it's a win-win situation for chunk proposers and transaction sender, at a loss for the protocol.

If we give rewards to the chunk producer, they are more incentivised to just follow the rules without off-chain shenanigans.

birchmd:
> If validators collude, the entire system is compromised, isn't it?

The level of cooperation between validators in this case is lower than something extreme like including an incorrect state transition. It's more like [price fixing](https://www.investopedia.com/terms/f/fixing.asp) in oligopoly situations. Each validator can notice the minimum priority fee being accepted by other validators and raise the minimum fee they accept to be in line. In this way the validators can come to an unspoken agreement to profit at users' expense.

> chunk producer ultimately hold the power over which transactions are included on chain at all.

Yes, this is the whole idea with price fixing we are thinking about. The validators would simply censor incoming transactions lower than the minimum priority fee they have chosen; including in the case where such a fee is not needed to control congestion.

> off-chain shenanigans

Yes, this also came up during our discussion. And I agree we do not want any kind of secondary market on blockspace. But at the same time, I think it is important to recognize that if validators profit off congestion then they will have an incentive to create congestion (perhaps even artificially, by sending the transactions themselves). Instead of giving no part of the priority fee to validators, another idea we had was to make the proportion of the priority fee the validator receives be a function with diminishing returns (eg log or sqrt) so that maybe there would be an optimal amount of congestion for validators. More detailed analysis of this is needed to see if it makes sense.

We also talked about how Ethereum faced this same issue with their gas price auction and they moved to a base-fee + tip model to make gas price fixing by validators less of a concern. We can't directly apply Ethereum's unsharded, synchronous execution setting directly to Near of course, but I think it does provide strong evidence that a pure auction is not the ideal model for users.

neps/nep-0541.md
bowenwang1996: Value is a `64` (index of delayed receipt)

mfornet:
> If validators collude, the entire system is compromised, isn't it?

To add to @birchmd's point, validators don't need to collude. They can raise the minimum accepted fee, and as long as enough of them have done that, submitters using a low fee will notice a lower throughput.

> But then chunk producer have no incentive to actually include higher priority fee transactions over normal fee transactions they simply "like" more for one reason for another.

One of the ideas discussed is that the system will operate at a desired capacity (which could be half of the total capacity). To include a transaction that uses above half of the capacity, a fee needs to be attached. The required fee will increase exponentially as more transactions go beyond the desired capacity.

garikbesson:
Hey @bowenwang1996, do you still want to proceed with this NEP?

garikbesson:
Because there has been no activity from the author for more than a year, I'm marking the proposal as RETRACTED. Please feel free to open it again if you'd like to continue working on it.

PR #545, opened 2024-05-08
Transactions.md redirect

ewiner:
https://nomicon.io/DataStructures/Transaction is blank.  I'm adding a link to the Transactions spec page.

We also could delete this page but I'm choosing not to because:
1. There might be links to this page that would break.
2. Many chains consider a Transaction to be a core data structure - so this blank page is the first place I looked to learn about the NEAR tx format.  I think having a link here that points people in the right direction could be nice for people new to NEAR.


bowenwang1996 (approved):
Thanks!

PR #547, opened 2024-06-13
Add information about state witness size limits to NEP 509

jancionear:


neps/nep-0509.md
```diff
@@ -128,6 +128,51 @@ TODO: This is essentially going to be describing the exact structure of `ChunkSt
 
 The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.]
 
+## State witness size limits
+
+Some limits are introduced to keep the size of `ChunkStateWitness` reasonable.
```
wacban, line 133: nit: I would rephrase to something like: "A number of new limits will be introduced in order to ..."

neps/nep-0509.md
```diff
@@ -128,6 +128,51 @@ TODO: This is essentially going to be describing the exact structure of `ChunkSt
...
+`ChunkStateWitness` contains all the incoming transactions and receipts that will be processed during chunk application and in theory a single receipt could be tens of megabatytes in size. Distributing a `ChunkStateWitness` this large would be troublesome, so we limit the size and number of transactions, receipts, etc. The limits aim to keep the total uncompressed size of `ChunkStateWitness` under 16MiB.
+
```
wacban, line 135: Can you also mention that this is a breaking change but that we don't predict it to affect more than x% of transactions and receipts? Can you also mention that the contract developers can work around the limits by reducing how much data is accessed in a single call (am I getting this right?)?

neps/nep-0509.md
```diff
@@ -128,6 +128,51 @@ TODO: This is essentially going to be describing the exact structure of `ChunkSt
...
+In total that gives 2 MiB + 500 KiB + 7MB + 5*100 KiB + 4.5 MiB ~= 14 MiB of maximum witness size
+
+### Validating the limits
+
+Chunk validators have to verify that chunk producer respected all of the limits while producing the chunk. This means that validators also have to keep track of recorded storage proof by recording all trie accesses and they have to enforce the limits.
+If it turns out that some limits weren't respected, the validators will generate a different result of chunk application and they won't endorse the chunk.
+
+### Missing chunks
+
+When a chunk is mising on a shard, this shard will receive receipts from more than one block height. This could lead to large `source_receipt_proofs` so a mechanism is added to reduce the impact. If there are two or more missing chunks in a row,
```
wacban, line 173: nit: When a shard is missing some chunks, the following chunk on that shard will receive receipts from multiple blocks.

wacban (approved):
LGTM

neps/nep-0509.md
```diff
@@ -128,6 +128,51 @@ TODO: This is essentially going to be describing the exact structure of `ChunkSt
...
+
+The limits are:
+* `max_transaction_size = 1.5 MiB`
+    * All transactions must be below 1.5 MiB, otherwise they'll be considered invalid and rejected.
+    * Previously was 4MiB, now reduced to 1.5MiB
+* `max_receipt_size - 4 MiB`:
+    * All receipts must be below 4 MiB, otherwise they'll be considered invalid and rejected.
+    * Previously there was no limit on receipt size. Set to 4MiB, might be reduced to 1.5MiB in the future to match the transaction limit.
+* `combined_transactions_size_limit - 2 MiB`
+    * Hard limit on total size of transactions from this and previous chunk. `ChunkStateWitness` contains transactions from two chunks, this limit applies to the sum of their sizes.
```
dndll, line 148: Is this calldata? How does this work in practice? Does this mean that If we have 2 separate transactions at `max_transaction_size`,  we have to wait since they cannot be included consecutively, since there is 3mb of calldata? What happens to the transactions that are larger than that, do they get dropped?

neps/nep-0509.md
```diff
@@ -128,6 +128,51 @@ TODO: This is essentially going to be describing the exact structure of `ChunkSt
 
 The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.]
 
+## State witness size limits
+
+Some limits are introduced to keep the size of `ChunkStateWitness` reasonable.
+`ChunkStateWitness` contains all the incoming transactions and receipts that will be processed during chunk application and in theory a single receipt could be tens of megabatytes in size. Distributing a `ChunkStateWitness` this large would be troublesome, so we limit the size and number of transactions, receipts, etc. The limits aim to keep the total uncompressed size of `ChunkStateWitness` under 16MiB.
```
firatsertgoz, line 134: Do we have any hard evidence that can back this claim? I think it would be great to see the limits of the current system in this NEP. It would greatly increase the justification.

neps/nep-0509.md
```diff
@@ -128,6 +128,51 @@ TODO: This is essentially going to be describing the exact structure of `ChunkSt
...
+    * It's a soft limit, which means that the total size of storage proof could reach 7 MB (2.99MB + one receipt which generates 4MB of storage proof)
+    * Due to implementation details it's hard to find the exact amount of storage proof generated by a receipt, so an upper bound estimation is used instead. This upper bound assumes that every removal generates additional 2000 bytes of storage proof, so receipts which perform a lot of trie removals might be limited more than theoretically applicable.
+* `outgoing_receipts_usual_size_limit - 100 KiB`
+    * Limit on the size of outgoing receipts to another shard. Needed to keep the size of `source_receipt_proofs` small.
+    * On most block heights a shard isn't allowed to send receipts larger than 100 KiB to another shard.
+* `outgoing_receipts_big_size_limit - 4.5 MiB`
+    * On every block height there's one special "allowed shard" which is allowed to send larger receipts, up to 4.5 MiB in total.
+    * A receiving shard will receive receipts from `num_shards - 1` shards using the usual limit and one shard using the big limit.
```
firatsertgoz, line 162: How is this shard determined? Is it going to be randomly picked every block? Is it going to be a gateway shard that allows for big receipts?

neps/nep-0509.md
jancionear: If someone submits two transactions at `max_transaction_size` then one of them will be included and the other one will stay in the transaction pool.

It would look like this:
* User submits 2 transactions, each of them at `max_transaction_size`
* Height 1 - transaction 1 is included
* Height 2 - no transaction is included because the previous chunk had 1.5MiB of transactions
* Height 3 - transaction 2 is included, as the last chunk didn't contain any transactions, so it fits under the limit.

neps/nep-0509.md
jancionear: We observed missing chunks when witness size reached ~16MB. @shreyan-gupta then implemented a new distribution mechanism, but I'm not sure if there have been any tests of the throughput with the new mechanism.

I believe we need some limits, for example I was able to create a 300TGas receipt which produced 24MB of storage proof. If there are 3 such receipts in a chunk, then that would add up to 72MB of state witness size. Apart from that we could have a lot of incoming receipts, there is no receipt size limit, but assuming that each receipt is 5MiB in size, and we received receipts from all 6 shards, that would add up to 30MB in incoming receipts. This means that without limits witness size could easily reach 100MB of size. We had problems with 16MB, I'm sure 100MB would be even worse.

neps/nep-0509.md
jancionear: It's the `allowed_shard` from congestion control NEP. I'll add a paragraph about it.

neps/nep-0509.md
jancionear: Added a section about breaking changes. I prepared a list of transactions larger than 1.5MiB. Contract developers can take a look at this list and see if their contract will be affected by the new limits.

PR #548, opened 2024-06-14
future possibilities and consequences

wacban:


neps/nep-0509.md
```diff
@@ -473,34 +473,34 @@ This makes chunk validator selection algorithm correctness critical for the secu
...
 ### Neutral
 
-- n1
+- The current approach to resharding will need to be revised to support generating state witness. 
+- The security assumptions will change. The responsibility will be moved from block producers to chunk validators and the security will become probabilistic. 
 
 ### Negative
 
-- n1
+- The hardware requirements will increase. In particular the network bandwidth and memory requirements will be higher.
```
walnut-the-cat, line 501: hmm wonder if we should mention potential decrease of hardware requirement (disk requirement?) as well in positive section

neps/nep-0509.md
```diff
@@ -473,34 +473,34 @@ This makes chunk validator selection algorithm correctness critical for the secu
...
-- n1
+- The current approach to resharding will need to be revised to support generating state witness. 
+- The security assumptions will change. The responsibility will be moved from block producers to chunk validators and the security will become probabilistic. 
 
 ### Negative
 
-- n1
+- The hardware requirements will increase. In particular the network bandwidth and memory requirements will be higher. 
+- Additional limits will be put on the size of transactions, receipts and, more generally, cross shard communication.
+- The dependency on cloud state sync will increase the centralization of the blockchain. This will be resolved separately by the decentralized state sync.
```
walnut-the-cat, line 503: should we specify what it starts having a dependency on state sync?

Longarithm: Well, it's mentioned above on explanation of validator roles.

neps/nep-0509.md
Longarithm: +1 - more specifically, small-stake nodes won't have to store state and apply chunks for all shards as they currently do.

wacban:
@walnut-the-cat and @Longarithm I added the suggestion and I'm going to merge it now. Feel free to add any more comments and suggestions and I'll just do it in a follow up PR.

PR #555, opened 2024-07-22
[NEP-518] Web3-Compatible Wallets Support

birchmd:
This PR contains the description originally posted in #518 

--

**Status**: Review

SME reviews:
* [ ] @akhi3030 
* [ ] @nagisa


neps/nep-0518.md
```diff
@@ -0,0 +1,337 @@
...
+## Future Opportunities
+
+- **Expansion of Ethereum Standards**: A key future development involves simulating other Ethereum standards such as ERC-721, ERC-1155, or ERC-3643. This would enhance NEAR's frontend compatibility with various Ethereum use cases.
+- **Account Abstraction and Extended Features**: This proposal lays the groundwork for account abstraction, potentially leading to the introduction of advanced features in the Wallet Contract. These could include native support for multi-signatures, time-locks, social recovery, and alternative signature algorithms compatible with secure elements in widespread devices.
+- **Aurora Integration**: Exploring integration with Aurora to enable interactions with Aurora-hosted contracts through NEAR, utilizing Ethereum transaction formats and processes.
+- **Bridging and Oracle Onboarding**: This model could facilitate simplified bridging and oracle integration, contributing to the expansion of NEAR's infrastructure.
+
+## Scope of the Project
+
+The scope of this project is extensive, encompassing several critical components:
```
nagisa, line 303: Nit: NEP is probably not the best place for project sizing and tracking. One thing to keep in mind is that `nearcore` is just *an* implementation of the NEAR Protocol, and NEPs are meant to describe the changes to *the protocol*, rather than a specific implementation of it.

neps/nep-0518.md
```diff
@@ -0,0 +1,337 @@
...
+
+4. **Transaction Relayer (TR)**: Ethereum wallets cannot generate NEAR-compatible transactions. Instead, Ethereum-compatible transactions produced by them are processed by the Transaction Relayer, which embeds it into a NEAR transaction and forwards it to the user’s Wallet Contract. Again, while operated in conjunction with the RPC-Translator, the Transaction Relayer is distinct in its role.
+
+5. **[COMPLEX] Ethereum Translation Contract (ETC)**: Addressing the incompatibility between NEAR's human-readable account names and Ethereum's cryptographic hash-based addresses, the ETC functions as an on-chain mapping system. This system records NEAR-compatible input values (like NEAR account names and smart contract function names) and maps them to their corresponding Ethereum-compatible cryptographic hashes. This feature is vital for preserving familiar user experiences, such as recognizing ft_transfer operations in NEP-141 as fungible token (ERC20) transfers, rather than generic contract calls, and ensuring that fungible token balances are displayed in Web3 wallets.
```
nagisa, line 59: Since we're in the normative section of the NEP here, is the list of translations a part of the NEAR protocol? Is the ETC itself a part of the NEAR protocol? More generally "such as", "for instance" and similar wordings read as somewhat suspect in the normative section of the protocol specification.

nagisa (commented):
Did a first read-through. Maybe its just on me for not being familiar with ETH, that I had to re-read many parts of the text multiple times over to get any semblance of comprehension, so bear with me…

Regardless I think there are some changes that can be made to the NEP text so that it is more approachable to the more casual readers like myself:

1. Add a single glossary section where you'd define all terms and abbreviations and make sure that every term with a remote chance of ambiguity would be defined there;
2. Speaking of terms, it would help comprehension if the terms were disambiguated a little more, especially when they refer to specific fields within an ETH-formatted transaction vs. a NEAR-formatted transaction. So for instance all ETH-specific terms could be prefixed with `Eth` or whatever. Otherwise readers are forced to try and infer the meaning of many words in the text either from their (possibly spotty) knowledge or from the context. (Although not mentioned in the text, I at some point realized that capitalized terms generally refer to ETH transaction fields, whereas the lowercase ones would be from the NEAR-land. But then some ETH functions are lowercase and NEAR actions are upper-cased...)
3. Many of the "what" would benefit from substantiation by a "why" even if by a link to text elsewhere in the NEP or the broader internet that would explain the reasoning or logic. This was especially apparent in the latter sections describing the behaviour of separate components (what is ethereum decimal count that's forcing RPC to truncate? Why does RPC always estimate 30M? Why are self-transactions special-cased?)
4. As I was reading this, I felt that large chunks of the NEP text are not at all a part of the NEAR **Protocol** and rather described either how a component remotely related to a Protocol feature would work or be implemented. If nothing else it would be great to clearly delineate what parts are actually protocol things, and which ones are supporting background information.

neps/nep-0518.md
```diff
@@ -0,0 +1,337 @@
...
+- `Chain ID`,
+- `Data field`,
+- `Signature` (yielding the public key and `From` address).
+
+For function calls to a smart contract, the data field contains:
+
+- The `Function Selector`,
+- Function call parameters.
+
+Both `To` and `From` addresses, encoded in hexadecimal and prefixed with “0x”, comprise 20 bytes. The `From` address is the right-most 20 bytes of the **Keccak-256** hash of the binary public key of the transaction sender (an **EOA**, externally owned address). The To address signifies either the recipient EOA or an on-chain contract address.
```
nagisa, line 131: Side note: I don't see https://github.com/near/NEPs/issues/518#issuecomment-1853942353 having been addressed in the NEP (collection of NEPs should be in some ways as self-contained as feasible; keeping motivation as a comment somewhere is not discoverable to most readers).

birchmd:
Thanks for the detailed review @nagisa ! I agree the idea proposal didn't end up working all that well as an NEP after all. I'll spend some time giving it a pretty significant re-write. To confirm that I understand your comments correctly, are you are suggesting the following changes?

1. Write the document from a Near perspective, defining the EVM concepts clearly and assuming the reader is no familiar with them.
2. Remove (perhaps replace with a link to the original 518 issue) much of the descriptions of components that are not direct protocol changes (i.e. RPC, wallet selector).
3. Ensure remaining descriptions (of the wallet contract and protocol-level handling of eth-implicit accounts) are specifications with precise descriptions of behavior (though the reasons for this behavior can still be motivated by higher level discussions of how the wallet contract fits into the broader solution).

Let me know if I have misunderstood the changes you would like to see.

nagisa:
> Write the document from a Near perspective, defining the EVM concepts clearly and assuming the reader is no familiar with them.

Yes, please.

> Remove (perhaps replace with a link to the original 518 issue) much of the descriptions of components that are not direct protocol changes (i.e. RPC, wallet selector).

Yes.

> Ensure remaining descriptions (of the wallet contract and protocol-level handling of eth-implicit accounts) are specifications with precise descriptions of behavior (though the reasons for this behavior can still be motivated by higher level discussions of how the wallet contract fits into the broader solution).

Sure, so long as those higher level discussions are either part of this NEP or are referenced here in some way or form.

birchmd:
@nagisa I started doing the rewrite. Some sections are still left with `TODO`, but I wanted to give you a chance to take a look at what I have so far! I'll work on filling in the remaining sections if you think the style and structure of the rewrite looks good.

akhi3030:
Thank for you for the initial review @nagisa.  I agree with your assessment and that a rewrite is needed. 

Thank you for starting the rewrite @birchmd.  To set expectations, I will be on PTO all of next week.  So if the NEP is not quite ready to reviewed before EOW, then my review will be delayed.  I hope the timeline works fine?

birchmd:
@akhi3030 thanks for the heads up. I probably will not have time to finish this by EOW. But each section is reasonably self-contained, so I encourage you to review what I have already written so far this week. Then I can address your comments and you can review those changes along with the rest of the new sections after you return from PTO.

neps/nep-0518.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Specification
+
+### Eth-implicit accounts
+
+**Definition**: An eth-implicit account is a top-level Near account with ID of the form `^0x[a-f0-9]{40}$` (i.e. 42-characters with `0x` as a prefix followed by 40 characters of hex-encoded data which represents a 20-byte address).
+
+Eth-implicit accounts, as the name suggests, are implicit accounts on Near. This means MUST be automatically created by a `Transfer` action if the target account ID does not exist. This includes being created even if the amount being transferred is zero (per the prior [NEP on zero balance accounts](https://github.com/near/NEPs/blob/master/neps/nep-0448.md)). Eth-implicit accounts represent an Ethereum **EOA** and therefore are controlled via the Ethereum private key corresponding to the address contained in the account ID (see Appendix B for a description of how 20-byte addresses are derived from a private key in the Ethereum ecosystem). To enable this control, eth-implicit accounts all have a smart contract deploy to them called the wallet contract (specification in the next section).
```
akhi3030, line 38:
```suggestion
Eth-implicit accounts, as the name suggests, are implicit accounts on Near. This means that if the traget account ID does not exist during a `Transfer` action, then they MUST be automatically created. This includes being created even if the amount being transferred is zero (per the prior [NEP on zero balance accounts](https://github.com/near/NEPs/blob/master/neps/nep-0448.md)). Eth-implicit accounts represent an Ethereum **EOA** and therefore are controlled via the Ethereum private key corresponding to the address contained in the account ID (see Appendix B for a description of how 20-byte addresses are derived from a private key in the Ethereum ecosystem). To enable this control, eth-implicit accounts all have a smart contract deploy to them called the wallet contract (specification in the next section).
```

akhi3030, line 38:
```suggestion
Eth-implicit accounts, as the name suggests, are implicit accounts on Near. This means MUST be automatically created by a `Transfer` action if the target account ID does not exist. Eth-implicit accounts represent an Ethereum **EOA** and therefore are controlled via the Ethereum private key corresponding to the address contained in the account ID (see Appendix B for a description of how 20-byte addresses are derived from a private key in the Ethereum ecosystem). To enable this control, eth-implicit accounts all have a smart contract deploy to them called the wallet contract (specification in the next section).
```

akhi3030: Is this information crucial to be included?  If it is, then why?  If not, then I propose to remove it.

akhi3030: Minor grammar improvement.

neps/nep-0518.md
```diff
@@ -0,0 +1,158 @@
...
+
+Currently, the Ethereum ecosystem is a leading force in the smart contract blockchain space, boasting a large user base and extensive installations of Ethereum-compatible tooling and wallets. However, a significant challenge arises due to the incompatibility of these tools and wallets with NEAR Protocol. This incompatibility necessitates a complete onboarding process for users to interact with NEAR contracts and accounts, leading to confusion, decreased adoption, and the marginalization of NEAR Protocol.
+
+Implementing Web3 wallet support in NEAR Protocol, with an emphasis on user experience continuity, would significantly benefit the entire NEAR Ecosystem.
+
+## Specification
+
+### Eth-implicit accounts
+
+**Definition**: An eth-implicit account is a top-level Near account with ID of the form `^0x[a-f0-9]{40}$` (i.e. 42-characters with `0x` as a prefix followed by 40 characters of hex-encoded data which represents a 20-byte address).
```
akhi3030, line 36: can this scheme be extended in future if we need to add support for different types of account ids?

akhi3030 (commented):
incomplete review... just covered a small section.

neps/nep-0518.md
birchmd: I thought it would be good to call out explicitly the interaction with zero balance accounts. Why do you think it is better if we do not include this information?

neps/nep-0518.md
birchmd: It's not clear how this could be extended to support other types of account IDs.

During the design we thought about having a chain-specifying prefix (e.g. `eth`), or having the addresses be sub-accounts of a chain-specifying account (again e.g. `eth`), but we ultimately chose to just have the account ID be equal to the address. The reasons for this choice are (1) with the number of EVM chains that exist now it doesn't really make sense to have a chain specifier with an address because they exact same key could control the same address on multiple chains (e.g. Etheruem, Aurora, Arbitrum, BNB, etc) -- including now the Near chain with this proposal; (2) it simplifies how users can look up their accounts in Near explorers (they just search for their address exactly as they would in etherscan for example, though with the caveat that they cannot use the address checksum casing because Near requires account IDs to be lowercase).

neps/nep-0518.md
birchmd: Fixed in https://github.com/near/NEPs/pull/555/commits/da5b9024950a7dde5133b388ca7015e1da10cf5a

birchmd:
@nagisa @akhi3030 This PR is ready for further review

neps/nep-0518.md
akhi3030: It felt a bit out of place to me.  I wasn't sure if there are some subtle issues that I didn't understand.  If you think it is good to clarify this, then it is fine.

neps/nep-0518.md
```diff
@@ -0,0 +1,221 @@
...
+
+When an eth-implicit account is created the runtime MUST set the contract code equal to specific "magic bytes". These bytes come from a UTF-8 encoded string which is equal to the constant `near` appended with the base-58 encoding of the sha2 hash of the wallet contract code. This constant allows the contract runtime to lookup the full contract code without needing it to be stored multiple times in the state. As well as being more efficient for the protocol, this setting the code equal to a hash of the contract instead of the contract itself is what keeps the storage requirements of a new eth-implicit account small enough to be a zero balance account.
```
akhi3030, line 40:
```suggestion
When an eth-implicit account is created the runtime MUST set the contract code equal to specific "magic bytes". These bytes come from a UTF-8 encoded string which is equal to the constant `near` appended with the base-58 encoding of the sha2 hash of the wallet contract code. This constant allows the contract runtime to lookup the full contract code without needing it to be stored multiple times in the state. As well as being more efficient for the protocol, setting the code equal to a hash of the contract instead of the contract itself keeps the storage requirements of a new eth-implicit account small enough to be a zero balance account.
```

minor nits.

neps/nep-0518.md
```diff
@@ -0,0 +1,221 @@
...
+
+The wallet contract has two state variables: the nonce, a 64-bit number; and a boolean flag indicating if a transaction is currently in progress. As with nonce values on Near access keys, the purpose of the wallet contract nonce is to prevent replaying the same Ethereum transaction more than once. The boolean flag prevents multiple transactions from being in-flight at the same time. The reason this is needed is because of the asynchronous nature of Near as compared with the synchronous nature of the **EVM**. On Ethereum if two transactions are sent (they must have sequential nonces per the Ethereum standard) then the all actions of the first will always happen before all the actions of the second. However, on Near there is no guarantee of the order of execution for receipts in different shards. Therefore, the only way to ensure that all actions from the first transaction are executed before all the actions of the second transaction is to prevent the second transaction from starting its execution until after the first one entirely finishes.
```
akhi3030, line 61:
```suggestion
The wallet contract has two state variables: the nonce, a 64-bit number; and a boolean flag indicating if a transaction is currently in progress. As with nonce values on Near access keys, the purpose of the wallet contract nonce is to prevent replaying the same Ethereum transaction more than once. The boolean flag prevents multiple transactions from being in-flight at the same time. The reason this is needed is because of the asynchronous nature of Near as compared with the synchronous nature of the **EVM**. On Ethereum if two transactions are sent (they must have sequential nonces per the Ethereum standard) all actions of the first will happen before all actions of the second. However, on Near there is no guarantee of the order of execution for receipts in different shards. Therefore, the only way to ensure that all actions from the first transaction are executed before all the actions of the second transaction is to prevent the second transaction from starting its execution until after the first one entirely finishes.
```

neps/nep-0518.md
```diff
@@ -0,0 +1,221 @@
...
+
+#### Details of `rlp_execute`
+
+This function is named after the **RLP** standard in Ethereum. In particular, the `tx_bytes_b64` argument is be parsed into bytes from base-64 then those bytes are parsed into structured data assuming it is RLP encoded. The structure the data is parsed into is an Ethereum transaction. Ethereum transactions can have multiple different forms since the Ethereum protocol has evolved over time (there are "legacy" transactions, [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) type transactions, [EIP-2930](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2930.md) transactions). All these different forms are supported by the wallet contract (they are distinguished based on the "type byte" which starts the encoding as per [EIP-2718](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2718.md)) and are ultimately all transformed into a common data structure with the following fields:
```
akhi3030, line 65:
```suggestion
This function is named after the **RLP** standard in Ethereum. In particular, the `tx_bytes_b64` argument is parsed into bytes from base-64; then the bytes are parsed into structured data assuming it is RLP encoded; then the structured data is parsed into an Ethereum transaction. Ethereum transactions can have multiple different forms since the Ethereum protocol has evolved over time (there are "legacy" transactions, [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md) type transactions, [EIP-2930](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2930.md) transactions). All these different forms are supported by the wallet contract (they are distinguished based on the "type byte" which starts the encoding as per [EIP-2718](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2718.md)) and are ultimately all transformed into a common data structure with the following fields:
```

neps/nep-0518.md
```diff
@@ -0,0 +1,221 @@
...
+##### Converting Ethereum transaction into Near actions
+
+Each Ethereum transaction is converted to a single Near action (batch transactions are not supported) based on the `data` field. Following the Solidity convention of the first four bytes of the data being a **method selector**, the wallet contract checks the first four bytes of the `data` to see if it is a known Near action. The **method selectors** for Near actions supported by the wallet contract are determined by mapping the actions to an equivalent Solidity function signature as follows:
+
+- `functionCall(string,string,bytes,uint64,uint32)`
+- `transfer(string,uint32)`
+- `addKey(uint8,bytes,uint64,bool,bool,uint128,string,string[])`
+- `deleteKey(uint8,bytes)`
+
+Note that the `uint32` fields in `functionCall` and `transfer` contain the amount of yoctoNEAR tha cannot be included in the Ethereum transaction's `value` field due to the difference in decimal places (see **Wei** definition in Appendix A), therefore the value there is always less than `1_000_000` so it will easily fit in a 32-bit number. These type signatures then hash to the **method selectors**:
```
akhi3030, line 99:
```suggestion
Note that the `uint32` fields in `functionCall` and `transfer` contain the amount of yoctoNEAR that cannot be included in the Ethereum transaction's `value` field due to the difference in decimal places (see **Wei** definition in Appendix A), therefore the value there is always less than `1_000_000` so it will easily fit in a 32-bit number. These type signatures then hash to the **method selectors**:
```

neps/nep-0518.md
akhi3030, line 61: Maybe it's explained further below but leaving a comment for now in case I forget.  How is the boolean flag used exactly?  What happens when it indicates that a transaction is in flight?  Users cannot control in which order their transactions are getting processed, will their contracts be rejected arbitrarily?  Is it possible that a user has submited transactions 1,2,3.  1 is being processed so 2 gets rejected and then when 1 is processed, 3 arrives and gets processed?

akhi3030 (commented):
Another incomplete review.  Will try to get through more of it later this week.

neps/nep-0518.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Summary
+
+This NEP describes the protocol changes needed to support the usage of Ethereum-compatible wallets (Web3 wallets), for example Metamask, on Near native applications. That is to say, with this protocol change all Metamask users can become Near users without installing any additional software; from their perspective Near will appear as just another network they can choose from (similar to Aurora today).
+
+This is accomplished through two key protocol changes:
+
+1. Ethereum-like addresses (i.e. account IDs of the form `^0x[a-f0-9]{40}$`) are implicit accounts on Near (i.e. can be created via a `Transfer` action). We call these "eth-implicit accounts".
+2. Unlike the current implicit accounts (64-character hex-encoded), eth-implicit accounts do not have any access keys added to them on creation. Instead, these accounts will have a special contract deployed to them automatically called the "wallet contract". This wallet contract enables the owner of the Ethereum address corresponding to the eth-implicit account ID to sign transactions with their Ethereum private key, thus providing similar functionality to the default access key of 64-character implicit accounts.
```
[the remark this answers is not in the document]
birchmd: Currently it is impossible to deploy a different contract to a new eth implicit account. This could be enabled if the wallet contract was upgraded to allow users to take the `DeployContract` action or if the wallet contract allowed adding full access keys. However, this is a dangerous feature because it could result in users no longer being able to control their account using Ethereum-like transactions.

I cannot think of good use cases where the benefits of being able to deploy a different contract are worth the risk of a user accidentally locking themselves out of their account. But if there is one in the future then we can consider upgrading the wallet contract in another protocol change.

neps/nep-0518.md
birchmd: If `rlp_execute` is called while the boolean flag is set (so there is already a transaction in flight) then that `rlp_execute` call will immediately fail with an error about not allowing simultaneous transactions.

In your example where transaction 2 is rejected because 1 is still executed then 3 will also be rejected because the wallet contract follows the Ethereum standard of requiring nonces to be sequential. Transaction 3 will be rejected for having the wrong nonce until transaction 2 executes.

neps/nep-0518.md
birchmd: Done in https://github.com/near/NEPs/pull/555/commits/1f64d1fec30ef7e70a6a0675eb459a3eae1a6315

neps/nep-0518.md
birchmd: Done in https://github.com/near/NEPs/pull/555/commits/1f64d1fec30ef7e70a6a0675eb459a3eae1a6315

neps/nep-0518.md
birchmd: Done in https://github.com/near/NEPs/pull/555/commits/1f64d1fec30ef7e70a6a0675eb459a3eae1a6315

neps/nep-0518.md
birchmd: Done in https://github.com/near/NEPs/pull/555/commits/1f64d1fec30ef7e70a6a0675eb459a3eae1a6315

krylov-krylov (approved):

gagdiez:
I think the functionality explained in this PR has already been merged into the protocol, right? Is the explanation on this NEP close enough to the actual implementation @birchmd @akhi3030 @nagisa ?

Please let us know, because if this NEP explains functionality that exists, then we could simply accept it and merge it

gagdiez:
@birchmd we cannot fix the `lint` error ourselves as we do not have write access to the branch. Could you either fix them, or give write access to the PR? My guess is that if we simply merge the new `master` all `lint` errors will desapear

gagdiez (approved):

PR #563, opened 2024-09-10
[NEP] Add Deterministic Derivation Paths for NEAR MPC Network with Chain ID and Nonce

BenKurrek:
This pull request introduces a new standard for deterministic derivation paths in the NEAR MPC network. The proposed method allows wallets to regenerate foreign blockchain addresses (e.g., Ethereum, Avalanche) from a NEAR account using a combination of the NEAR account ID, chain ID, and a nonce that starts at 1. This structure supports multiple foreign addresses on the same chain and enables efficient account management across different blockchains.

### DIFF
--- a/neps/nep-tbd.md
+++ b/neps/nep-tbd.md
@@ -1,0 +1,170 @@
+---
+NEP: TBD
+Title: Deterministic Derivation Paths for MPC Network
+Authors: Ben Kurrek <benjaminkurrek@gmail.com>
+Status: New
+DiscussionsTo: TBD
+Type: Standards Track
+Version: 1.0.0
+Created: 2024-09-10
+LastUpdated: 2024-09-10
+---
+
+## Summary
+
+This NEP proposes a standard for deterministic derivation paths for NEAR accounts interacting with the Multi-Party Computation (MPC) network. The goal is to enable any wallet or service to deterministically regenerate addresses on foreign blockchains (e.g., Ethereum, Avalanche) from a NEAR account without requiring external information or tracking. This will simplify the process of creating and managing multi-chain accounts while ensuring that all accounts derived from a single NEAR account are easily reproducible across different applications and wallets.
+
+## Motivation
+
+As the NEAR MPC network becomes more widely adopted, there is a need for a standardized approach to generate multi-chain addresses from a NEAR account. Without a deterministic standard, it becomes difficult to infer the derivation paths for multi-chain addresses, making it challenging to manage multi-chain accounts consistently across different platforms.
+
+For example, a user with a NEAR account and an Ethereum account derived through MPC has no direct way of exposing or sharing their derivation path in a straightforward, repeatable manner. Wallets and services that need to access the user’s Ethereum account from the NEAR account would face difficulties without knowing how the Ethereum address was derived. By introducing a deterministic derivation path standard, any wallet or service can consistently regenerate foreign accounts from the NEAR account without additional tracking or custom configurations.
+
+This proposal seeks to prevent fragmentation by establishing a clear, scalable solution for deriving accounts across chains before multiple, incompatible solutions start emerging in the ecosystem.
+
+## Specification
+
+### Derivation Path Structure
+
+The derivation of foreign addresses will follow a deterministic method based on the NEAR account ID, a unique chain identifier, and a nonce. This structure allows for multiple foreign addresses to be derived from a single NEAR account on the same blockchain. The formula for deriving addresses is as follows:
+
+```
+{NEAR_ACCOUNT_ID} + {CHAIN_IDENTIFIER} + {NONCE} = {FOREIGN_ADDRESS}
+```
+
+Where:
+
+- `{NEAR_ACCOUNT_ID}` is the user’s NEAR account (e.g., example.near).
+- `{CHAIN_IDENTIFIER}` is a unique string representing the target blockchain (e.g., ethereum for Ethereum, avalanche for Avalanche).
+- `{NONCE}` is an incrementing number starting from 1, allowing for multiple foreign addresses to be derived for the same NEAR account on the same blockchain.
+
+This structure ensures that wallets can deterministically derive all foreign addresses associated with a NEAR account on any supported blockchain by iterating through the nonces.
+
+Example:
+
+- example.near + ethereum + 1 = 0x123... (Ethereum Address 1)
+- example.near + ethereum + 2 = 0x456... (Ethereum Address 2)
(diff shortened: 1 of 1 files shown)


BenKurrek:
> Any reason for having the the near account id in the derivation path? could be a bit redundant since you will already know which accountId you're deriving to call the mpc contract

Oh sorry perhaps it was written in a confusing way. I meant the full derivation path the MPC contract uses so technically what you would pass in is ethereum,1 but MPC would use account + ethereum,1

ewiner:
Some scattered thoughts:

1. There's a few existing standards for "chain IDs" that could serve us well to adopt:
   1. [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) from the creators of the Trezor hardware wallet.  Probably the most comprehensive chain directory out there.
   1. [CAIP-2](https://chainagnostic.org/CAIPs/caip-2), which seems more fork-resistant and well-defined than SLIP-44, and includes testnets as well as mainnets.  But it's not very human-readable. and I'm not sure whether it's well-defined for all known chains, including NEAR.
   1. (**EVM-only**) [ERC-3770](https://eips.ethereum.org/EIPS/eip-3770) relies on the human-readable `shortName`s in the directory at [ethereum-lists/chains](https://github.com/ethereum-lists/chains).   Each filename in that directory is a CAIP-2 chain ID and most listed chains (maybe all of them?) also reference their SLIP-44 chain ID.  I really like the idea of using a standardized but human-readable chain ID, but I'm not sure if there's an `ethereum-lists` equivalent that includes non-EVM chains, other than the SLIP-44 markdown file.

1. Sometimes you want to use the same derivation path on multiple chains.  For example, it can be very useful to have one Ethereum wallet address across all EVM-compatible chains.  We could have derivation paths that are specific to the literal chain (e.g. "Aurora") or to a whole architecture ("EVM chains").  Some people also just assume that any Ethereum mainnet address (CAIP2: `eip155:1`, SLIP44: 60, ERC-3770: `eth`) could exist on any EVM chain, which is pretty reasonable too.

1. For UTXO-based coins like Bitcoin, it's somewhat common to use a new address for each UTXO.  You could just increment the `nonce` each time, but [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) specifies a more detailed hierarchy of `account (nonce) / change (bool) / index (nonce)` which might be useful.  I'm curious how [SatoshiBridge](https://www.satos.network/) is doing address derivation but not sure who to ping on that.

esaminu:
@BenKurrek A few comments since we have had various discussions on this when implementing https://github.com/near/multichain-tools with @DavidM-D and @Pessina :

1. Using canonical-json [RFC-8785](https://www.rfc-editor.org/rfc/rfc8785) for serialization: [required serialization properties outlined here](https://github.com/near/near-fastauth-wallet/blob/2084bfd62d565b0cdd39d0ec44cc60ec19074905/docs/chain_signature_api.org#serialization). String concatenation meets some of the properties but not all
2. Use of [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) for `chainId`: avoids implementation errors and is more concise - also using `60` for all EVM chains since the same address is often used between EVM chains
3. Use of `meta` key to allow for multiple keys per chain serves the same purpose as `nonce` while allowing further properties to be added without breaking older derivations
4. Use of `domain` keys to allow for signing without prompting the user when the requesting domain can be identified (e.g. in iframe wallets)

[This PR](https://github.com/near/near-fastauth-wallet/blob/2084bfd62d565b0cdd39d0ec44cc60ec19074905/docs/chain_signature_api.org) documents the technical decisions we made when implementing the [multichain-tools](https://www.npmjs.com/package/multichain-tools) package

DavidM-D:
One thing that @ilblackdragon pointed out to me is that SLIPP-44 is a token identifier not a network identifier, note the lack of Base network because they don't have a native token. Mea culpa!
Network identifiers are defined in [EIP-3770](https://eips.ethereum.org/EIPS/eip-3770) which is also the spec used by the [omnitoken standard](https://hackmd.io/_MBniS8yTUKJCDHwDCc3-w). @esaminu do you think it's worth changing things on our end since people are still converging on a standard or is it more trouble than it's worth?

ewiner:
EIP-3770 only covers EVM chains, though.  For other chains... make our own?  Or use CAIP-2 (https://chainagnostic.org/CAIPs/caip-2 / https://docs.reown.com/cloud/chains/chain-list)?

GaloisField2718:
Hello, 
Is there a topic open on a forum? 

Because I don't understand why to not use BIP32 standards for ethereum and avalanche. And not sure if it's the right place.

ewiner:
> Hello, Is there a topic open on a forum?
> 
> Because I don't understand why to not use BIP32 standards for ethereum and avalanche. And not sure if it's the right place.

I think the discussion is mostly happening here.  BIP32 is relevant but it's not obvious to me how that helps us – what did you have in mind?

DavidM-D:
> why to not use BIP32 standards

The reason is that the BIP32 derivation standard is harder to do homomorphically as part of the MPC signing process than the derivation standard we settled on. We also have a proof of security for our derivation standard with MPC and presigning that we don't have for the BIP32 standard.

 What is your use case, maybe it's possible to add later on.

flmel:
Hi @BenKurrek – thank you for starting this proposal. As the moderator, I labeled this PR as "Needs author revision" because we assume you are still working on it since you submitted it in "Draft" mode.

Please ping the [@near/nep-moderators](https://github.com/orgs/near/teams/nep-moderators) once you are ready for us to review it.

GaloisField2718:
> > why to not use BIP32 standards
> 
> 
> 
> The reason is that the BIP32 derivation standard is harder to do homomorphically as part of the MPC signing process than the derivation standard we settled on. We also have a proof of security for our derivation standard with MPC and presigning that we don't have for the BIP32 standard.
> 
> 
> 
>  What is your use case, maybe it's possible to add later on.

Thanks for your answer. 
For now I'm just looking new updates on HDwallet for personal projects.  
But it could be important for future cross-chain feature. I'm trying to stay aware. 
I'll take a look on MPC thank you!

GaloisField2718:
> > Hello, Is there a topic open on a forum?
> 
> > 
> 
> > Because I don't understand why to not use BIP32 standards for ethereum and avalanche. And not sure if it's the right place.
> 
> 
> 
> I think the discussion is mostly happening here.  BIP32 is relevant but it's not obvious to me how that helps us – what did you have in mind?

Cool thanks. Just to not flood a wrong place

garikbesson:

PR #568, opened 2024-10-24
NEP-568: Resharding V3

wacban:
NEP Status (Updated by NEP Moderators)

Status: Approved

SME reviews:
- [X] @birchmd https://github.com/near/NEPs/pull/568#issuecomment-2672771979
- [X] @mfornet https://github.com/near/NEPs/pull/568#pullrequestreview-2602405518

* [👍] @birchmd https://github.com/near/NEPs/pull/568#issuecomment-2672771979 
* [👍] @mfornet As a working group member I support approving this proposal.
* [ ] @mm-near
* [👍] @bowenwang1996 https://github.com/near/NEPs/pull/568#issuecomment-2672940286


flmel:
Hi @wacban – thank you for starting this proposal. As the moderator, I labeled this PR as "Needs author revision" because we assume you are still working on it since you submitted it in "Draft" mode.
Please ping the [@near/nep-moderators](https://github.com/orgs/near/teams/nep-moderators) once you are ready for us to review it

neps/nep-0568.md
```diff
@@ -205,6 +205,46 @@ The change we propose is to move the initial state download point to one in the
...
+could theoretically include the outgoing receipts from their shared ancestor
+chunk. However, this naive approach would lead to the duplication of receipts,
+which must be avoided.
+
+The proposed solution is to reassign the outgoing receipts from the parent chunk
+to only one of the child shards. Specifically, the child shard with the lower
+shard ID will claim all outgoing receipts from the parent, while the other child
+will receive none. This ensures that all receipts are processed exactly once. 
+
+#### Incoming Receipts
```
wacban, line 231: ref: `get_incoming_receipts_for_shard`

neps/nep-0568.md
```diff
@@ -205,6 +205,46 @@ The change we propose is to move the initial state download point to one in the
...
 
+When the shard layout changes, it is crucial to handle cross-shard traffic
+correctly, especially in the presence of missing chunks. Care must be taken to
+ensure that no receipt is lost or duplicated. There are two important receipt
+types that need to be considered - the outgoing receipts and the incoming
+receipts.
+
+Note - this proposal reuses the approach taken by ReshardingV2. 
+
+#### Outgoing Receipts
```
wacban, line 216: ref: `reassign_outgoing_receipts_for_resharding_v2`

neps/nep-0568.md
Trisfald, line 1: LGTM (commit 09dc287fc2d5c75727f6d876ff4077627e79a1f7)

walnut-the-cat:
As the moderator,  

I believe this proposal is ready for SME review. 

@near/wg-protocol , could you help assign SMEs who can review the proposal?

Thank you.

bowenwang1996:
As a working group member, I nominate @birchmd and @mfornet as SME reviewers.

neps/nep-0568.md
```diff
@@ -0,0 +1,629 @@
...
+  }
+  ```
+
+  The `get` function retrieves data using the resolved `ShardUId` and key, while `increment_refcount_by` manages reference counts, ensuring correct tracking even when accessing data under an ancestor shard.
+
+#### Mapping Retention and Cleanup
+
+Mappings in `DBCol::ShardUIdMapping` persist as long as any descendant relies on an ancestor’s data. To manage this, the `set_shard_uid_mapping` function in `TrieStoreUpdateAdapter` adds a new mapping during resharding:
+
+```rust
```
birchmd, line 510: If I understand this mapping correctly, it is applied to all keys for the child shards, not just bridging to historical data. By which I mean even new keys for the child shard will be stored with the ancestor's shard UID.

I think this means the mapping needs to be taken into account when assigning the child. For example, suppose shard A splits into B and C, then shard B splits into D and E. In this case, D and E need to map to A instead of B because there are not any actual B keys in the database in the first place. Is that right or have I misunderstood?

neps/nep-0568.md
```diff
@@ -0,0 +1,629 @@
...
+* **Chain Forks Handling**: Resharding should work correctly in the presence of chain forks.
+* **Retired Shards Cleanup**: Retired shards should be cleaned up.
+
+Note that the Flat States of the newly created shards will not be available until resharding is completed. This is acceptable because the temporary MemTries are built instantly and can satisfy all block processing needs.
+
+The main component responsible for carrying out resharding on Flat State is the [FlatStorageResharder](https://github.com/near/nearcore/blob/f4e9dd5d6e07089dfc789221ded8ec83bfe5f6e8/chain/chain/src/flat_storage_resharder.rs#L68).
+
+#### Flat State's Status Persistence
+
+Every shard's Flat State has a status associated with it and stored in the database, called `FlatStorageStatus`. We propose extending the existing object by adding a new enum variant named `FlatStorageStatus::Resharding`. This approach has two benefits. First, the progress of any Flat State resharding is persisted to disk, making the operation resilient to a node crash or restart. Second, resuming resharding on node restart shares the same code path as Flat State creation (see `FlatStorageShardCreator`), reducing code duplication.
```
birchmd, line 384: If the node crashes during resharding then the `HybridArena` MemTrie objects also need to be reconstructed to resume block processing. How does this recovery happen? I assume it is easy to reconstruct the `FrozenArena` from the Flat state, but what about the changes on top of it? Are they persisted somewhere? If they are not then I suppose the node would need to recreate the MemTrie by re-applying the chunks that have happened so far in the epoch.

birchmd (commented):
As a reviewer I would like to thank the team for putting together such a well-written and detailed proposal! Designing this feature was no easy task, and the proposal here seems to address all the intricacies nicely.

I have left two questions, but otherwise the design looks good to me.

neps/nep-0568.md
wacban: @staffik Can you reply please?

neps/nep-0568.md
wacban: @Trisfald Can you reply please?

staffik: > even new keys for the child shard will be stored with the ancestor's shard UID.

Yes.



> suppose shard A splits into B and C, then shard B splits into D and E. In this case, D and E need to map to A instead of B because there are not any actual B keys in the database in the first place. Is that right or have I misunderstood?

That's right if a node did two splits.
Another case would be a node did state sync after resharding and downloaded the state of shard B directly. Then it won't have the mapping from B to A.

birchmd: Thanks for the clarification @staffik ! Should the code block be updated so that the `parent_shard_uid` is replaced with it's mapped value (if present)?

> Another case would be a node did state sync after resharding and downloaded the state of shard B directly. Then it won't have the mapping from B to A.

That's quite interesting! It means two nodes may represent the same state differently depending on their shard tracking history. I don't think it's a problem, but it does feel like an unexpected consequence to me.

staffik: Ah yes. Updated here: https://github.com/near/NEPs/pull/583

neps/nep-0568.md
Trisfald: Yes the HybridArena Memtrie will need to be reconstructed upon crash or even a simple restart.

I think at the moment stage of `nearcore/master` we don't support seamless node restarts yet. There are a couple possible solution to this problem since all changes to the state are always persisted, in a way or another. In increasing order of complexity:
- Wait for flat storage to complete resharding and load the memtrie for children shards as usual
- Recreate the memtrie of the parent shard as it was at the resharding block, split it again, and apply changes
- Create memtries from current flat storage state of children (even if not complete) and apply remaining changes
- More complex solutions which leverage decoupling of memtrie from flat storage

birchmd: Thanks for the clarification @Trisfald . I brought this up here because crash recovery is mentioned as part of the motivation for the design. Maybe it is worth adding a section to the MemTrie specification about how restarts during resharding will be handled.

walnut-the-cat:
@mfornet , could you be able to review this NEP?

mfornet (approved):
Hi @wacban, @walnut-the-cat, sorry for the delay reviewing this PR.

I really appreciate all the details layed out in this document, given the complexity of the proposal. The questions I was annotating were answered in the document itself. 

The design looks good.

flmel:
@near/wg-protocol Can you please fully read this NEP and comment in the thread if you are leaning towards approving or rejecting it? Please make sure to include your rationale and any feedback that you have for the author.

birchmd:
As a subject matter expert and working group member I lean towards approving this proposal. You can see more detailed comments from me in my SME review.

bowenwang1996:
As a working group member I lean towards approving this proposal. Thank you for writing such a detailed proposal!

mfornet:
As a working group member I support approving this proposal.

flmel (approved):

PR #570, opened 2024-10-28
Add flat state specs to resharding NEP

Trisfald:
Filling the specification section about flat state.

### DIFF
--- a/neps/nep-0568.md
+++ b/neps/nep-0568.md
@@ -86,8 +86,7 @@ the latest block.
 
 For the reason stated above, the chosen strategy is to reshard Flat State in a
 long-running background task. The new shards' states must converge with their
-Mem Tries representation in a reasonable amount of time, indicatively by the
-first half of the new epoch.
+Mem Tries representation in a reasonable amount of time.
 
 Splitting a shard's Flat State is performed in multiple steps:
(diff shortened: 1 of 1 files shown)


neps/nep-0568.md
```diff
@@ -73,6 +73,38 @@ post-processing, as long as the chain's view reflects a fully resharded state.
...
+Flat State main purposes are allowing the creation of State Sync snapshots and
+the construction of Mem Tries. Fortunately, these two operations can be delayed
+until resharding is completed. Note also that with Mem Tries enabled the chain
+can move forward even if the current status of Flat State is not in sync with
+the latest block. 
+
+For the reason stated above, the chosen strategy is to reshard Flat State in a
+long-running background task. The new shards' states must converge with their
+Mem Tries representation in a reasonable amount of time, indicatively by the
+first half of the new epoch.
```
wacban, line 90: Why the first half of the new epoch? It's just for my information, no need to add it here. I don't think it's too important so maybe skip this part (the "indicatively by the first half of the new epoch").

Trisfald: Removed!

It was badly worded; the flat state must converge with memtrie, at some point in time.

The time consideration was to have flat storage ready soon enough so that state parts for the new shards can be distributed before the next epoch.

PR #571, opened 2024-11-05
State Storage - State

staffik:


wacban (commented):
Looks great overall! 

I think the implementation details can go into the `reference implemention` section. Can you add a more high level overview in specification and move the rest? Can you also mention what the naive / brute force approach would be and why it is not feasible?

neps/nep-0568.md
```diff
@@ -108,6 +108,35 @@ Splitting a shard's Flat State is performed in multiple steps:
...
+these adapters check `ShardUIdMapping` to access data under the relevant ancestor’s `ShardUId` where necessary,
+applying this behavior to both read and write operations.
+
+#### Mapping retention and cleanup
+
+Mappings in `ShardUIdMapping` persist as long as any descendant shard still references the ancestor shard’s data.
+When no descendants rely on a particular ancestor shard, its mapping entry can be removed, allowing the ancestor’s data to be garbage collected.
+If a full state sync is performed for a child shard (downloading and storing all its data independently),
+the mapping to the ancestor shard can also be removed, as the child’s data is now directly stored under its own `ShardUId`.
+
```
wacban, line 135: The cleanup shouldn't depend on state sync. I think we should cleanup the mapping when the node stops tracking all children / descendant shards. Also can you mention that the mapping will be permanent on archival nodes?

wacban (commented):
Once we have a good design for the integration with cold storage let's also add a short description of how that will work.

wacban (approved):
looks great, thanks!

wacban:
Once merged into the main PR, please check if there any any lints that need fixing and address those - feel free to commit directly, no need for another PR.

PR #579, opened 2024-12-13
state sync: reword to make clear it's not just for convenience

marcelo-gonzalez:


wacban:
Allow me to go ahead and merge it before day dawns down under ;)

PR #591, opened 2025-02-11
[WIP]: Global Contracts

pugachAG:


flmel:
Hi @pugachAG – thank you for starting this proposal. As the moderator, I labeled this PR as "Needs author revision" because we assume you are still working on it since you submitted it in "Draft" mode.
Please ping the [@near/nep-moderators](https://github.com/orgs/near/teams/nep-moderators) once you are ready for us to review it.

neps/nep-0591.md
```diff
@@ -0,0 +1,190 @@
...
+With the advent of chain signatures, the smart contract wallet use case will become more ubiquitous.
+As a result, it is very desirable to be able to reuse already deployed contract without having to pay for the storage cost again.
+
+Additionally global contracts cover the underlying use case for [NEP-491](https://github.com/near/NEPs/pull/491): https://github.com/near/nearcore/pull/12818.
+
+## Specification
+
+Global contract can be deployed in 2 ways: either by its hash or by owner account id.
+Contracts deployed by hash are effectively immutable and cannot be updated.
+When deployed by account id the owner can redeploy the contract updating it for all its users.
```
walnut-the-cat, line 37: Can we talk about when to use each way? For instance, when should a user use hash instead of account id, and vice versa?

neps/nep-0591.md
```diff
@@ -0,0 +1,190 @@
...
+
+One potential issue is increasing infrastructure cost for global contracts with growing number of shards.
+A global contract is effectively replicated on every shard, so with increase in number of shards each global contract uses more storage.
+This can be potentially addressed in the future by making deployment costs parametrized with the number of shards in the current epoch, but it still wouldn't address the issue for the already deployed contracts.
+
+## Alternatives
+
+In [the original proposal](https://github.com/near/NEPs/issues/556) we considered storing global contracts in a separate global trie (managed at the block level) and introducing a dedicated distribution mechanism.
+We decided not to proceed with this approach as it requires significantly higher effort to implement and also introduces new critical dependencies for the protocol.
+
```
walnut-the-cat, line 162: Did you intentionally skip Future possibilities and Consequences section? (e.g. https://github.com/near/NEPs/blob/global-contracts/neps/nep-0539.md#future-possibilities)

neps/nep-0591.md
pugachAG: Sure, added in bfad2e3

neps/nep-0591.md
pugachAG: Yes, I don't have anything to say in those section, so it was omitted

pugachAG: I've added Future possibilities section in 053cb24 mentioning sharded contracts

walnut-the-cat:
Thank you @pugachAG and @stedfn for submitting this NEP.

<details>

Technical Summary guidelines:

```

### Recommendation

Add recommendation

### Benefits

* Benefit

* Benefit

### Concerns

| - | - | - | - |

```

</details>

bowenwang1996:
As a working group member, I nominate @akhi3030 and @wacban as SME reviewers

akhi3030:
To help me schedule the work here, what is the deadline for this review?

neps/nep-0591.md
akorchyn: Sorry to jump in, but I’m curious, will the deployed-by account ID receive a hash?

I don’t see a reason to allow someone to change my code and potentially take everything from my contract. But if a globally deployed account also gets a hash that I can lock onto (and update later if needed), that would be a nice DevX improvement.

I do see a downside, though that every update essentially creates a new copy of the code instead of replacing it. So I’m curious, how is this going to be addressed?

pugachAG:
@akhi3030 testnet release is scheduled for March 17th, so it would be nice to get this NEP reviewed by that that date.

neps/nep-0591.md
```diff
@@ -0,0 +1,207 @@
...
+## Motivation
+
+A common use case on NEAR is to deploy the same smart contract many times on many different accounts. For example, a multisig contract is a frequently deployed contract.
+However, today each time such a contract is deployed, a user has to pay for its storage and the cost is quite high. For a 300kb contract the cost is 3N.
+
+With the advent of chain signatures, the smart contract wallet use case will become more ubiquitous.
+As a result, it is very desirable to be able to reuse already deployed contract without having to pay for the storage cost again.
+
+Additionally global contracts cover the underlying use case for [NEP-491](https://github.com/near/NEPs/pull/491): https://github.com/near/nearcore/pull/12818.
```
wacban, line 31: Can you add a short context and explanation?

neps/nep-0591.md
```diff
@@ -0,0 +1,207 @@
...
+struct GlobalContractData {
+    code: Vec<u8>,
+    id: GlobalContractIdentifier,
+}
+```
+
+`GlobalContractDistribution` receipt is generated as a result of processing `DeployGlobalContractAction`.
+Receipt distribution logic is updated to route such receipts to all shards.
+So effectively it is a part of `ShardProof` (incoming receipts) for each shard, but occurs only once in the outgoing receipts of that chunk.
+Applying `GlobalContractDistribution` receipt updates the corresponding `TrieKey::GlobalContractCode` in the trie.
```
wacban, line 117: It would be good to test that this logic works well in case of resharding. One particular corner case would be:
* `#10` - deploy smart contract in shard X
* `#11` - shard Y and shard Z are both missing chunk
* `#12` - new shard layout after resharding where Y was split into Y1 and Y2

For X = Y, X = Z, X != Y && X != Z

neps/nep-0591.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+Applying `UseGlobalContractAction` updates user account `contract` field accordingly.
+
+`FunctionCall` action processing is updated to respect global contracts. This includes updating [contract preparation pipeline](https://github.com/near/nearcore/blob/fb95d7b7740d1fda9245afa498ce4e9ac145c8af/runtime/runtime/src/pipelining.rs#L24) as well as [recording of the executed contract to be included in the state witness](https://github.com/near/nearcore/blob/fb95d7b7740d1fda9245afa498ce4e9ac145c8af/core/store/src/trie/update.rs#L338).
+
+### Costs
+
+For global contracts deployment we burn tokens for storage instead of locking like what we do regular contracts today.
+The cost per byte of global contract code `global_contract_storage_amount_per_byte` is set as 10x the storage staking cost `storage_amount_per_byte`.
```
wacban, line 140: Is it set to 10x symbolically or do you just set its value to be 10x? I would suggest decoupling them so that they can be updated independently.

neps/nep-0591.md
```diff
@@ -0,0 +1,207 @@
...
+
+Additionally we add action costs for the global contract related actions:
+
+* `action_deploy_global_contract` is exactly the same as `action_deploy_contract`
+* `action_deploy_global_contract_per_byte`:
+  * send costs are the same as `action_deploy_contract_per_byte`
+  * execution costs should cover distribution of the contract to all shards:
+    * this is pretty expensive for the network, so want want to change significant amount of gas for that
```
wacban, line 148: typo: change -> charge

neps/nep-0591.md
```diff
@@ -0,0 +1,207 @@
...
+
+enum GlobalContractDeployMode {
+    /// Contract is deployed under its code hash.
+    /// Users will be able reference it by that hash.
+    /// This effectively makes the contract immutable.
+    CodeHash,
+    /// Contract is deployed under the owner account id.
+    /// Users will be able reference it by that account id.
+    /// This allows the owner to update the contract for all its users.
+    AccountId,
```
wacban, line 57: Does deploying a global contract by account id deploy said contract to the owner account as well?

neps/nep-0591.md
```diff
@@ -0,0 +1,207 @@
...
+    AccountId(AccountId),
+}
+```
+
+The value is contract code bytes, similar to `TrieKey::ContractCode`.
+
+### Distribution
+
+Global contract has to be distributed to all shards after being deployed.
+This is implemented with a dedicated receipt type:
```
wacban, line 100: Is this accounted for in the BandwidthScheduler?

wacban (commented):
I had the first look, looks good so far, I'll come back to it some time this week.

neps/nep-0591.md
pugachAG: @staffik can you please help me here

neps/nep-0591.md
```diff
@@ -0,0 +1,207 @@
...
+    CodeHash,
+    /// Contract is deployed under the owner account id.
+    /// Users will be able reference it by that account id.
+    /// This allows the owner to update the contract for all its users.
+    AccountId,
+}
+```
+
+Also new action is added for using previously deployed global contract:
+
```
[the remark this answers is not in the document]
pugachAG: As far as I know it is not possible to unlink the regular contract, so we don't have to solve that for the global contracts. The same applies for the state of the contracts.

neps/nep-0591.md
akhi3030: +1 to this comment.

akhi3030 (approved):
generally looks good to me and I am happy for it to proceed with the current spec.  I will do a detailed SME review soon.

neps/nep-0591.md
staffik:
```suggestion
Additionally, global contracts cover the underlying use case for [NEP-491](https://github.com/near/NEPs/pull/491): [#12818](https://github.com/near/nearcore/pull/12818).
Some businesses onboard their users by creating an account on their behalf and deploying a contract to it. Such a use case is susceptible to refund abuse, where there is a financial incentive to repeatedly create and destroy an account, cashing in on the storage fee that was initially paid by the business to deploy the contract.
```
cc @pugachAG

neps/nep-0591.md
```diff
@@ -0,0 +1,207 @@
...
+    /// This effectively makes the contract immutable.
+    CodeHash,
+    /// Contract is deployed under the owner account id.
+    /// Users will be able reference it by that account id.
+    /// This allows the owner to update the contract for all its users.
+    AccountId,
+}
+```
+
+Also new action is added for using previously deployed global contract:
```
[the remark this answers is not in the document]
pugachAG: addressed in c528f22

neps/nep-0591.md
pugachAG: `global_contract_storage_amount_per_byte` is a separate runtime parameter completely independent from `storage_amount_per_byte`, here I meant that initial values is set to be `10 * storage_amount_per_byte`

neps/nep-0591.md
pugachAG: nope, it does not do that

neps/nep-0591.md
pugachAG: implemented as part of https://github.com/near/nearcore/pull/13147

neps/nep-0591.md
pugachAG: Nope, in this case we store account id, so the owner can update contract for all current users. There are underlying use cases for that, for example application developers onboarding users into crypto. 
Also it is possible to lock such contract if owner deletes all keys from the account, effectively making it impossible to make any changes.

neps/nep-0591.md
```diff
@@ -0,0 +1,221 @@
...
+    target_shard: ShardId,
+    already_delivered_shards: Vec<ShardId>,
+    code: Arc<[u8]>,
+}
+```
+
+`GlobalContractDistribution` receipt is generated as a result of processing `DeployGlobalContractAction`.
+Such receipt contains destination shard `target_shard` as well as list of already visited shard ids `already_delivered_shards`.
+Applying `GlobalContractDistribution` receipt updates the corresponding `TrieKey::GlobalContractCode` in the trie for the current shard.
+It also generates distribution receipt for the next shard in the current shard layout. This process continues until `already_delivered_shards` contains all shards.
```
wacban, line 128: One consequence of this approach is that global contract deployment will take at least num_shards blocks to get onto all shards. In the future we may need some way of checking if the contract is fully deployed / updated. Nothing to worry about now, I'm sure the contract developers will ask about it once they need it.

neps/nep-0591.md
```diff
@@ -0,0 +1,221 @@
...
+`AccountV2` is introduced changing `code_hash: CryptoHash` field to more generic `contract: AccountContract`:
+
+```rust
+enum AccountContract {
+    None,
+    Local(CryptoHash),
+    Global(CryptoHash),
+    GlobalByAccount(AccountId),
+}
+```
```
wacban, line 145: Will it be possible to deploy / use global contracts from accounts created before AccountV2 is released? Is there any migration / lazy update necessary?

wacban (approved):
As a subject matter expert I approve this NEP. 

It adds an important feature, it is well designed, it strikes a great balance between added complexity and practical aspects and it well integrates with the rest of nearcore. Thank you @bowenwang1996, @pugachAG and @stedfn for this great proposal.

neps/nep-0591.md
pugachAG: Indeed with more shards in the future propagating global contract updates might take quite some time. 
We can extend the functionality in the future by adding some deployment progress indicator as part of owners account id data if that would be required.

neps/nep-0591.md
```diff
@@ -0,0 +1,221 @@
...
+    contract_identifier: GlobalContractIdentifier,
+}
+
+enum GlobalContractIdentifier {
+    CodeHash(CryptoHash),
+    AccountId(AccountId),
+}
+```
+
+This action is similar to deploying a regular contract to an account, except user does not cover storage deposit.
```
birchmd, line 77: I think it is worth discussing possible errors arising from this action. If a code hash is given that does not correspond to a known global contract does the action fail? Similarly, if there is no global contract associated with the given account does the action fail?

This is important to know because global contracts take time to propagate to all shards and so a code hash which is unrecognized at one time might become recognized later.

neps/nep-0591.md
```diff
@@ -0,0 +1,221 @@
...
+    /// This effectively makes the contract immutable.
+    CodeHash,
+    /// Contract is deployed under the owner account id.
+    /// Users will be able reference it by that account id.
+    /// This allows the owner to update the contract for all its users.
+    AccountId,
+}
+```
+
+User pays for storage by burning NEAR tokens from its balance depending on the deployed contract size.
```
birchmd, line 62: Error cases should be discussed here too. If the given code is not valid wasm are the Near tokens still burned? Or is a different amount of gas charged instead?

birchmd (approved):
As a working group member, I lean towards approving this proposal. In addition to the reduced cost of the "factory pattern", I especially look forward to how this feature lays the foundation for "sharded contracts" in the future.

I've left a few comments where I think this specification needs a little more detail.

akhi3030 (approved):
As a SME reviewer, I also lean towards approving this NEP.

Pros:
- it reduces the costs for deploying the same wasm module on multiple accounts making it cheaper for many people to use the network.
- it is a necessary requirement for sharded contracts

Cons:
- There are concerns about potential state not being cleaned up from the chain.  However, given the current storage staking incentives, we have similar problems already and in this instance the problem is less severe as the costs for deploying a global contract are high.

neps/nep-0591.md
pugachAG: good point, mentioned in 57545f524d3bf7eb9ed60c4db1cd64b28f5b6138

neps/nep-0591.md
pugachAG: added description in dcfdcd3bfffe253234cf1bf68e760f27330df50f, thanks!

pugachAG:
@birchmd thanks for your review!

> It might be worth adding a note on how this routing interacts with the new instant-resharding feature.

I've added description for that in 6bf2c85c893dbac416eb8c99201541834e02310a

neps/nep-0591.md
pugachAG: added resharding description in 6bf2c85c893dbac416eb8c99201541834e02310a

flmel:
@near/wg-protocol Can you please comment in the thread if you are leaning towards approving or rejecting this NEP? Please include your rationale and/or any feedback that you have for the authors.

neps/nep-0591.md
```diff
@@ -0,0 +1,224 @@
...
+    already_delivered_shards: Vec<ShardId>,
+    code: Arc<[u8]>,
+}
+```
+
+`GlobalContractDistribution` receipt is generated as a result of processing `DeployGlobalContractAction`.
+Such receipt contains destination shard `target_shard` as well as list of already visited shard ids `already_delivered_shards`.
+Applying `GlobalContractDistribution` receipt updates the corresponding `TrieKey::GlobalContractCode` in the trie for the current shard.
+It also generates distribution receipt for the next shard in the current shard layout. This process continues until `already_delivered_shards` contains all shards.
+This way we ensure that at any point in time there is at most one `GlobalContractDistribution` receipt in flight for a given deployment and eventually it will reach all shards.
```
mfornet, line 131: > This way we ensure that at any point in time there is at most one `GlobalContractDistribution` receipt in flight

What is the motivation for having at most one such receipt, instead of deploying it in parallel to all shards, or at least using some other strategy that allows faster deployment.

neps/nep-0591.md
```diff
@@ -0,0 +1,224 @@
...
+}
+```
+
+Applying `UseGlobalContractAction` updates user account `contract` field accordingly.
+
+`FunctionCall` action processing is updated to respect global contracts. This includes updating [contract preparation pipeline](https://github.com/near/nearcore/blob/fb95d7b7740d1fda9245afa498ce4e9ac145c8af/runtime/runtime/src/pipelining.rs#L24) as well as [recording of the executed contract to be included in the state witness](https://github.com/near/nearcore/blob/fb95d7b7740d1fda9245afa498ce4e9ac145c8af/core/store/src/trie/update.rs#L338).
+
+### Costs
+
+For global contracts deployment we burn tokens for storage instead of locking like what we do regular contracts today.
```
mfornet, line 156: Do we burn the tokens for storage on every deployment? even when we are using deploying by account id which overrides and removes the previous one?

neps/nep-0591.md
```diff
@@ -0,0 +1,224 @@
...
+    id: GlobalContractIdentifier,
+    target_shard: ShardId,
+    already_delivered_shards: Vec<ShardId>,
+    code: Arc<[u8]>,
+}
+```
+
+`GlobalContractDistribution` receipt is generated as a result of processing `DeployGlobalContractAction`.
+Such receipt contains destination shard `target_shard` as well as list of already visited shard ids `already_delivered_shards`.
+Applying `GlobalContractDistribution` receipt updates the corresponding `TrieKey::GlobalContractCode` in the trie for the current shard.
```
mfornet, line 129: Just to clarify, this means that a contract becomes available in a shard after processing the `DeployGlobalContractAction` action, so the contract can be used in some shards before other ones right?

mfornet: In case this is correct, some "funny" side effect, is that two contracts living in different shards, pointing at the same global contract, could run an action on the same block, and be using two different contracts, since the update is still in flight.

By "funny" I mean that it is hard to foresee side effects of this behavior, or if it could be exploited.

mfornet (approved):
Congrats on delivering this NEP. 

As a working group member I lean toward approving this NEP, and personally I'm looking forward to being released.

I left a few comments and questions.

neps/nep-0591.md
pugachAG: Having at most one receipt allows us to easily handle edge case when deployment happens at the resharding boundary and we need to make sure the contract code reaches shards introduced in the new layout.
When deploying to all shards at once (in parallel) resharding would require either special handling or major changes to how we route buffered receipts. Specifically when buffered deployment receipt is targeted to an old shard, we can only re-route it to only one of the child shards after resharding, hence special logic would be required when applying that receipt to further forward it to the other child shard.
To summarise: we considered multiple options and decided to proceed with the current approach as it was the easiest to implement and reason about

neps/nep-0591.md
pugachAG: Yes, in the initial implementation we decided to proceed with the simplest approach and always burn tokens regardless of previously deployed global contract under that account id. This also indirectly covers distribution costs that cannot be fully accounted in gas costs due to chunk gas limit.

neps/nep-0591.md
pugachAG: You are right, this can result in such "funny" behaviours. Effectively during upgrade 2 versions of the contract will be active on different shards for some range of blocks. That is something contract developers should be aware of. Bowen also raised that point and we reached a conclusion that it should not be a big issue in practice.

neps/nep-0591.md
bowenwang1996: The previous one isn't really removed. It will still be there

pugachAG: In case of deployment by account_id the previously deployed code will be replaced by the new one

bowenwang1996:
As a working group member, I approve this NEP

pugachAG:
@flmel I guess this can be merged now?

flmel (approved):

PR #605, opened 2025-04-05
initial commit

akhi3030:


neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,182 @@
...
+### Upgrading a sharded contract
+
+Upgrading a contract normally requires two steps:
+
+- Upgrading the contract code and
+- Upgrading the actual contract state
+
+For normal contracts, the upgraded contract code contains sufficient logic to upgrade the associated contract state.  Once the upgraded contract code is deployed, it can [potentially over multiple calls] upgrade the associated contract state.  Further, the owner can hold off on additional upgrades till the existing upgrade has finished.
+
+In the sharded contract scenario, upgrading the contract code is straight forward if we use the `AccountId` mode of global contract but upgrading the contract state is no long as straight forward.  The actual state is fully distributed on all the user accounts of the contract.  Further, the contract owner may not even have a list of all the user accounts that are using an instance of the sharded contract.  So the only entities that can actually upgrade the state are the users themselves and they would only upgrade the state the next time they choose to use the contract.
```
wacban, line 107: typo: no long -> no longer

wacban (commented):
looks good so far, I'm looking forward to more :) 

I'm curious what type of contracts will not be easily shardable using this approach. This design is perfect for the ft use case, what would be the opposite?

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,182 @@
...
+- In the centralised situation, all function calls took place on a single shard, assuming a uniform distribution of accounts across the network, the 2 function calls will be uniformly distributed across all the shards of the network.  This implies that the maximum TPS of this application scenario will be the sum of the TPS of all the shards on the network.
+
+In order to enable the above scenario, we need the following new primitives in the NEAR protocol:
+
+- Global contract code: Each user account needs to deploy the sharded version of the FT contract on their account.  If the application has many users, doing this deployment can become quite expensive and it also feels unnecessary as each user is using exactly the same contract code.  Here, we propose to rely on the ongoing work of GlobalContracts.  More specifically, we will be using the `AccountId` version of this work.  This is because when the owner of the application wants to upgrade their contract code, they need all the users to upgrade as well otherwise their users will be fragmented between different versions of the contract.
```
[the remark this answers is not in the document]
akhi3030: Yeah, agreed.  I am thinking MVP and iterative development so I think we do not need to enable multiple deployments on a single account in this NEP but a future one.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,182 @@
...
+
+- The sender sends a transaction with a function call action to the contract to transfer the tokens from one account to another.
+- This transaction is routed to the shard where the user's account lives where it is converted to a receipt.
+- Then the receipt is routed to the shard where the FT contract lives.
+- Once it arrives on the FT contract's shard, the receipt is executed, the function call is performed, and the transfer is performed.
+
+In the lucky case, the sender's account resides on the same shard as the FT contract.  In this case, depending on the load on the network, the transaction may be converted into a receipt and the receipt is executed in the same block.  Assuming a uniform distribution of user accounts across the network, such cases will only happen for every `N`th transaction where `N` is the number of shards on the network.  In other cases, the minimum latency is 2 blocks.  In the first block, the transaction is converted into a receipt and transferred to the receiver's shard.  And in the second block, the receipt is processed.
```
bowenwang1996, line 38: > In the lucky case, the sender's account resides on the same shard as the FT contract. In this case, depending on the load on the network, the transaction may be converted into a receipt and the receipt is executed in the same block.

Not true. The user account is different from the ft contract account, so even if they are on the same shard, the receipt to call ft contract will *not* be executed in the same block

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,182 @@
...
+### Sharded FT contract
+
+In the centralised FT contract above, all the state is stored in a single centralised location.  The opposite extreme of this approach would be to store all the state in as distributed a manner as possible i.e. the account balances of each user is stored locally on their accounts instead.
+
+If the state were stored in such a distributed manner, performing a FT transfer would require the following steps.
+
+- The sender would send a transaction to their account to transfer some tokens to the receiver.
+- The transaction is converted to a receipt and the receipt is guaranteed to execute on the same shard as the account to execute it against is the same account that is used to convert the transaction into a receipt.
+- When the receipt is executed, the sender's balance is deducted and a cross contract receipt is sent to the receiver's account to increase its balance.
+- The receipt arrives at the receiver's shard and is executed, which increases the receiver's account balance.
```
bowenwang1996, line 55: This is still somewhat vague. Please be more precise with sender being `alice.near` and receiver being `bob.near`. What exactly is the function call and where are the balances stored?

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,182 @@
...
+
+Turns out that the NEAR protocol is already solving a version of this problem.  State from all accounts is stored together in the same `Trie` data structure.  So in theory, if one account could figure out under what key another account has stored some value, it could issue a `storage_write` for that key and make unauthorised modifications to it.  To prevent this type of attack, when accounts issue a `storage_write`, the protocol prepends the `AccountId` of the account that is currently executing to the key.  This creates a unique namespace for each account that is inaccessible to all other accounts.
+
+We propose a similar approach for sharded contracts as well.  Instead of using the `AccountId` namespace, when an account starts using a sharded contract code, the protocol can create a new `ShardedContractInstanceId` for it.  This id can then be prepended to all storage accesses and just like above, this creates an isolated namespace for that specific sharded contract that the account cannot access.
+
+If an account stops using a sharded contract code (e.g. by deploying another contract code on itself), then the associated `ShardedContractInstanceId` can be removed.
```
bowenwang1996, line 83: I am not sure what this means. The sharded contract code should be for a specific namespace and there can be many namespaces. They are independent from what contract is deployed on the account itself

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,182 @@
...
+
+The second case is trickier to address.  In order to ensure that no tokens are maliciously minted, when a receiver is executing the call to receive tokens, it needs a guarantee that the sender of the message is indeed another instance of the sharded contract and that it has indeed decremented its local balance appropriately.  The minimum amount of information that the receiver needs to check to get this guarantee is that the receiver needs to know that the sender of the cross contract call was an instance of the sharded contract.
+
+Currently, the receiver of a message can check the `AccountId` of the sender of the message.  In this case that is not sufficient as the receiver does not know if the sender is indeed using an instance of the sharded contract or not.  Therefore, we propose that when an instance of a sharded contract sends a cross contract receipt to another contract, on top of including the `AccountId` of the actual account, we also include the `AccountId` of the sharded contract code.  Then the receiver of the cross contract receipt can check this bit of information and convince itself that this is a legitimate token transfer.
```
bowenwang1996, line 96: It is more subtle than what is described here:
* For writes, it should be coming from another instance of the same contract (or the sharded contract account itself)
* For reads, it should allow any other contracts to call directly

bowenwang1996 (commented):
I'd like to see the designs of the namespaces be more fleshed out with reference-level spec details on how exactly the work and how runtime may be affected as a result. What is currently written still feels too high-level to me

neps/nep-0605-sharded-contracts.md
akhi3030: Yes, reading https://github.com/near/nearcore/blob/ed1a56d4d33436c59ad1415b1722518ce34bfee2/runtime/runtime/src/lib.rs#L1529 more carefully, I see you are right.  I will fix this comment.

neps/nep-0605-sharded-contracts.md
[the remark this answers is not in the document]
wacban: Agreed, but I would suggest giving it a thought - it may be easier to do it upfront than to perform a migration later.

neps/nep-0605-sharded-contracts.md
wacban: The most generic approach would be to specify explicit access for each method. For each method the developer could specify that it can be called by some subset of (anyone, owner, other contract instance).

akhi3030:
Summary of discussion between Jakob, Anton P., and myself.

- separate the concept of global and sharded contracts
- an account can have at most 1 local or global contract and any number of sharded contracts
- sharded contracts need to be explicitly deployed to ensure no malicious state creation
  - we also considered not needing explicit deployment and deploying lazily when the sharded contract on an account is called.  However, this has the problem of maliciously creating state.
- user account pays for storage of sharded contract state
- sharded contract state deletion is not really figured out.  Maybe user can delete it somehow but that can be awkward for sharded contract developer.
- instead of changing receipt, add a new function call action to call sharded contracts.
- `create_storage_key()` will take an argument: ShardedOrNot.  In the non sharded case, we use `TrieKey::ContractData` as currently.  In the sharded case, we use new prefixes.
- sharded contracts should support both types of global contracts

I will update the NEP with the above discussions.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,477 @@
...
+        account_id: AccountId,
+        // Which sharded contract code on the destination account to call
+        sharded_contract_account_id: AccountId,
+        // In between a sender sending the message and the receiver executing
+        // it, the contract code could have been upgraded.  The two fields below
+        // allow the sender to specify which versions the receiver can be in
+        // when called.  If the receiver's version is outside the specified
+        // range, then the call is rejected.
+        minimum_version: Option<u64>,
+        maximum_version: Option<u64>,
```
jakmeier, line 114: Could this also be left to the contract code (SDK)? I'm not sure what we gain by adding this code to the runtime spec, it only makes things less flexible.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,477 @@
...
+
+Hence, we need a way to enable storage namespaces which guarantees that whatever state a sharded contract code is storing on an account, it cannot be modified by another contract code on that account.
+
+### Detailed specification
+
+With the high level requirements and the pseudocode presented, we can discuss the specification of the new primitives been proposed.
+
+#### `DeployShardedContractAction`
+
+This action is similar to the `DeployGlobalContractAction`.  Note that an account can have only a single sharded contract code deployed on it.  Processing this action will generate a `GlobalContractDistributionReceipt`.  We propose the following modifications to this receipt.
```
jakmeier, line 289: I think we should explore if we can make global vs local contract code completely orthogonal to sharded contracts. I feel it might be nicer.

We could probably get away without touching global contract distribution code. I think when a user uses `UseShardedContractAction`, all the network has to check is if there is a global contract with that contract identifier. That shouldn't require anything special on the code distribution side.

In that case, the difference between a global sharded contract and a global non-sharded contract is simply whether the user has submitted `UseShardedContractAction` or not.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,477 @@
...
+}
+```
+
+By prepending the account id of the current account to the key, each account gets a separate namespace which ensures that no other account can access their data.
+
+We propose extending this namespace to support per sharded contract namespaces.
+
+```rust
+enum ContractType {
+    LocalOrGlobal,
```
jakmeier, line 369: `LocalOrGlobal` isn't great, since Sharded contracts are also global. Maybe `NonSharded `?

jakmeier (commented):
@akhi3030  I left some comments inline that I think we should discuss.

Also, here are some suggestions from my first in-detail pass through: https://github.com/akhi3030/NEPs/pull/2
PTAL

neps/nep-0605-sharded-contracts.md
akhi3030: This is how I am thinking about this.

There are two types of versioning constraints apps might want to enforce.

1. The receiver wants to ensures that it is executing msgs from a sender in some version range.
2. The sender wants to ensure that only a receiver in some version range will execute msgs from it.

Further, as long as the receiver completely under our control, 2 can be achieved via 1.  However, this requires that the smart contract developers are very careful about upgrades and that when they upgrade, they make sure they update the constraints on 1 properly.

Further, this API allows one to call any sharded contract not just an instance of my sharded contract.  In this case, it is possible to imagine application scenarios where a receiver might intentionally not enforce 1 above to create corruption on the sender side.

Given these considerations, I see this as a second layer of defense.  

Having said all of that, I don't feel very strongly about this.

neps/nep-0605-sharded-contracts.md
akhi3030: I agree, if we could simplify that, that will be ideal.  The challenge is distributing the `version`.  

The more I think about it, the more I feel like `version` is adding a lot of additional complexity.  

One way, contracts could handle `version` entirely in the application space is by adding the version to the function names e.g. `receive_tokens_v1()`.

So when a sender sends a message, they are sending it to a specific version: `receive_tokens()_vX`.  And when a receiver upgrades, it is "rejecting" messages sent by earlier versions of the sender by replacing `receive_tokens_v[X]` by `receive_tokens_v[X+1]`.  

Do you think, the above would be good enough for apps to use and then we can kill the whole `version` business?

neps/nep-0605-sharded-contracts.md
jakmeier: > 2. The sender wants to ensure that only a receiver in some version range will execute msgs from it.

Good point, I did not think about that. That's interesting. 

Yeah, I think you are right, we want a strict way of checking that, especially when calling a sharded contract that is not in our control. The potential to swap out the code could be abused. Only the protocol can guarantee that there is no code swap.

This case seems quite tricky to me. Specifically, our current model seems unable to simultaneously solve all of these:

1. No malicious code swap
2. Updates are possible
3. Updates have no downtime for users

Number 1 is covered by global contracts defined by code hash. But then an update would mean we have to use an entirely different global contract, which means separate state namespaces for both versions. I think that breaks number 2. (Maybe reading state from other sharded contracts would make the update still possible?)

With your model of specifying the min & max version on the caller side, I think we can solve 1 + 2. But then I don't see how downtime can be avoided, since I have to wait until the new version of the sharded contract is distributed before I can update the code that's calling it.

Maybe we have to accept downtime for such cases? Not sure.

The same problem arises in an SDK solution, too. But the thing is, if we put something in the protocol, we should ensure it's really solid.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,493 @@
...
+    // TODO: Should we emit an ft_transfer events here?  Or at the receiver?  Both?
+
+    // This pseudocode assumes that `receive_tokens()` always succeeds.  In a
+    // more complete version, if `receive_tokens()` fails, then this function
+    // should undo the balance decrement above to ensure that tokens are not
+    // lost.
+}
+
+// A smart contract function used to receive FT from a transfer.
+fn receive_tokens(amount: Balance) {
```
akhi3030, line 180: I think I found a problem here.  What if the account owner end user calls this function directly?  They will be using the `ShardedFunctionCallAction` and at least how we have specified the checks here, they will pass.  We need to think about what `current_sharded_contract_info()` returns if the caller is an end user.  

CC: @jakmeier

neps/nep-0605-sharded-contracts.md
jakmeier: Nice idea. I believe such version naming convention would work, yeah. Might be more effective than having versions be part of the protocol.

The only downside I see is that we basically push the problem to developers, without much help. Solving it in the SDK in such a way might be possible but sounds quite hacky...

I'll try to think of alternative solutions before our next call.

neps/nep-0605-sharded-contracts.md
jakmeier: If they call it directly, wouldn't `predecessor_sharded_contract_info` return None? (Or panic)

neps/nep-0605-sharded-contracts.md
jakmeier: Also, this naming convention wouldn't protect from malicious code swap, right?
In that case, anytime a sharded contract uses a global code by account id, it has to fully trust the owner of that account, I guess.

akhi3030:
for those following the progress on the NEP here, we discovered an issue with the specification and are brainstorming about it [here]([#general > Sharded contracts @ 💬](https://near.zulipchat.com/#narrow/channel/295302-general/topic/Sharded.20contracts/near/513650664).

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,494 @@
...
+## Reference Implementation
+
+TODO
+
+## Security Implications
+
+TODO
+
+## Alternatives
+
```
attila0x2A, line 438: One interesting alternative for data storage (that you're likely aware of) is to take an approach similar to solana's with accounts having owners. In that case sharded contract can own a bunch of different account (e.g. one per user) and store relevant information there instead of storing it on the user's account itself. (I have no clear idea on the complexity involved here though.)

https://solana.com/docs/core/accounts

akhi3030: thanks for the link and the idea.  We have a similar account model too where a single account or access key can own many different accounts.  

there are some subtle problems that would have to be solved with this approach though:

- The FT contract has created account `A` for me to use.  So as a user, I need to remember that `A` is for me or I need to ask the FT contract for that info when I want to use the account.  So either, users need to store additional data on top of the usual keys or we still have a "centralised" database in the main account that users need to query often / all the time creating potential bottlenecks.
- if I want to send you tokens, then either I need to figure out the mapping between your account id and the id of the account created for you or my account does.  This will again require querying the centralised database.

I suppose we could use some kind of hashing to derive the account id?  If we hash (account_id of central FT contract + account_id of the user) and use that to create the new account, would that work?  As long as it is "guaranteed" that there will never be any collisions I suppose.

attila0x2A: > If we hash (account_id of central FT contract + account_id of the user) and use that to create the new account, would that work? As long as it is "guaranteed" that there will never be any collisions I suppose.

Are collisions really a problem? - I think it may be OK to store data for several users on the same data account. Maybe it's something that would even be beneficial in some cases (e.g. to make it easier to examine state for somehow related users).

akhi3030: To circle back on this, I had to do some digging but the main issue with this idea is that on NEAR one account cannot have access keys to another account.  So even if we get the FT contract to create accounts for the users, the users will still have to manage multiple set of access keys.  And one of the main high level requirements is that users should not need to have to maintain multiple set of access keys.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,701 @@
...
+Since all the account balances are stored in a single contract, this one contract has to be invoked in order to make any transfers.  Therefore, the maximum TPS of this contract is the maximum TPS of the shard it is deployed on.  The only way to increase this capacity would be to increase the capacity of the shard.  Adding more shards does not help.
+
+
+## High-level explanation
+
+In the centralised FT contract above, all the state is stored in a single centralised location.  The opposite extreme of this approach would be to store all the state in as distributed a manner as possible i.e. the account balances of each user is stored locally on their accounts instead.  This distributes it across shards and allows for horizontal scaling.
+
+Of course, we cannot just put the state in normal user storage without a way to mitigate them manipulating their FT balance.  We need a way to ensure only the real FT contract code can modify this state.
+
+To enable this isolation, we introduce a new action `SwitchContextAction`.  Within a receipt, all actions that follow a context switch are executed in a different context that has a separate storage namespace.
```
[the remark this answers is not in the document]
akhi3030: OK, I guess if there are other such limits in place, we are really making it clear that all the subcontracts / or whatnot will live together.  So this is fine as is.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,701 @@
...
+
+### Pseudocode for a sharded FT contract
+
+We start by showing what a sharded version of the above FT contract would look like if it were using our proposed changes.
+
+```rust
+// Pseudocode interface to various host functions
+trait HostFunctions {
+    /// Returns the account id of the parent account that created this sharded
+    /// subordinate account.  If this function is called by an account that is
```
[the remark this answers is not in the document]
akhi3030: yeah I think subcontract might be the way to go.

luciotato:
gm! my 2 cents:
Are you considering using this opportunity of sharded contracts to create a full decoupling of state and code?

With this MVP, would we be able to deploy a single FT code globally, and then execute the same code for all FTs?

Meaning that in order to create a new token, you only need to call the global FT code to create a "Mint" (state) and later use the "Mint" as the 'state' argument, to transfer those particular tokens.

e.g.
call ft.global.near using-state(usdt.global.near) method:ft_transfer(10000000, 'bob.near') signed by alice.near
where the args are: using-state(mint) (amount, destination)
This will imply a mechanism to store 'state' in accounts ('usdt.global.near') and make the the contract ft.global.near the 'owner' of the account (meaning the state inside usdt.global.near can only be modified by ft.global.near)
This also implies some mechanism to easily pass one or more 'state-accounts' to a contract (as long as the contract is the owner of the state-account)

With this structure, I guess you need less modification to the existing account model. 
1) The global code is immutable and has no state
2) The transaction executes in the shard of the state account
3) The state account still stores hashmaps with the balance of each user, so the code can do atomic transfers between users. Code does not change much vs. the actual NEP-141 FT token
4) predecessor is still alice.near, the code is the same, but the state changes if she's transferring usdt.global.near vs stnear.global.near (same predecessor, same code, different mint/state)

PRO: using this model (with state-accounts owned by stateless code) maybe it becomes easier to interact with legacy NEP-141 tokens

jakmeier:
Hi @luciotato , thanks for joining the discussion!

> With this MVP, would we be able to deploy a single FT code globally, and then execute the same code for all FTs?

Yes, you can do this already today. Using https://github.com/near/NEPs/pull/591 you can already share code between all FTs, if you don't have custom logic. (In practice, you probably have custom logic for your FT on top of the standard)

> Meaning that in order to create a new token, you only need to call the global FT code to create a "Mint" (state) and later use the "Mint" as the 'state' argument, to transfer those particular tokens.

Interesting idea! 

> The transaction executes in the shard of the state account

Uuh, that directly contradicts the whole reason we started this project :sweat_smile: 

The number 1 goal is to use bandwidth of more than one shard for a single FT. We are talking about the gas limit per shard here, not the storage bytes. That's where the big players are hitting limits.

> My Proposal V2 ...

Your second proposal has conceptual similarities with the current proposal. I think what you describe as `using-state`, is essentially the same as we call `SwitchContext` in the NEP text.

> ...
> In this case, the global code ft2.global.near receives 2 state arguments that it can modify. The 2 states "belong" to ft2.global.near so only the code at ft2.global.near can modify those states.

On which shard is this executed? In the NEP description, we would execute one part on `alice.near`'s shard and one part on `bob.near`'s shard in two separate steps. Did you have something else in mind?

> ...
> In this case, the code can alter the state at usdt2.global.near, incrementing the total token supply and also increment the balance at claude.near+usdt2.global.near by the same amount.

Again, on which shard are we operating here? The one where`claude.near` lives or the one where `usdt2.global.near` lives? You fundamentally can't modify state in both at once because all state in Near is sharded by accounts.

In our proposal, the global code can run on every account that is an FT holder. The global code can modify its protected share of state on each account. Accounts can call each other to send tokens between them, but every call needs to preserve the total_supply. Minting still has to originate from a central account but it can fan out efficiently if needed. (e.g. for sweat)

Can you please elaborate how your second suggestion differs from this proposal written down in the NEP?

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,695 @@
...
+    - `SwitchContextAction(Sharded("ft.near"))`
+    - `UseGlobalContractAction("ft.near")`
+    - `SetContextPermissionsAction(Limited { reserved_balance: 0 })`
+3. When `alice.near` wants to transfer tokens to `bob.near`, Alice calls the `send_tokens()` function on the sharded FT contract on her account using
+    - (`receiver_id="alice.near"`)
+    - `SwitchContextAction(Sharded("ft.near"))`
+    - `FunctionCallAction("send_tokens", balance=1, ...)`
+4. `send_tokens()` ensures that caller is `alice.near` and has 1 yocto NEAR attached, as full access to `alice.near` is required to initiate transfers.
+5. Next, the contract decrements the balance.  We will discuss access control issues to storage in the namespace section below.
+6. Then, it sends a sharded cross contract call from `alice.near` to `bob.near`
```
birchmd: I'm curious how sending tokens to an account that has never interacted with the FT before will work. Today you can cover the storage deposit on behalf of another account using the `storage_deposit` method on the FT contract. To mimic this functionality, will it be possible to create a new sharded context on another account other than your own by attaching a balance?

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,695 @@
...
+}
+```
+
+With the above in place, following is the flow for how a sharded FT contract would be used.
+
+1. The owner of `ft.near` uses `DeployGlobalContractAction` to deploy the code under the `ft.near` name.
+2. Before `alice.near` can use the sharded FT contract `ft.near`, she has to enable it on her account.
+    - (`receiver_id="alice.near"`)
+    - `SwitchContextAction(Sharded("ft.near"))`
+    - `UseGlobalContractAction("ft.near")`
```
birchmd: Is this action necessary? Since it is only possible to use the global contract code for `ft.near` in the `ft.near` context, it seems redundant.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,695 @@
...
+    - `SetContextPermissionsAction(Limited { reserved_balance: 0 })`
```
birchmd: It feels a little weird to me `SetContextPermissionsAction` is used after `SwitchContextAction` to set the permissions for that context; as opposed to directly saying which context should be affected in the `SetContextPermissionsAction` itself.

I thought the idea of `SwitchContextAction` was to allow reusing `FunctionCallAction` without needing to invent a new `ShardedFunctionCallAction` that duplicates most of the fields. For new actions (like `SetContextPermissionsAction`) I think it makes sense for them to be context-aware from the beginning instead of having to switch contexts to use them.

I imagine only the account owner (root context) should be able to set the permissions of sharded context, and validating that condition might be easier if you didn't need to switch contexts first.

birchmd (commented):
Thanks for continuing to write this up @jakmeier ! I like the direction it is going.

neps/nep-0605-sharded-contracts.md
jakmeier: That's one of the details I've been trying to figure out while implementing a real sharded FT. Haven't found a satisfying solution, yet.

First, I thought something analogue to `storage_deposit` work through a full access module that manages the deployment of FTs on every account (could be same as router). But then users have to set that up that full access module first, so it's not really a good solution IMO.


> To mimic this functionality, will it be possible to create a new sharded context on another account other than your own by attaching a balance?

I like this direction! I think we could say it's possible to create a sharded context on foreign account with limited access. By default the storage bytes above the ZBA limit would be 0 but we could even allow setting another limit with `SetContextPermissionsAction` within the boundary of the attached deposit.

The problem: Currently, this NEP assumes that an account can't be deleted if they are using a sharded contract with state. Thus, if we allow creating contexts on other accounts, one can also prevent other accounts from being deleted.

If we allow deleting, some other things become more complicated. For example, I can't think of a way to track `total_supply` in a sharded FT properly if users can delete their share of the state.

But even if we get around that, I don't know how we could handle deletion of a sharded contract context with more state than can be deleted in a single transaction. It would fail like DeleteAccount can fail today. And anybody could create such state on any other account. So again, attackers can prevent any account to be deleted. Not sure why they would, other than for trolling... But maybe some factory contract out there rely on deletion? :shrug: 


@birchmd Curious to hear your thoughts on how to handle this!

neps/nep-0605-sharded-contracts.md
jakmeier: It's not strictly necessary. But if the work of `UseGlobalContractAction` is merged into `SwitchContextAction`, it means the gas cost for executing `SwitchContextAction` has to include the creation of a sharded context. And as explained in the other comment, explicitly including this action would be how one would decide to pay for the creation of the context, if it doesn't exist yet.

We could of course use a new action that doesn't redundantly specify the global account id again. But I thought it's nicer to reuse the existing `UseGlobalContractAction`. It also keeps the door open to later allow non-sharded modules per account without inventing a new API. It would simply use something like: 

- `SwitchContextAction( Loca("sync-module-xzy" ) + DeployContractAction(...)` to deploy code by binary
- `SwitchContextAction( Loca("sync-module-xzy" ) + UseGlobalContractAction("extension.xyz.near" )` to use the global contract code optimization, without the sharded contract protections / limitations

neps/nep-0605-sharded-contracts.md
jakmeier: Yeah, I think you are right. We can have instead:

- (`receiver_id="alice.near"`)
- `SetContextPermissionsAction { context: Sharded("ft.near"), permission: Limited { reserved_balance: 0 }) }`
- `SwitchContextAction(Sharded("ft.near"))`
- `UseGlobalContractAction("ft.near")`

I didn't like that now I have to specify `"ft.near"` three times, the way I used the other actions.
Maybe we can simplify it:

- (`receiver_id="alice.near"`)
- `SetContextPermissionsAction { context: Sharded("ft.near"), permission: Limited { reserved_balance: 0 }) }`

To create a context in another account, we could say `SetContextPermissionsAction` is allowed by a foreign caller, if the context currently has no permissions defined.

I think this can work. It means that the gas fees for `SetContextPermissionsAction` need to cover the work of using a global contract as sharded contract module, which is essentially the fees of existing `UseGlobalContractAction` plus the burnt gas to cover the ZBA storage. This makes changing permissions later more expensive. But maybe nobody cares about that, as I don't think changing permissions will be a frequent action.

I think I'll update the NEP to follow this design.

neps/nep-0605-sharded-contracts.md
jakmeier: As mentioned in the other comment, I think we can just use `SetContextPermissionsAction` to "activate" a sharded context with the correct code, no need to reuse `UseGlobalContractAction` with a redundant argument.

The API for non-sharded (local) contexts could still be as I mentioned above, I don't see any conflicts there.

jakmeier:
Quick update on the sharded FT contracts: I have added a detailed explanation of how a sharded FT contract could work over at https://github.com/near/NEPs/issues/614, including a reference implementation that uses the new protocol API as described in this NEP.

Kind of mind-blowing: We can actually execute a sharded version of ft_transfer_call() in 1 fewer block latencies compared to NEP-141! :exploding_head: 

cc @graphite @birchmd  @bowenwang1996

neps/nep-0605-sharded-contracts.md
birchmd: > By default the storage bytes above the ZBA limit would be 0 but we could even allow setting another limit with SetContextPermissionsAction within the boundary of the attached deposit.

I think we need to be careful with this because if we do allow external contracts to create sharded instances then we want to make sure they are paying for the storage they use. Otherwise a troll could use up another account's ZBA without their permission. I think we would need to enforce the attached balance be greater than or equal to the amount needed to cover the ZBA and also set the permissions limit to 0 so that it is impossible to use up the free space on another account.

> an account can't be deleted if they are using a sharded contract with state.

So once an account has one sharded contract state then adding more does no harm. Maybe that means we should allow external accounts to create (by paying for) sharded state as long as there is already at least one sharded context present. The user of a new account would need to add the first sharded contract themselves, but I think that is ok (it can be part of some welcome to Near on boarding process). At least then most of the time receiving a new token will be relatively smooth.

neps/nep-0605-sharded-contracts.md
jakmeier: updated just now

neps/nep-0605-sharded-contracts.md
jakmeier: updated just now

neps/nep-0605-sharded-contracts.md
jakmeier: Yeah, making the caller cover for all costs is important. Good point about using up the ZBA limit, too.

> The user of a new account would need to add the first sharded contract themselves, but I think that is ok (it can be part of some welcome to Near on boarding process)

Hm, that makes the first experience of using NEAR quite a bit more involved than I'd like it to be. Ideally, one should be able to receive an FT from a CEX directly onto an implicit account, without creating the account, or even following an onboarding process.

What about something like:

- Allow users to delete sharded context state. But before it's deleted, a clean up function `on_delete` on the WASM contract is called. This function could emit events and make cross contract calls.
- `on_delete` can use at most 10 Tgas (Exact number can vary) and this is already paid for when the sharded context gets created. Hence, triggering deletion can be for free (aside from the receipt creation base cost of 0.2 Tgas).
- Now when calling `ft@alice.near` from `ft@***` (same sharded context), we allow creation of the context with limited access, even without the user consent. But that's okay because removing undesired sharded contexts can be one simple and cheap click in a wallet.

I don't necessarily like the added complexity for an MVP.... But I think it might be worth it. It keeps the user in control of deleting state of their account, while also giving sharded contracts a way to be protected from corrupted state. Without this, I see no good story around deletion of contexts, which then also limits how we can do creation of contexts.

WDYT @birchmd?

wacban:
I don't know about the necessity of addressing by contract hash but:
* If we don't add it now and need to add it later, it'll likely be even worse
* In the option with supporting both, rather than duplicating the method per type of contract can you add an enum { AccountId, ContractHash } as first argument?

neps/nep-0605-sharded-contracts.md
birchmd: I like this idea! Keeping the on-boarding simple while allowing users to delete unwanted data and still giving developers control of how to handle deleted together form a strong enough argument to warrant the additional complexity in my opinion.

birchmd:
> Low-level host functions of the NearVM are gonna be more numerous if we want to support both ways.

Why do we need to have different versions of the host functions for global contracts by account id as opposed to by hash? Couldn't the ID used in the sharded context be an enum which can encode either an account id or hash?

graphite:
I think the by hash approach allows an immutable out of the box implementation, so this is a use case that I would rather keep

jakmeier:
Okay, I will keep the sharded contract by hash code in that case. Thanks for all the feedback! :)


Answering both 

@wacban
> In the option with supporting both, rather than duplicating the method per type of contract can you add an enum { AccountId, ContractHash } as first argument?

@birchmd
> Why do we need to have different versions of the host functions for global contracts by account id as opposed to by hash? Couldn't the ID used in the sharded context be an enum which can encode either an account id or hash?

We need to represent the enums `enum ContractContext` somehow, which has `enum GlobalContractIdentifier` nested.

```rust
pub enum ContractContext {
    Root,
    Sharded { code_id: GlobalContractIdentifier },
}

pub enum GlobalContractIdentifier {
    CodeHash(CryptoHash),
    AccountId(AccountId),
}
```

which can also be flattened to

```rust
pub enum ContractContext {
    Root,
    ShardedByAccountId { account_id: AccountId },
    ShardedByCodeHash { code_hash: CryptoHash },
}
````


For context:

- We haven't used enums in any near host functions so far. Generally speaking, we try to keep the interface types at numbers, strings, and black-box `Vec<u8>` only. 
- `promise_batch_action_stake` and a few other methods are exceptions that use the borsh encoding of `PublicKey`.
- The `alt_bn128` family of host functions are another exception, using a simple ad-hoc encoding, independent of borsh. The same goes for `ecrecover`.
- Most recently we added `promise_batch_action_deploy_global_contract` and `promise_batch_action_deploy_global_contract_by_account_id` as two host functions, rather than using `GlobalContractIdentifier` as an enum on the ABI level. If we now also add an enum for `GlobalContractIdentifier` as part of `enum ContractContext`, it will be inconsistent unless we also change those other host functions.

For NEP-605, we could extend the Near VM ABI to have a general notion of enums. Maybe embrace borsh for it and lock it in as a dependency for smart contract?

```rust
/// writes `enum ContractContext` to register using borsh encoding
fn current_sharded_context(register_id: u64);
```

Replayability and compatibility when adding new enum variants could be fun in this one. But probably solvable.

Or we could do something ad-hoc once again, like this:

```rust
/// The return value defines the result written to register
/// 
/// 0:           context is root and nothing was written to the register
/// 1:           context is sharded and `AccountId` was written to the register
/// 2:           context is sharded and `CryptoHash` was written to the register
/// >3:         unknown future context
fn current_sharded_context(register_id: u64) -> Result<u64>;
```

I'm just not sure if this is better or worse than multiple host functions with simple typing on the ABI-level.

```rust
/// writes `AccountId` to register
/// Returns a bool indicating success (1) or failure (0) as a `u64`.
/// Failures means this is not a sharded context by account id.
fn current_sharded_context_by_account_id(register_id: u64) -> Result<u64>;

/// writes `CryptoHash` to register
/// Returns a bool indicating success (1) or failure (0) as a `u64`.
/// Failures means this is not a sharded context by hash.
fn current_sharded_context_by_code_hash(register_id: u64) -> Result<u64>;

/// Returns a bool indicating true (1) or false (0) as a `u64`.
fn current_sharded_context_is_root() -> Result<u64>;
```

Note that all of these can be wrapped the same way with SDK-level types anyway.

Maybe we should consult @nagisa. What are your thoughts on adding an enum type in host functions?

nagisa:
> Maybe we should consult @nagisa. What are your thoughts on adding an enum type in host functions?

You can't really do that in literal sense. The best way you get an "enum" is an argument for discriminant and another multipurpose argument(s) that gets appropriately encoded based on the first argument. For `ContractContext`-like thing the 2nd argument can be a simple pointer to bytes in memory (or `0`.)

I encourage not looking for ways to make this generic. Definitely not borsh or any other serialization format. JSON being a dependency for contracts by default is already a pain enough. I can imagine it involving passing a pointer to a `#[repr(C)]` structure in wasm memory which would have to be manually decomposed on the host side (as `#[repr(C)]` don't have to mean the same thing for both targets,) though.

wacban:
Brace yourself, noob question coming :) 

Why can't we use regular rust enums?

nagisa:
https://doc.rust-lang.org/reference/type-layout.html is a great reference here. In particular

> There are no other guarantees of data layout made by this [Rust] representation.

Besides the fact that Rust does not guarantee layout unless `#[repr(C)]` or similar are used, we also need to worry about the fact that these host functions are at a boundary of two very different targets, so for the most part the same `repr(C)` type definition does not mean the same thing on `wasm32-unknown-unknown` side as it does on `x86_64-unknown-linux-gnu`.

jakmeier:
> I encourage not looking for ways to make this generic. Definitely not borsh or any other serialization format. JSON being a dependency for contracts by default is already a pain enough. 

Agreed.

> I can imagine it involving passing a pointer to a `#[repr(C)]` structure in wasm memory which would have to be manually decomposed on the host side (as `#[repr(C)]` don't have to mean the same thing for both targets,) though.

So, my idea was to use a discriminant that breaks it down to a `Vec<u8>` which can be interpreted as string or hash. It was my impression that's what we have in most existing host functions. In that case, we don't really have to worry about `#[repr(C)]`. We just keep using the encoding of ptr + len that we have always used. 

Further, to stay consistent with other methods, I suggest we pass values from host to guest in registers, like we have  in [`current_account_id()`](https://github.com/near/nearcore/blob/bd137861ff32e3d96c3d1f1def792f188d7aee49/runtime/near-vm-runner/src/logic/logic.rs#L557).

That gives this interface that I suggested.

```rust
/// The return value defines the result written to register
/// 
/// 0:           context is root and nothing was written to the register
/// 1:           context is sharded and `AccountId` was written to the register
/// 2:           context is sharded and `CryptoHash` was written to the register
/// >3:         unknown future context
fn current_sharded_context(register_id: u64) -> Result<u64>;
```

The other direction, reading from guest to host will be necessary for promise batch functions. I would do it as it's already done for example in [`value_return`](https://github.com/near/nearcore/blob/bd137861ff32e3d96c3d1f1def792f188d7aee49/runtime/near-vm-runner/src/logic/logic.rs#L3001) which reads a `Vec<u8>` from memory or register using a magic value of `len == u64::MAX` as follows:

```rust
pub(super) fn get_memory_or_register<'a>(
    gas_counter: &mut GasCounter,
    memory: &'a Memory,
    registers: &'a Registers,
    ptr: u64,
    len: u64,
) -> Result<Cow<'a, [u8]>> {
    if len == u64::MAX {
        registers.get(gas_counter, ptr).map(Cow::Borrowed)
    } else {
        memory.view(gas_counter, MemSlice { ptr, len })
    }
}
```

Does that make sense to you @nagisa ?

jakmeier:
@pugachAG is helping me to finalize the design. We just had a great discussion with the an outcome on how to simplify the MVP further.

1. Instead of reserved balance for limited subcontracts, we can use non-refundable storage for subcontracts. (Conceptually what #491 would have introduced, just limited to subcontracts.) Specifically, in the `SwitchContext` action, we can add a token amount, which is permanently burned and permanently increases the storage allowance of the subcontract. The same concept can work for limited and full access subcontracts. No more storage sharing between main- and subcontract.
3.  Thanks to (1), all zero-balance considerations are void, we can remove it from the NEP. As a side-effect, this makes lazy creation cheaper, as it does not need to pay the full ZBA storage cost.
4. Let's disallow deletion of sharded contracts. It makes implementing sharded contracts much simpler. And with (1) in place, users cannot get a refund for the storage allowance anyway.
5. Thanks to (3), the on_delete_hook is no longer necessary.

This makes sharded contracts easier to implement, at the cost of never refunding storage and disallowing deletion in general. Given how rare the deletion case is, we believe this is a worth while trade-off. However, if someone comes out with a strong use case for deletion, we are willing to reconsider and add back in the more complex solutions.

jakmeier:
@near/wg-protocol @near/nep-moderators  I believe this proposal is at a stage where we can start the review and voting process. Can someone of you please initiate it?

The only detail unspecified in the proposal are the exact gas prices. I will add them within the next week but I don't expect this to play a big role in the design discussion.

The reference implementation is not ready as a real PR to be merged to nearcore but it definitely shows that this proposal can be implemented as described. I believe that was the standard we need for a WG voting? Otherwise, just let me know what more is needed.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/0000
+Type: Protocol
+Version: 0.0.0
+Created: 2025-04-07
+LastUpdated: 2025-07-11
+---
+
+## Summary
+
+Today, a single contract is limited to the transactions per second throughput (TPS) of a single a shard, ergo a contract already at the TPS limit of a single shard cannot benefit from increase in TPS made possible by increasing the number of shards.  This NEP builds on top of the global contracts NEP-591 and adds the necessary tools to enable sharded contracts.
```
nagisa: Ergo is a nice fancy word, but it also is quite unnecessarily fancy. It being rare may force readers to grab their copy of oxford dictionary if they want to be sure they understood the sentence correctly.

```suggestion
Today, a single contract is limited to the transactions per second throughput (TPS) of a single a shard, therefore a contract already at the TPS limit of a single shard cannot benefit from increase in TPS made possible by increasing the number of shards.  This NEP builds on top of the global contracts NEP-591 and adds the necessary tools to enable sharded contracts.
```

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+A sharded contract spreads its state across all user accounts, rather than storing it all in one place.  This becomes possible to implement securely once users can have multiple isolated subcontracts in one account that are guaranteed to run the same sharded contract code.
```
nagisa: Nit: wording precision.

```suggestion
A sharded contract spreads its state across shards of accounts that use it, rather than storing the entire state in a single shard. This becomes possible to implement securely once users can have multiple isolated subcontracts in one account that are guaranteed to run the same sharded contract code.
```

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+With that, sharded contracts will be able to seamlessly scale to use the entire TPS capacity of the network.
```
nagisa: Wording nit: keep tenses uniform (previous paragraph talks about sharded contracts in the present tense.)


```suggestion
With that, sharded contracts are able to seamlessly scale to use the entire TPS capacity of the network.
```

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+
+## Motivation
+
+As a single contract is deployed on a single shard, the maximum TPS that it can have is the maximum TPS of the single shard.  Horizontally scaling (i.e. increasing the number of shards) a blockchain is easier than vertically scaling (i.e. increasing the TPS of a single shard).  Once, all the software bottlenecks are addressed, the only way to vertically scale a shard is by requiring the validators to use faster machines.  Faster machines are more expensive and thereby hurts decentralisation.
+
+Without spreading the contract across multiple accounts, a single contract will therefore remain bound by the TPS throughput of a single shard regardless of how many shards are added to the network.  But with today's protocol, the options to spread state across multiple accounts are very limited.
```
nagisa: Nit: this sentence does not pass scrutiny as the previous paragraph already presented an alternative mechanism to continue scaling vertically albeit with its own caveats. Its also somewhat of a jump of thought that makes the section somewhat difficult to follow. How about piggy-backing on vertical scaling line of thought:


```suggestion
Vertical scaling opportunities are limited, imposing a hard upper limit on the throughput of a single-shard contract. This limit remains an issue for the contract regardless of how many shards are added to the network. This is especially true with today's protocol; options to spread state across multiple shards (i.e. to scale horizontally) are very limited.
```

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+1. One could deploy multiple worker contracts on different accounts to benefit from multi-shard throughput.  In that case, developers need an off-chain load-balancing solution to distribute users among the workers.  Workers among themselves need to communicate with cross-contract function calls to execute any operation that involves user state from two different user groups.
+2. State could also be stored on every user account, which needs to run the shared contract code.  Global contracts NEP-591 makes this economically feasible.  However, this means one account can only use one contract.  For example, if the accounts holds an FT that's sharded in this way, the user cannot buy an NFT and hold it on the same account.
+
+Neither of these architectures is deemed satisfying.
```
nagisa:
```suggestion
Neither of these architectures is satisfying.
```

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+This NEP proposes solving this problem by introducing new protocol level primitives which are necessary to get around these limitations.
```
nagisa: Nit: please clarify what "this problem" and "these limitations" refer to. Is it the scaling/throughput limitations? The limitations of contract developers sharding their contract to enable horizontal scaling? Something else?

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+Discussion https://github.com/near/NEPs/issues/614 has a concrete example of how to implement a sharded FT contract using the proposed changes.
```
nagisa: Seems like this paragraph would be more appropriate for the Usage Guide section.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+
+## Specification
+
+
+### Subcontracts
+
+A *subcontract* is an isolated module inside an account.  Aside from the main contract, an account can have up to `N` subcontracts deployed, each with separate code and state.
```
nagisa: `N` is not referenced again anywhere in the text in context of limits. The only other time `N` appears as a variable is in the discussion of subcontract upgrades. Please clarify this part of the text. Even better, cross-link to the parts of the specification that describe the limit on the number of subcontracts.

(And separately a cross-link to the part that describes what an "isolated" module means; probably "Isolation requirements"?)

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+
+### Contract Context
+
+Every subcontract is created inside a certain *context*.  Contexts are defined by a pair of an account id and a contract context.  The main contract uses the root context, while subcontracts have to use a different context.
```
nagisa: > Contexts are defined by ... a contract context.

seems recursive. Is a context a different concept from a contract context? Any chance there is a different term we could use to differentiate the two better? Especially since the following paragraph talks about "contexts" but none of the variants are a pair of an account id and something else.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+This NEP introduces three contexts:
+
+- `Root` - Runs the code deployed directly on the account. Reserved is for the main contract, which is technically not a *subcontract*.
+- `ShardedByAccountId { account_id: AccountId }` - Runs code deployed globally under the given account id.
+- `ShardedByCodeHash { code_hash: CryptoHash }` - Runs code deployed globally under the given code hash.
+
+For sharded contexts, the code is implicitly defined by the context fields.  An account can use any global contract code in this way  Even if this is against the original intention of the smart contract developer.  We do not see any security concerns with allowing this.
```
nagisa:
```suggestion
For sharded contexts, the code is implicitly defined by the context fields. An account can use any global contract as a sharded contract, even if doing so was not intended by the smart contract developer.  We do not see any security concerns with allowing this.
```

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+While not necessary for sharded contracts, this design leaves the door open to add more contexts in the future.  A future proposal could add a way to deploy multiple local modules without the need to deploy the code globally first.  For example, using a context like `LocalSubcontract { module_name: String }`.  The code for such a subcontract could be deployed using `DeployContractAction` inside the context.
+
+
+### Isolation Requirements
+
+This proposal allows holding multiple subcontracts in the same account.  But if every subcontract has the same tools available as the main contract, this makes building secure sharded contracts difficult if not impossible.
```
nagisa: Please clarify what 

> holding multiple subcontracts in the same account

means exactly; and/or crosslink to document(-ation) describing what it means to **hold** a (sub-)contract.

My current intuition reading the NEP top-to-bottom is that there is neither a concept of deploying or otherwise associating the subcontract to an account. Only that there is some sort of a context determined by the pair of the account's account id and the subcontract's account id/hash (which then probably allows isolating the state jointly owned by the account and the subcontract?)

EDIT: though that intuition is overriden a couple paragraphs later. Lets make sure to always use "deploy" related wording such as ("deploy multiple contracts to the same account", "have multiple contracts deployed in the same account", etc.)

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+A bit of necessary background.  Code running inside the main contract has full access on the account.  It can, among other things, send all tokens to an arbitrary address, delete itself, or even redeploy its own code.  The limitations are only set by the source code of the contract.  This makes sense if users only deploy code to their account that they have either written themselves or vetted thoroughly.
+
+With the introduction of sharded contracts, users will deploy **untrusted** subcontracts on their account on a regular basis.  For example, receiving an FT requires installing the corresponding subcontract on the user account.  It is not feasible to assume users vet the code of every meme coin they get airdropped.  Therefore, the first isolation requirement is that the permissions of subcontracts can be limited.
```
nagisa: > installing

Similar nit, please stick to "deploy" everywhere, unless "installing" is meaningfully a different operation from "deploying" in which case lets use "install" or whatever other name consistently throughout to mean that distinct operation.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+Deleting subcontracts is forbidden, to completely shut off state tampering.
+
+
+#### Subcontract Permissions
+
+Most subcontracts only need to access their own state and make calls to other subcontracts on other users.  However, it would be too limiting to disallow all other actions entirely.  We propose a binary permission system, akin to how access keys in NEAR Protocol work today.
+
+- `FullAccess` - A subcontract with this permission has the same capabilities as the main contract.
+- `Limited` - A subcontract with this permission is forbidden to use most actions in outgoing receipts.  Only function calls to non-root contexts are allowed.  No NEAR balance can be attached to outgoing calls.
```
nagisa: > Only function calls to non-root contexts are allowed.

nit: this is motivated in the next section so maybe some pointer/x-ref to that section would be useful.

nagisa: Side note: does this permission cover other function-call-like actions (such as yield) or not? If not, any reason why?

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+#### Cross-Contract Call Authorization
+
+Today, for cross contract calls, the `predecessor_id` is used for authorization in the receiver's code.
+
+With this proposal, subcontracts making cross-contract calls will have the same `predecessor_id` as calls from the main contract on the same account.  However, this should not allow to act on behalf of the main account on other contracts.  The perfect example is an existing FT contract.  A limited subcontract must be prevented from calling `ft_transfer` on a `usdt.tether-token.near`, for example.
+
+For new code that is aware of subcontracts, we propose a new host function `predecessor_context()`.  (Exact definition in the detailed specification section.)  If this is not the root context, the call comes from a subcontract and should not have access to NEP-141 fungible tokens.
+
+Crucially, however, contracts deployed prior to the introduction of contract contexts did not check the predecessor's context.  Therefore, the proposal disallows calling root contracts from a limited permission subcontract.  Allowing to call non-root is considered save since they can only be deployed after the introduction of contexts.
```
nagisa:
```suggestion
Crucially, however, contracts deployed prior to the introduction of contract contexts did not check the predecessor's context.  Therefore, the proposal disallows calling root contracts from a limited permission subcontract. Restricting calls to non-root contexts is considered safe because they can only be deployed after the introduction of contexts.
```

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+
+As described in specification below, we decided to forbid all balance access to limited-access subcontract, while giving access to the main account balance to any subcontract with full access.
+
+
+#### Isolation of Storage limits
+
+Subcontracts require storage for their context meta data, as well as for the namespaced state modified by WASM code.  How exactly this is paid for is not essential for the core functionality of sharded contracts.  After considering different options, we propose to use a non-refundable storage model, where tokens to cover storage are permanently burnt.
+
+This is in contracts to existing (main) accounts.  They initially did not burn anything. Instead, they must always hold a certain amount in NEAR balance to cover their storage cost.  However, this came with the problem of refund attacks on sponsored account creation.  Therefore, a zero-balance account (ZBA) limit of 770 bytes per account was introduced in #448.  This makes every account burn enough tokens to cover 770 bytes upon creation.  The refundable storage model is only applied to accounts that use more than that limit.
```
nagisa:
```suggestion
This is in contrast to existing (main) accounts.  They initially did not burn anything. Instead, they must always hold a certain amount in NEAR balance to cover their storage cost.  However, this came with the problem of refund attacks on sponsored account creation.  Therefore, a zero-balance account (ZBA) limit of 770 bytes per account was introduced in #448.  This makes every account burn enough tokens to cover 770 bytes upon creation.  The refundable storage model is only applied to accounts that use more than that limit.
```

But I would also look at rewriting the entire paragraph or even removing it entirely. It adds some background information but does not really do anything to motivate (or demotivate) the non-refundable model for the subcontracts which require it due to the requirement that subcontract state is never removed. So lets just mention that and move along.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+Aside from the isolation requirements described above, there are a couple additional things needed to make sharded contracts work smoothly.
+
+
+#### Guaranteeing the Existence of the Receiver
+
+We propose that any subcontract to subcontract call can opt-in to pay for the subcontract creation (with limited access) if the receiver account exists but the subcontract has not been created, yet.
+
+Without this, imagine a user wants to transfer an amount of FT from a centralized exchange to their self-custody near account.  This user would have to manually enable the token on their account, after they created the account.  This is not intuitive and not how FT's work today. Not on NEAR Protocol, nor on other popular chains.
+
+With the proposed opt-in lazy subcontract creation, an exchange can first try sending tokens the cheap way (no lazy creation)  If that fails they can repeat the same call but opt-in this time and pay the increased gas cost to cover the subcontract creation.
```
nagisa: What does subcontract creation mean here? It seems to suggest that something more than just `call_subcontract("account.near", subcontract_hash, function_name, ...)` happens. Meanwhile my model is that deployment cost at least is obviated due to global contracts being used for the sub-contracts. The only extra cost I can imagine is that of calling the `init`-like method on the contract to initialize its state.

nagisa: I guess another case is when `account.near` does not exist at all... but doesn't creating an account itself require sending NEAR, not just FTs & attaching gas?

nagisa:
Some thoughts I had as I was reading that I'm writing down before they get lost:

* what happens if near is attached to a sub-contract call?
* what happens when an account is deleted? will that retain the sub-contract state? What happens if a new account is later created with the same account name? Does it inherit the subcontract state of the deleted account?
* root contracts being on the hook to "not forget" to check `predecessor_context` is a footgun. Seems difficult to make sure that footgun does not fire – devs won't be actively thinking about it all the time, for instance. Is there really no way to somehow mark contracts as "subcontract aware" instead and only allow calls to those contracts? The contract would only be marked as such with an explicit action by the contract developer.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+enum SubcontractPermission {
+    FullAccess,
+    Limited,
+}
+```
+
+The detailed rules for permissions are:
+
+- A contract deployed with `FullAccess` permissions can do anything the main account can do.  This includes all actions, without additional limits.  Tokens attached to a function call or in a `TransferAction` are taken from the main account's balance.
+- A contract deployed with `Limited` access can only produce outgoing receipts with `FunctionCallAction` and `SwitchContextAction`.
```
nagisa, line 178: I asked elsewhere, but any reason to not support other `FunctionCallAction`-like actions (such as yield-resume?)

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+    }
+    FunctionCallAction { ... }
+] 
+```
+
+The called contract can use the `current_context()` host function to read the own contract context (`target`) and the `predecessor_context()` host function to read the calling contract's context (`caller`).  The only reason why `caller` exists on `SwitchContextAction` is to enable this host functions.  Outgoing receipt validation must ensure the caller is correct.
+
+The rules for using `SwitchContextAction` are:
+
+- Receipts created from a transaction must always set `caller = ContractContext::Root`.
```
nagisa, line 233: Currently all transactions only ever produce a single receipt. One or two things in the implementation of the transaction runtime depend on this detail. Is this NEP suggesting that transactions that deal with subcontracts would produce more than 1 receipt (given that `SwitchContextAction` is not particularly useful by itself)?

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+    FullAccess,
+    Limited,
+}
+```
+
+The detailed rules for permissions are:
+
+- A contract deployed with `FullAccess` permissions can do anything the main account can do.  This includes all actions, without additional limits.  Tokens attached to a function call or in a `TransferAction` are taken from the main account's balance.
+- A contract deployed with `Limited` access can only produce outgoing receipts with `FunctionCallAction` and `SwitchContextAction`.
+  - Any `FunctionCallAction` must be in a non-root context.
```
nagisa, line 179: This effectively means that `subcontract` in a context with `Limited` access cannot create `SwitchContextAction` with `target: ContractContext::Root`, correct? `FunctionCallAction` does not change and it has no concept of the context by itself?

I ask this because both of this line not being super clear but especially

> - Any context can be entered without special permissions.

below which seems to contradict my understanding of the requirements and the implementation approach (so far.)

EDIT: I see this clarified later, in which case make sure these points agree with each other or remove duplication.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+- Receipts created from a transaction must always set `caller = ContractContext::Root`.
+- Receipts created from a sharded contract must always set `caller = ContractContext::ShardedBy*` with their respective code id.
+- Any context can be entered without special permissions.
+- Anyone can create any subcontract on any account by setting `create_missing_subcontract` to true.  (See [Implicit Subcontract Creation](#implicit-subcontract-creation))
+- `SwitchContextAction` will fail if the subcontract does not exist and `create_missing_subcontract` is set to false.
+
+
+The rules inside a sharded context are:
+
+- Inside a `ContractContext::ShardedBy*` context with limited permissions, the only allowed actions are `FunctionCallAction`, `SwitchContextAction`.
```
nagisa, line 242: Does this mean that it would be an invalid receipt even if the `Root` context with full permissions were to create a batch of `SwitchContextAction` and some other action that's not permitted here (e.g. `TransferAction`?) Seems somewhat awkward to have to switch back, but I guess I can understand wanting to have defense in depth.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+- Inside a `ContractContext::ShardedBy*` context with limited permissions, `SwitchContextAction` cannot target a `Root` context.  (This is to prevent calling context-unaware contracts from a sharded context. Those contracts only check predecessor_id and generally assume the caller has full access on that account.)
```
nagisa, line 243: Does this mean it would be an invalid for a `Root` context to construct a batch receipt that first switches to a `ShardedBy` context and then right back to `Root`? How would a `Root` contract make a x-contract call to a subcontract in that case and then tack on follow-up actions?

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+        context_len: u64,
+        context_ptr: u64,
+    ) -> Result<()>;
+```
+
+
+#### Changes to Existing Host Functions
+
+- `storage_usage()` in any subcontract will return the storage usage of just subcontract, without the main account.
+- `storage_usage()` in the main account will include the storage usage of all subcontracts usage above their respective ZBA limit.
```
nagisa: What's the reasoning for this? It'll make this host call pretty expensive as it has to walk an entire subtrie of a certain key prefix.

neps/nep-0605-sharded-contracts.md
```diff
@@ -0,0 +1,838 @@
...
+- `account_balance()` in a limited access subcontract will always return 0.
+- `account_locked_balance()` in a limited access subcontract will always return 0.
```
nagisa: Why not an error? Misuse of other host functions does properly err.

neps/nep-0605-sharded-contracts.md
jakmeier: Great point! I simple haven't thought about yields much. Akhi suggested we avoid them in an MVP due to added complexity. But now I think it would be a weird omission. I will need to look at it closer but my intuition is that yields are the one other action we should allow even from limited contexts.

neps/nep-0605-sharded-contracts.md
jakmeier: We don't create the account if it is missing.

Creating a subcontract is basically just creating a `Subcontract` object that tracks meta data about the subcontract (permissions, storage usage, and storage allowance) and storing it in the trie.

If you want to see it in code, here are the relevant lines in reference implementation:
https://github.com/near/nearcore/compare/master...jakmeier:nearcore:sharded-contracts?expand=1#diff-318756fbcf83731e0211cbc53dad981140af78c2b06dce9de383c06cbd93d6efR1117-R1126

To pay for the required storage allowance, the sender indeed will attach NEAR, not only gas. The amount is specified in the `added_storage_balance` field of `SwitchContextAction` that triggers the implicit creation.

Now that you pointed this out... I realise this means a limited subcontract cannot use this field, as it cannot spend NEAR... This was not the intention. I think this should better be converted to gas cost on the caller side. I will need to think about it some more to understand the consequences of that change but probably I will update the proposal to pay for the storage in the equivalent gas amount.

neps/nep-0605-sharded-contracts.md
jakmeier: > This effectively means that subcontract in a context with Limited access cannot create SwitchContextAction with target: ContractContext::Root, correct? FunctionCallAction does not change and it has no concept of the context by itself?

Yes

> I ask this because both of this line not being super clear but especially
>
> Any context can be entered without special permissions.
>
> below which seems to contradict my understanding of the requirements and the implementation approach (so far.)

Good point, I tried making it clearer now, saying that this is only true for non-root contexts.

neps/nep-0605-sharded-contracts.md
jakmeier: Transactions should still only produce one receipt. A receipt is for one account but it can touch multiple subcontracts and make calls on each of them.

The runtime already allows making more than one function call in the same receipt, which already prepares the Wasm code multiple times. I don't see a problem with switching the context to select a different Wasm code and storage namespace between the calls. But please let me know if I am missing something!

neps/nep-0605-sharded-contracts.md
jakmeier: Yes that is he current proposal. But I don't have a good reason why it could not be opened more. I just don't see the value of it.

neps/nep-0605-sharded-contracts.md
jakmeier: Yes, kind of, that's what it means.

But I think you are missing that whatever "follow-up" action you need, the caller can put it in the batch before the context switch. As far as I know, there is no difference in the ordering at all.

- There is no way for different actions to depend on each other's outputs.
- All actions inside a receipt are atomically successful or if any actions fails, they are all reverted.

We could probably allow switching back to root but I simply don't see any use cases for that.

neps/nep-0605-sharded-contracts.md
jakmeier: So IMO this number should be however much storage a user needs to pay for in one way or another.

I think I wrote this before changing to non-refundable storage. Then all bytes above the ZBA limit used by the subcontracts also needed to be covered by the main balance.

The way I implemented it, it didn't require to walkt the subtrie, though. I just maintained the total storage usage on the account field. ([See reference implementation](https://github.com/near/nearcore/compare/master...jakmeier:nearcore:sharded-contracts?expand=1#diff-318756fbcf83731e0211cbc53dad981140af78c2b06dce9de383c06cbd93d6efR361-R383))

But now with non-refundable storage, I believe the main account should not be concerned about the storage usage of subcontracts at all. I guess that means `storage_usage()` for the main contract should return the sum of the usage of all account related trie values, except the new ones added in this proposal. I.e. main contract state, main contract code, account meta data, and access keys. But not subcontract code or state.

I've updated the proposal accordingly. The reference implementation will also need an update.

neps/nep-0605-sharded-contracts.md
jakmeier: It's not a misuse in my mind. But thinking again, I see no reason to return 0 rather than the correct value.

Just to avoid confusion, the naming is a bit bad here. "locked" has nothing to do with covering storage usage. This is how much is locked due to staking, thus only relevant for validators and block/chunk producers.

So, the caller of this function wants to know how much was staked by the account. Limited contracts cannot use staking actions but they might still want to know how much was staked by the account.

I would argue this host function should return the correct value for the account, independent of the context. I updated the proposal accordingly. I will keep the host function here in the list of "changes to existing host functions" to make it explicit that is is unaffected.

neps/nep-0605-sharded-contracts.md
jakmeier: Tried my best in commit [70b2627](https://github.com/near/NEPs/pull/605/commits/70b2627fec6c3fabe61d864a799ef7341e5f9d53)

neps/nep-0605-sharded-contracts.md
jakmeier: moved

neps/nep-0605-sharded-contracts.md
jakmeier: Done in commit [70b2627](https://github.com/near/NEPs/pull/605/commits/70b2627fec6c3fabe61d864a799ef7341e5f9d53)

neps/nep-0605-sharded-contracts.md
jakmeier: Tried my best in commit [70b2627](https://github.com/near/NEPs/pull/605/commits/70b2627fec6c3fabe61d864a799ef7341e5f9d53)

neps/nep-0605-sharded-contracts.md
jakmeier: Changed in commit [70b2627](https://github.com/near/NEPs/pull/605/commits/70b2627fec6c3fabe61d864a799ef7341e5f9d53)

neps/nep-0605-sharded-contracts.md
jakmeier: Changed in commit [70b2627](https://github.com/near/NEPs/pull/605/commits/70b2627fec6c3fabe61d864a799ef7341e5f9d53)

neps/nep-0605-sharded-contracts.md
jakmeier: Yes makes sense. I removed the paragraph in commit [70b2627](https://github.com/near/NEPs/pull/605/commits/70b2627fec6c3fabe61d864a799ef7341e5f9d53).

jakmeier:
Thanks @nagisa for reviewing!

I've answered the in-line comments already. Let me also address these questions.

> Some thoughts I had as I was reading that I'm writing down before they get lost:
> 
>     * what happens if near is attached to a sub-contract call?

Like a call to the main account, any NEAR attached to actions in the receipt is added to the balance of the account. The subcontract can even use `attached_deposit` to check how much was paid by the caller.

>    * what happens when an account is deleted? will that retain the sub-contract state? What happens if a new account is later created with the same account name? Does it inherit the subcontract state of the deleted account?

Deleting accounts with a subcontract is not allowed. Any attempts will fail with an error. (The specific error is still missing in the 

>     * root contracts being on the hook to "not forget" to check `predecessor_context` is a footgun. Seems difficult to make sure that footgun does not fire – devs won't be actively thinking about it all the time, for instance. Is there really no way to somehow mark contracts as "subcontract aware" instead and only allow calls to those contracts? The contract would only be marked as such with an explicit action by the contract developer.

The options considered so far are mentioned in the future possibilities section:

> ### Extend interoperability with non-sharded contracts
> 
> We could add support for sharded contracts to call non-sharded contracts later.  Possible ways to keep this secure are:
> 
> - The non-sharded contract has to somehow flag that is accepts sharded calls.  For example, with an explicit host function like `enable_contract_feature(sharded_to_nonsharded)`, or even just by enabling it as soon as a contract has called `predecessor_contract_context`.
> - We could make any sharded to non-sharded contract call fail if there is no host function call to `predecessor_contract_context`.
> - We could make calls to `predecessor_id` fail in sharded to non-sharded contract calls and provide a `predecessor_id_and_context` host method to use instead.
> 
> All these options are a bit hacky and are not currently part of the MVP proposal.  But if the limitation on sharded to nonsharded contracts becomes an issue, these solutions can be added on top the proposed design without backwards compatibility issues.

I'm not opposed to including a proper solution in the proposal. I agree that it would remove an ugly footgun. It just moves further away from the MVP ethos this project was started with. But I guess that ship has sailed a long time ago :sweat_smile: 

My favourite would be to have a special host function `enable_contract_feature(feature: u64)` that the runtime can call to check if the contract is aware of the sharded contract feature. But I suggest if we want to add such a feature, we should ask contract developers / SDK maintainers what their opinion is.

jakmeier:
A general note on this proposal.

There is now an alternative to this proposal with #616 which arguably proposes a cleaner architecture for sharded contracts.

For that reason, I will not spend more time on the reference implementation of this proposal until we have a decision on which proposal is chosen. I will keep the proposal description itself up to date, however.

nagisa:
Lets involve `frol` on the question of `enable_contract_feature` if/when we come back to this. From the compiled WASM perspective I can imagine it wouldn't be too hard to add a sub-command that introduces, say, a custom section or something that lists all different runtime features a contract can understand.

PR #611, opened 2025-05-30
NEP-611: Pending Transaction Queue

robin-near:
In the near future of the Near blockchain, we foresee that via the SPICE project, transaction and receipt execution will become decoupled from the blockchain itself; they will no longer run in lockstep. Instead, transactions will be included in the blocks first, and then execution will follow later.

This inherently introduces a problem that we must accept transactions before we know whether they are valid. Today, when a chunk producer produces a chunk containing a transaction, it can verify using the current shard state that the transaction has a valid signature, has enough balance, and a valid nonce. But as execution becomes asynchronous, we no longer have the current shard state to verify the
transactions against.

This NEP proposes a mechanism called the Pending Transaction Queue to solve this problem.


neps/nep-0611.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Access Key vs Gas Key
+We introduce a new transaction version, `TransactionV2`, which is able to either specify an access key
+or a gas key. Any older transaction version is equivalent to a `TransactionV2` that specifies an access
+key.
+
+#### Access Key Parallelism Restriction
+We now restrict the ability to send multiple parallel pending transactions with Access Keys.
+
+Specifically, for any given account $A$ with any number of access keys, the total number of access key transactions in the pending transaction queue whose sender is $A$ cannot exceed $P_{\mathrm {max}}$, a
```
birchmd: Is this enough to prevent the attack described above? Instead of having one account they send many transactions to, it could be many accounts they send only a few transactions to each.

neps/nep-0611.md
```diff
@@ -0,0 +1,233 @@
...
+(account ID, transaction type), and further indexed by the block hash. Furthermore, the pending
+transaction queue is stored per block per shard, not as a single data structure. The contents of the queue
+is exactly the pending transactions according to the definition above.
+
+The constraints are enforced as described above; we reiterate it here:
+* There cannot be more than 4 access key transactions under the same account in the queue;
+* There cannot be more gas key transactions under the same gas key in the queue whose total transaction
+  cost exceed the gas key's balance.
+
+The constraints are maintained at the time of chunk production: when producing a chunk, we only accept
```
birchmd, line 312: To check the constraints hold, it sounds like this means chunk producers need to include a state witness to validators along with the proposed chunk, just like they do today. Does this work against the idea of SPICE separating the consensus and execution? I suppose maybe if this state witness is very small relative to today then SPICE could still be effective.

birchmd (commented):
Thanks for bringing this up! It is good to think about how to prevent possible DoS attacks under SPICE.

neps/nep-0611.md
robin-near: The idea is that the amount of block space the attacker consumes vs amount of gas the attacker has to pay, i.e. the "waste amplification factor", is unbounded in the attack mentioned in the NEP, but only O(1) with a maximum parallelism. It's then still possible to attack, but the attack would cost a proportional amount of gas - just cheaper than it is today - which isn't a very great incentive to mount the attack (given that the attack is only a DoS).

neps/nep-0611.md
robin-near: Ah, good point. We are only addressing the attack surface from (unauthenticated) users, not attacks from chunk producers. Chunk producers today (and in SPICE) are high-stake entities because they are also block producers, and mounting this attack as a chunk producer is not even very effective because you can only affect the chunks you produce. It is also already possible today to simply produce a chunk that is empty.

garikbesson:
Because there has been no activity from the author for more than 3 months, I'm marking the proposal as RETRACTED. Please feel free to open it again if you'd like to continue working on it.

neps/nep-0611.md
```diff
@@ -0,0 +1,352 @@
...
+
+### Negative
+
+* Adds complexity to protocol and implementation.
+
+### Backwards Compatibility
+
+* Adding `TransactionV1` does not change the semantics for `TransactionV0`. Users can continue to submit their transactions as such.
+* Adding new keys to the trie may require modifications for downstream parsers, indexers.
+* Existing use cases may assume they can submit an unbounded number of in-flight transactions (for example, they may rely on them being in the mempool). As changes only impact accounts with contracts, the impact of this is assumed to be low.
```
darioush:
```suggestion
* Existing use cases may assume they can submit an unbounded number of in-flight transactions. As changes only impact accounts with contracts, the impact of this is assumed to be low.
```

neps/nep-0611.md
```diff
@@ -0,0 +1,352 @@
...
+
+* Account must already exist.
+* Check that the same key does not already exist as a gas key. Notably, the same key can be added as an access key and a gas key as their namespace is separate.
+* Requested number of nonces must be less than or equal to `GasKey::MAX_NONCES` (currently suggested as 1024). This is to limit the number of trie operations for a single action.
+* Increases storage usage of associated account,
+* Check that if the permission is a `FunctionCallPermission`, the allowance is `None` (unlimited
+  allowance).
+* A new `GasKey` entry is added to the trie, keyed by (gas key prefix, account ID, public key). The
+  gas key prefix is a new trie prefix.
+* For each nonce ID (from 0 to the number of nonces minus 1), store the default nonce at the trie
```
nagisa: This is mildly problematic for scaling/perf reasons. We're spending a humongous amount of time accessing the account + accesskey during transaction validation in tx runtime in some workloads already. This is adding another walk through a trie.

OTOH if we make nonces array be inline the `GasKey` struct, then every time we load/deserialize the `GasKey` we have to read out and deserialize all them nonces.

---

I see that this multiple-nonce mechanism is meant to enable something outside of this NEP's primary motivation. This is okay, but we should introduce proper motivation for adding this feature in the NEP text.

I also wouldn't lock in a specific implementation here for now. I suspect that with `MAX_NONCES` scaled down to a smaller number (e.g. 16) it might become feasible to keep `GasKey` as a monolithic type.

Another reason to keep MAX_NONCES low initially is that it is trivial to increase the limit in the future, but decreasing it would be effectively impossible.

neps/nep-0611.md
```diff
@@ -0,0 +1,352 @@
...
+  * See below for "gas key transactions": it is also possible to fund a gas key using a gas key; so this
+    amount is not necessarily deducted straight from the account.
+* To apply this action, the balance on the gas key is increased by the same amount.
+
+`DeleteGasKey` is verified as executed as follows:
+
+* The gas key must exist under the account.
+* To apply this action, all relevant trie nodes are deleted,
+* Decreases storage usage of associated account,
+* The remaining balance left in the key is **burned**. (This prevent the key deletion attack.)
```
nagisa: I think there is a possible middle ground that does not make gas keys be this risky.

Indeed, if this were to transfer the balance out of the gas key, we'd have to prevent the issue described in the motivation section (with `send_near` replaced with gas key deletion.) Even in that context a middle ground is still achievable by e.g. treating this action specially and:

1. Whenever we put a transaction containing this action into the pending transaction queue, reserving all the remaining balance inside of it;
2. Not accepting any further transactions (cause the chunk producers see 0/negative remaining balance due to the reservation above.)
3. (It might be necessary to make this action only valid as the only action of the transaction, i.e. not possible to construct via promises.)

neps/nep-0611.md
```diff
@@ -0,0 +1,352 @@
...
+* For user's benefit we may consider failing delete account action if there are associated gas keys with a total balance exceeding a certain amount.
+
+Note the reference implementation [already](https://github.com/near/nearcore/blob/136acc3a524575aae1300e26901e664adb521a6f/runtime/runtime/src/actions.rs#L662) prohibits deleting accounts over a certain storage threshold, which prevents a single action from doing an unbounded amount of work.
+
+#### Cost of Gas Key actions
+
+* Cost of `AddGasKey` will be based on the cost of adding a new access key, as it is the most similar existing action. In addition, the following fees will be charged per nonce to account for further trie operations:
+  * `storage_write_base`: Represent cost of adding a separate key to the trie,
+  * `storage_write_key_byte * len(Some(u32))`: Notably we only charge for the nonce portion of the key, as the trie nodes share a prefix with the gas key, and was already accounted for.
+  * `storage_write_value_byte * len(Nonce)`
```
nagisa: If we decide to have separate trie keys for each nonce, then each of those trie keys should be charged for as its own write operation, rather than only for the byte-length of the conceptual vector of nonces.

neps/nep-0611.md
```diff
@@ -0,0 +1,352 @@
...
+
+This allows creating [balance](https://github.com/near/nearcore/blob/136acc3a524575aae1300e26901e664adb521a6f/core/primitives/src/receipt.rs#L602) and [gas](https://github.com/near/nearcore/blob/136acc3a524575aae1300e26901e664adb521a6f/core/primitives/src/receipt.rs#L655-L656) refunds with either a `Transfer` or `TransferToGasKey` action, depending on the variant of `TransactionKey` used to originate the action.
+
+##### Refunds alternatives
+
+1. We could route balance and gas refunds to the account balance. This trades-off user experience for simplicity of implementation and protocol: no specific changes to receipts would be needed, however the user would have to "top-off" the gas key balance more frequently.
+
+2. Addition of `ActionV3` and `PromiseYieldV3` requires careful consideration of possible interactions with `Delegate` action and `refund_to`. It may be simpler to track the gas key for refunds by adding a `Receipt` variant, however this seems a bit out of place (as `Receipt` currently only tracks `predecessor_id`, where `ActionReceipt` tracks `signer_id` and `signer_public_key` i.e, access key).
```
nagisa: More specifically the current contract-host interface now has a hard assumption that refunds can (only) be specified to go to an `AccountId`, which strongly twists our hand towards only ever refunding to `AccountId`. There are some issues that stem largely from the fact that its the caller's decision whether to use an AccessKey or a GasKey and contracts (especially contracts that are "locked") can't upgrade to code that's GasKey aware.

So the solution space starts looking like:

1. If we force the refund provenance (i.e. refunds are only possible to the same type of balance from which it was charged from) and an incompatible contract calls `refund_to`, we panic for calls that use `GasKey`?
2. If we do not force the refund provenance (i.e. refunds can go to either account or gas key) then we still force the "inconvenient" behaviour of having to recharge the gas key more frequently than otherwise required;
3. If we force refunds to go to the account ID, the contract-host interface does not change in any way and gas keys are transparent and a strictly additive feature from the contract's standpoint. I think both of these properties are quite high priority if possible.

---

I'd also say there is a very good reason to force frequent gas key recharges anyway. Similar to how forcing short-lived certificates force infra engineers to stay on their toes and "simply" set up infrastructure to automatically renew certificates, gas keys refunding to an account unconditionally would force infra engineers for the contract to do the same. That way they won't run out of balance in 10 years unexpectedly breaking their production and long after everybody involved have left the company :)

neps/nep-0611.md
darioush: I intended to charge each trie key as its own operation, by charging per nonce: `storage_write_base` + appropriate key and value size multiplied by `storage_write_key_byte` and `storage_write_value_byte` respectively.

For the value of each nonce, `len(Nonce)` seems the correct amount to charge (`Nonce = u64` here).
For the key length it seems more appropriate to charge for the common prefix (`public_key`) part only once, which is accounted for already in the fee that is similar to adding an access key.

Are you suggesting to charge the common key prefix also `num_nonces` times?

neps/nep-0611.md
darioush: There is a compounding effect here with deleting the account (which can be done programmatically).

If we do a refund on delete for GasKey then we need to do something like not allowing accounts with GasKey to be deleted.

neps/nep-0611.md
darioush: Using gas keys would add one additional trie access (to load the nonce from a separate key).

However I am not sure it is a major performance concern as gas keys are only needed by accounts with contracts that wish to have more than 4 in-flight transactions (or by users that issue parallel transactions using multiple nonces).

There is a proof size tradeoff as well: if vector approach is used, value of all nonces will be included in the proof vs. just the accessed nonce.

I agree about keeping MAX_NONCES low initially.

darioush: @robin-near can you please weigh in on the choice about making each nonce its own trie key?

darioush: Actually it seems with some of the recent [caching efforts](https://github.com/near/nearcore/pull/14521/files#diff-f44da223e1ccef43ec41bb7d1cf9fb0a12ac91d2033673c34661a5960bee6209R1754-R1759) we can cache fetching the "main part" of the GasKey (with balance, permissions etc). so an additional trie access only happens only for the first access of each GasKey (then accessing other nonce indexes of the same gas key become cheaper for the same chunk).

neps/nep-0611.md
darioush: I am tending to agree with this because of the above complications and the user will need some mechanism to top up their balance periodically regardless.
Also it will discourage users from attaching too much gas to transactions.

walnut-the-cat:
Thank you @darioush for re-submitting this NEP.

As a moderator, I reviewed this NEP and it contains enough information to initiate SME review process based on [the proposed template guidelines](https://github.com/near/NEPs/blob/master/nep-0000-template.md), therefore I am moving this NEP to [the REVIEW stage](https://github.com/near/NEPs/blob/master/neps/nep-0001.md#nep-lifecycle).

I would like to nominate @nagisa and @matejpavlovic as SMEs for this NEP to complete a technical review (see expectations below).

<details>

Technical Summary guidelines:

```

### Recommendation

Add recommendation

### Benefits

* Benefit

* Benefit

### Concerns

| - | - | - | - |

```

</details>

neps/nep-0611.md
nagisa: > then accessing other nonce indexes of the same gas key become cheaper for the same chunk

I'm not sure that holds true with memtries at least. Accesses to memtrie don't really retain any memory of the prior accesses and so any new access will start walking from the trie root until it gets to the value.

Conceptually, though, this could be possible to make optimal by changing the memtrie. It would just need a much more involved API that allowed you to pass in a node from previous lookup to start walking from and only use the tail of the key, rather than the entire key.

neps/nep-0611.md
nagisa: Hm, fair enough. It could be quite nasty if an account deletion with a boatload of gas keys required a bunch of book-keeping on the chunk producer side as well.

I think it would be a fair requirement to impose that if you want to delete an account you have to get rid of all gas keys first and use the access key to sign the action. We already have a limit on the amount of storage for the contract already, this wouldn't be an unusual requirement.

neps/nep-0611.md
nagisa: Yes, I believe the key fees should be charged for the common key prefix as accessing/writing to each key is going to (currently) walk the memtrie from the root each time. Doing anything more efficient would require a fancy API to allow looking up *nodes* rather than values and doing lookups starting at arbitrary provided nodes.

This is a good API to add regardless for many reasons (e.g. isolation of contract data accesses,) but right now it does not exist AFAIK.

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+Type: Protocol
+Version: 1.0.0
+Created: 2025-05-28
+LastUpdated: 2025-11-21
+---
+
+## Summary
+
+In the near future of the Near blockchain, we foresee that via the SPICE project, transaction and receipt
+execution will become become decoupled from the blockchain itself; they will no longer run in lockstep.
```
matejpavlovic:
```suggestion
execution will become decoupled from the blockchain itself; they will no longer run in lockstep.
```

neps/nep-0611.md
darioush: I also agree with the alternative of forbidding a deletion of accounts with gas keys.

This is because if we try to add some protection from deleting high balances, then we have to check an unspecified number of gas keys to do 1 account delete, which makes charging for it difficult.

This makes it such that the user has to first delete each of the gas keys and pay for the nonce deletion accordingly, in a separate action.

neps/nep-0611.md
darioush: Updating this to account for key len corresponding to the trie key len.

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+#### Cost of Gas Key actions
+
+* Cost of `AddGasKey` will be based on the cost of adding a new access key, as it is the most similar existing action. In addition, the following fees will be charged per nonce to account for further trie operations:
+  * `storage_write_base`: Represent cost of adding a separate key to the trie,
+  * `storage_write_key_byte * key_len`, where `key_len = MAX_ACCOUNT_ID_LEN + len(ACCOUNT_DATA_SEPARATOR) + len(PublicKey) + len(Some(u32))`: Corresponds to length of the associated trie entry's key.
+    * Note `len(ACCOUNT_DATA_SEPARATOR) = 1`.
+    * We use `MAX_ACCOUNT_ID_LEN` here as actual account id is not available in the action.
+  * `storage_write_value_byte * len(Nonce)`: Corresponds to the length of associated trie entry's value.
+* Cost of `DeleteGasKey` will be based on the cost of deleting an access key, as it is the most similar existing action. In addition, the following fees will be charged per nonce to account for further trie operations:
+  * `storage_remove_base`: Represent cost of adding a separate key to the trie,
```
matejpavlovic: Adding or removing? (Copy-paste mistake?)

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+With gas keys, this may become the public key of a gas key or an access key.
+
+As alternative, we could provide more context to the VM, allowing contracts to distinguish gas keys and access keys. Doing so provides a richer context to applications, and trades-off simplicity and future flexibility. Without having concrete use cases or limitations, the additional VM changes seem unnecessary or best left for a future NEP.
+
+### Gas Key Pending Transaction Constraints
+
+Unlike access key transactions, gas key transactions are not limited in parallelism; they are only limited
+by the amount of gas these transactions consume. Specifically, for a gas key $G$, the total cost of 
+the transactions signed with $G$ in the pending transactions must not exceed the balance of $G$
+(according to the state that the pending transactions are calculated based on).
```
matejpavlovic:
```suggestion
(according to the state that the pending transactions are based on).
```

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+transaction queue is stored per shard, not as a single data structure. The contents of the queue
+is exactly the pending transactions according to the definition above.
+
+The constraints are enforced as described above; we reiterate them precisely here. For each account, Let
+$T_A$ be the set of transactions in the queue signed with any access key of this account, and let
+$T_G$
+be the set of transactions in the queue signed with any gas key of this account. The following constraints
+must hold true:
+
+* If the account has a contract deployed (as of the latest available state), then $|T_A| \le 4$.
```
matejpavlovic:
```suggestion
* If the account has a contract deployed (as of the latest available state), then $|T_A| \le P_{\mathrm {max}}$.
```

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+* If any transaction $t\in T_A\cup T_G$ contains a `DeployContract` action, then $|T_A| \setminus \{t\}=\emptyset$. In other words, deploying a contract cannot be done in parallel with any access key
+  transactions.
+* The sum of the costs of all transactions in $T_A$ does not exceed the balance of the account.
```
matejpavlovic: I assume that by "costs" of transactions we mean gas limits, right? Since the tXs are not yet executed, we don't know their actual costs.

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+"execution head" (how far the execution has caught up) is defined:
+
+* It can be defined locally as the progress of execution at a node, but this will be different between nodes.
+* It can be defined deterministically as the last block whose execution result is certified by consensus.
+
+For the purpose of this NEP, we use the latter definition, so that the notion of pending transactions is
+consistent across all nodes and the determination of what transactions are eligible to be included by
+a chunk producer can be verified -- even though we do not plan to implement this verification right now.
+
+Another note is that the notion of pending transactions is anchored at a specific chunk that is being produced. In case of forks, we use the block that the chunk is being produced on top of to compute the
```
matejpavlovic, line 91: "The notion of pending transactions is anchored at a specific chunk that is being produced." Agree.

The tricky part is that the chunk producer is required to know about all the transactions included in the previous chunks all the way up to the chunk that it is producing. I.e., producing a chunk for height h requires the knowledge of all transactions associated with heights until h-1. (We're considering a single shard.)

In SPICE we count on decoupling (from consensus) not just the execution, but data availability (even though we are not implementing it in SPICE v1). That is, making transaction data available will also be asynchronous. A chunk producer with the right of producing a chunk associated with height h might actually have a window of a few blocks after h to make sure the data is available, during which the data owners will certify (on chain) having received their respective data parts. This means that the availability certificate for a chunk associated with height h might only occur in a block at height h+a (a being the number of blocks it takes to ensure the availability). The next chunk producer (producing a chunk for h+1), however, might not yet know the contents of the chunk associated with h, and thus not have a precise view of the pending transaction set. This is a more general problem that also has other  implications (such the problem with transaction duplication), but an incomplete view of the pending transaction set is also an issue.

I could imagine a solution that requires the chunk producer for h to post at least a commitment to the chunk on chain as soon as possible, and the pending transaction set at h+1 would be defined through these commitments. The chunk producer for h+1 would then need to obtain the transaction data associated with h before proposing its own chunk. Ideally the chunk producer at h would directly (with high priority) send its chunk to the producer at h+1. If the former does not send the chunk, the latter would need to retrieve the chunk from the data owners, slowing down the chunk production.

neps/nep-0611.md
matejpavlovic: As I understand it, the requirement for burning the remaining balance stems from the fact that a gas key (or the whole account) can be deleted through an access key, i.e. not involving a transaction signed by the gas key. Is this the case? Even if we disallow deleting accounts with gas keys?

If yes, would it make sense (or even be possible) to enforce the same restrictions for gas key deletions as we do for transfers? Putting it differently, can we treat a gas key deletion the same way as we treat a withdrawal, with the exception that the gas key would be deleted afterwards? The NEP later says that "	contract execution cannot create receipts that withdraw from gas keys; only gas key transactions can". Is there a fundamental reason to restrict this to withdrawals?

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+    amount is not necessarily deducted straight from the account.
+* To apply this action, the balance on the gas key is increased by the same amount.
+
+`DeleteGasKey` is verified as executed as follows:
+
+* The gas key must exist under the account.
+* To apply this action, all relevant trie nodes are deleted,
+* Decreases storage usage of associated account,
+* The remaining balance left in the key is **burned**. (This prevent the key deletion attack.)
+  * For user's benefit we may consider failing this action if the balance exceeds a certain threshold.
```
matejpavlovic: Would it make sense to consider an additional "force" flag in the `DeleteGasKey` action to give the user the possibility to override this check?

matejpavlovic (commented):
Well written. I added some comments inline. Regarding performance issues and the section on nonces, I'm not able to make properly informed opinion - my knowledge of the implementation is unfortunately not deep enough to meaningfully contribute to the discussion.

neps/nep-0611.md
darioush:
```suggestion
  * `storage_remove_base`: Represent cost of removing a key from the trie,
```

neps/nep-0611.md
darioush: Yes good point. Here the costs refer to the sum of near and gas attached (ie, maximum cost) of the transaction.

It is possible that not all the gas would be used in execution, and that the user would receive a refund later.

Adding some clarification for this.

neps/nep-0611.md
darioush: Adding a note for this as an implementation consideration.

neps/nep-0611.md
darioush: > i.e. not involving a transaction signed by the gas key. Is this the case? Even if we disallow deleting accounts with gas keys?

This is correct. Notably, account deletion can also occur programmatically (as a result of contract execution).

> If yes, would it make sense (or even be possible) to enforce the same restrictions for gas key deletions as we do for transfers?

We could force the same restriction, i.e., disallow programmatic deletion of gas keys. However, contracts may benefit from the flexibility of programmatically adding/deleting gas keys.

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+    amount is not necessarily deducted straight from the account.
+* To apply this action, the balance on the gas key is increased by the same amount.
+
+`DeleteGasKey` is verified as executed as follows:
+
+* The gas key must exist under the account.
+* To apply this action, all relevant trie nodes are deleted,
+* Decreases storage usage of associated account,
+* The remaining balance left in the key is **burned**. (This prevent the key deletion attack.)
+  * For user's benefit we may consider failing this action if the balance exceeds a certain threshold. Implementation may include a `force` flag to overrides this check.
```
matejpavlovic:
```suggestion
  * For user's benefit we may consider failing this action if the balance exceeds a certain threshold. The implementation may include a `force` flag that overrides this check.
```

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+
+The constraints are enforced as described above; we reiterate them precisely here. For each account, Let
+$T_A$ be the set of transactions in the queue signed with any access key of this account, and let
+$T_G$
+be the set of transactions in the queue signed with any gas key of this account. The following constraints
+must hold true:
+
+* If the account has a contract deployed (as of the latest available state), then $|T_A| \le P_{\mathrm {max}}$.
+* If any transaction $t\in T_A\cup T_G$ contains a `DeployContract` action, then $|T_A| \setminus \{t\}=\emptyset$. In other words, deploying a contract cannot be done in parallel with any access key
+  transactions.
```
jakmeier: To be clear, I assume the same would be enforced for "deploy-like" actions, right?

`UseGlobalContract`, `DeterministicStateInit`, and `Delegate` with inner deploy-like actions all should have the same restrictions, I think.

neps/nep-0611.md
jakmeier: Throwing in another idea: Why not make gas keys just another type of access key and simply refer to them by public key?

Today, gas refunds of transactions created with a function access keys already restore the balance allowance on the corresponding key. (Unless it has been deleted while the transaction's receipts DAG was executing.) This works by recognizing a refund receipt by its sender being "system" plus checking that the transaction signer (always preserved across all receipts of a transaction) is a function access key.

See here:
https://github.com/near/nearcore/blob/be8d8cc7acf716c3c084d1a0fe323109aa32d478/runtime/runtime/src/lib.rs#L2481-L2482

It should be possible to use the same mechanism to detect gas refunds from a transaction that was initiated by a gas key. But only if access keys and gas keys share the same namespace. I think they should.

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+
+This NEP adds Gas Keys, **conceptually** defined as the following:
+
+```rust
+struct ConceptualGasKey {
+    public_key: PublicKey,
+    nonces: Vec<Nonce>,
+    balance: Balance,
+    permission: AccessKeyPermission,
+}
```
jakmeier, line 137: As I wrote in another comment to make refunds easier, can we consider another representation?

Specifically, if we can store gas keys the same way we store the two existing access key types, then this would let them share the same namespace. Which makes it possible to detect refunds that should go to a gas key by using the public key.

So I would suggest something more like this:

```diff
pub enum AccessKeyPermission {
    FunctionCall(FunctionCallPermission),

    /// Grants full access to the account.
    /// NOTE: It's used to replace account-level public keys.
    FullAccess,

+   // Same access as FullAccess but with separate gas accounting
+   ConceptualGasKey,
}

struct ConceptualGasKey {
-   public_key: PublicKey,
    nonces: Vec<Nonce>,
    balance: Balance,
-   permission: AccessKeyPermission,
}
```

This would be slightly less flexible, as it doesn't allow to limit methods callable by the gas key. We could add a `FunctionCallGasKey` as another variant if that's needed.

Of course, this has pretty large ramifications for technical code details. We might not even need `AddGasKey` and `DeleteGasKey`  actions and  instead just reuse `AddAccessKey` and `RemoveAccessKey`.

Maybe this is going too far off the intended design. But from my point of view, this seems to be an alternative that simplifies a bunch of things by staying closer to the existing concepts of access key types.

neps/nep-0611.md
```diff
@@ -0,0 +1,370 @@
...
+The constraints are enforced as described above; we reiterate them precisely here. For each account, Let
+$T_A$ be the set of transactions in the queue signed with any access key of this account, and let
+$T_G$
+be the set of transactions in the queue signed with any gas key of this account. The following constraints
+must hold true:
+
+* If the account has a contract deployed (as of the latest available state), then $|T_A| \le P_{\mathrm {max}}$.
+* If any transaction $t\in T_A\cup T_G$ contains a `DeployContract` action, then $|T_A| \setminus \{t\}=\emptyset$. In other words, deploying a contract cannot be done in parallel with any access key
+  transactions.
+* The sum of the costs of all transactions in $T_A$ does not exceed the balance of the account. Here, the cost of a transaction refers to the sum of attached gas and NEAR deposit (as calculated by `total_cost` defined [here](https://github.com/near/nearcore/blob/61c087c5250566070684a624a5ea50e796f61e5f/runtime/runtime/src/config.rs#L332)). Importantly, this cost is knowable by just inspecting the actions contained in the transaction and does not depend on state or execution.
```
jakmeier: >  Importantly, this cost is knowable by just inspecting the actions contained in the transaction and does not depend on state or execution.

It does depend on the gas price though, doesn't it? The gas price is part of the block, so not state. I'm not very familiar with SPICE. Is the block header used for execution already known when we do these checks?

jakmeier (commented):
This is a high quality proposal, thanks to everyone who brought it so far!

I left an idea to potentially simplify the refund story in the comments.

neps/nep-0611.md
jakmeier: Note that https://github.com/near/NEPs/pull/522 is also open and might have a different solution to parallelization. But I'm not sure if it is compatible with the exact limit we want to enforce here.

neps/nep-0611.md
darioush: Yes, that's right, it depends on gas price. I will clarify that it is needed as well.

For Spice, we have not yet figured out the details of propagating the gas price. As execution would be a few chunks behind, we could use the latest executed chunk as a basis for the gas price of the next chunk. There is likely some nuanced tradeoffs which we have to consider.

Notably, this proposal doesn't require spice. For this NEP, we can assume the chunk producer would have a way to know the appropriate gas price, like it does currently.

neps/nep-0611.md
darioush: I agree, how can we find the full list of these actions? (are there any others?)

neps/nep-0611.md
darioush: This is an interesting proposal, and indeed a valid approach to dealing with duplicate transactions.
I don't agree with the part about writing state into the trie (this seems very expensive, to account for transactions in recent blocks)

This means the nodes would have to re-compute the new "recently accepted transactions" data structure. In spice we are trying to avoid this type of dependency on past chunks / blocks.

neps/nep-0611.md
jakmeier: As of now, that should be the full list. Any potential future action that can also set the contract field of an account should be included, though.

neps/nep-0611.md
darioush: Note: this approach may help avoid require separate Add/Remove actions

neps/nep-0611.md
```diff
@@ -0,0 +1,459 @@
...
+1. Limiting the number of in-flight transactions for accounts with contracts: This aims for backwards compatibility for most use cases and also limits the amount of spam they can include on chain.
+
+2. Allowing users with sophisticated issuance needs to prepay for gas and have many in-flight transactions. This NEP additionally includes a way to associate a single signing key and balance with an array of nonces, to provide a convenient way for parallel transaction issuance for such users.
+
+## Specification
+
+To solve this problem, we present two critical components which work together to ensure that all accepted
+transactions are valid. At a high level, they are:
+
+* **Access Key vs Gas Key**: In addition to Access Keys, we introduce a second kind of signing key called the Gas Key. A gas key can be funded with NEAR, and when issuing a transaction using a gas key, the transaction only consumes gas from the gas key, not from the account.
```
jakmeier: Without changing anything around the implementation, wouldn't it make more sense to describe gas keys as a new type of access key? There are already two types of access keys (full-access and function call access) that smart contract developers are familiar with. Adding gas keys as a entirely new concept seem more complicated than necessary.

I get that an additive approach is easier from a coding perspective. But any added new concept has a long tail of changes needed in NEAR tooling (wallets, explorers, SDKs, RPCs, etc.) and documentation around it.

For example, if you look at the [documentation for the RPC API](https://docs.near.org/api/rpc/access-keys), how would you say gas keys fit in? Will they be listed on a `view_access_key_list` call? Can their changes be watched with `single_access_key_changes`?

I think it would be easier to integrate with the existing infrastructure, if we add new variants to an existing concept, rather than introducing a new one. Or is there a reason that I'm missing, as to why "gas keys" should be a new concept?

neps/nep-0611.md
```diff
@@ -0,0 +1,459 @@
...
+`DeleteAccount` action will fail if the account has any associated gas keys.
+
+`AddKey` action will fail if the public key is already registered as a gas key.
+
+
+#### Cost of Gas Key actions
+
+* Cost of `AddGasKey` will be based on the cost of adding a new access key, as it is the most similar existing action. In addition, appropriate fees will be charged per nonce to account for further trie operations.
+* Cost of `DeleteGasKey` will be based on the cost of deleting an access key, as it is the most similar existing action. In addition, appropriate fees will be charged per nonce to account for further trie operations.
+* Cost of `TransferToGasKey` and `TransferFromGasKey` will be the same as `Transfer`, as it is the most similar existing action.
```
jakmeier: `Transfer` is the cheapest action in terms of gas because it only needs to look at the account meta data.  `TransferFromGasKey` and `TransferToGasKey`  would need to do additional trie modifications on sending and on execution, respectively. That should be included in the gas cost parameters.

neps/nep-0611.md
```diff
@@ -0,0 +1,459 @@
...
+
+## Reference Implementation
+
+* [Initial implementation of gas key trie modifications](https://github.com/near/nearcore/pull/13687)
+* [Initial implementation of gas key actions](https://github.com/near/nearcore/pull/14532)
+* [Work in progress implementation of actions and refund receipts](https://github.com/near/nearcore/pull/14521)
+
+## Alternatives
+
+One alternative to the gas key design is to introduce a `SenderReservedBalance`: under this model, the sender's account reserves a portion of its NEAR balance specifically for gas on future transactions.
```
jakmeier, line 402: Now that we have found more complications with the gas keys approach, can we talk about the sender reserved balance approach again?

I will add comments per listed point against it.

neps/nep-0611.md
```diff
@@ -0,0 +1,459 @@
...
+
+The runtime would check that the sender has enough reserved funds to cover both the gas cost and the required minimum balance, thereby enforcing validity without requiring additional trie keys. While initially it seems simpler, this approach comes with significant and subtle drawbacks:
+
+* Modifies the semantics of the account's main balance, requiring checks that contract execution doesn't deplete the balance lower than `SenderReservedBalance`, potentially breaking existing assumptions about how balance operates.
```
jakmeier, line 410: For better or worse, we already have `storage_usage` on an account which defines a minimum of the balance that must be maintained. The runtime already needs to check it after every transaction or receipt execution. And users already have to deal with the fact that they can't access the full balance.

So yeah, we have to be careful. But I don't think we need to be afraid too much to change the account fields. (It is even versioned now! :D)

neps/nep-0611.md
```diff
@@ -0,0 +1,459 @@
...
+* Contract execution can lead to deletion of access keys and accounts (which transfers the balance to a beneficiary). The chunk producer cannot predict this. Further, such actions can be originated from a different account.
```
jakmeier: How is the deletion of access keys relevant? I think only deleting accounts matters.

The account deletion, we can solve it the same way as it would be solved with the gas key approach. We either don't allow deleting when there is reserved balance, or we just burn the reserved balance on deletion. Or maybe burn it if it is below a threshold, block deletion it otherwise.

neps/nep-0611.md
```diff
@@ -0,0 +1,459 @@
...
+* When transferring balance to another account, the `Transfer` transaction's cost currently includes the amount being transferred, since the conversion to a receipt requires deducting that amount of balance. With sender reserved balance, we must change it so that if a transaction includes a `Transfer` action but there's not enough balance to cover the amount being transferred, the gas portion of the transaction is still deducted.
```
jakmeier: Isn't the same true for the gas key approach?

neps/nep-0611.md
```diff
@@ -0,0 +1,459 @@
...
+
+In contrast, the gas keys solution is semantically additive, does not modify semantics of existing account balance, and also provides improved user experience for concurrent transaction issuance via `nonce_index`.
```
jakmeier: Hm, if we keep access keys the same and add a `gas_reserved_balance` field to accounts, wouldn't that be a much smaller change overall? I guess we still need new actions to add and subtract from that field... But otherwise, what else needs to change?

@darioush Can you maybe outline what changes would be necessary? It would really help me understand the trade-off.

Again, an additive change might be easier on the implementation. But I don't think it is necessarily better for the ecosystem. The more of the complexity we can deal with inside the protocol and hide from the users and contract devs, the better.

And about the developer experience, do we know if that's actually what devs want? Did anyone ask for this?

I only know that in #522, developers asked for a way to avoid nonce conflicts. Specifically, they don't want to be forced to sync up with the latest nonce on chain before signing a transaction. Here, we double-down on clients needing to syncing the nonces before they can sign a transaction.

jakmeier (commented):
As I am now assigned as an SME, I have reviewed the full NEP in more details.

I have a couple of questions before I will write my summary and recommendation. @darioush  Please take a look at the inline comments I left with this round of review.

neps/nep-0611.md
darioush: I agree with this framing. It is closer to the updated trie storage model, and as you point out better compatible with the existing RPCs.

To answer the RPC question, I added a new [section](https://github.com/near/NEPs/pull/611/files#diff-950ac706323fb324c543b37585e73c72ec58a52c6fa39934b8096fce34ac5523R353-R359). I wouldn't say the implementation needs to strictly adhere to this section, however since there is some questions about it I will include it for clarity and discussion.

Please let me know what you think.

neps/nep-0611.md
darioush: Yes, this is now true of the gas key approach after updating it to only deduct gas from gas keys and use account's balance for the deposit. I will remove this as it is no longer relevant.

neps/nep-0611.md
darioush: This is a great question, as the problem with deleting keys was not explained properly here before. Please see the updated paragraph.

You're right that we can solve the account deletion problem with some care. However, a major issue with the `SenderBalanceReserve` approach is that the keys can be deleted via contract execution, which makes using a shared balance difficult.

neps/nep-0611.md
darioush: I agree that changing runtime and `Account` are relatively minor disadvantages for the `SenderReserveBalance` approach. 

I moved it below the main issue. It is still a non-trivial change for the ecosystem to have to support `AccountV2`.

neps/nep-0611.md
darioush: > Can you maybe outline what changes would be necessary? It would really help me understand the trade-off.

In terms of an outline, I will not get too detailed because of the deleted key shared balance issue, however:
- Adding AccountV2 with the new field,
- Adding actions for manipulating the field,
- Introduce something like `ActionErrorKind::LackBalanceForGasReserve`, and `InvalidTxError::LackBalanceForGasReserve`, which we can check after each transaction and receipt.
- Potential updates for [Rosetta](https://github.com/near/nearcore/blob/e730cc30f985d79c7b3962b82888b4d69ce36730/chain/rosetta-rpc/src/utils.rs#L257-L275)

> Again, an additive change might be easier on the implementation

I'm removing the wording about additive since it's somewhat subjective and the current suggestion is not as additive as the original one.

> And about the developer experience, do we know if that's actually what devs want? Did anyone ask for this? I only know that in #522, developers asked for a way to avoid nonce conflicts. Specifically, they don't want to be forced to sync up with the latest nonce on chain before signing a transaction

I don't know of any incoming request for 2D nonces specifically. [Here](https://github.com/near/NEPs/pull/522#issuecomment-3319172723) in one example use of >100 access keys is mentioned. While not as convenient as setting a random nonce, the `nonce_index` can be a replacement of using many access keys.

Unfortunately to avoid nonce synchronization entirely, we would need to track "recently observed transactions" for de-duplication either as part of state or a separate data structure requiring warm-up.

neps/nep-0611.md
darioush: Good point. Even though I think `Transfer` also sometimes updates the `AccessKey` on `send` (specifically in the case of allowances), it would be good to charge for the additional accesses.

If we want to base the additional cost on use `wasm_storage_{read/write}` costs, we would have to use an approximate or minimum value as the actual value size of gas keys is variable (may contain `FunctionCallPermission` which contains `String`). The value size is needed for use with `wasm_storage_{read/write}_value_byte`.

Please let me know if you think this is a good direction or if you have an alternative suggestion here.

neps/nep-0611.md
darioush: Thanks! Based on your [comment](https://github.com/near/NEPs/pull/611#discussion_r2688256693), I realized the reason access key deletion matters was not properly described in the NEP before.

This is a main reason to not use the `SenderReservedBalance` approach, please see relevant Zulip discussions (from the same thread):
[#core/spice > SPICE - Transaction Validity Problem @ 💬](https://near.zulipchat.com/#narrow/channel/503182-core.2Fspice/topic/SPICE.20-.20Transaction.20Validity.20Problem/near/521360878)
[#core/spice > SPICE - Transaction Validity Problem @ 💬](https://near.zulipchat.com/#narrow/channel/503182-core.2Fspice/topic/SPICE.20-.20Transaction.20Validity.20Problem/near/520992569)

I updated the NEP accordingly. Please let me know if you still would like to discuss the `SenderReserveBalance` approach further.

neps/nep-0611.md
jakmeier: Yeah that's perfect, thank you! The NEP doesn't need to specify the RPC changes but it is good to have thought about it.

neps/nep-0611.md
jakmeier: Yes, that makes sense. Base the costs on a normal `Transfer` plus the additional trie accesses.

For context, a `Transfer` today only modifies the access keys on a gas refund from a function call access key. This cost is indeed not well represented and we should not copy previous mistakes in new implementations.

But if you look closer, it gets worse. `Transfer` costs are significantly more expensive when the receiver has an implicit account id, since it could trigger an implicit account creation. Except in case of refunds, which can never trigger implicit account creation. So for a refund `Transfer`, the gas amount charged is the same for all receivers. But also, then the gas price is always zero anyway because it is a refund... So yeah, let us not copy any of that, either. :)

neps/nep-0611.md
jakmeier: Thank you for the links and the new explanation in the NEP! Now I understand the main concern with the alternative approach.

neps/nep-0611.md
jakmeier: I see, thank you for the additional explanation! That's more than enough for an outline. :)

jakmeier:
As an SME, I have reviewed the proposal. Kudos to the team behind it, it is well designed thought through. Special thanks to @darioush for finalizing the NEP and being very responsive to any questions and suggestions I had.

## Technical summary

The NEP builds the basis for a new concept, the pending transaction queue (PTQ). This queue is for transactions that are included for execution but have not been applied to the latest known blockchain state, yet.

The PTQ allows decoupling consensus on new chunks from transaction execution. Transactions to be included in a new chunk are verified against the state of an older block height, plus all pending transactions.

The challenge with this approach is that a transaction may look valid when it enters the PTQ but is no longer valid once it gets executed. This is especially problematic if the signer no longer has the balance to pay for the gas cost it owes for handling the transaction, or if the signing key got removed and it is no longer known whether it was a valid signature in the first place.

To address these concerns, a new type of access key is introduced, called gas key. Gas keys have a balance reserved for gas that cannot be removed without it being visible in the PTQ. Therefore, it is possible to verify ahead of time that a transactions signed with a gas key will have the balance to cover the gas cost once it executes.

For transactions signed by normal access keys, such a verification is not possible. Each account will be limited to a small number (4) of transactions per account in the PTQ to reduce the number of transactions that may go through the PTQ without being charged in the end.

### Recommendation

I recommend accepting this proposal to move SPICE development forward.

See benefits and concerns below for the arguments in both directions.

### Benefits

* Enables decoupling of consensus from execution


### Concerns

| - | - | - | - |
| 1 | Existing signing workflows may be broken. | This would only become a problem later when SPICE is implemented. For now the PTQ is always empty. Even then, probably only a few users would be affected and they can easily migrate to gas keys. | Has a good solution with plenty of time. |
| 2 | NONCE handling may not be what developers want. | ? | Still open. See #522 for what some developers wanted. |
| 3 | Normal access keys can still be used to submit transactions that fail and can't be charged. | By limiting the number of such transactions per account, such an attack would require creating many accounts. This itself is a cost, making the attack no longer free. | Addressed in the proposal. |
| 4 | This change is almost exclusively to enable SPICE, which itself is not fully designed, yet. Moving ahead with this implementation now has little benefit on its own. There is a risk we add features and change APIs that we need to revert later. | ? | Still open. |

neps/nep-0611.md
```diff
@@ -0,0 +1,470 @@
...
+2. Allowing users with sophisticated issuance needs to prepay for gas and have many in-flight transactions. This NEP additionally includes a way to associate a single signing key and balance with an array of nonces, to provide a convenient way for parallel transaction issuance for such users.
+
+## Specification
+
+To solve this problem, we present two critical components which work together to ensure that all accepted
+transactions are valid. At a high level, they are:
+
+* **Access Key vs Gas Key**: We introduce a new type of access key, which we will call "Gas Key". A gas key can be funded with NEAR, and when issuing a transaction using a gas key, the transaction only consumes gas from the gas key, not from the account.
+* **Pending Transaction Queue**: Chunk producers keep track of pending (accepted into a block, but not 
+  executed) transactions, and ensures that transactions sent with access keys are limited in parallelism,
```
matejpavlovic:
```suggestion
  executed) transactions, and ensure that transactions sent with access keys are limited in parallelism,
```

neps/nep-0611.md
matejpavlovic: Adding to the above: A weaker form of this problem still exists even before we decouple transaction availability, because even then the chunk producer for height h+1 might not be aware of the transactions in the chunk for height h (same shard) by the time it needs to produce its chunk (for h+1). In other words, the definition of pending transactions as "included on chain but not yet executed" is not sufficient to fully solve the problem, because there exist transactions that should be considered pending, but are not yet included on the chain. Let's call these transactions "in-flight". It seems that unless those in-flight transactions (or at least their gas limits) are not directly communicated to the chunk producer, it is impossible to completely prevent the no-gas-attack.

neps/nep-0611.md
```diff
@@ -0,0 +1,470 @@
...
+* The remaining balance left in the key is **burned**. (This prevent the key deletion attack, in case in the future gas keys can be deleted programmatically.)
+  * For user's benefit we may consider failing this action if the balance exceeds a certain threshold. The implementation may include a `force` flag that overrides this check.
+  * Note that this does not mean that all balance in a gas key is eventually burned; one can withdraw from
+    a gas key using the `TransferFromGasKey` action.
+
+#### Modifications to existing actions
+
+`DeleteAccount` action will fail if the account has any associated gas keys.
+
+`DeleteKey` is not allowed to delete gas keys.
```
matejpavlovic: Would it make sense to not introduce a separate `DeleteGasKey` action and instead modify the `DeleteKey` action and define it as follows?

"If the key is an access key, <whatever the current definition of `DeleteKey` is>. If the key is a gas key, <definition from above>"

I imagine the separate action for deleting gas keys might be an artifact from the times where we considered gas keys and access keys to have different name spaces. Now that gas keys share the namespace with access keys, the two separate actions don't seem necessary. Moreover, I'd personally argue that a single `DeleteKey` is much more intuitive from a UX perspective.

matejpavlovic: Realizing that another difference between these two actions is nonce deletion, I see that the proposed simplification might not work. I guess this UX hiccup is something we'll have to live with.

darioush: ~The other (main) reason to keep it separate is that contracts can currently request `DeleteKey` via the [`promise_batch_action_delete_key`](https://github.com/near/nearcore/blob/e730cc30f985d79c7b3962b82888b4d69ce36730/runtime/near-vm-runner/src/imports.rs#L234-L238) host call.~
^ This was not correct.

Actually, we can go with this approach based on this feedback and feedback from others.
The downside is we cannot charge gas for deletion accurately, but this seems not a practical issue.
We can use compute costs to prevent too much work in one chunk, and we could collect the gas ahead of time during addition if we really care about it.

gessellericauerta145-star (approved):

neps/nep-0611.md
```diff
@@ -0,0 +1,469 @@
...
+
+## Summary
+
+In the near future of the Near blockchain, we foresee that via the SPICE project, transaction and receipt
+execution will become decoupled from the blockchain itself; they will no longer run in lockstep.
+Instead, transactions will be included in the blocks first, and then execution will follow later.
+
+This inherently introduces a problem that we must accept transactions before we know whether they are
+valid. Today, when a chunk producer produces a chunk containing a transaction, it can verify using the
+current shard state that the transaction has a valid signature, has enough balance, and a valid nonce.
```
matejpavlovic: Nit: technically, it's the account that has the balance, not the transaction.
```suggestion
current shard state that the transaction has a valid signature and a valid nonce, and that the corresponding account has enough balance.
```

neps/nep-0611.md
```diff
@@ -0,0 +1,469 @@
...
+    balance to $B$.
+  * Right after, it floods the blockchain by signing and submitting many (arbitrary) transactions as $A$.
+    Because execution is asynchronous, the chunk producers think that there is still enough balance in
+    $A$'s account, so these transactions are accepted into chunks.
+  * Some blocks later, the execution catches up, and the `send_near` function drains $A$'s account.
+  * Subsequent executions of the following transactions all fail because $A$ has insufficient balance.
+  * After this is done, the attacker repeats but with $A$ and $B$ swapped.
+
+This attack can be carried out with a very simple script and requires no cost other than a single contract
+call every few blocks, but this ends up filling up the blockchain, denying legitimate transactions from
```
matejpavlovic: Nit:
```suggestion
call every few blocks, but this ends up filling up the blockchain, preventing legitimate transactions from
```

neps/nep-0611.md
```diff
@@ -0,0 +1,469 @@
...
+#### Access Key Parallelism Restriction
+
+We now restrict the ability to send multiple parallel pending transactions with Access Keys, from
+accounts that also have a contract deployed.
+
+Specifically, for any given account $A$ that has a contract deployed, the total number of access key transactions (across all access keys in the account) in the pending transaction queue whose sender is $A$
+cannot exceed $P_{\mathrm {max}}$, a constant determined by the epoch; we propose $P_{\mathrm {max}} = 4$.
+
+In other words, with traditional access keys, one cannot send more than 4 transactions from an account
+with a contract deployed, before they are executed. If one wishes the send more transactions in parallel,
```
matejpavlovic: Nit:
```suggestion
with a contract deployed, before they are executed. If one wishes to send more transactions in parallel,
```

neps/nep-0611.md
```diff
@@ -0,0 +1,469 @@
...
+Since gas keys are a kind of access key, they share a namespace. This means that the user is not allowed to create a gas key with the same public key as an existing access key. This is important in refund handling.
+
+### Gas Key Actions
+
+`AddKey` when adding a gas key is verified and executed as follows:
+
+* Account must already exist.
+* Check that the same key does not already exist as a gas key or access key.
+* Requested number of nonces must be less than or equal to `GasKey::MAX_NONCES` (currently suggested as 1024). This is to limit the number of trie operations for a single action.
+* Increases storage usage of associated account,
```
matejpavlovic: Nit:
```suggestion
* Increases storage usage of the associated account,
```

neps/nep-0611.md
```diff
@@ -0,0 +1,469 @@
...
+* Check that if the permission is a `FunctionCallPermission`, the allowance is `None` (unlimited
+  allowance).
```
matejpavlovic, line 158: Is this the case for access keys as well? If yes, I guess we require this for consistency - all good. If not, is there a specific reason we don't allow limiting the allowance for keys with `FunctionCallPermission`?

neps/nep-0611.md
```diff
@@ -0,0 +1,469 @@
...
+is exactly the pending transactions according to the definition above.
+
+The constraints are enforced as described above; we reiterate them precisely here. For each account, Let
+$T_A$ be the set of transactions in the queue signed with any access key of this account, and let
+$T_G$
+be the set of transactions in the queue signed with any gas key of this account. The following constraints
+must hold true:
+
+* If the account has a contract deployed (as of the latest available state), then $|T_A| \le P_{\mathrm {max}}$.
+* If any transaction $t\in T_A\cup T_G$ contains a `DeployContract` action, then $|T_A| \setminus \{t\}=\emptyset$. In other words, deploying a contract cannot be done in parallel with any access key
```
matejpavlovic: Nit:
```suggestion
* If any transaction $t\in T_A\cup T_G$ contains a `DeployContract` action, then $|T_A|=\{t\}$. In other words, deploying a contract cannot be done in parallel with any access key
```

neps/nep-0611.md
```diff
@@ -0,0 +1,469 @@
...
+
+
+The constraints are maintained at the time of chunk production: when producing a chunk, we only accept
+transactions that would maintain these constraints. For limiting gas key transactions, we always query the
+balance of the gas key from the executed state (as opposed to storing it in the queue).
+
+To compute the pending transaction queues (one for each tracked shard) for a new block,
+
+* Start from the queue from the previous block;
+* Subtract the transactions that are included in each newly certified block;
```
matejpavlovic, line 319: The reader needs to know how SPICE works to understand this. Maybe we should mention that and repeat the remark that if this NEP is implemented on the current system, the pending transaction queue will always be empty anyway.

matejpavlovic (approved):
Nice!

All the comments I added are just nit picking.

neps/nep-0611.md
darioush: Ordinary access keys can use the allowances.

One reason is to avoid having to track them in the pending transaction queue (similar to balances). 
I don't see a fundamental reason, so I guess we could add it later on if it was a request or requirement.

neps/nep-0611.md
```diff
@@ -0,0 +1,471 @@
...
+* Decreases storage usage of associated account,
+* The remaining balance left in the key is **burned**. This prevents an attack where an adversary floods the chain with transactions, then deletes the key to reclaim the balance before the transactions execute.
+  * For user's benefit we may consider failing this action if the balance exceeds a certain threshold. The implementation may include a `force` flag that overrides this check.
+  * Note that this does not mean that all balance in a gas key is eventually burned; one can withdraw from
+    a gas key using the `WithdrawFromGasKey` action.
+* To deter creating many keys and then deleting them all at once (which could consume excessive computation in a single chunk), a per-nonce compute cost is charged during execution. Combined with the 1024 max limit on number of nonces, this provides sufficient protection against computational DoS.
+
+#### Modifications to existing actions
+
+`AddKey` action will fail if the public key is already registered as a gas key (unless adding a gas key with that public key).
```
birchmd: Is the statement in paratheses here meant to imply that the action to add a gas key is idempotent? I think it makes sense to fail a duplicate add transaction instead of having it silently do nothing.

neps/nep-0611.md
```diff
@@ -0,0 +1,471 @@
...
+* The specified gas key must have sufficient balance to perform the transfer from.
+* To apply this action, the balance on the gas key is decreased by the specified amount, which is credited to the account balance.
+
+`DeleteKey` when deleting a gas key is verified as executed as follows:
+
+* The gas key must exist under the account.
+* To apply this action, all relevant trie nodes are deleted,
+* Decreases storage usage of associated account,
+* The remaining balance left in the key is **burned**. This prevents an attack where an adversary floods the chain with transactions, then deletes the key to reclaim the balance before the transactions execute.
+  * For user's benefit we may consider failing this action if the balance exceeds a certain threshold. The implementation may include a `force` flag that overrides this check.
```
birchmd: I think having this threshold is a good idea and should be included officially in the NEP rather than being left as an ambiguous maybe.

neps/nep-0611.md
```diff
@@ -0,0 +1,471 @@
...
+As an alternative, we could use the [estimator](https://github.com/near/nearcore/blob/master/docs/architecture/gas/estimator.md) to calculate these fees. However, this is known to not be accurate and ignored in favor of consistency with other fees in recent additions ([example](https://github.com/near/nearcore/issues/14160)).
+
+#### Gas Key Transactions
+
+`TransactionV1` reuses the fields of `TransactionV0` except replacing the `nonce` field with an enum:
+
+```rust
+enum TransactionNonce {
+    Nonce { nonce: Nonce },
+    GasKeyNonce { nonce: Nonce, nonce_index: u32 }
```
birchmd: Nit: if we the limit on parallel nonces will be 1024 then we could use `u16` here for the nonce index instead of `u32` just to save a few bytes of space in the serialized form.

neps/nep-0611.md
```diff
@@ -0,0 +1,471 @@
...
+  transactions, except
+  * The gas key cost is deducted from the gas key instead of the account. (Deposit cost is still deducted from account).
+  * The new nonce is written for the specific nonce ID under the gas key.
+
+
+##### What happens if account cannot cover deposit?
+
+Transaction processing is extended with a new failure case: `InvalidTxError::NotEnoughBalanceForDeposit`. In case the account programmatically sends away its balance we need to be able to charge the user a fee, otherwise we will have the spam issue this NEP intends to prevent. With honest chunk producers, this case only happens when using a gas key. 
+
+*Note*: As of the current date, `nearcore` ignores failed transactions in processing. After `ProtocolFeature::InvalidTxGenerateOutcomes`, invalid transactions impact the outcome but not the state. Therefore, this NEP also introduces the change that failed transactions may update the state (to charge for gas).
```
birchmd: Do I understand correctly that this statement means the gas key balance will be reduced even in the case of a failed transfer due to insufficient balance? I think that does make sense for DoS prevention, I am just double checking because it's not fully clear to me in the text as written.

neps/nep-0611.md
```diff
@@ -0,0 +1,471 @@
...
+
+Unlike access key transactions, gas key transactions are not limited in parallelism; they are only limited
+by the amount of gas these transactions consume. Specifically, for a gas key $G$, the **gas key cost** of 
+the transactions signed with $G$ in the pending transactions must not exceed the balance of $G$
+(according to the state that the pending transactions are based on).
+
+This constraint should be good enough to cover cases of adding, funding, removing, or withdrawing from gas
+keys as well. For adding and funding, we do need the execution to catch up before the newly available
+balance can be used for pending transactions, which is suboptimal but correct. 
+Note that contract execution cannot create receipts that withdraw from gas keys; only gas key
```
birchmd: This is because we are intentionally not updating the promise-creating host functions in the VM to include the new action? I'm not sure where it is appropriate to do so, but it would be good to document that these actions are intentionally excludes as opposed to missed by an oversight so that it is not accidentally added in the future.

birchmd (commented):
As a protocol working group member, I lean towards approving this proposal.

I have left some minor comments, but overall I think this proposal is well-motivated and technically sound. Rolling out SPICE-related changes as small, self-contained and backwards-compatible pieces will ease the transition for all participants in the Near ecosystem.

neps/nep-0611.md
darioush: Good call, we will discuss in the SPICE implementation group on Monday and update the NEP with a suitable value.

Please let me know if you have an opinion about the `force` flag as well.

neps/nep-0611.md
darioush: The duplicate add transaction is intended to fail. The writing here is confusing, I updated the paragraph.

neps/nep-0611.md
darioush: Good point, with u16 we can still increase the max to ~65k nonces.
I will also discuss this on Monday with the implementation group.

neps/nep-0611.md
darioush: That's correct. Updating to capture this comment.

neps/nep-0611.md
darioush: Yes, that's correct. In this scenario, the gas key balance is reduced to charge for gas even though the transaction fails.

I'll clarify this in the text.

neps/nep-0611.md
matejpavlovic: Do you mean that we'd need to track the allowance changes? Would the problematic case be wrongly accepting a transaction that exceeds the allowance because the allowance has only recently been decreased by some transaction in the pending queue? If that's the idea, then it makes sense to me - feel free to resolve this comment.

neps/nep-0611.md
darioush: We suggested 1 NEAR as this threshold. 

Additionally the `force` option is a leftover from a previous revision where we introduced separate actions to delete gas keys. To keep the same action, we can't add such option, so I removed its mention here.

neps/nep-0611.md
darioush: Changed to u16.

gagdiez:
Hi everyone! seeing that some members of the WG voted, and by the request of @walnut-the-cat, we are moving forward to meet and vote this NEP

@robin-near, @jakmeier, @matejpavlovic, @darioush if you could share your availability [here](https://doodle.com/group-poll/participate/elkD64jd) we will try to find some time that works for everyone

@robin-near / @darioush would be awesome if one of you could present a high level explanation of the NEP. No need to have a full presentation, just 1 - 3 slides, enough overview for people to understand the gist of it.

Thank you everyone!

gagdiez (approved):

Issue #607, opened 2025-05-05
[DISCUSSION] Sharded contracts design


jakmeier:
This issue is to outline several considered designs for [NEP-605](https://github.com/near/NEPs/pull/605) introducing sharded contracts.

The primary goal of that NEP is to create on-chain applications on NEAR that are able to utilize the throughput of more than one shard.

# Related issues

- Account Extensions discussion #478
    - Resulting NEP from that discussion #480
- Wasm submodules NEP #481

# Overview of design proposal

All currently discussed designs make use of the recently added [global contract feature](https://github.com/near/NEPs/pull/591) to have the code available on all shards. That code can either be mutable and defined by an account id, or immutable and defined by contract hash. Both should be supported with sharded contracts.

Further, all discussed designs assume that the state of the application is spread across user accounts, rather than application accounts. Each user holds the state for their own usage of the application on their own account. There is no shared application state between accounts.

For the two main designs in considerations, every user of an application must opt-in by executing a new action (`UseShardedContractAction`) on their account with the application id as an argument.

For these two design proposals, we have also decided that this state should be namespaced to isolate it per application. Technically, this means the `TrieKey::ContractData` for namespaced accounts needs to include an application id. We propose to use the account id  for dApps with mutable contract code and the contract code hash for dApps with immutable contract code.

*Note: Everything described so far essentially follows the same implementation as also proposed in #478 to enable Account Extensions. The two main differences are that (1) we now already have global contract code and (2) sharded contracts don't necessarily need synchronous execution. Thus limiting the total scope significantly.*

On top of what is described above, sharded contracts require an answer to the following three questions, in which the currently considered designs differ.

1. Should namespaced accounts have a public account ID that can be used as receipt predecessor and receiver?
2. How can a namespaced contract check that it was called by code of the same application?
3. How to prevent namespaced accounts from accessing unrelated assets?

### Open Issues for all designs

There are also still open issues which need to be solved in all proposed designs.

- How to update a sharded contract without inconsistencies during the update?
- How to enable full compatibility with existing contract standards, such as calling `ft_transfer_call` without additional requirements on the receiving contract?
- How will event emitting and tracking work for sharded contracts?

We shall put those issues aside for now and focus on the bigger design trade-offs first. These details are unlikely to influence this decision.

## Solution 1: Namespaced account IDs

1. **Separate ID?** Yes, we create a separate account ID per namespaced account.
2. **Same code check?** Contract code can parse the `predecessor_id` to check if the caller runs the same code.
[...]

PR #616, opened 2025-07-15
Deterministic AccountIds

mitinarseny:


fusede:
Hey @garikbesson, thanks for pointing these out! Fixed ✅

mfornet:
Is it possible to solve this problem today with a smart contract. Say you deploy a contract `deterministic.near`, with the necessary logic to make all deterministics deployments, and make it such that it has no access keys.

Deployed contracts will be subaccounts of this account.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+
+```rust
+/// Initialization state for non-existing contract with
+/// deterministic account id, according to NEP-616.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct StateInit {
+    /// Code to deploy
+    pub code: ContractCode,
+    /// Optional key/value pairs to populate to storage on first initialization
+    pub data: BTreeMap<Vec<u8>, Vec<u8>>,
```
mfornet, line 67: A more general pattern is to indicate wether or not there is a init method that needs to be called, and populates the state. It is arguably easier, and allows more complex, general, and safer initializations.

mitinarseny: Hey @mfornet, thank you for diving into proposal! Your concern is valid and it has been discussed [here](https://github.com/near/NEPs/issues/607#issuecomment-3073896730). 
TLDR: calling a method doesn't ensure deterministic state (due to availability of non-deterministic inputs like `env::block_height()`) and doesn't provide guarantees for gas consumption, along with creating more challenges for optimizations on validators by forcing them to actually load WASM and execute `init()` function to clean up local state. 

Do you still think it might be better to call `init()` method?

fusede:
> Is it possible to solve this problem today with a smart contract. Say you deploy a contract deterministic.near, with the necessary logic to make all deterministics deployments, and make it such that it has no access keys. Deployed contracts will be subaccounts of this account.

@mfornet You’re right, but then every creation of such account would need to go through a single `deterministic.near` account, which ruins the whole idea of sharing. Deployment of new instances of sharded deterministic contracts will happen much more often than it seems. See `escrow.near` example [here](https://github.com/near/NEPs/issues/607#issuecomment-3071350472).

neps/nep-0616.md
mitinarseny: Initialization state for sharded contracts is supposed to be simple enough, so that in majority of cases it will be stored in a single (empty) storage key. Here is an [example](https://github.com/mitinarseny/near-sdk-rs/blob/d9950de34084a352bd58bbb0872e686fc7d40d04/near-contract-standards/src/sharded_fungible_token/wallet.rs#L124-L145) of such init state construction for Sharded Fungible Token.

garikbesson:
Thank you @mitinarseny for submitting this NEP. I would now request @near/wg-protocol to assign two reviewers.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+pub enum ContractCode {
+    /// Reference global contract's code by its hash
+    GlobalCodeHash(CryptoHash),
+    /// Reference global contract's code by its [`AccountId`]
+    GlobalAccountId(AccountId),
+}
+```
+
+Now, let's [derive](https://github.com/mitinarseny/near-sdk-rs/blob/50eb6e68544b75288145f7f0d07068a482b9e3fc/near-sdk/src/state_init.rs#L189-L203)
+`AccountId` deterministically as `"0s" .. hex(keccak256(state_init)[12..32])`:
```
birchmd, line 81: Pure bikeshedding here, but I wonder if we could pick a different prefix to make these addresses more recognizable (i.e. not creating confusion with eth-implicit addresses). Maybe something like `d-`, where `d` means "deterministic".

It should be noted that because of [NEP-492](https://github.com/near/NEPs/blob/master/neps/nep-0492.md), no user is allowed to create top-level accounts via the `CreateAccount` action; this is important for the security of protocols that depend on deterministic addresses for verifying the code of predecessor calls.

However, it is worth double checking that no account IDs already exist on mainnet which happen to match whatever format for deterministic addresses we choose. Prior to NEP-492, users were allowed to create top-level accounts so it is possible a legacy account exists. This was an issue for eth-implicit accounts for example.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+mechanics where one wants to sponsor the deployment but doesn't bother to
+actually initialize it.
+
+To support this case, we say that each deterministic account can be in one of
+following states:
+
+* `nonexist`: there were no accepted receipts on this account, so it doesn't
+  have any data (or the contract was deleted). Initially all deterministic
+  accounts are in this state.
+* `uninit`: account has some data, which contains balance and meta info. An
```
birchmd, line 220: This possibility of having deterministic addresses that are not yet initialized adds some complexity to this proposal, which I am not sure is necessary. The only use case this enables is partial pre-funding of deterministic accounts because if the account was being fully funded then we might as well enforce the usage of the full `StateInit` action. So my question is whether partial pre-funding is important enough functionality to warrant the added complexity.

For example, allowing deterministic address accounts to be created by a `Transfer` action means these deterministic addresses must be classified as a kind of [implicit account](https://github.com/near/near-account-id-rs/blob/main/src/account_id_ref.rs#L52). This has downstream impacts on tooling such as wallets.

On the other hand, if we do not allow creation by `Transfer` then these accounts do not need to be considered implicit and would only be created by the new `StateInit` action.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+contracts, which will emit these events in the future.  
+
+However, to properly track these cross-contract calls they would need parse
+function names (e.g. [`sft_transfer()`](https://github.com/mitinarseny/near-sdk-rs/blob/50eb6e68544b75288145f7f0d07068a482b9e3fc/near-contract-standards/src/sharded_fungible_token/wallet.rs#L26-L55))
+and their args, while this information combined with receipt status already
+contains all necessary info for indexing.
+
+## Unresolved Issues
+
+1. Can we leverage [zero-balance accounts](https://github.com/near/NEPs/blob/master/neps/nep-0448.md)
```
birchmd: I don't think we need to do anything special to leverage ZBAs. It is simply part of the protocol that all accounts are allowed to have up to 770 bytes of state without holding any balance. So if an account created by the `StateInit` action had state (including code, keys and values) which totaled less than 770 bytes then that account can be created for free.

If more state than that is needed then the action must also attach enough base token to cover the storage staking for the full account state (i.e. the ZBA limit is not subtracted from the amount of storage that must be paid for).

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+signatures and tracking nonces by themselves instead of relying on
+[access keys](https://nomicon.io/DataStructures/AccessKey), then it would open
+up doors for upgradable [wallet implementations](https://docs.ton.org/v3/documentation/smart-contracts/contracts-specs/wallet-contracts)
+with [arbitrary plugins](https://docs.ton.org/v3/documentation/smart-contracts/contracts-specs/wallet-contracts/#wallet-v5)
+such as 2FA, social recovery, gasless transactions and [more](https://github.com/ton-blockchain/wallet-contract-v5/blob/main/README.md#suggested-extensions).  
+This can be seen as an alternative to [contract namespaces](https://gov.near.org/t/proposal-account-extensions-contract-namespaces/34227).
+
+### Co-location
+
+It might be possible to implement some analog of [`closeTo`](https://docs.ton.org/v3/documentation/smart-contracts/tolk/tolk-vs-func/create-message#sharding-deploying-close-to-another-contract)
```
birchmd, line 461: For what it is worth, I am not sure if something like this will be implemented on Near. It has been a core principle from the beginning of Near that the details of the shard layout are intentionally obscured from users (this is why calls to other contracts are always treated as if the receiving contract is in another shard, even when it is not). The idea behind this principle is that dynamic resharding should not break any user expectations about where contracts are located in the real shard layout. And dynamic resharding has been a goal for Near since day one. Note that if dynamic resharding is implemented well then users should never need to care about co-locating contracts because real usage patterns will dictate which contracts end up in the same shard.

neps/nep-0616.md
mitinarseny: >  So my question is whether partial pre-funding is important enough functionality to warrant the added complexity

@birchmd Great question!
I would recommend not to limit ourselves and allow for "implicitness" and accepting the incoming transfers without `StateInit` action. This becomes extremely useful if we ever come to [Wallet Extensions](https://github.com/mitinarseny/NEPs/blob/deterministic-account-ids/neps/nep-0616.md#wallet-extentions):
> If we allow deterministic accounts to handle external messages by verifying signatures and tracking nonces by themselves instead of relying on [access keys](https://nomicon.io/DataStructures/AccessKey), then it would open up doors for upgradable [wallet implementations](https://docs.ton.org/v3/documentation/smart-contracts/contracts-specs/wallet-contracts) with [arbitrary plugins](https://docs.ton.org/v3/documentation/smart-contracts/contracts-specs/wallet-contracts/#wallet-v5) such as 2FA, social recovery, gasless transactions and [more](https://github.com/ton-blockchain/wallet-contract-v5/blob/main/README.md#suggested-extensions).
This can be seen as an alternative to [contract namespaces](https://gov.near.org/t/proposal-account-extensions-contract-namespaces/34227).

Suppose that we ever let contracts to handle "external messages". I understand that this might be very different from current transaction model, but it's worth to think about future extensibility using this feature.  
So, from the top of my head, this is how the high-level implementation can look like:
1. We define a new `TransactionV2` which allows specified contract to handle creation of the first receipt(s) instead of Access Key.
1. We define `ReceiptV2` which has `predecessor_id: Option<AccountId>`, where empty predecessor means that this receipt was created by "external message".

Here is how an implementation for wallet contract can look like (simplified, inspired by [wallet-v5](https://github.com/ton-blockchain/wallet-contract-v5/tree/main)):
```rust
#[near(contract_state)]
pub struct WalletData {
    pub wallet_id: u32, // a single public key can have multiple wallets
    pub nonce: u64,
    pub public_key: [u8; 32], // only one EdDSA key for illustration
    // a set of plugins that can also authenticate txs (e.g. 2FA, social recovery)
    pub extensions: BTreeSet<AccountId>,
}

impl WalletData {
    pub fn init(wallet_id: u32, public_key: [u8; 32]) -> Self {
        Self {
            wallet_id,
            nonce: 0,
            public_key,
            extensions: BTreeSet::new(), // by default, no plugins installed
        }
    }
}

pub struct WalletExternalMessage {
    pub wallet_id: u32,
    pub nonce: u64,
    pub actions: Vec<Action>, // actions to be executed
    pub signature: [u8; 64], // only EdDSA signature for simplicity
}

impl WalletExternalMessage {
    pub fn hash() -> CryptoHash {
        // construct envelope and calculate hash
    }
}

#[near]
impl WalletData {
    pub fn handle_external_message(&mut self, msg: WalletExternalMessage) {
        require!(env::predecessor_account_id().is_none(), "not external message");

        require!(env::ed25519_verify(&msg.signature, &msg.hash(), &self.public_key), "invalid signature");

        require!(msg.wallet_id == self.wallet_id, "wrong wallet");

        require!(msg.nonce == self.nonce, "wrong nonce");
        self.nonce += 1;

        for action in tx.actions {
            self.handle_action(action);
        }
    }
}
```
So, if we ever do that, then users can create accounts deterministically derived from their public key!
But, in order to send txs, they would need someone to fund it first. it will be much simpler to withdraw NEAR from CEX to this address, rather than asking CEX to also deploy it via `StateInit` action which has user's public key, etc... Instead, CEX will just send NEAR to this account, and later my wallet will embed `StateInit` in the first tx that I send from this wallet. 

What do you think about such potential use case in the future?

neps/nep-0616.md
mitinarseny: Very cool! So, since [wallet-contract](https://github.com/near/near-sdk-rs/pull/1376) implementation for sFT takes only ~310 bytes, does it mean they can be created for free?

birchmd: Yes, it should work. The ZBA feature is also why you can create an eth-implicit account with a transfer of 0 $NEAR ([exmaple](https://testnet.nearblocks.io/txns/FHa2cpLBhNfKXaoyiV9Lto7Gz3tWnvRXbbSqJKXFmcPK)).

neps/nep-0616.md
birchmd: Funding a deterministic address by withdraw from a CEX is a compelling use case.

Alright, if we are going to make these deterministic addresses implicit accounts then we also need to define the mechanics of the creation. We currently have two kinds of implicit accounts: Near implicit account and eth-implicit accounts each with different actions on creation. Near implicit accounts are defined by hex encoding a full public key as the account ID and when a new Near implicit account is created that public key is automatically added as a full access key on the new account. Eth-implicit accounts are meant for Ethereum compatibility, so the account IDs look like Ethereum addresses and the eth-implicit wallet contract is automatically deployed when the account is created.

What should happen when a deterministic address is created by a transfer? I think it makes sense for nothing to happen (no access keys added and no code deployed either) because the code and state will show up later in a `StateInit` action.

mitinarseny: You're right, nothing should be done except for recording the balance on this account.

Here is a much simpler example - NEAR lockup deterministic contract:
```rust
#[near(contract_state)]
pub struct LockupData {
    pub beneficiary_id: AccountId,
    pub unlock_after: BlockHeight,
}

#[near]
impl LockupData {
    pub fn claim(self) -> Promise {
        require!(env::block_height() >= self.unlock_after, "too early");
        Promise::new(env::current_account_id()).delete_account(self.beneficiary_id)
    }
}
```

Someone can fund it by a simple NEAR withdrawal from CEX, and for the beneficiary it only makes sense to  initialize this deterministic account via `StateInit` and claim the locked NEAR tokens after `unlock_after` block.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+pub enum ContractCode {
+    /// Reference global contract's code by its hash
+    GlobalCodeHash(CryptoHash),
+    /// Reference global contract's code by its [`AccountId`]
+    GlobalAccountId(AccountId),
+}
+```
+
+Now, let's [derive](https://github.com/mitinarseny/near-sdk-rs/blob/50eb6e68544b75288145f7f0d07068a482b9e3fc/near-sdk/src/state_init.rs#L189-L203)
+`AccountId` deterministically as `"0s" .. hex(keccak256(state_init)[12..32])`:
```
jakmeier, line 81: To verify a predecessor id, a contract needs access to the full `StateInit` structure, right?

For example, to check if an incoming FT transfer is valid, I need to reconstruct the struct and insert the user account id of the sender, hash it, and compare it against the predecessor.

What if large parameters are stored in init? Did you also consider hashing as `hash( hash(state_init.code), hash(state_init.data) )`? Or even `hash( hash(state_init.code), hash(state_init.data[0]), hash(state_init.data[1]), ... hash(state_init.data[n]) )`?

Not sure if this is better than what is in the proposal. But I think we should consider this well before we lock in a specific hashing schema.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+```rust
+/// Initialization state for non-existing contract with
+/// deterministic account id, according to NEP-616.
+#[derive(BorshSerialize, BorshDeserialize)]
+pub struct StateInit {
+    /// Code to deploy
+    pub code: ContractCode,
+    /// Optional key/value pairs to populate to storage on first initialization
+    pub data: BTreeMap<Vec<u8>, Vec<u8>>,
+}
```
jakmeier, line 68: Would it make sense to add a version field to `StateInit` that's also hashed in for the deterministic account id? This way, we could change the structure of `StateInit` in a protocol update without worrying about collisions of old and new borsh representations of the data that would lead to conflicting account ids.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+> NEAR today by using `DeleteAccount` action with non-existing `beneficiary_id`.
+
+### Other host functions
+
+We would also need to implement a couple of other trivial host functions, which
+are required for contracts to function properly:
+
+```rust
+/// Returns code deployed on current contract being executed.
+/// For now, only references to globally deployed contracts are supported.
```
jakmeier, line 346: Is the reason for only supporting that returning the full contract code would be too expensive? If so, we could return the code hash (that's the only thing ~actually stored on chain~ directly stored on the account anyway) and that would also be quite cheap.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+pub fn promise_result_length(result_idx: u64) -> Result<u64, PromiseError>;
+
+// Currently constants, but better to be exposed as host functions,
+// so that changing the protocol config doesn't brick existing contracts.
+pub fn storage_byte_cost() -> NearToken;
+pub fn storage_num_bytes_account() -> StorageUsage;
+pub fn storage_num_extra_bytes_record() -> StorageUsage;
+```
+
+## Reference Implementation
```
jakmeier, line 366: note: We will also need a reference implementation of nearcore at some point.

neps/nep-0616.md
mitinarseny: Yes, I was referring to your [comment](https://github.com/near/NEPs/issues/607#issuecomment-3068826686):
> [...] Note that when executing a WASM module, we usually don't load the raw WASM, we only load the prepared executable which is stored by hash of the raw WASM.

Are you suggesting adding `ContractCode::InlineHash(CryptoHash)` variant, too? It would work for retrieving current contract's code. However, this enum is also used in `StateInit`, so receiver's shard would also need to have this (optionally, prepared) WASM to be present in its local state, which we cannot guarantee.

neps/nep-0616.md
mitinarseny: I was also thinking about it. The first field is `ContractCode` enum, so that the first byte denotes its variant. So, if we ever want to introduce a new version for `StateInit`, we can use `u8::MAX` for making it prefixed and the next byte would be the version.

Do you think keeping this as-is is a good idea or we should explicitly include the version from the beginning?

Note that if we ever introduce a new version, which only extends these fields, then this newer version can (and should) then be used by new contracts to deploy new deterministic contracts. However, serialized representation would be different, and, thus, the hash... So, there is no way for a new contract to find out which version should be used to derive deterministic account id of another contract it interacts with.

neps/nep-0616.md
mitinarseny: > Or even hash( hash(state_init.code), hash(state_init.data[0]), hash(state_init.data[1]), ... hash(state_init.data[n]) )?

At first, I was thinking it might be actually a good idea: verifying contract would need only to have hashes of some data, while calculating other part of the data on-the-fly.

However, the contracts would still need to actually have this data in order to compose `StateInit` from it. So, how else would we benefit from it?

It's worth to mention here, that TVM stores all data (including contracts' bytecode) in [cells](https://docs.ton.org/v3/documentation/data-formats/tlb/cell-boc#cell): tree-like data structure, where nodes are identified by their [hashes](https://docs.ton.org/v3/documentation/data-formats/tlb/cell-boc#cell-hash). Global contracts are achieved by pushing [library cells](https://docs.ton.org/v3/documentation/data-formats/tlb/library-cells) containing the contract's code to MasterChain, but the same can be done for any other data. So, if we ever adapt similar model in Near and allow referencing storage key-value pairs by their global hashes rather than always inlining them, then it can reduce network and storage overhead, while making transactions cheaper for end users.

jakmeier: Yeah you are right, to initialise a new contract, we need the full data anyway. I guess the only benefit would be for contracts that are called by a sharded contract, such as the receiver of sft_transfer_call.

mitinarseny: I think now I get it: using hashes can optimize verification part, not the deployment. Maybe should we treat `state_init.data` as Merkle tree? So that only its root hash in needed to calculate the top-level hash for `StateInit`: `hash(hash(state_init.code) .. data_root_hash)`

neps/nep-0616.md
jakmeier: > I was also thinking about it. The first field is ContractCode enum, so that the first byte denotes its variant. So, if we ever want to introduce a new version for StateInit, we can use u8::MAX for making it prefixed and the next byte would be the version.

We sometimes applied this trick to do upgrades where we didn't think about upgrade ahead of time. but it always feels hacky to me. It requires manually altering the usually-derived (de-)serialization code, which is easy to get wrong. Putting this kind of code in smart contracts isn't pretty IMO.

> Do you think keeping this as-is is a good idea or we should explicitly include the version from the beginning?

I'd personally prefer an explicit version but I have no super strong opinion if you have arguments against it.

> Note that if we ever introduce a new version, which only extends these fields, then this newer version can (and should) then be used by new contracts to deploy new deterministic contracts. However, serialized representation would be different, and, thus, the hash... So, there is no way for a new contract to find out which version should be used to derive deterministic account id of another contract it interacts with.

Hm, I initially assumed that most existing contract will keep using the old version. A new version would be introduced to address a specific problem. Applications that don't have this problem can keep using the old version, I don't see a problem with that. It's a bit like IPv4 vs IPv6, or HTTP/2 vs HTTP/3. I don't think the old version will ever be fully replaced but some applications can benefit from features in the new version.

For applications that started with version 1 and want to upgrade to version 2, the solutions will probably look different for each case. But since using a new version will change the derived account it, they would be forced to create a new derived account id for each user with the new StateInit version. They would need to find a way to migrate, which should be possible with some effort but most likely would discourage most applications from upgrading unless they really need to. But again, I think that's fine.

On the off-chain /  indexer side, I think here any tool already needs to keep the full `StateInit` anyway to link a derived account id to its owner. This is just one more field on that struct. Having a version field would fit in well there. IMO this is better than shoehorning in a version into the existing enum somehow later.

neps/nep-0616.md
jakmeier: > Are you suggesting adding ContractCode::InlineHash(CryptoHash) variant, too? It would work for retrieving current contract's code.

Yes, this would make the new host function more complete. But I have to admit, it might not be of much use. :shrug: 

> However, this enum is also used in StateInit, so receiver's shard would also need to have this (optionally, prepared) WASM to be present in its local state, which we cannot guarantee.

Just because the host function has this variant, it doesn't mean we have to also support it in `StateInit` for derived account ids. (We can use different types under the hood.) But if we do, then it would probably be the contract's responsibility to get the original source code and send it inside `StateInit` message. But I'm not suggesting to add that, I'm only talking about the host function in this comment here.

neps/nep-0616.md
jakmeier: Yes I think that's a good idea! Not sure if it is necessary, so if you think it would be more complex than needed, then we can keep it as it is. But if you also think it's worth the flexibility on the verifier, I'm in favour of treating it as Merkle tree.

neps/nep-0616.md
mitinarseny: It does make sense if the traffic itself will dictate how close accounts should be located, not the applications itself. However, regular users usually send tokens no more than couple of times a day, so this dynamic resharding algorithm should account for not so active accounts, too, which contradicts the idea of "adapting to real traffic". Designing such algorithm seems to be extremely non-trivial. Am I missing something here?

It's worth to mention that TON Core team has just [released](https://t.me/toncore/74) Jetton 2.0 smart-contracts that utilizes `closeTo` for faster finality.

neps/nep-0616.md
mitinarseny: > But since using a new version will change the derived account it, they would be forced to create a new derived account id for each user with the new StateInit version. They would need to find a way to migrate, which should be possible with some effort but most likely would discourage most applications from upgrading unless they really need to.

The main issue is that two account ids corresponding to different `StateInit` versions are indistinguishable between each other just by looking at them. So, if we upgrade at least one sharded standard implementation to use a new version, then it might brick the entire ecosystem of sharded contracts using the old version: they will no longer be able to verify or deploy such contracts, and will no longer be composable with contracts using the new version (e.g. how  `.sft_receive()` can [deploy](https://github.com/mitinarseny/near-sdk-rs/blob/327ebe55e97e0f3f168679d0ed5b733a13658581/examples/sharded-fungible-token/wallet/src/lib.rs#L191-L215) its owner). Moreover, even for new contracts it will be unfeasible be compatible with both new and old versions, since the number of version combinations grows exponentially with number of composed interactions between different contracts with deterministic account ids.

> I initially assumed that most existing contract will keep using the old version. A new version would be introduced to address a specific problem.

It's always hard to predict the direction for future improvements, but, as for now, I can't imagine what other data, besides code and storage, can be used to dictate behavior of the contract. If we ever want to derive account ids deterministically from some other data, then it might be even not `StateInit` but something with completely different semantics, so we would need a separate proposal for it in the future.

I would recommend for new versions of deterministic account ids to be distinguishable between each other by encoding version byte into account id itself. For example, use a different `0t` prefix or any [other](https://github.com/near/NEPs/pull/616#discussion_r2236858857). New actions and host functions can be introduced, similar to how `function_call() -> function_call_weight()` migration was done.

jakmeier: > The main issue is that two account ids corresponding to different `StateInit` versions are indistinguishable between each other just by looking at them. So, if we upgrade at least one sharded standard implementation to use a new version, then it might brick the entire ecosystem of sharded contracts using the old version: they will no longer be able to verify or deploy such contracts, and will no longer be composable with contracts using the new version (e.g. how `.sft_receive()` can [deploy](https://github.com/mitinarseny/near-sdk-rs/blob/327ebe55e97e0f3f168679d0ed5b733a13658581/examples/sharded-fungible-token/wallet/src/lib.rs#L191-L215) its owner).

Wouldn't this be an argument in favour of adding an explicit version number? This way, the caller can set the version right there in the `notify: Option<TransferNotification>` in the example of `sft_receive`, which covers the deployment.

For the verification, I'm not sure I understand your issue. Can't a contract call simply add a version parameter if it wants to support more than one versions of the standard? Then the callee can verify the predecessor id as usual, using the version that the caller provided.

>  Moreover, even for new contracts it will be unfeasible be compatible with both new and old versions, since the number of version combinations grows exponentially with number of composed interactions between different contracts with deterministic account ids.

Can you make an example how this leads to exponential growth? I'm not following, in my mind, each specific function call of each contract has experiences only a linear growth of cases to handle. I'm unable to see the combinatoric effect. Probably I am misunderstanding the composition you are talking about. Please help me understand if this is the case.

> It's always hard to predict the direction for future improvements, but, as for now, I can't imagine what other data, besides code and storage, can be used to dictate behavior of the contract.

Deterministic account ids are tightly linked to the account model. Many changes on accounts can have an effect. Here is a short list things that I can think of that would require an update:

- Adding access keys as part of determinstic deployment, maybe for linking it to an implicit account on another chain or other creative use cases people come up with.
- Introduction of submodules after the fact for things like synchronous execution, which adds more than one code per account, which means the `InitState` has to assign the code to a submodule.
- Adding other runtimes than WASM, which means we also have to specify the code type rather than just the bytes.
- New storage models, such as some form of global state values or cells, which makes the single data field insufficient.
- Feature flags that contracts can enable or disable on deployment.
- Shard colocation requirements that contracts can enable or disable on deployment.
- A new field on the account to manage hooks, as proposed from time to time, for example [here](https://github.com/near/NEPs/issues/478#issuecomment-1552908785).

Probably all of these can be worked around in one way or another. While it's hard to predict the exact requirements of future changes, I know that I have wished many times that there was a version field when there wasn't one.

> I would recommend for new versions of deterministic account ids to be distinguishable between each other by encoding version byte into account id itself. For example, use a different 0t prefix or any https://github.com/near/NEPs/pull/616#discussion_r2236858857.

That's certainly also possible. This is just a personal preference but I don't like encoding more and more information in account ids.

How about this? You add a section that explain how new requirements for state initialization are handled in your proposal (e.g. one more prefix for each change) and you add the idea to add a version field as an alternative. You can argue that you think it's unlikely to ever be needed, so the version field is no necessary. Then other people can chime in with their opinions.

neps/nep-0616.md
birchmd: > Designing such algorithm seems to be extremely non-trivial

Yes, it certainly is. I expect dynamic resharding is still a long way off. My only point is that many design decisions have been made previously to leave the door open to dynamic resharding and I anticipate that trend will continue.

neps/nep-0616.md
birchmd: I agree with @jakmeier that including a version number as the first field (allowing this struct to be an enum in the future) is a good idea to make future improvements much less painful. And I don't think there is a drawback to including it now.

neps/nep-0616.md
mitinarseny: Sure, I'm not against dynamic resharding. It's the opposite: it seems to me as good way of "hinting" the algorithm that these two should be located close to each other. In addition, such design nicely incorporates paying fees for these complications: the closer you want to deploy your contract to another, the higher the fee is. We can even try to make this rent-like fee [instead](https://github.com/mitinarseny/NEPs/blob/deterministic-account-ids/neps/nep-0616.md#storage-rent-vs-fixed-storage-staking) of one-time fixed payment.

I'm just curious how does `closeTo` prevent from future implementation of dynamic resharding?

birchmd: Sure, using it just as a suggestion from the user as opposed to a requirement could work well. The general ethos has been to hide the underlying details of the shard layout from users because this allows dynamic resharding to freely change the layout without disrupting user expectations. As long as any `closeTo` implementation keeps to that ethos then maybe it can be added to the protocol some day.

neps/nep-0616.md
mitinarseny: > > Moreover, even for new contracts it will be unfeasible be compatible with both new and old versions, since the number of version combinations grows exponentially with number of composed interactions between different contracts with deterministic account ids.
>
> Can you make an example how this leads to exponential growth? I'm not following, in my mind, each specific function call of each contract has experiences only a linear growth of cases to handle. I'm unable to see the combinatoric effect.

Sure. Imagine a DeFi protocol that wraps sFT tokens `A` received from user in another sFT token `wrapA` (e.g. to indicate supplied collateral in lending protocol). Now, let's say that it wants to treat sharded and non-sharded (i.e. NEP-141) versions of the any token in the same way: i.e. apply same LTV ratios, share supply caps, etc...
When receiving (e.g. withdrawing collateral) `wrapA` tokens in [`.sft_on_receive()`](https://github.com/mitinarseny/near-sdk-rs/blob/327ebe55e97e0f3f168679d0ed5b733a13658581/near-contract-standards/src/sharded_fungible_token/receiver.rs#L6-L29), the contract needs to:
1. Compose `StateInit` for his wallet-contract for minter `wrapA` and verify that its derived account id matches the predecessor.
1. Compose `StateInit` for `wrapA` minter taking into account original account id of `A` and verify that its derived account id matches `minter_id` of his wallet-contract.
1. Compose `StateInit` for [Ft2Sft](https://github.com/mitinarseny/near-sdk-rs/blob/327ebe55e97e0f3f168679d0ed5b733a13658581/near-contract-standards/src/sharded_fungible_token/minter/ft2sft.rs#L18-L26) minter for original NEP-141 representation of `A` and verify that it matches account id of `A`.

If any of these could have used `StateInit::V2 { code, data, /* additional fields */ }` for deployment and address derivation, then the worst-case number of combinations that need to be checked before address verification succeeds is `2^3 = 8`. If we introduce `StateInit::V3`, then this number will be `3^3 = 27` and so on. So, the final formula will be `v ^ n`, where `v` is number of versions for `StateInit` and `n` is number of composable interactions between different contracts.

That's why I believe that we should use derivation scheme with a different prefix for each new `StateInit` version, so that verifying contract can distinguish between them rather than iterating over exponential number of combinations.

Let me address these:
> * Adding access keys as part of determinstic deployment, maybe for linking it to an implicit account on another chain or other creative use cases people come up with.

This can already be done by encoding these access keys into `StateInit` storage and exposing a permissionless method that adds these keys:
```rust
#[near(contract_state)]
pub struct Contract {
    pub init_access_keys: Vec<AccessKey>,
    // other fields ...
}

#[near]
impl Contract {
    pub fn init(&mut self) -> Option<Promise> {
        if self.init_access_keys.is_empty() {
            return None;
        }
        let mut p = Promise::new(env::current_account_id());
        for access_key in mem::take(&mut self.init_access_keys) {
            p = p.add_access_key(access_key);
        }
        Some(p)
    }

    // other methods ...
}
```

However, I don't see a case where access keys can be ever needed in deterministic account ids. I also do believe that all authentication logic should be performed by [contracts themselves](https://github.com/mitinarseny/NEPs/blob/deterministic-account-ids/neps/nep-0616.md#wallet-extentions), while having access keys implemented on protocol level limits the flexibility and seems to me as a rudimentary design.

> * Adding other runtimes than WASM, which means we also have to specify the code type rather than just the bytes.

This can be achieved by adding new variants in `ContractCode` enum.

> * Feature flags that contracts can enable or disable on deployment.

Feature flags should be encoded into `StateInit` storage, so it's up to the implementation how to handle them.

> * Shard colocation requirements that contracts can enable or disable on deployment.

That's a very good one, but it should be applied on top of `StateInit` and shouldn't be part of derivation scheme. See [`closeTo`](https://github.com/mitinarseny/NEPs/blob/deterministic-account-ids/neps/nep-0616.md#co-location) implementation in TON.

> * A new field on the account to manage hooks, as proposed from time to time, for example [here](https://github.com/near/NEPs/issues/478#issuecomment-1552908785).
> > A receipt issuance and receival hooks can allow to implement things like:
> > * block receival of not interesting assets or from non whitelisted/black listed accounts
> > * prevent account from transferring out specific types of assets (eg borrowed NFTs)
> > * Specify usage limits on the account

Receipt hooks is a very powerful feature and I do believe it's necessary! It allows for more flexibility, especially when it comes to [handling](https://github.com/mitinarseny/NEPs/blob/deterministic-account-ids/neps/nep-0616.md#prior-art) native refunds.

However, I also believe that there is no way to prevent receiving some assets, since they can be represented as completely different token standard, which does not support any ABI for transfer notifications like `.sft_on_receive()` or `.ft_on_transfer()`. Other features like imposing limits on storage usage and transferring out specific asset types should be implemented on the level of corresponding contracts.

> This is just a personal preference but I don't like encoding more and more information in account ids.

I think account ids is the best place to do so, since they are already included in all receipts as `predecessor_id` and we should utilize them as much as possible to avoid the need to forward additional info like "sharded contexts" and etc... At least, that's the whole idea of deterministic account ids.

To sum up: we can add versioning to `StateInit` now, but I really hope we will never need it.

jakmeier: Thank you for the detailed reply, I really appreciate it!

> If any of these could have used StateInit::V2 { code, data, /* additional fields */ } for deployment and address derivation, then the worst-case number of combinations that need to be checked before address verification succeeds is 2^3 = 8. If we introduce StateInit::V3, then this number will be 3^3 = 27 and so on. So, the final formula will be v ^ n, where v is number of versions for StateInit and n is number of composable interactions between different contracts.

This makes it very clear where the exponential number of combinations comes from. I'm still no sure if it's necessary to iterate through all of them, though. It seems to me the caller can always provide the version info, so there won't be any guession. But I can see how having it encoded in the account id prefix would make it easier.

Regarding the reasons for changing `StateInit`, I don't think it's worth discussing them all in detail. I admit these can all be worked around with necessarily changing `StateInit`, with certain assumptions on how these are implemented. Your assumptions are all reasonable. But for example with feature flags, it might be that these protocol level feature flags that I meant must be set at deployment and cannot be updated from Wasm. We just don't know yet what exact requirements there will come in the future.

My point is, if we make modifications to `StateInit` near impossible, we drastically limit the freedoms for future changes to the account state. That's why I suggest we leave the door open to upgrades as much as we can.

> I think account ids is the best place to do so, since they are already included in all receipts as predecessor_id and we should utilize them as much as possible to avoid the need to forward additional info like "sharded contexts" and etc... At least, that's the whole idea of deterministic account ids.

Absolutely! This thinking fits well with this new (to Near) concept of deterministic account id, I agree. I just dislike how we go further and further away from the simple account id model that Near used to be known for. 

From a purely technical perspective, I have no issues with this approach. From a "I need to explain what this address is to my grandmother" perspective, I don't like it. Every Near user has to interact with account ids, so that's why I dislike additional complexity exposed on that level.

But again, it's just an opinion. These opinionated product-level design decisions are tricky to agree within a community and I'm certainly not going to claim that my opinion is somehow better than anyone else's.

Also, if we can come up with an easy-to-understand scheme like for example `d0-*`, `d1-*`, `d2-*`, ... for different versions of deterministic account ids, I think that would be essentially the same mental load on account ids that we are adding anyway. Maybe then adding the field on `StateInit` is unnecessary. I just want us to have a good plan for how to the case where we want to add more information into `StateInit`.

> To sum up: we can add versioning to `StateInit` now, but I really hope we will never need it.

Fair enough, I fully agree with this!

I still think it's a good idea to write down the version field as one option and also have another option in the proposal. Then the WG-members can form their own opinion and agree on what they like best.

mitinarseny: > I'm still no sure if it's necessary to iterate through all of them, though. It seems to me the caller can always provide the version info, so there won't be any guession. But I can see how having it encoded in the account id prefix would make it easier.

You're right. In fact, this combinatorial issue will be present for sFT anyway due to different possible implementations of wallet- and minter- contracts. So, the contracts are most likely to [forward](https://github.com/mitinarseny/near-sdk-rs/blob/327ebe55e97e0f3f168679d0ed5b733a13658581/near-contract-standards/src/sharded_fungible_token/receiver.rs#L13-L18) global contract id of specific wallet-contract implementation and its corresponding minter-contract implementation in `msg` for `.sft_on_receive()`. Here is a real-world [example](https://github.com/evaafi/contracts/blob/40a0e8bb32f88df8e09def536371192a824d1c3d/contracts/master.fc#L381-L389) of how TON contracts deal with it.

> I still think it's a good idea to write down the version field as one option and also have another option in the proposal. Then the WG-members can form their own opinion and agree on what they like best.

Agreed and thank you for this explanation, I appreciate it! I'll add section about `StateInit` versioning to this NEP in next couple of days.

neps/nep-0616.md
mitinarseny: That's, of course, a nice addition. If I understand it correctly, [ZBAs](https://github.com/near/NEPs/blob/master/neps/nep-0448.md) and [non-refundable storage staking](https://github.com/near/NEPs/blob/master/neps/nep-0491.md) were introduced as a mitigation to remove economical incentive to delete accounts, which were created and sponsored by third-party relayers.

However, in case of sharded contracts with deterministic account ids, it's up to implementation whether to allow deletion of itself or not. I assume that the majority of sharded contracts will be using a small amount of storage. So, if ZBA rules apply to deterministic accounts, too, then I'm afraid we might end up with 90% of accounts in Near are not paying for their storage, which might disincentivize validators.

@birchmd @jakmeier what do you guys think if we explicitly say that ZBA rules are not applied to deterministic account ids? This will still let other projects to fund onboarding users via [wallet implementations](https://github.com/mitinarseny/NEPs/blob/deterministic-account-ids/neps/nep-0616.md#wallet-extentions) with "delete me" functionality removed or by utilizing [non-refundable storage staking](https://github.com/near/NEPs/blob/master/neps/nep-0491.md) to fund these accounts (I prefer the latter more).

PS: Should the same reasoning apply to contracts deployed via `UseGlobalContract` action from [NEP-591](https://github.com/near/NEPs/blob/master/neps/nep-0591.md)?

birchmd: I think it is ok for deterministic account ids to be allowed to have zero balance. We should encourage developers to automatically delete accounts that have served their purpose, but even if that does not happen I think the node implementation can be made more efficient to deal with it. For example, state that has not been used in a long time can be archived to some cheaper storage solution and only restored again if it is requested.

770 bytes is not very much in the grand scheme of things. I think @jakmeier previously did some analysis on storage cost and concluded that the gas cost to create an account was enough to pay for storing a ZBA for many years on the cheapest cloud storage. If we are really worried about it we could make the gas cost for the `StateInit` action higher than it would need to be otherwise to make up for the potential future storage cost. This is the approach that was also taken for storing global contracts which also do not require a storage deposit.

jakmeier: So, yes, the specific problem that ZBAs address isn't as relevant for deterministic account id. We could make an exception for them and disable the ZBA feature for those accounts. But I don't think we have to.

Regarding non-refundable storage, just to clarify, this has never been actually included in the protocol. But #491 describes the concept that I think could be applied here, if we need it.

Also, in practice, the cost of holding a tiny amount of NEAR on the account has been proven to be too small to be a real incentive to delete accounts, despite the [cost per byte being quite high](https://github.com/near/nearcore/issues/8492#issuecomment-1419385080). So, the final result of non-refundable storage, a ZBA limit, or the storage staking model is always kind of the same anyway - people create once and they never delete.

Perhaps the more relevant consequence for this proposal comes from whether state is paid in gas or in NEA. Passing along a NEAR balance across calls can be much more annoying than gas.

- With ZBAs, it could all be handled with gas costs as long as the account stays below 770 bytes. This is great because the signer of the transaction pays the gas upfront.
- However, if `InitState` requires a NEAR balance to create the account with the deterministic id, then it can be awkward when one sFT contracts initializes another, since the balance needs to be paid by the calling contract. 

For that reason, I think just using the existing ZBA limit will be the easiest.

neps/nep-0616.md
jakmeier: I will start working on the reference implementation tomorrow

jakmeier: Progress can be observed here: https://github.com/near/nearcore/pull/14080

@mitinarseny you can link to that PR in the proposal if you want

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+    Note over C: Failed
+
+    C -->>+ B: failed receipt
+    B -->>- A: refund???
+
+    C -)- B: refund: attached_deposit
+```
+
+To remove this limitation let's add an optional `refund_to` field to [`Receipt`](https://github.com/near/nearcore/blob/685f92e3b9efafc966c9dafcb7815f172d4bb557/core/primitives/src/receipt.rs#L81-L93)
+and a host function for setting it:
```
jakmeier: I suggest we put it on ActionReceipt rather than Receipt. Data receipts don't really need a refund id.

I already have a first implemented for this in https://github.com/near/nearcore/pull/14080/commits/cd91f5bc795b0d444855613fcbf75a2780dca0ca

As expected, this is was sizeable change, even if it only affects action receipts.

neps/nep-0616.md
```diff
@@ -0,0 +1,546 @@
...
+All existing rules for current account model apply to deterministic
+accounts, too. In particular, only the contract itself is allowed to perform
+these actions on himself:
+
+* `CreateAccount`: for sub-accounts like `sub.0s123...`
+* `DeleteAccount`
+* `DeployGlobalContract`
+* `UseGlobalContract`
+* `DeployContract`, `AddKey`, `DeleteKey`: doesn't make sense and should be
+  discouraged by implementations, but still allowed
```
jakmeier, line 203: @mitinarseny a small nitpick: `Stake` is missing in this list. Can you add it for completeness?

neps/nep-0616.md
mitinarseny: > That's why I believe that we should use derivation scheme with a different prefix for each new StateInit version, so that verifying contract can distinguish between them rather than iterating over exponential number of combinations.

@jakmeier I thought about it once more, and it now seems to me that you were right: it doesn't make sense to use different prefixes for derivation schemes of all future `StateInit` versions. Instead, it makes sense to unify all deterministic account ids as it better aligns with deterministic account ids core idea, where account id is derived from its initialization state. So, the particular `StateInit` version that should have been used to initialize some particular contract can be thought as part of its initialization state. What do you think?

jakmeier: Makes sense to me! If a future revision is needed, an explicit version field in the `StateInit` data seems cleaner to me than an implicit version by account id prefix. That way, it's all contained in `StateInit` and therefore implied by the hash.

neps/nep-0616.md
mitinarseny: I gave it a shot and it seems that it wouldn't bring much value for verifying contracts: even if they store pre-calculated Merkle root for some subtree of `data: BTreeMap<Vec<u8>, Vec<u8>>`, then inserting a new key/value pair might require re-calculation of root hash for this subtree. So, the contract would still need to have all leaf elements at hand, which makes such optimization useless, unless newly inserted element is the new minimum.

If we take an assumption that the majority of sharded contracts would store minimal amount of data, then it should not be a problem to re-create entire `data: BTreeMap<Vec<u8>, Vec<u8>>` in runtime and it doesn't add too much complexity to derivation logic. What do you think?

jakmeier: Thanks for looking into this!

Yeah that makes sense. Let's keep it simple then.

Keeping `StateInit` small is probably best practice anyway. Optionally deploying with a `StateInit` action before a function call can become quite expensive otherwise, since all action gas costs must be burnt ahead of execution and that gas cost will depend on how much data it includes.

I think it might be better to think of other ways to reduce the StateInit size for contracts that need large data. For example, a two-stage initialization, where the second stage requires a contract-defined function call to be called before it's ready for other function calls. This call could include merkle tree logic or something similar. That way, only the contracts that really need it have this added complexity. And we don't have to worry about it on the protocol level.

jakmeier:
Thanks for the responses. It seems nobody is missing the previous proposal, so let's go ahead with the new one.

@near/wg-protocol @near/nep-moderators can we get this on the agenda for voting please? (with necessary SME review etc)

neps/nep-0616.md
jakmeier: @mitinarseny I've just finished my analysis for gas parameter values.

The TLDR is that the cost for a state initialization is: 

> That translates to 7.93 Tgas base cost + 0.2 Tgas per entry + 0.000_142 Tgas per byte.
>
> With 10 entries and 10 kB of data, that would be 11.35 Tgas.

If that seems to high or is somehow problematic, please let me know and we can discuss how we could lower those numbers.

The full report is here: https://github.com/near/nearcore/issues/14160

jakmeier: Oh and if you think these numbers are fine, they should probably be included in the NEP.

mitinarseny: > With 10 entries and 10 kB of data, that would be 11.35 Tgas.

I assume majority of sharded contracts that use deterministic AccountIds will use < 1kB of storage, at least for their initialization state. So, ~10 TGas for deploying an sFT wallet-contract seems quite reasonable!

Can you please also provide a bit more details about ZBAs in your estimations?
Based on your [comment](https://github.com/near/NEPs/pull/616/files#r2258217476), if I understand it correctly, if `StateInit` translates to <= 770 bytes of storage, then we pay for storage with gas, while if it's > 770 bytes, then we pay with NEAR, right? If so, wouldn't it be a headache for smart-contract developers to check for `StateInit` storage usage and branch their execution path every time they want to initialize another sharded contract?
```rust
let mut p = Promise::new(state_init.derive_account_id());

if state_init.storage_usage() <= 770 {
    // ZBA: storage is paid with gas
    p = p.state_init(state_init, NearToken::from_yoctonear(0));
} else {
    // storage is paid with NEAR
    p = p.state_init(state_init, state_init.storage_cost());
}
```
This gets even worse when multiple interactions between different sharded contracts are composed within a single transaction.

jakmeier: Sure, I can try to explain ZBA a bit better.

Basically, every account creation always pays extra gas ahead of time for the initial 770 bytes. There is no way to opt-out of this. `Action::CreateAccount` includes this cost. `Action::Transfer` includes this cost if it targets an implicit account. Deterministic account ids also count as implicit account for this. Therefore, `Action::DeterministicStateInit` also **always** includes the initial cost in the form of increased base gas cost.

On top of the initial cost, after applying an action receipt, the runtime checks if the used storage is above 770 bytes. If yes, then it also checks if the balance if high enough. Specifically, 0.00001NEAR per byte. So, if you hold 1 kB, the account must hold at least 0.01 NEAR. Otherwise, the action receipt fails and any changes are reverted, including the account creation itself.

So, if you initialize a deterministic account that ends up with 771 bytes and not enough NEAR is sent along to cover for 771 bytes, then initialization will fail. (Unless you have sent a Transfer ahead of time that added a balance already.)

> If so, wouldn't it be a headache for smart-contract developers to check for StateInit storage usage and branch their execution path every time they want to initialize another sharded contract?

Yes, if they want to know ahead time, they need to check. But as defined in the proposal, the caller can always attach some NEAR just to be sure. And anything that's not used will be refunded. At least that's how I understood this:

> ```
> /// * The remaining amount is transferred back to predecessor or
> ///   [refund_to](promise_set_refund_to) if set.
> ```

So I would expect smart contract devs either know they stay below 770 bytes, or they always send some NEAR along, just to be sure. But that second case seems quite unpractical anyway. Somehow, you have to ensure each contract has enough balance not only for itself, but also to fund other contracts. At that point, if I would implement it, I would always initialize an account externally and fail on calls from a deterministic account to another if the target does not exist.

> This gets even worse when multiple interactions between different sharded contracts are composed within a single transaction.

Exactly. That's also related to what I wrote in the other [comment](https://github.com/near/NEPs/pull/616/files#r2258217476).

> However, if InitState requires a NEAR balance to create the account with the deterministic id, then it can be awkward when one sFT contracts initializes another, since the balance needs to be paid by the calling contract.

Disabling ZBA would make this even worse, as we now always have this problem. ZBA is a good solution if we assume we can stay below the 770 bytes.

The only other alternative discussed so far would be to rewrite the proposal to include a solution similar to the older [non-refundable storage](https://github.com/near/NEPs/blob/master/neps/nep-0491.md) proposal (which was never implemented). But I'm not even sure if this makes things better, since a `ReserveStorage` action would also require a balance on the caller.

I've also thought about a solution to pay the `DeterministicStateInit::amount` from gas rather than from balance. It can make certain cases easier, while it will break down for a chain of calls that require a total larger than 300 Tgas. Conceptually, it also doesn't fit the current gas model at all.

For me, it all boils down to this: Is 770 bytes good enough for currently known use cases? If so, let's go with the current proposal. Otherwise, we have to go back and redesign the storage handling part. Probably go back and finally tackle the whole "Near storage is a mess" problem fundamentally. But to be honest, I'd rather avoid that for the scope of this feature.

Does that clear things up? Which approach do you want to follow @mitinarseny ?

mitinarseny: @jakmeier Thanks for explaining these in detail! I like how `Action::DeterministicStateInit` **always** includes the initial cost in the form of increased base gas cost, so there is no need to branch execution on smart-contract level.

> So I would expect smart contract devs either know they stay below 770 bytes, or they always send some NEAR along, just to be sure. But that second case seems quite unpractical anyway. Somehow, you have to ensure each contract has enough balance not only for itself, but also to fund other contracts. At that point, if I would implement it, I would always initialize an account externally and fail on calls from a deterministic account to another if the target does not exist.

I think it's necessary to keep an option to initialize other deterministic accounts within the same transaction, otherwise it worsens UX a lot. I suggest we design sharded contract standards in such a way that it will allow for passing around some sort of [`StateInitActionParams { state_init, amount }`](https://github.com/mitinarseny/near-sdk-rs/blob/327ebe55e97e0f3f168679d0ed5b733a13658581/near-contract-standards/src/sharded_fungible_token/wallet.rs#L151-L155) struct between function calls and keep them `#[payable]`, so that we put responsibility of calculating `StateInit` storage usage and checking for ZBA limits off-chain. This will make it more developer-friendly and allow for more composability between different contract standards.

> I've also thought about a solution to pay the `DeterministicStateInit::amount` from gas rather than from balance. It can make certain cases easier, while it will break down for a chain of calls that require a total larger than 300 Tgas. Conceptually, it also doesn't fit the current gas model at all.

AFAIK, there were a lot of discussions lately within the community around:
* NEAR <-> Gas conversions
* Removing 300 TGas limit
* [Moving](https://github.com/mitinarseny/NEPs/blob/deterministic-account-ids/neps/nep-0616.md#storage-rent-vs-fixed-storage-staking) towards storage **rent** model instead of one-time fixed payment

Keeping an option to pass around NEAR as attached deposit to function calls and forward it to `env::state_init(state_init, amount)` makes it future-compatible with these changes if they ever get implemented. What do you think?

jakmeier: > I think it's necessary to keep an option to initialize other deterministic accounts within the same transaction, otherwise it worsens UX a lot.

Sure, I never intended to take away the option. I just think it can be impractical in some cases.

> I suggest we design sharded contract standards in such a way that it will allow for passing around some sort of [StateInitActionParams { state_init, amount }](https://github.com/mitinarseny/near-sdk-rs/blob/327ebe55e97e0f3f168679d0ed5b733a13658581/near-contract-standards/src/sharded_fungible_token/wallet.rs#L151-L155) struct between function calls and keep them #[payable], so that we put responsibility of calculating StateInit storage usage and checking for ZBA limits off-chain. This will make it more developer-friendly and allow for more composability between different contract standards.

Great idea! I like the approach to do calculations off-chain and passing along the balance accordingly. We don't have to work out the exact contract standard at this point but it's great to have an understanding of how it can tackle the indirect-initialization problem. What you describe makes sense to me.

And yes, it is compatible with potential changes around gas and storage.

So in summary, it seems we agree to keep the proposal as it is today with regards to paying for storage / initialization. Contracts and relevant contract standards can bridge the gap where the current model is impractical.

neps/nep-0616.md
```diff
@@ -0,0 +1,563 @@
...
+* Who is allowed to perform different actions (deploy, call, delete, etc.) on
+  different sharded contract instances?
+* Who pays for storage costs of deploying new sharded contract instances?
+* How does it get refunded if the contract turned out to be already deployed?
+* How can a contract verify that the caller was executing the same code?
+
+This NEP proposes enabling sharded contracs design by introducing
+fully backwards-compatible [deterministic AccountIds](#deterministic-accountids)
+and new protocol-level primitives required to deploy such contracts, addressing
+concerns above.
```
mfornet, line 43: On my second read of the PR, I re-read summary and motivation over and over, and it was not enough to understand the high level idea of the proposal or the motivation for it. My suggestion is to add a use case that is not possible today without it, and becomes possible after this proposal.

neps/nep-0616.md
mfornet: At the point were you are interacting with a sharded contract, you are already *trusting* its source code to behave in an honest way. Every method you call can behave in a non-determinstic way, for example on first invocation it can dump all initial state and re-initialize using non-deterministic sources, so I don't see this as a compelling argument.

neps/nep-0616.md
```diff
@@ -0,0 +1,563 @@
...
+pub enum ContractCode {
+    /// Reference global contract's code by its hash
+    GlobalCodeHash(CryptoHash),
+    /// Reference global contract's code by its [`AccountId`]
+    GlobalAccountId(AccountId),
+}
+```
+
+Now, let's [derive](https://github.com/mitinarseny/near-sdk-rs/blob/50eb6e68544b75288145f7f0d07068a482b9e3fc/near-sdk/src/state_init.rs#L189-L203)
+`AccountId` deterministically as `"0s" .. hex(keccak256(state_init)[12..32])`:
```
mfornet, line 81: In a discussion with @birchmd we talked about the following proposal:

Similar to this NEP and [Solution 1 in this discussion](https://github.com/near/NEPs/issues/607).

There is a global contract deployed: `my_sharded_ft.near`, and then we can deploy `my_sharded_ft.near` at `alice.near`, `bob.near`, etc... These are each independent accounts, that uses the global contract. This can be implemented in a way that each contract knows its "namespace" (the contract where it was deployed at), and other contracts can check they are running the expected code, and runs under given account id by looking at the predecessor account id.

In practice I don't think this alternative is too different from what is being proposed here.

birchmd: The key difference between the proposals is in composition of sharded contracts. This was originally discussed [here](https://github.com/near/NEPs/issues/607#issuecomment-3057266803), but I'll summarize for ease of following the discussion.

Since deterministic account IDs are real account IDs, it is possible to derive new deterministic account IDs from other deterministic account IDs; effectively nesting the sharded functionality of one contract inside another. This kind of nesting is not possible with the "sharded contexts" solution. It can be worked around, but it either creates bottlenecks that are not sharded or increases the complexity of the two contracts that are trying to work together.

I think the reason the deterministic accounts proposal is more appealing is because of the composition it enables.

mfornet (approved):
I was discussing a bit more with @birchmd, and I see that my alternative won't work right away for [fully sharded nested sub-accounts](https://github.com/near/NEPs/issues/607#issuecomment-3057266803). The comment I linked is a great motivation for the current design, about using hashes for account ids.

In this case if we are creating the account internet.near @ alice.near it will be the hash of the contract at internet.near + some extra initial state that mentions alice.near. Further nesting will pass the new account id in the initial state, so we can have sharded.ft @ (internet.near @ alice.near).

I still lean toward using `init` method, that receives some arguments, where these arguments serve a similar purpose as the initial state, and they contribute to the hash that is used as account id. I think the `init` method allows writing cleaner code, without having to ever mess up with state keys. I don't think it makes a huge difference in the end though.

neps/nep-0616.md
mitinarseny: @mfornet Thank you for coming back to this! I get the reasoning behind determinism and [cleanness](https://github.com/near/NEPs/pull/616#pullrequestreview-3224676068) of code.

Do I understand correctly that you're suggesting replacing `data: BTreeMap<Vec<u8>, Vec<u8>>` with:
```rust
#[derive(BorshSerialize, BorshDeserialize)]
pub struct StateInitV1 {
    /// Code to deploy
    pub code: ContractCode,
    /// Optional FunctionCall to make on first initialization
    pub data: Option<FunctionCallArgs>,
}

#[derive(BorshSerialize, BorshDeserialize)]
pub struct FunctionCallArgs {
    name: String,
    args: Vec<u8>,
}
```
If so, then we also need to pass gas-related arguments to the [host function](https://github.com/mitinarseny/NEPs/blob/db95e5082e49819c3ef01014ab9f9cb2633f9a01/neps/nep-0616.md#stateinit-action), since we don't know in advance how much gas would be needed for `#[init]` function to succeed:
```rust
pub fn promise_batch_action_state_init(
    promise_index: PromiseIndex,
    state_init: &StateInit,
    amount: NearToken,
    gas: Gas,
    weight: GasWeight,
);
```

The implementation of such `#[init]` function inside a smart-contract might create promises. A common use-case for that would be to refund excessive `attached_deposit` that was attached to this only-once initialization call. The refund should go to the same account which was previously set for this receipt via [`env::set_refund_to()`](https://github.com/mitinarseny/NEPs/blob/db95e5082e49819c3ef01014ab9f9cb2633f9a01/neps/nep-0616.md#refunds) or `env::predecessor_account_id()` otherwise. So, we might need to introduce another host function `pub fn env::get_refund_to() -> AccountId;` for that.

> By the wat, this can be a good addition anyway, so that regular contracts and functions can use `env::get_refund_to()` for refunds, too.

Now, a more important question is: do we want to wait for these promises to finish before proceeding to the next action in current receipt? Of course, we can't do that because it breaks atomicity of receipt execution. But what if the contract wants to, let's say, create another account on its first initialization? And what if next regular `FunctionCall` action (the one after `StateInit` in the current receipt) is going to create another `FunctionCall` calling this freshly created account assuming it was deployed already? If so, then should we apply [`.then()`](https://docs.rs/near-sdk/latest/near_sdk/struct.Promise.html#method.then) semantics for all promises created in all actions following `StateInit` one?

This might also bring another issue: being unable to optimize validator node and [clean up the unchanged local state](https://github.com/near/NEPs/issues/607#issuecomment-3073490770) efficiently.

Having this said, I don't think the flexibility worth the complexity here: the contract would still be able to do post-initialization in subsequent FunctionCalls to it anyway, while the exact `StateInit` can be dynamically [calculated](https://github.com/mitinarseny/near-sdk-rs/blob/327ebe55e97e0f3f168679d0ed5b733a13658581/examples/sharded-fungible-token/wallet/src/lib.rs#L89) by the caller contract if needed.

@jakmeier @birchmd What do you think about this change in general? Am I missing something? Can you think of any other caveats or advantages of this approach?

birchmd: I agree with @mitinarseny that the possibility of using a function call to initialize a contract creates more complexity on the protocol level and this complexity is not needed. While it is true that the low-level interface of storage slots might be challenging for developers, they do not need to be aware of those details because the contracts SDK can simplify the interface, as @mitinarseny showed.

jakmeier: I agree with @mitinarseny.

My perspective is, if an `init()` call is what a developer wants, they don't need a protocol standard at all. A contract standard can serve the same use case. But there are important cases not covered by that.

Specifically, it becomes really hard to lazy-initialize a receiver. With a function call, the caller must estimate the gas cost and handle the case of the init function running out of gas or failing otherwise.

With the proposed `BTreeMap<Vec<u8>, Vec<u8>>` inside `StateInitAction`, the caller knows the exact gas cost ahead of time. Furthermore, the required balance for storage staking is fixed by the fields of `StateInitAction`, too. So, the protocol can fully validate `StateInitAction` including hash, gas and state costs, all on the sender side and guarantee it will always succeed on the receiver, if receipt creation doesn't fail.

Achieving the same reliability with a contract standard might be possible but it's tricky to get right with potential gas parameter changes in mind and considering the fact that the same Rust contract code compiled with different compilers will end up using different amount of gas.

Using function calls might be easier for prototyping. But for building something reliable, I believe `BTreeMap<Vec<u8>, Vec<u8>>` is easier on contract developers and for all future contract standard designs.

mitinarseny:
Hey @gagdiez, sure, I'd love to! I'm on vacation until Monday, so feel free to invite me to a call starting from Sep 22nd. Would it work for you?

bowenwang1996:
@mitinarseny thank you for this great proposal. I approve this proposal.

mitinarseny:
Hey guys, thanks for the call and all of your time and effort! ❤️
I'm super excited about so many things that can be built on top of this NEP 🚀

@jakmeier @birchmd I've added one more host function `fn get_refund_to() -> AccountId` and mentioned ZBA limits in the NEP. Can you please take a look and confirm that it's aligned with what we've discussed?

neps/nep-0616.md
```diff
@@ -313,6 +313,17 @@ sequenceDiagram
...
+account was set for the current receipt by its predecessor via `promise_set_refund_to()`,
+so that they can use it according to its business logic instead of duplicating
+it in a FunctionCall arguments. So, we might need another host function for that:
+
+```rust
+/// Returns the [`AccountId`] that was set for the current receipt by its
+/// predecessor via [`set_refund_to()`], or [`predecessor_account_id()`] otherwise.
+pub fn get_refund_to() -> AccountId;
+```
+
```
jakmeier, line 331: Makes sense to me!

Just one bikeshedding: No existing host function uses the `get_*`prefix. I think a more consistent name would be `refund_to_account_id`.

jakmeier: Oh and the host function for setting this is named `promise_set_refund_to` in my implementation. Maybe update the comment accordingly.

mitinarseny:
Just to reference it here: there is an ongoing discussion on decoupling `.state_init()` and `.top_up_for_storage()` actions: https://github.com/near/nearcore/pull/14307#discussion_r2378330478

PR #619, opened 2025-08-04
Update nep-0001.md

gagdiez:
This PR proposes updates to the current `nep-0001.md` - the NEP that defines the `NEP Purpose and Guidelines` - with the aim of simplifying it and fixing some minor issues:

#### 1. Too Verbose
The current `nep-001` is too verbose, to the point where simple things seams harder than intended (e.g. how to create a NEP)

#### 2 . Extinct Dev Gov Gigs Board
`nep-001` contains mentions to `Dev Gov Gigs` board, which no longer exists.

The good news is that actually, most of the NEPs followed the classic Github process of creating an issue, debating on it, creating a PR and eventually resolve it (with a `merge` or  a `decline`)

I have updated the `nep-0001` to better reflect this naturally happening flow, moving away from "ideating and sharing in the devgovgigs" to "create and discuss ideas on github issues"

#### 3. Removed `tool` NEP category
As far as I can tell, not a single `tools` NEP was created, and that makes sense, as tools tend to reflect opinionated takes on how to realize / tackle a specific task.

I would go as long as to argue that it is **best** for the community to not standardize tools on any way.

### DIFF
--- a/neps/nep-0001.md
+++ b/neps/nep-0001.md
@@ -1,70 +1,68 @@
 ---
 NEP: 1
 Title: NEP Purpose and Guidelines
-Authors: Bowen W. <bowen@near.org>, Austin Baggio <austin.baggio@near.org>, Ori A. <ori@near.org>, Vlad F. <frol@near.org>;
-Status: Approved
+Authors: Bowen W. <bowen@near.org>, Austin Baggio <austin.baggio@near.org>, Ori A. <ori@near.org>, Vlad F. <frol@near.org>, Guillermo G. <guillermo@near.dev>;
+Status: Draft
 DiscussionsTo: https://github.com/near/NEPs/pull/333
 Type: Developer Tools
-Version: 1.1.0
+Version: 2.0.0
 Created: 2022-03-03
-Last Updated: 2023-03-05
+Last Updated: 2025-08-04
 ---
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


frol (approved):
These are very welcome changes to the language to make NEPs friendlier to contributors instead of creating a mystery aura of complexity (there is no complexity in the process itself, in fact, it was designed to be straightforward and avoid ambiguity)

neps/nep-0001.md
```diff
@@ -1,70 +1,68 @@
...
+Version: 2.0.0
 Created: 2022-03-03
-Last Updated: 2023-03-05
+Last Updated: 2025-08-04
 ---
 
 ## Summary
 
-A NEAR Enhancement Proposal (NEP) is a design document that specifies reusable and interoperable components integrated across the NEAR ecosystem.
+NEAR Enhancement Proposals (NEPs) are design document that describe standards for the NEAR platform, including core protocol specifications, contract standards, and wallet APIs. Each NEP provides concise technical specifications and the rational behind the proposed enhancement.
```
garikbesson: NEAR Enhancement Proposals (NEPs) are design documents that describe standards for the NEAR platform, including core protocol specifications, contract standards, and wallet APIs. Each NEP provides concise technical specifications and the rationale behind the proposed enhancement.

(`design document` > `design documents`, `the rational` > `the rationale`)

garikbesson:
@ori-near @bowenwang1996 could you review the suggested changes as well since you're authors of the original document?

PR #621, opened 2025-08-13
NEP-xxx: Tokenized Vault

edwardchew97:
Reference implementations: https://github.com/Meteor-Wallet/tokenized-vault-nep-implementation

### DIFF
--- a/neps/nep-xxxx.md
+++ b/neps/nep-xxxx.md
@@ -255,16 +255,12 @@ pub struct VaultWithdraw {
 
 ## Reference Implementation
 
-We have created an example implementation. [ [Github Link](https://github.com/Meteor-Wallet/near-erc4626-vault) ]
-
-Besides that, we have also written a Rust Trait and Events that we wish to be merge with the near-contract-standards repo. [ [Contract Standards](https://github.com/Meteor-Wallet/near-erc4626-vault/tree/main/src/contract_standards) ]
-
-We are still working hard on writting test case to test our example implementation
+Example implementation can be found [here](https://github.com/Meteor-Wallet/tokenized-vault-nep-implementation). 
 
 ## Security Implications
 
 ### Exchange Rate Manipulation
-Vaults allow dynamic exchange rates between shares and assets, calculated by dividing total vault assets by total issued shares. If the vault has a permissionless donation mechanism, it creates vulnerability to inflation attacks where attackers manipulate the rate by donating assets to inflate share values, potentially stealing funds from subsequent depositors. Vault deployers can protect against this attack by making an initial deposit of a non-trivial amount of the asset, such that price manipulation becomes infeasible.
+Vaults allow dynamic exchange rates between shares and assets, calculated by dividing total vault assets by total issued shares. If the vault has a permissionless donation mechanism, it creates vulnerability to inflation attacks where attackers manipulate the rate by donating assets to inflate share values, potentially stealing funds from subsequent depositors. Vault deployers can protect against this attack by making an initial deposit of a non-trivial amount of the asset, such that price manipulation becomes infeasible. The sample implementation doesn't allow donations to the vault, but if future implementations do, they should implement a mechanism to prevent inflation attacks.
 
 ### Cross-contract Calls
 Redeem and withdraw functions perform cross-contract calls to transfer fungible tokens, creating opportunities for reentrancy attacks and state manipulation during asynchronous execution. Vaults should implement reentrancy protection through proper state management, proper callback security, and rollback mechanisms for failed operations.
@@ -277,12 +273,16 @@ Vaults that rely on external price oracles or cross-contract calls for exchange 
 
 ## Alternatives
 
-[Explain any alternative designs that were considered and the rationale for not choosing them. Why your design is superior?]
+1. **Custom Per-Protocol Vault Implementations**  
+   - While possible, this leads to fragmentation, increases integration costs, and reduces composability between protocols.
+2. **Direct ERC-4626 clone Without NEAR Adjustments**  
+   - Rejected because NEAR’s asynchronous execution model makes a one-to-one ERC-4626 clone inefficient. This proposal integrates share and vault logic in a single contract to avoid unnecessary cross-contract calls.
+
 
 ## Future possibilities
 
 ### NEP-245 Multi Token Support
-Future vault implementations could extend this standard to support NEP-245 Multi Token contracts as underlying asset.
+Future vault implementations could extend this standard to support NEP-245 Multi Token contracts as underlying asset. We have also created a minimal implementation of this [here](https://github.com/Meteor-Wallet/tokenized-vault-mt-nep-implementation).
 
 ### Multi-Asset Vault Extensions
 Future extensions could allow vaults to accept multiple assets for deposit and withdrawal. This would enable the standardization of LP vaults.
@@ -292,23 +292,24 @@ Future vault standards could introduce asynchronous deposit and withdrawal patte
 
 ## Consequences
 
-[This section describes the consequences, after applying the decision. All consequences should be summarized here, not just the "positive" ones. Record any concerns raised throughout the NEP discussion.]
-
 ### Positive
-
--   p1
+- Enables a unified, predictable vault interface.
+- Simplifies integration for wallets, DEXs, and aggregators.
+- Improves security through consistent design and accounting.
+- Encourages reuse of tooling, libraries, and audits.
 
 ### Neutral
-
--   n1
+- Standard defines interface, not yield strategy — implementation remains flexible.
+- Protocols may implement only relevant parts of the interface initially.
 
 ### Negative
+- Migration overhead for existing vault implementations to become compliant.
 
--   n1
 
 ### Backwards Compatibility
+- No breaking changes to NEP-141 itself, but existing vault-like contracts that don’t conform will need to add or rename methods to comply.
+- Share tokens must be NEP-141 compliant, meaning non-NEP-141 share implementations require migration.
 
-[All NEPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. Author must explain a proposes to deal with these incompatibilities. Submissions without a sufficient backwards compatibility treatise may be rejected outright.]
 
 ## Unresolved Issues (Optional)
 
@@ -319,8 +320,6 @@ Future vault standards could introduce asynchronous deposit and withdrawal patte
 -   What related issues do you consider out of scope for this NEP that could be addressed in the future independently of the solution that comes out of this NEP?]
 
 ## Changelog
-
-[The changelog section provides historical context for how the NEP developed over time. Initial NEP submission should start with version 1.0.0, and all subsequent NEP extensions must follow [Semantic Versioning](https://semver.org/). Every version should have the benefits and concerns raised during the review. The author does not need to fill out this section for the initial draft. Instead, the assigned reviewers (Subject Matter Experts) should create the first version during the first technical review. After the final public call, the author should then finalize the last version of the decision context.]
 
 ### 1.0.0 - Initial Version
 
--- a/neps/nep-xxxx.md
+++ b/neps/nep-xxxx.md
@@ -1,0 +1,331 @@
+---
+NEP: 0
+Title: Vault NEP
+Authors: JY Chew <edwardchew97@gmail.com>, Lee Hoe Mun <leehoemun@gmail.com>, Wade <wz.lim.00@gmail.com>, Steve Kok <kokchoquan@gmail.com>
+Status: New
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/0000
+Type: Contract Standard
+Requires: 141
+Version: 1.0.0
+Created: 2025-08-08
+LastUpdated: 2025-08-08
+---
+
+## Summary
+
+This NEP proposes a standardized interface for implementing vault contracts on the NEAR Protocol, drawing inspiration from the ERC-4626 standard widely used on Ethereum. A vault contract allows users to deposit an underlying fungible token (FT) into the vault, in exchange for which the vault issues shares that represent proportional ownership of the vault’s assets.
+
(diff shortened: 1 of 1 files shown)


garikbesson:
Hi @edwardchew97, thank you for bringing new ideas and helping to push the entire ecosystem forward!

As a NEP moderator, I reviewed your proposal, and there is only one thing I would like to ask you - please, remove the `Unresolved Issues (Optional)` section from the proposal, since it's optional and hasn't been filled in.

Other than that, everything looks fine, and as soon as you push the requested changes, I will ask WG to assign SMEs to review the proposal.

garikbesson:
As a moderator, I'm asking @near/wg-contract-standards working group members to assign 2 Technical Reviewers to complete a technical review of the proposal.

IkerAlus (commented):
Hi @edwardchew97 thanks for the good work. I have done a general review of the proposal (not the implementation) and compared it with original [ERC4626 specification](https://eips.ethereum.org/EIPS/eip-4626), and I have a couple of comments:

- The proposal is missing `deposit` and `mint` methods. Is there any reason for this? AFAIK, these two methods are main state-changing flows in ERC4626.
- In ERC4626, `totalAssets` view method states that it MUST be inclusive of any fees that are charged against assets in the Vault. Can you add this statement too?
- Can you add the event emitted to the corresponding method description? For example, `withdraw` method should emit `withdraw` event.
- Same you did for `max_redeem`, with "Implementations should return `0` if...", can you add a similar statement for `maxWithdraw` method as done in ERC4626?
- Can you add when a method must or not revert (like it is done in the ERC4626)?
- Finally, in "Alternatives" it is mentioned this spec is not a "direct clone ERC-4626 clone without NEAR Adjustments".  Are this adjustments purely in the implementation? If they affect the spec, can you add a section stating which methods differ from  ERC-4626 spec and how?

edwardchew97:
@IkerAlus Thanks a lot for the thorough review! Addressed all your points below:

> * The proposal is missing `deposit` and `mint` methods. Is there any reason for this? AFAIK, these two methods are main state-changing flows in ERC4626.
- Good catch — deposits/mints happen through ft_on_transfer. That was already in the implementation but not explicitly mentioned in the NEP. We’ve now added it to the spec for clarity.

> * In ERC4626, `totalAssets` view method states that it MUST be inclusive of any fees that are charged against assets in the Vault. Can you add this statement too?
- Added this to the spec, thanks!

> * Can you add the event emitted to the corresponding method description? For example, `withdraw` method should emit `withdraw` event.
- Great suggestion — events are now included in the method descriptions.

> * Same you did for `max_redeem`, with "Implementations should return `0` if...", can you add a similar statement for `maxWithdraw` method as done in ERC4626?
- Added that as well.

> * Can you add when a method must or not revert (like it is done in the ERC4626)?
- Updated the spec accordingly (no changes in the reference implementation).

> * Finally, in "Alternatives" it is mentioned this spec is not a "direct clone ERC-4626 clone without NEAR Adjustments".  Are this adjustments purely in the implementation? If they affect the spec, can you add a section stating which methods differ from ERC-4626 spec and how?
- Yes, most of the adjustments are only in the implementation and not shown on the interfaces. The differences are primarily due to the asynchronous nature of NEAR Protocol. This doesn’t necessarily prevent ERC-4626 from working but introduces some implementation-level nuances (e.g. `ft_on_transfer`). Beyond that, most choices are left to the developer, so we’ve opted to keep these adjustments in the implementation rather than the core spec. That said, we’ve added further clarification under the section. Appreciate you pointing this out too.

neps/nep-0621.md
```diff
@@ -0,0 +1,491 @@
...
+    /// Implementers are free to define their own schema for `msg`, but the vault must
+    /// correctly handle `ft_on_transfer` according to NEP-141:
+    /// - Return `PromiseOrValue::Value(0)` if all tokens were accepted.
+    /// - Return a non-zero `U128` to indicate the number of tokens to refund.
+    fn ft_on_transfer(&mut self, sender_id: AccountId, amount: U128, msg: String) -> PromiseOrValue<U128>
+}
+````
+
+```rust
+pub trait VaultOptionalSlippageWithdraw {
```
IkerAlus: are these two new methods mandatory in the NEP or just recommended? 

I see they add a very valid use case but they are not part of ERC4626

neps/nep-0621.md
```diff
@@ -0,0 +1,491 @@
...
+    /// Optional memo provided by the redeemer for off-chain use.
+    pub memo: Option<String>,
+}
+```
+
+Deposit and Mint
+
+````rust
+pub trait FungibleTokenReceiver {
+    /// According to the NEP-141 standard, all fungible token transfers to a contract
```
IkerAlus: can we take out this explanation (from line 324 to 349) and place it just below "Deposit and Mint" subsection?

neps/nep-0621.md
```diff
@@ -0,0 +1,491 @@
...
+
+    /// Amount of underlying assets withdrawn from the vault.
+    pub assets: U128,
+
+    /// Optional memo provided by the redeemer for off-chain use.
+    pub memo: Option<String>,
+}
+```
+
+Deposit and Mint
```
IkerAlus: shouldn't this section be placed inside the "Contract Interface" section and not inside "Events" section?

neps/nep-0621.md
```diff
@@ -0,0 +1,491 @@
...
+    ///     /// If `true`, the transfer is treated as a donation and no shares are minted.
+    ///     donate: Option<bool>,
+    /// }
+    /// ```
+    ///
+    /// Implementers are free to define their own schema for `msg`, but the vault must
+    /// correctly handle `ft_on_transfer` according to NEP-141:
+    /// - Return `PromiseOrValue::Value(0)` if all tokens were accepted.
+    /// - Return a non-zero `U128` to indicate the number of tokens to refund.
+    fn ft_on_transfer(&mut self, sender_id: AccountId, amount: U128, msg: String) -> PromiseOrValue<U128>
```
IkerAlus: Both `deposit` and `mint` should emit a `VaultDeposit` event.  Something you indeed do in the ref impl in https://github.com/Meteor-Wallet/tokenized-vault-nep-implementation/blob/5952f914a4f1601aa73ac142d77af45685309933/src/lib.rs#L276 

This should be specified in the spec doc too for completeness.

IkerAlus (changes requested):
hey @edwardchew97 thanks for the thorough and detailed answers. I think the spec looks almost ready to go. I only have one main comment, the rest are format changes and clarifications.

neps/nep-0621.md
```diff
@@ -0,0 +1,490 @@
...
+    /// Returns the [`AccountId`] of the underlying asset token contract.
+    ///
+    /// The asset **must** be NEP-141 compliant.
+    /// Implementations should store this as an immutable configuration value.
+    fn asset(&self) -> AccountId;
+
+    /// Returns the total amount of underlying assets represented by all shares in existence.
+    ///
+    /// **Important:**
+    /// - Represents the vault's *total managed value*, not just assets held in the contract.
```
fadeevab, line 74: I understand where it comes from, but two "asset"/"assets" terms create a wild confusion. "Asset" means FT address, and "assets" means "amount of that one underlying asset (FT token)". Perhaps, it could be better addressed, e.g. "assets" could be called as "underlying asset's amount" in the docs comments.

neps/nep-0621.md
```diff
@@ -0,0 +1,490 @@
...
+### Optional Slippage Control:
+
+-   To mitigate the risk of unfavorable execution due to block delays, vault contracts may optionally implement the following methods, which act as "slippage-protected" counterparts to `redeem` and `withdraw`.
+
+```rust
+pub trait VaultOptionalSlippageWithdraw {
+    /// Redeem a specific number of shares, **only if** they are worth
+    /// at least `min_amount` of assets.
+    /// This acts as a slippage-protected version of `redeem`.
+    fn redeem_with_slippage(
```
fadeevab: It seems both `redeem` and `withdraw` must have slippage control via `min_amount` parameter by default on NEAR, without "optional for security" `redeem/withdraw_with_slippage` counterparts. Because NEAR is fundamentally asynchronous in execution, and because this NEP is adaptation to NEAR's reality, it's worth to introduce secure by default methods, not vice versa.

fadeevab (commented):
Hi! I left two suggestions, check it out.

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
...
+    /// ERC4626 original function name: previewMint(uint256 shares)
+    ///
+    /// Differs from [`Self::convert_to_assets`] by accounting for:
+    /// - Per-user minting limits.
+    /// - Protocol-specific minting fees.
+    ///
+    /// Useful for frontends to estimate the cost of minting shares.
+    ///
+    /// # Panics / Fails
+    /// - The `preview_mint` method must never panic.
```
fadeevab: Maybe `preview_mint` was not so bad (not so long) :) Otherwise, change the name here as well

```suggestion
    /// - The `preview_asset_amount_required_to_mint_shares` method must never panic.
```

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
...
+    ///
+    /// ERC4626 original function name: previewRedeem(uint256 shares)
+    ///
+    /// Differs from [`Self::convert_to_assets`] by factoring in:
+    /// - The caller's current share balance.
+    /// - Vault withdrawal limits.
+    /// - Applicable fees or penalties.
+    ///
+    /// # Panics / Fails
+    /// - The `preview_redeem` method must never panic.
```
fadeevab: Same here, you got inconsistency once you renamed functions here and there
```suggestion
    /// - The `preview_redeem_amount` method must never panic.
```

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
...
+    ///
+    /// ERC4626 original function name: previewWithdraw(uint256 assets)
+    ///
+    /// Differs from [`Self::convert_to_shares`] by factoring in:
+    /// - The caller's current share balance.
+    /// - Vault withdrawal limits.
+    /// - Applicable fees or penalties.
+    ///
+    /// # Panics / Fails
+    /// - The `preview_withdraw` method must never panic.
```
fadeevab:
```suggestion
    /// - The `preview_shares_deducted_for_withdraw` method must never panic.
```

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
...
+    ///   It should return the number of shares required to withdraw the given amount of assets,
+    ///   or `0` if withdrawals are not currently allowed.
+    ///
+    /// Useful for frontends to preview required shares for a given withdrawal.
+    fn preview_shares_deducted_for_withdraw(&self, assets: U128) -> U128;
```
fadeevab:
```suggestion
    fn preview_shares_deducted_for_withdraw(&self, asset_amount: U128) -> U128;
```

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
...
+```rust
+/// Specification for a fungible token vault that issues NEP-141 compliant shares.
+///
+/// A FungibleTokenVault accepts deposits of an underlying NEP-141 compliant asset
+/// and issues NEP-141 compliant "shares" in return. These shares can be transferred
+/// and traded like any other NEP-141 token.
+///
+/// This trait extends:
+/// - [`FungibleTokenCore`] to provide NEP-141 functionality for shares.
+/// - [`FungibleTokenReceiver`] to receiver the underlying NEP-141 assets
```
fadeevab:
```suggestion
/// - [`FungibleTokenReceiver`] to receive the underlying NEP-141 assets
```

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
...
+```
+
+### Deposit and Mint
+
+```rust
+pub trait FungibleTokenReceiver {
+    /// # Events
+    /// - The vault **must** emit a `VaultDeposit` event upon a successful deposit.
+    /// - The vault should also emit a `FtMint` event for the share they issued.
+    fn ft_on_transfer(&mut self, sender_id: AccountId, amount: U128, msg: String) -> PromiseOrValue<U128>
```
fadeevab:
```suggestion
    fn ft_on_transfer(&mut self, sender_id: AccountId, amount: U128, msg: String) -> PromiseOrValue<U128>;
```

fadeevab (approved):
@gagdiez @edwardchew97 Sorry for delay, I missed the message. Besides of these 2 typos, it seems good to me, everything is fixed, and the long method names are very clear in meaning despite deviation from the original ERC :)

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
+---
+NEP: 621
+Title: Vault NEP
+Authors: JY Chew <edwardchew97@gmail.com>, Lee Hoe Mun <leehoemun@gmail.com>, Wade <wz.lim.00@gmail.com>, Steve Kok <kokchoquan@gmail.com>
+Status: New
```
gagdiez:
```suggestion
Status: Approved
```

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
...
+
+### 1.0.0 - Initial Version
+
+> Placeholder for the context about when and who approved this NEP version.
+
+#### Benefits
+
+> List of benefits filled by the Subject Matter Experts while reviewing this version:
+
+-   Benefit 1
```
gagdiez:
```suggestion
-   Would benefit the ecosystem by introducing a standardized way to create token vaults. Indeed, multiple projects currently exists (e.g. metapool, more markets, etc) but none of them follow a single standard making it hard to interconnect with them 
```

neps/nep-0621.md
```diff
@@ -0,0 +1,497 @@
...
+-   Benefit 2
+
+#### Concerns
+
+> Template for Subject Matter Experts review for this version:
```
gagdiez: We can just note here that no major concerns were raised, besides the need to need to be careful with the `#security implications` sections described above

PR #628, opened 2025-09-24
Add missing NEPs

bucanero:
Some standards are only available under the `Standards` folder, but they're not properly listed as NEPs and are not shown on the main table.


gagdiez:
Amazing, thanks! I think for NEP245 we better create a folder in the `neps` (i.e. `neps/nep-0245/`), and we need to make sure that all the files are reachable (I think currently there is no way to get to `Multi Token Enumeration`)

gagdiez (approved):

PR #635, opened 2026-01-25
NEP-635: P256 host function

bowenwang1996:
Proposal to add ECDSA P256 signature verification as a host function in the runtime. Implementation in https://github.com/near/nearcore/pull/14927

### DIFF
--- a/neps/nep-0635.md
+++ b/neps/nep-0635.md
@@ -3,7 +3,7 @@ NEP: 622
 Title: P-256 ECDSA Signature Verification Host Function
 Authors: Bowen Wang <bowen@nearone.org>
 Status: Draft
-DiscussionsTo: https://github.com/near/NEPs/pull/0000
+DiscussionsTo: https://github.com/near/NEPs/pull/0635
 Type: Runtime Spec
 Category: Protocol
 Version: 1.0.0
--- a/neps/nep-0635.md
+++ b/neps/nep-0635.md
@@ -94,7 +94,7 @@ pub fn p256_verify(signature: [u8; 64], message: [u8; 32], public_key: [u8; 33])
 
 The reference implementation in `nearcore`:
 
-- Uses the RustCrypto `p256` crate to parse `Signature::from_slice` and `VerifyingKey::from_sec1_bytes`.
+- Uses the RustCrypto `p256` crate - [audited here](https://reports.zksecurity.xyz/reports/near-p256/) - to parse `Signature::from_slice` and `VerifyingKey::from_sec1_bytes`.
 - Returns `0` for parsing failures or verification failures, and raises `P256VerifyInvalidInput` for invalid lengths.
 - Charges `p256_verify_base` once and `p256_verify_byte` per message byte, in addition to the standard input costs.
 
--- a/neps/nep-0635.md
+++ b/neps/nep-0635.md
@@ -49,7 +49,10 @@ extern "C" {
   ///
   /// - `signature` MUST be 64 bytes encoded as `r || s` (32 bytes each, big-endian).
   /// - `public_key` MUST be a 33-byte compressed SEC1 encoding.
-  /// - `message` is the exact byte sequence to verify (no hashing is performed).
+  /// - `message` is the pre-hashed digest to verify (no hashing is performed by the host function).
+  ///   Callers MUST hash the signed data with an appropriate cryptographic hash function before
+  ///   calling `p256_verify` (e.g., WebAuthn uses SHA-256). If `message` is not exactly 32 bytes,
+  ///   it will be truncated or zero-padded to 32 bytes to match the P-256 field size.
   ///
   /// Returns:
   /// - 1 if the signature verifies
--- a/neps/nep-0635.md
+++ b/neps/nep-0635.md
@@ -84,7 +84,7 @@ The gas cost MUST be computed as:
 Once enabled, the runtime SHOULD expose a higher-level helper in `near-sdk` as:
 
 ```rust
-pub fn p256_verify(signature: &[u8], message: &[u8], public_key: &[u8]) -> bool;
+pub fn p256_verify(signature: [u8; 64], message: [u8; 32], public_key: [u8; 33]) -> bool;
 ```
 
 ## Reference-level specification
--- a/neps/nep-0622.md
+++ b/neps/nep-0622.md
@@ -1,0 +1,114 @@
+---
+NEP: 622
+Title: P-256 ECDSA Signature Verification Host Function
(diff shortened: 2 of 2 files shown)


neps/nep-0635.md
```diff
@@ -0,0 +1,114 @@
...
+The gas cost MUST be computed as:
+
+`input_cost(num_bytes_signature) + input_cost(num_bytes_message) + input_cost(num_bytes_public_key) + p256_verify_base + p256_verify_byte * num_bytes_message`
+
+### SDK exposure
+
+Once enabled, the runtime SHOULD expose a higher-level helper in `near-sdk` as:
+
+```rust
+pub fn p256_verify(signature: &[u8], message: &[u8], public_key: &[u8]) -> bool;
```
mitinarseny: To avoid unnecessary conversions on the SDK side, I suggest accepting byte arrays of exact length:
```suggestion
pub fn p256_verify(signature: [u8; 64], message: [u8; 32], public_key: [u8; 33]) -> bool;
```

neps/nep-0635.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Host function
+
+```rust
+extern "C" {
+  /// Verify a P-256 ECDSA signature for a message and public key.
+  ///
+  /// - `signature` MUST be 64 bytes encoded as `r || s` (32 bytes each, big-endian).
+  /// - `public_key` MUST be a 33-byte compressed SEC1 encoding.
+  /// - `message` is the exact byte sequence to verify (no hashing is performed).
```
mitinarseny: AFAIK, all ECDCA signatures MUST be calculated over a digest of cryptographic hash  function, while it's up to signing standard which hash function to use: e.g. webauthn [uses](https://w3c.github.io/webauthn/#sctn-verifying-assertion) SHA-256. Do you think it's worth to mention it here?

mitinarseny (commented):
Hey @bowenwang1996, thank you for this proposal!

We (Near Intents team) are super-excited to have it on mainnet soon, as it would drastically reduce gas costs of verifying p256 signatures in WASM, while leaving more gas for other operations.

I left a couple of comments, would appreciate your feedback.

neps/nep-0635.md
```diff
@@ -0,0 +1,114 @@
...
+
+```rust
+pub fn p256_verify(signature: &[u8], message: &[u8], public_key: &[u8]) -> bool;
+```
+
+## Reference-level specification
+
+The reference implementation in `nearcore`:
+
+- Uses the RustCrypto `p256` crate to parse `Signature::from_slice` and `VerifyingKey::from_sec1_bytes`.
```
IkerAlus: For the reviewer of this NEP, I think it would be nice to mention that this crate was audited recently: https://reports.zksecurity.xyz/reports/near-p256/

Particularly since the maintainers of this crate still claim this crate has never been independently audited: https://github.com/RustCrypto/elliptic-curves/tree/master/p256#%EF%B8%8F-security-warning 

even though they have added the suggested changes from the audit: https://github.com/RustCrypto/elliptic-curves/pull/1155

neps/nep-0635.md
mitinarseny: As far as I checked: the value will be [truncated or padded with zeros](https://github.com/RustCrypto/signatures/blob/3db77e32fa27b82ab71313458405e6a69740ded6/ecdsa/src/hazmat.rs#L47-L77) inside of [`<VerifyingKey<C> as PrehashVerifier<Signature<C>>>::verify_prehash`](https://github.com/RustCrypto/signatures/blob/89232d6a962a199fd8211a117db74408353e4383/ecdsa/src/verifying.rs#L163-L166) if the length is not equal to `<C as Curve>::FieldBytesSize`, which [is](https://github.com/RustCrypto/elliptic-curves/blob/6e6886563bb066cbaea4f59f89459341af6879af/p256/src/lib.rs#L101-L102) 32 bytes for NistP256

birchmd (commented):
@gagdiez we had a discussion about this NEP at our monthly Protocol Working Group review call. We decided that it is ok to skip the separate SME review for this proposal and move straight to voting. This because it is using a commonly used and well-specified signature algorithm while also making use of a library that has been audited. Essentially, we fell the SME review has implicitly already taken place in the form of the popularity of this algorithm (e.g. in passkeys) and in the audit of the crate.

To that end, as my role as a WG member, I vote to approve this proposal. My only minor comment is that we should accept @mitinarseny 's comments about making clear the `msg` passed to the verify function should come from some cryptographically secure hash function (the exact one chosen is dependent on the signature scheme), and  enforcing input lengths at the SDK level.

neps/nep-0635.md
r-near:
```suggestion
  /// - `message` is the pre-hashed digest to verify (no hashing is performed by the host function).
  ///   Callers MUST hash the signed data with an appropriate cryptographic hash function before
  ///   calling `p256_verify` (e.g., WebAuthn uses SHA-256). If `message` is not exactly 32 bytes,
  ///   it will be truncated or zero-padded to 32 bytes to match the P-256 field size.
```

graphite (approved):

neps/nep-0635.md
```diff
@@ -0,0 +1,117 @@
...
+
+```rust
+pub fn p256_verify(signature: [u8; 64], message: [u8; 32], public_key: [u8; 33]) -> bool;
+```
+
+## Reference-level specification
+
+The reference implementation in `nearcore`:
+
+- Uses the RustCrypto `p256` crate to parse `Signature::from_slice` and `VerifyingKey::from_sec1_bytes`.
```
gagdiez:
```suggestion
- Uses the RustCrypto `p256` crate - [audited here](https://reports.zksecurity.xyz/reports/near-p256/) - to parse `Signature::from_slice` and `VerifyingKey::from_sec1_bytes`.
```

neps/nep-0635.md
gagdiez: I feel this would also be relevant for future readers, I will go ahead and add a small comment on the NEP itself

gagdiez:
Both @birchmd [1] and @graphite [2] who are members of the @near/wg-protocol gave their approval, which means we have reached enough quorum for it to be accepted as a Protocol NEP

For anyone reading this in the future, we did skip the protocol meeting to vote, as this NEP proposes to add a well-known cryptographical operation using audited `crates`, which means that it has been implicitly reviewed by multiple subject matter experts

- [1] https://github.com/near/NEPs/pull/635#pullrequestreview-3930588115
- [2] https://github.com/near/NEPs/pull/635#issuecomment-4179142966

gagdiez (approved):

PR #639, opened 2026-04-09
NEP-639: Dynamic Resharding

Wiezzel:
This proposal introduces dynamic resharding: a protocol-level mechanism that enables automatic splitting of shards based on their state size, without requiring a protocol upgrade. It replaces the static approach used by Resharding V2 (`NEP-508`) and Resharding V3 (`NEP-568`), where shard layout changes were hardcoded into specific protocol versions and required coordinated binary releases.

### DIFF
--- a/neps/nep-0639.md
+++ b/neps/nep-0639.md
@@ -24,7 +24,7 @@ With dynamic resharding, each shard proposes a split during chunk application wh
 As the NEAR ecosystem grows, the network needs the ability to scale out automatically. Dynamic resharding addresses this by:
 
 - **Eliminating manual coordination**: Shard splits are proposed and executed by the protocol itself, based on observed state size.
-- **Reacting to demand**: Overloaded shards are detected and split within two epochs (~24 hours on mainnet), rather than waiting weeks for a protocol upgrade cycle.
+- **Reacting to demand**: Overloaded shards are detected and split within two epochs (~14.5 hours on mainnet), rather than waiting weeks for a protocol upgrade cycle.
 - **Enabling continuous scaling**: The network can grow from its current shard count toward a configurable maximum without any human intervention.
 
 ## Specification
@@ -331,10 +331,10 @@ A non-deterministic split algorithm would cause validators to disagree on the sh
 
 ### Validation of Split Proposals
 
-Chunk producers could attempt to forge `proposed_split` values to manipulate which shard gets split or the boundary account. This is prevented by:
+Both chunk producers and block producers could attempt to forge their respective resharding fields to manipulate which shard gets split or the boundary account:
 
-1. Chunk validators recompute `proposed_split` from the state witness and reject mismatches (`InvalidChunkHeaderShardSplit`).
-2. Block validators recompute `shard_split` from the chunk headers in the block and reject mismatches (`InvalidBlockHeaderShardSplit`).
+1. **Chunk header forgery**: A chunk producer could embed a false `proposed_split` in the chunk header (e.g., proposing a split for a shard that doesn't meet the threshold, or proposing a wrong boundary account). This is prevented by chunk validators, who recompute `proposed_split` from the state witness and reject mismatches (`InvalidChunkHeaderShardSplit`).
+2. **Block header forgery**: A block producer could embed a false `shard_split` in the block header (e.g., selecting a different shard than the one with the highest memory usage, or ignoring a valid proposal). This is prevented by block validators, who recompute `shard_split` from the chunk headers in the block and reject mismatches (`InvalidBlockHeaderShardSplit`).
 
 ### DoS via Threshold Manipulation
 
@@ -384,7 +384,7 @@ Allowing more than one split per epoch was considered but rejected to limit the 
 
 - The network can scale automatically in response to demand without protocol upgrades.
 - Eliminates the operational burden of manual shard layout analysis and deployment.
-- Reduces reaction time from weeks (manual protocol upgrade) to ~24 hours (two epochs on mainnet).
+- Reduces reaction time from weeks (manual protocol upgrade) to ~14.5 hours (two epochs on mainnet).
 - Fully compatible with stateless validation -- split proposals are verifiable through state witnesses.
 - Builds directly on the proven Resharding V3 execution machinery.
 
@@ -392,6 +392,7 @@ Allowing more than one split per epoch was considered but rejected to limit the 
 
 - The split algorithm performs a greedy descent that is equivalent to binary search on a monotonic cumulative distribution, so it finds the optimal boundary. However, a perfectly balanced split is not always achievable due to inherent constraints (indivisible accounts, minimum boundary length).
 - Thread pool sizes are based on `max_number_of_shards` rather than the current shard count, which is slightly wasteful but safe.
+- Mainnet and testnet shard layouts will diverge, since splits are driven by actual state size which differs between the two networks. This is expected and by design, but means tooling and tests cannot assume a fixed shard layout for either network.
 
 ### Negative
 
@@ -409,9 +410,6 @@ External tooling that queries shard layouts via the `EXPERIMENTAL_protocol_confi
 
 ## Unresolved Issues
 
-- **Shard assignment stability across resharding**: Validator-to-shard assignment currently uses `ShardIndex`, so a resharding event that changes shard indices may cause excessive validator reassignment. The assignment algorithm should be made resharding-aware to minimize unnecessary churn.
-- **Shadow validation across resharding boundaries**: Shadow validation currently breaks at resharding boundaries.
-- **Fork network tooling**: The fork-network tool does not yet fully support dynamic resharding.
 - **SPICE integration**: Multiple integration points between the SPICE feature and resharding transitions remain incomplete (chunk executor, data distributor, chunk validation, chunk application).
 
 ## Changelog
--- a/neps/nep-0639.md
+++ b/neps/nep-0639.md
@@ -77,7 +77,9 @@ pub struct DynamicReshardingConfig {
     pub max_number_of_shards: NumShards,
 
     /// Minimum epochs between consecutive resharding events.
-    /// A value of 0 allows resharding every epoch.
+    /// A value of 0 allows resharding every epoch (gap of 1 epoch between
+    /// effective layouts); a value of N enforces a gap of at least N+1 epochs
+    /// between effective layouts.
     pub min_epochs_between_resharding: EpochHeight,
 
     /// Shards to force-split regardless of memory thresholds.
@@ -101,7 +103,19 @@ pub enum ShardLayoutConfig {
 
 When `ShardLayoutConfig::Dynamic` is active, the shard layout is stored in `EpochInfo` rather than derived from the protocol version. This allows different epochs at the same protocol version to have different shard layouts.
 
+#### Source of Configuration Values
+
+`DynamicReshardingConfig` values (thresholds, limits, force/block lists) are part of `EpochConfig` and are loaded from per-protocol-version JSON files baked into the `neard` binary at build time, located under `core/primitives/res/epoch_configs/{mainnet,testnet}/<protocol_version>.json`. They are not exposed as node-operator-tunable settings and cannot be overridden at runtime.
+
+Updating any value therefore requires a protocol upgrade: a new protocol version is added with a new epoch config JSON file, and the network adopts the change once enough nodes vote to upgrade. This is intentional -- thresholds and limits affect consensus (every validator must agree on whether a split should be proposed), so they MUST be uniform across the network. Initial mainnet values will be set conservatively (see [Limiting the Blast Radius](#limiting-the-blast-radius)) and tuned through subsequent protocol upgrades as operational experience accumulates.
+
 With dynamic resharding, per-shard default cache sizes are no longer derived from the protocol version, since the shard layout is not known at compile time. Nodes use uniform default cache sizes across all shards. Performance testing on mainnet state has shown no significant impact from this change. Node operators can still override cache sizes via manual configuration if needed.
+
+### Limiting the Blast Radius
+
+When dynamic resharding is first released on mainnet, `max_number_of_shards` will be set to one more than the current shard count. This caps the damage from any single resharding event: at most one new shard can be created before further splits are blocked, regardless of memory thresholds. Successive protocol upgrades will raise the limit gradually as confidence in the feature grows. With this conservative rollout, in practice we expect that disabling dynamic resharding will never be necessary -- the shard limit itself acts as the safety mechanism.
+
+In the unlikely event that the shard limit proves insufficient (for instance, a previously-allowed split turns out to be problematic and we want to prevent any further splits while we investigate), an emergency mitigation is to ship a protocol upgrade that sets `block_split_shards` to include every existing shard ID. This blocks all further splits while leaving the rest of the protocol unchanged. We note, however, that if a protocol upgrade is needed anyway, it is generally preferable to use that upgrade to fix the underlying issue rather than to disable resharding.
 
 ### Finding the Optimal Split Point
 
@@ -164,6 +178,17 @@ During chunk application, the runtime checks whether a shard should be split. Th
    - If `total_mem_usage >= memory_usage_threshold`: compute the split via `find_trie_split`. If both children would have at least `min_child_memory_usage`, **propose the split**.
 
 The result (`Option<TrieSplit>`) is stored in `ChunkExtra` and embedded in the chunk header via the new `ShardChunkHeaderInnerV5.proposed_split` field.
+
+#### Behavior with Missing Chunks
+
+When a chunk is missing in a block, the previous chunk header for that shard is reused (standard NEAR behavior). The reused header carries whatever `proposed_split` was set on the last produced chunk. This has the following implications for dynamic resharding:
+
+- If the shard had at least one chunk produced within the "possibly-last-block" window, its `proposed_split` is valid (the shard's state cannot change while its chunks are missing) and is correctly carried forward through the missing-chunk headers. The shard is considered normally for selection.
+- If the shard's last produced chunk predates that window, its `proposed_split` is `None` (the proposal is only computed near the epoch boundary). The shard is silently skipped this epoch and will be reconsidered in the next.
+
+A subtle consequence of the second case: if multiple shards are above the split threshold and the **largest** one is the one whose chunks went missing during the window, its `None` excludes it from the picker and a smaller-but-eligible shard gets picked instead. The split of the larger shard is then delayed by at least one epoch (and longer if `min_epochs_between_resharding > 0`, since the cooldown will apply to the smaller shard's split that just occurred). The split chosen is still legitimate (the smaller shard is above threshold by construction), so this is a degradation in priority, not in correctness.
+
+In all cases, correctness is preserved: chunk-header validation compares `proposed_split` against the matching `ChunkExtra` from when the chunk was applied, and these agree by construction. Missing chunks can only delay or substitute a split, never produce an incorrect one.
 
 ### Selecting the Split at Epoch Boundary
 
@@ -350,10 +375,14 @@ An attacker could try to force unnecessary resharding by inflating a shard's sta
 
 Resharding requires additional memory and disk I/O (for flat state migration). The `max_number_of_shards` parameter is used to size thread pools conservatively, preventing resource exhaustion from unexpected shard count increases.
 
+The compute cost of running the split-search algorithm itself is negligible: tests on mainnet state with a release-optimized binary and memtries enabled showed `find_trie_split` completing in under 1 ms for every shard.
+
 The peak memory usage during resharding deserves particular attention:
 
 - **Hybrid MemTrie sharing**: When the parent memtrie is frozen at the resharding boundary, its memory is wrapped in an `Arc<STArenaMemory>` and shared between both child shards' `HybridArena`s. There is **no immediate duplication** of memory at the moment of the split.
-- **Gradual growth post-split**: After resharding, each child's `HybridArena` accumulates new owned memory as state changes. The shared frozen parent memory is retained until the background flat-state resharding completes and the children's memtries are rebuilt from scratch (typically within the same epoch). At that point the frozen parent memory is dropped.
+- **Frozen-layer orphan accumulation**: As state mutates after the split, modifications allocate new nodes in each child's owned layer. Nodes in the frozen (shared) layer that become unreachable are **not** reclaimed: the trie's refcount logic intentionally short-circuits for shared memory (`add_ref` and `remove_ref` are no-ops when the node is in the frozen layer; see `core/store/src/trie/mem/node/encoding.rs`), so orphaned frozen nodes remain resident. The owned layer grows in parallel with each modification, so the **total** memory occupied by a child's hybrid memtrie keeps growing for as long as the node tracks that child shard. We recommend operators provision an additional **10–20%** of the largest shard's memtrie size as headroom for this effect. This range is an engineering estimate, not empirically measured -- it should be refined once mainnet measurements are available. Two mechanisms release the hybrid memtrie (and its frozen layer) entirely:
+  - **Automatic, via assignment changes**: when a node's assignment changes such that it stops tracking a child shard, that shard's hybrid memtrie is dropped at the epoch boundary (`retain_memtries`).
+  - **Manual, via node restart**: an operator who observes elevated RAM usage can restart the node. After restart, the memtrie is reloaded from flat storage as a plain `STArena` with no frozen layer, recovering the orphaned memory.
 - **Pre-loading is the dominant cost**: Because dynamic resharding can split any shard at any epoch boundary, a validator whose assignment changes such that it must track a child of a soon-to-be-split shard needs to pre-load the parent memtrie before the split executes. During this preparation window (epoch N+1, between the split decision in epoch N's last block and execution at the N+1 → N+2 boundary), the validator holds **two memtries simultaneously**: its currently-tracked shard plus the parent shard being prepared.
 
 In the pessimistic case, validators MUST be provisioned with enough RAM to hold the two largest shards' memtries concurrently. This is the fundamental constraint dynamic resharding adds compared to [Resharding V3][NEP-568], where the parent shard to be loaded was known weeks in advance via the protocol upgrade schedule. Operators SHOULD treat "2× the size of the largest current shard's memtrie" as the minimum RAM headroom available beyond their currently-tracked shard.
@@ -373,10 +402,6 @@ Instead of memory usage, gas (compute) utilization could drive resharding decisi
 - Gas usage is more volatile and harder to threshold reliably.
 - Memory (state size) is the primary constraint for node operators (RAM for memtries).
 - Gas-based resharding could be added as a complementary signal in a future extension.
-
-### Shard Merging
-
-This NEP only supports splitting. Merging underutilized shards would allow the network to reclaim resources, but introduces significant additional complexity (reversing the split map, merging state tries, reassigning receipts). Merging is deferred to a future NEP as discussed in [Resharding V3's Future Possibilities][NEP-568].
 
 ### Split Multiple Shards per Epoch
 
--- a/neps/nep-0639.md
+++ b/neps/nep-0639.md
@@ -125,14 +125,14 @@ pub struct TrieSplit {
 }
 ```
 
-NEAR's state trie is a 16-ary Merkle-Patricia trie where each node stores a cumulative `memory_usage` for its subtree. Of the trie's column subtrees (identified by the first byte of the key), only four are keyed by account ID and therefore relevant to the split:
+NEAR's state trie is a 16-ary Merkle-Patricia trie where each node stores a cumulative `memory_usage` for its subtree. Several column subtrees (identified by the first byte of the key) are keyed by account ID -- the full list is in `COLUMNS_WITH_ACCOUNT_ID_IN_KEY`. Of these, the split algorithm considers only the four columns that dominate state size in practice:
 
 - **ACCOUNT** (column 0): account records
 - **CONTRACT_CODE** (column 1): WASM contract code
 - **ACCESS_KEY** (column 2): access keys
 - **CONTRACT_DATA** (column 9): contract key-value storage
 
-Other columns (delayed receipts, buffered receipts, promise yields, etc.) store shard-global state not keyed by account ID and are excluded from the split search.
+The remaining account-keyed columns (received data, postponed receipts, promise yield receipts, etc.) are typically small and including them in the split search would add overhead without meaningfully improving the split quality. Columns not keyed by account ID (delayed receipts, buffered receipts, promise yield indices, etc.) store shard-global state and are inherently excluded.
 
 The algorithm performs a greedy binary-search-like traversal, descending through all four subtrees simultaneously:
 
@@ -146,7 +146,7 @@ The algorithm performs a greedy binary-search-like traversal, descending through
 6. Track the best split (lowest `mem_diff = |left_memory - right_memory|`).
 7. Stop when the ACCOUNT subtree reaches a leaf or when further descent cannot improve the split.
 
-The resulting boundary account MUST be at least 2 bytes long and composed of valid UTF-8. If the natural descent terminates too early, the algorithm force-descends into the first available child until the minimum path length is met.
+The resulting boundary account MUST satisfy NEAR's existing account ID validity rules (minimum 2 characters, valid UTF-8). If the natural descent terminates too early to produce a valid account ID, the algorithm force-descends into the first available child until the minimum path length is met.
 
 The algorithm works with both in-memory tries (memtries) and disk-based tries, and is deterministic and reproducible from state witnesses (recorded trie storage), which is critical for stateless validation.
 
@@ -173,7 +173,7 @@ At the last block of each epoch, the block producer:
 2. Calls `get_upcoming_shard_split()`, which:
    - Verifies dynamic resharding is enabled (via `ShardLayoutConfig::Dynamic`).
    - Checks the resharding cooldown: `epoch_height - last_resharding >= min_epochs_between_resharding`.
-   - Calls `pick_shard_to_split()` to select the winning shard: forced shards have priority; otherwise the shard with the highest `total_memory()` wins.
+   - Calls `pick_shard_to_split()` to select the winning shard: forced shards have priority; otherwise the shard with the highest `total_memory()` wins. In the (extremely unlikely in production, but possible in tests) case of a tie, the shard with the highest `ShardId` wins -- the selection key is the tuple `(total_memory, shard_id)`, ensuring determinism.
 3. Embeds the result as `shard_split: Option<(ShardId, AccountId)>` in `BlockHeaderInnerRestV6`.
 
 Only the last block of each epoch MAY carry a non-None `shard_split`.
@@ -187,7 +187,8 @@ During `EpochManager::finalize_epoch()`, when finalizing epoch N:
    - **Static fallback**: If the next-next epoch config specifies a static layout, use it (backward compatibility).
    - **Transitional**: If the current epoch doesn't have dynamic resharding enabled, carry forward the existing layout.
    - **Dynamic**: If a split is present, derive a new `ShardLayoutV3` from the current layout. New shard IDs are assigned as `max(existing_shard_ids) + 1` and `max(existing_shard_ids) + 2`.
-3. Store the resulting layout in `EpochInfoV5.shard_layout` for epoch N+2.
+3. **Stale-split guard**: If `shard_split` references a shard that no longer exists in the next layout (e.g. it was already split in a previous epoch via a static-layout protocol upgrade), the split is skipped and the next layout is carried forward unchanged. This prevents inconsistencies when a static and dynamic split happen to target the same shard in adjacent epochs.
+4. Store the resulting layout in `EpochInfoV5.shard_layout` for epoch N+2.
 
 The two-epoch delay (split decided at end of epoch N, takes effect in epoch N+2) gives the network one full epoch (N+1) to prepare: validators are reassigned, state migration begins, and memtries are pre-loaded.
 
@@ -231,7 +232,7 @@ This section lists data structures introduced or modified relative to [Reshardin
 | `DynamicReshardingConfig` | **New** | Resharding thresholds and limits |
 | `ShardLayoutConfig` | **New** | Enum: `Static` (legacy) or `Dynamic` |
 | `EpochInfoV5` | **New version** | Adds `shard_layout` and `last_resharding` fields |
-| `BlockHeaderInnerRestV6` | **New version** | Adds `shard_split: Option<(ShardId, AccountId)>` |
+| `BlockHeaderInnerRestV6` | **New version** | Adds `shard_split: Option<(ShardId, AccountId)>`, removes deprecated challenges fields |
 | `ShardChunkHeaderInnerV5` | **New version** | Adds `proposed_split: Option<TrieSplit>` |
 | `BlockInfoV4` | **New version** | Adds `shard_split`, removes deprecated `slashed` map |
 | `ChunkExtraV5` | **New version** | Adds `proposed_split: Option<TrieSplit>` |
@@ -243,7 +244,7 @@ This section lists data structures introduced or modified relative to [Reshardin
 `ShardLayoutV3` replaces V2 with a layout that stores the **full cumulative split history**:
 
 - `shards_split_map: ShardsSplitMapV3` — records every split across all generations, not just the most recent one.
-- `shards_ancestor_map: ShardsAncestorMapV3` — enables O(1) lookup of all ancestors for any shard, which is needed for shard tracking (determining which current shards a node must track based on which historical shards it was assigned to).
+- `shards_ancestor_map: ShardsAncestorMapV3` — enables O(log N) lookup of all ancestors for any shard (it is a `BTreeMap<ShardId, Vec<ShardId>>`), which is needed for shard tracking (determining which current shards a node must track based on which historical shards it was assigned to).
 - `last_split: ShardId` — tracks the most recently split shard.
 
 This cumulative history is essential because dynamic resharding can produce long chains of splits (e.g., shard A → B,C → B,D,E → ...) and every node must be able to resolve any historical shard ID to its current descendants.
@@ -307,7 +308,7 @@ The implementation is gated behind the `ProtocolFeature::DynamicResharding` feat
 - `chain/chain/src/chain.rs`: `maybe_start_memtrie_preload_for_resharding`, `postprocess_ready_block`
 - `core/store/src/trie/shard_tries.rs`: `spawn_background_memtrie_loading_for_shard`, `try_finalize_background_memtrie_loading`, `retain_memtries`
 - `core/store/src/trie/mem/loading.rs`: `apply_deltas_to_memtries`
-- `core/store/src/flat/storage.rs`: `hold_flat_head`, `release_flat_head_hold`
+- `core/store/src/flat/storage.rs`: `hold_flat_head` (returns a `FlatHeadHold` guard that releases the hold when dropped)
 
 ### Shard Layout Derivation
 
@@ -347,7 +348,17 @@ An attacker could try to force unnecessary resharding by inflating a shard's sta
 
 ### Resource Consumption During Resharding
 
-Resharding requires additional memory (for hybrid memtries) and disk I/O (for flat state migration). The two-epoch delay ensures nodes have adequate preparation time. The `max_number_of_shards` parameter is used to size thread pools conservatively, preventing resource exhaustion from unexpected shard count increases.
+Resharding requires additional memory and disk I/O (for flat state migration). The `max_number_of_shards` parameter is used to size thread pools conservatively, preventing resource exhaustion from unexpected shard count increases.
+
+The peak memory usage during resharding deserves particular attention:
+
+- **Hybrid MemTrie sharing**: When the parent memtrie is frozen at the resharding boundary, its memory is wrapped in an `Arc<STArenaMemory>` and shared between both child shards' `HybridArena`s. There is **no immediate duplication** of memory at the moment of the split.
+- **Gradual growth post-split**: After resharding, each child's `HybridArena` accumulates new owned memory as state changes. The shared frozen parent memory is retained until the background flat-state resharding completes and the children's memtries are rebuilt from scratch (typically within the same epoch). At that point the frozen parent memory is dropped.
+- **Pre-loading is the dominant cost**: Because dynamic resharding can split any shard at any epoch boundary, a validator whose assignment changes such that it must track a child of a soon-to-be-split shard needs to pre-load the parent memtrie before the split executes. During this preparation window (epoch N+1, between the split decision in epoch N's last block and execution at the N+1 → N+2 boundary), the validator holds **two memtries simultaneously**: its currently-tracked shard plus the parent shard being prepared.
+
+In the pessimistic case, validators MUST be provisioned with enough RAM to hold the two largest shards' memtries concurrently. This is the fundamental constraint dynamic resharding adds compared to [Resharding V3][NEP-568], where the parent shard to be loaded was known weeks in advance via the protocol upgrade schedule. Operators SHOULD treat "2× the size of the largest current shard's memtrie" as the minimum RAM headroom available beyond their currently-tracked shard.
+
+Note that a transient case of holding two memtries also exists today whenever a validator's assignment changes between epochs (independently of resharding). Dynamic resharding makes this case more frequent and less predictable.
 
 ## Alternatives
 
--- a/neps/nep-0639.md
+++ b/neps/nep-0639.md
@@ -1,0 +1,448 @@
+---
+NEP: 639
+Title: Dynamic Resharding
+Authors: Adam Wierzbicki
+Status: Draft
+DiscussionsTo: https://github.com/near/nearcore/issues/14070
+Type: Protocol
+Version: 1.0.0
+Created: 2026-04-09
+LastUpdated: 2026-04-09
+Requires: 568
+---
+
+## Summary
+
+This proposal introduces dynamic resharding: a protocol-level mechanism that enables automatic splitting of shards based on their state size, without requiring a protocol upgrade. It replaces the static approach used by [Resharding V2][NEP-508] and [Resharding V3][NEP-568], where shard layout changes were hardcoded into specific protocol versions and required coordinated binary releases.
+
+With dynamic resharding, each shard proposes a split during chunk application when its memory usage exceeds a configurable threshold. The block producer for the last block of an epoch selects the best candidate, and the resulting shard layout takes effect two epochs later. The entire process is deterministic, validated through state witnesses, and requires no human intervention.
+
+## Motivation
+
+[Resharding V3][NEP-568] established the core infrastructure for splitting shards: instant MemTrie resharding, stateless validation of resharding proofs, state mapping for efficient storage, and flat state background resharding. However, V3 still requires the NEAR developer team to manually analyze shard utilization, determine boundary accounts, hardcode the new layout into the `neard` binary, and coordinate a protocol upgrade across the network. This process is slow, operationally expensive, and cannot react to sudden changes in demand.
+
+As the NEAR ecosystem grows, the network needs the ability to scale out automatically. Dynamic resharding addresses this by:
+
+- **Eliminating manual coordination**: Shard splits are proposed and executed by the protocol itself, based on observed state size.
+- **Reacting to demand**: Overloaded shards are detected and split within two epochs (~24 hours on mainnet), rather than waiting weeks for a protocol upgrade cycle.
+- **Enabling continuous scaling**: The network can grow from its current shard count toward a configurable maximum without any human intervention.
+
+## Specification
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+### Overview
+
+Dynamic resharding extends the [Resharding V3][NEP-568] infrastructure with three new protocol mechanisms:
+
+1. **Split proposal**: During chunk application, each shard computes a proposed split point based on trie memory usage.
+2. **Split selection**: The block producer for the last block of an epoch collects proposals and selects the best candidate.
+3. **Layout derivation**: The epoch manager derives a new `ShardLayoutV3` for two epochs ahead.
+
+The actual resharding execution (MemTrie splitting, flat state migration, state mapping) reuses the [Resharding V3][NEP-568] machinery without modification.
+
+### Key Differences from Resharding V3
+
+| Aspect | Resharding V3 | Dynamic Resharding |
+|--------|--------------|-------------------|
+| **Trigger** | Manual: hardcoded in binary, tied to protocol version | Automatic: threshold-based, per-epoch evaluation |
+| **Boundary account** | Determined offline by the developer team | Computed algorithmically from trie memory usage |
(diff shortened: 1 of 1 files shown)


walnut-the-cat:
Does dynamic resharding require validators to always be equipped with enough RAM for parent shard and child shards at all time?

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+An attacker could try to force unnecessary resharding by inflating a shard's state size. This is mitigated by:
+
+- The `min_child_memory_usage` threshold prevents splits that would create very small shards.
+- The `max_number_of_shards` cap prevents unbounded shard proliferation.
+- The `min_epochs_between_resharding` cooldown limits resharding frequency.
+- Storage staking costs make state inflation expensive.
+
+### Resource Consumption During Resharding
+
+Resharding requires additional memory (for hybrid memtries) and disk I/O (for flat state migration). The two-epoch delay ensures nodes have adequate preparation time. The `max_number_of_shards` parameter is used to size thread pools conservatively, preventing resource exhaustion from unexpected shard count increases.
```
Trisfald: Is there a practical recommendation we should give to node operators, to have their nodes ready for resharding?
Something along the lines of: "ensure your nodes have at least X GB of RAM available at all times, because resharding can happen"?

neps/nep-0639.md
Trisfald, line 1: Did we implement best-effort sticky assignment of child shards? Should it be mentioned in the NEP?

Wiezzel: Not yet, but it will be implemented.

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+| `ChunkExtraV5` | **New version** | Adds `proposed_split: Option<TrieSplit>` |
+
+#### ShardLayoutV3
+
+`ShardLayoutV2` (used by [Resharding V3][NEP-568]) stores only a single-generation split map: it records which parent shard was split into which children for the most recent resharding event. This is sufficient for manually scheduled, one-off resharding but breaks down with dynamic resharding, where repeated automatic splits create multi-generational parent-child chains.
+
+`ShardLayoutV3` replaces V2 with a layout that stores the **full cumulative split history**:
+
+- `shards_split_map: ShardsSplitMapV3` — records every split across all generations, not just the most recent one.
+- `shards_ancestor_map: ShardsAncestorMapV3` — enables O(1) lookup of all ancestors for any shard, which is needed for shard tracking (determining which current shards a node must track based on which historical shards it was assigned to).
```
staffik: nit: is not logN ?

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+### Validation
+
+- `chain/chain/src/validate.rs`: `validate_chunk_with_chunk_extra_and_receipts_root` (chunk header validation), `validate_block_shard_split` (block header validation)
+
+### Memtrie Pre-loading
+
+- `chain/chain/src/chain.rs`: `maybe_start_memtrie_preload_for_resharding`, `postprocess_ready_block`
+- `core/store/src/trie/shard_tries.rs`: `spawn_background_memtrie_loading_for_shard`, `try_finalize_background_memtrie_loading`, `retain_memtries`
+- `core/store/src/trie/mem/loading.rs`: `apply_deltas_to_memtries`
+- `core/store/src/flat/storage.rs`: `hold_flat_head`, `release_flat_head_hold`
```
staffik: I cannot find `release_flat_head_hold`

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+| Structure | Change | Description |
+|-----------|--------|-------------|
+| `ShardLayoutV3` | **New** | Shard layout with cumulative split history and ancestor maps (replaces `ShardLayoutV2`) |
+| `TrieSplit` | **New** | Split result: boundary account, left/right memory usage |
+| `DynamicReshardingConfig` | **New** | Resharding thresholds and limits |
+| `ShardLayoutConfig` | **New** | Enum: `Static` (legacy) or `Dynamic` |
+| `EpochInfoV5` | **New version** | Adds `shard_layout` and `last_resharding` fields |
+| `BlockHeaderInnerRestV6` | **New version** | Adds `shard_split: Option<(ShardId, AccountId)>` |
+| `ShardChunkHeaderInnerV5` | **New version** | Adds `proposed_split: Option<TrieSplit>` |
+| `BlockInfoV4` | **New version** | Adds `shard_split`, removes deprecated `slashed` map |
```
staffik: Should we mention about removed fields like `challenges` ?

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+2. Compute a threshold: `total_memory / 2 - left_memory`.
+3. Find the "middle child" -- the nibble where the cumulative sum first exceeds the threshold.
+4. Descend into the middle child across all subtrees.
+5. At each position, evaluate two candidate splits:
+   - **Split A**: Use the current path as the boundary (middle memory goes to the right shard).
+   - **Split B**: Append "-0" to the current path, pushing the current account and its attached data (access keys, contract data) to the left shard and child accounts to the right. This handles "big middle accounts" with large contracts or many data entries.
+6. Track the best split (lowest `mem_diff = |left_memory - right_memory|`).
+7. Stop when the ACCOUNT subtree reaches a leaf or when further descent cannot improve the split.
+
+The resulting boundary account MUST be at least 2 bytes long and composed of valid UTF-8. If the natural descent terminates too early, the algorithm force-descends into the first available child until the minimum path length is met.
```
staffik: I would rephrase it to tell that the 2 bytes is an existing restriction

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+
+### Selecting the Split at Epoch Boundary
+
+At the last block of each epoch, the block producer:
+
+1. Collects `proposed_split` values from all chunk headers in the block.
+2. Calls `get_upcoming_shard_split()`, which:
+   - Verifies dynamic resharding is enabled (via `ShardLayoutConfig::Dynamic`).
+   - Checks the resharding cooldown: `epoch_height - last_resharding >= min_epochs_between_resharding`.
+   - Calls `pick_shard_to_split()` to select the winning shard: forced shards have priority; otherwise the shard with the highest `total_memory()` wins.
```
staffik: Should we mention that highest shard_id wins in the theoretical case of equal  total_memory?

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+get_upcoming_shard_split() -- collects proposals from chunk headers, picks the best
+    |
+    v
+[BlockHeaderInnerRestV6.shard_split] -- embedded in block header
+    |
+    v
+[EpochManager::finalize_epoch()]
+    |
+    v
+next_next_shard_layout() -- derives ShardLayoutV3 for epoch N+2
```
staffik, line 343: There is condition "Skip the split if the shard no longer exists in the next layout" but not sure if we should mention it here

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+The result (`Option<TrieSplit>`) is stored in `ChunkExtra` and embedded in the chunk header via the new `ShardChunkHeaderInnerV5.proposed_split` field.
+
+### Selecting the Split at Epoch Boundary
+
+At the last block of each epoch, the block producer:
+
+1. Collects `proposed_split` values from all chunk headers in the block.
+2. Calls `get_upcoming_shard_split()`, which:
+   - Verifies dynamic resharding is enabled (via `ShardLayoutConfig::Dynamic`).
+   - Checks the resharding cooldown: `epoch_height - last_resharding >= min_epochs_between_resharding`.
```
staffik, line 203: To clarify the semantics: `min_epochs_between_resharding = 1` means the resharding can happen every second epoch, right?

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+The resulting boundary account MUST be at least 2 bytes long and composed of valid UTF-8. If the natural descent terminates too early, the algorithm force-descends into the first available child until the minimum path length is met.
+
+The algorithm works with both in-memory tries (memtries) and disk-based tries, and is deterministic and reproducible from state witnesses (recorded trie storage), which is critical for stateless validation.
+
+### Proposing a Shard Split
+
+During chunk application, the runtime checks whether a shard should be split. This check is performed by `compute_proposed_split`, which:
+
+1. Verifies the `DynamicResharding` protocol feature flag is enabled.
+2. Checks that the next block could be the last of the epoch (using `is_next_block_possibly_last_in_epoch`). This conservative check may produce false positives but never false negatives.
```
staffik, line 175: Maybe we should tell how it behaves with missing chnks

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+### DoS via Threshold Manipulation
+
+An attacker could try to force unnecessary resharding by inflating a shard's state size. This is mitigated by:
+
+- The `min_child_memory_usage` threshold prevents splits that would create very small shards.
+- The `max_number_of_shards` cap prevents unbounded shard proliferation.
+- The `min_epochs_between_resharding` cooldown limits resharding frequency.
+- Storage staking costs make state inflation expensive.
+
+### Resource Consumption During Resharding
```
staffik, line 413: Maybe we could paste some results from dry-run, how long `find_trie_split` took, etc

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+
+    /// Minimum epochs between consecutive resharding events.
+    /// A value of 0 allows resharding every epoch.
+    pub min_epochs_between_resharding: EpochHeight,
+
+    /// Shards to force-split regardless of memory thresholds.
+    pub force_split_shards: Vec<ShardId>,
+
+    /// Shards that MUST NOT be split regardless of memory thresholds.
+    pub block_split_shards: Vec<ShardId>,
```
staffik, line 92: Maybe we could describe what steps can be done in case we detect some issues with resharding and want to disable it.

neps/nep-0639.md
```diff
@@ -0,0 +1,446 @@
...
+    /// right shard gets this account and above.
+    pub boundary_account: AccountId,
+    /// Total memory_usage of the left part.
+    pub left_memory: u64,
+    /// Total memory_usage of the right part.
+    pub right_memory: u64,
+}
+```
+
+NEAR's state trie is a 16-ary Merkle-Patricia trie where each node stores a cumulative `memory_usage` for its subtree. Of the trie's column subtrees (identified by the first byte of the key), only four are keyed by account ID and therefore relevant to the split:
```
staffik: There are more in `COLUMNS_WITH_ACCOUNT_ID_IN_KEY`, but only these 4 are important, right?

neps/nep-0639.md
Wiezzel: Yes.

neps/nep-0639.md
Wiezzel: find_trie_split took <1ms for all mainnet shards

neps/nep-0639.md
Wiezzel: I have extended this section.

neps/nep-0639.md
Wiezzel: Fixed.

neps/nep-0639.md
Wiezzel: Added.

neps/nep-0639.md
Wiezzel: Yes, improved the description.

Wiezzel:
@walnut-the-cat 

> Does dynamic resharding require validators to always be equipped with enough RAM for parent shard and child shards at all time?

Not exactly, see the update "resource consumption" section.

neps/nep-0639.md
```diff
@@ -0,0 +1,457 @@
...
+Created: 2026-04-09
+LastUpdated: 2026-04-09
+Requires: 568
+---
+
+## Summary
+
+This proposal introduces dynamic resharding: a protocol-level mechanism that enables automatic splitting of shards based on their state size, without requiring a protocol upgrade. It replaces the static approach used by [Resharding V2][NEP-508] and [Resharding V3][NEP-568], where shard layout changes were hardcoded into specific protocol versions and required coordinated binary releases.
+
+With dynamic resharding, each shard proposes a split during chunk application when its memory usage exceeds a configurable threshold. The block producer for the last block of an epoch selects the best candidate, and the resulting shard layout takes effect two epochs later. The entire process is deterministic, validated through state witnesses, and requires no human intervention.
```
darioush, line 18: Can we specify a bit more concretely how the threshold is configured? I couldn't figure out where DynamicReshardingConfig values should come from.

neps/nep-0639.md
```diff
@@ -0,0 +1,457 @@
...
+
+### Dynamic Resharding Based on Gas Usage
+
+Instead of memory usage, gas (compute) utilization could drive resharding decisions. This was considered but deferred because:
+
+- Gas usage is more volatile and harder to threshold reliably.
+- Memory (state size) is the primary constraint for node operators (RAM for memtries).
+- Gas-based resharding could be added as a complementary signal in a future extension.
+
+### Shard Merging
```
darioush: nit: should this one be under future possibilities?

neps/nep-0639.md
Wiezzel: Comes from epoch config. Clarified.

neps/nep-0639.md
Wiezzel: Added a section about that.

neps/nep-0639.md
Wiezzel: Added.

walnut-the-cat:
Thank you @Wiezzel for submitting this NEP.

As a moderator, I reviewed this NEP and it meets the proposed template guidelines. I am moving this NEP to the REVIEW stage and suggest @Trisfald and @shreyan-gupta as SME reviewers to complete a technical review (see expectations). @near/wg-protocol working group members, please let us know if you have better suggestions. 

<details>

</details>

neps/nep-0639.md
Trisfald: Can you please update the NEP?
I think sticky assignment has been merged in master

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+| **Boundary account** | Determined offline by the developer team | Computed algorithmically from trie memory usage |
+| **Shard layout source** | `EpochConfig` (determined by protocol version) | `EpochInfo` (stored per-epoch) |
+| **Configuration** | `ShardLayoutConfig::Static` | `ShardLayoutConfig::Dynamic` |
+| **Shard layout struct** | `ShardLayoutV2` (single-generation split map) | `ShardLayoutV3` (cumulative split history) |
+| **Chunk headers** | V4 | V5 (adds `proposed_split: Option<TrieSplit>`) |
+| **Block headers** | V5 | V6 (adds `shard_split: Option<(ShardId, AccountId)>`) |
+| **EpochInfo** | V4 | V5 (adds `shard_layout` and `last_resharding` fields) |
+| **Resharding cooldown** | N/A (manual scheduling) | Enforced via `min_epochs_between_resharding` |
+
+The execution path after a split is decided (MemTrie splitting, flat state resharding, state mapping, stateless validation, receipt handling) is **identical** to Resharding V3. See [NEP-568][NEP-568] for details on those mechanisms.
```
shreyan-gupta, line 58: `StickyReshardingValidatorAssignment` (`core/primitives-core/src/version.rs:509`, PV 153) doesn't appear anywhere in the NEP. Without it, every layout change triggers from-scratch chunk-producer reassignment by `ShardIndex` — effectively a network-wide state-sync stampede on each split, so it's pretty load-bearing for operational viability.

Could you add a short subsection covering the strategies (`CarryOver` / `StickyResharding` / `Fresh`), the greedy stake-balanced bin-pack with imbalance bound ≤ max validator stake (`chain/epoch-manager/src/shard_assignment/sticky_resharding.rs:77`), the `chunk_producer_assignment_changes_limit` raise on sticky epochs, and the fall-through to `Fresh` when sticky construction fails? Worth noting that this is availability-only — misassignment causes missed chunks / kickouts, not consensus disagreement — and that merges aren't yet supported.

One related question: both `DynamicResharding` (PV 150) and `StickyReshardingValidatorAssignment` (PV 153) are still in nightly. Any thoughts on the intended ordering at release? Sticky before / with / after DR? Just want to make sure DR doesn't ship without sticky.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+    /// Maximum number of shards in the network. No further splitting
+    /// occurs once this limit is reached.
+    pub max_number_of_shards: NumShards,
+
+    /// Minimum epochs between consecutive resharding events.
+    /// A value of 0 allows resharding every epoch (gap of 1 epoch between
+    /// effective layouts); a value of N enforces a gap of at least N+1 epochs
+    /// between effective layouts.
+    pub min_epochs_between_resharding: EpochHeight,
```
shreyan-gupta: The doc-comment says "value of 0 means resharding can happen every epoch", but with `= 0` there's a latent issue worth flagging. Concretely:

1. The resharding manager (`chain/chain/src/resharding/manager.rs:256`) builds each child's first `ChunkExtra` by `clone()`-ing the parent's last `ChunkExtra`. If the parent shard was proposing yet another split in its last chunk of epoch N+1, the child inherits that `proposed_split = Some(...)`.
2. At the start of epoch N+2, the child shard's first chunk freshly computes `proposed_split = None` because the `is_next_block_possibly_last_in_epoch` short-circuit fires at epoch start (`chain/chain/src/runtime/mod.rs:565`).
3. The child's first chunk header therefore carries `proposed_split = None` while its first `ChunkExtra` still has `proposed_split = Some(...)`.
4. Validation at `chain/chain/src/validate.rs:175` compares header vs `ChunkExtra` and rejects with `InvalidChunkHeaderShardSplit`.

`min_epochs_between_resharding ≥ 1` makes this dormant — the cooldown prevents the parent's last `ChunkExtra` from carrying a fresh proposal. The `manager.rs:253` TODO ("set all fields of ChunkExtra") implicitly covers it. Could either require `≥ 1` in the spec, or reset `proposed_split = None` on the child clone before allowing `= 0`?

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+    /// Shards to force-split regardless of memory thresholds.
+    pub force_split_shards: Vec<ShardId>,
+
+    /// Shards that MUST NOT be split regardless of memory thresholds.
+    pub block_split_shards: Vec<ShardId>,
+}
```
shreyan-gupta, line 93: The NEP defers all four parameters to "initial mainnet values will be set conservatively" but doesn't commit to numbers, which makes "Limiting the Blast Radius" hard to evaluate. Could you fill in the planned values plus brief rationale? Something like:

- `memory_usage_threshold` ≈ 40 GB (~20 GB actual memtrie RAM, given the TRIE_COSTS ratio)
- `max_number_of_shards` = 10 (mainnet currently has 9 shards — matches the "current + 1" promise)
- `min_child_memory_usage`: ?
- `min_epochs_between_resharding`: ?

Also worth noting that as of master, no `core/primitives/res/epoch_configs/{mainnet,testnet}/*.json` has `ShardLayoutConfig::Dynamic` configured — the rollout JSON commit is a prerequisite for DR to actually take effect.

One specific tension for `min_child_memory_usage`: the first split candidate (shard 5) produces a fairly imbalanced result (~8 GB / ~32 GB) because one account dominates the shard. A value above ~8 GB would block that split. The imbalance is intentional — it's just the best boundary available given AccountID granularity, not an algorithm bug — so the parameter should be picked to allow this case rather than around it. Worth calling out so reviewers don't read the imbalance as a problem to design away.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+Updating any value therefore requires a protocol upgrade: a new protocol version is added with a new epoch config JSON file, and the network adopts the change once enough nodes vote to upgrade. This is intentional -- thresholds and limits affect consensus (every validator must agree on whether a split should be proposed), so they MUST be uniform across the network. Initial mainnet values will be set conservatively (see [Limiting the Blast Radius](#limiting-the-blast-radius)) and tuned through subsequent protocol upgrades as operational experience accumulates.
+
+With dynamic resharding, per-shard default cache sizes are no longer derived from the protocol version, since the shard layout is not known at compile time. Nodes use uniform default cache sizes across all shards. Performance testing on mainnet state has shown no significant impact from this change. Node operators can still override cache sizes via manual configuration if needed.
```
shreyan-gupta: Two small clarifications on this paragraph:

1. Which cache. This is the trie cache at the RocksDB trie-storage layer (`per_shard_max_bytes` in `trie_cache_config`, see `core/store/src/config.rs:320`) — not the memtrie or block cache. Worth naming explicitly so RPC operators reading the NEP know what to look for.
2. The per-shard cache is keyed by `ShardUId`, and new shards get IDs `max+1`/`max+2` allocated at split time — which operators can't pre-declare in config. So RPC nodes that rely on per-shard cache sizing end up with the default (500 MB + 50 MB per shard) for the new shards until they update config post-split. Validators run on memtries so this is RPC-side only, but it could affect query latency.

Could you expand the paragraph to mention both?

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+### Limiting the Blast Radius
+
+When dynamic resharding is first released on mainnet, `max_number_of_shards` will be set to one more than the current shard count. This caps the damage from any single resharding event: at most one new shard can be created before further splits are blocked, regardless of memory thresholds. Successive protocol upgrades will raise the limit gradually as confidence in the feature grows. With this conservative rollout, in practice we expect that disabling dynamic resharding will never be necessary -- the shard limit itself acts as the safety mechanism.
+
+In the unlikely event that the shard limit proves insufficient (for instance, a previously-allowed split turns out to be problematic and we want to prevent any further splits while we investigate), an emergency mitigation is to ship a protocol upgrade that sets `block_split_shards` to include every existing shard ID. This blocks all further splits while leaving the rest of the protocol unchanged. We note, however, that if a protocol upgrade is needed anyway, it is generally preferable to use that upgrade to fix the underlying issue rather than to disable resharding.
```
shreyan-gupta: The current wording calls the protocol-upgrade `block_split_shards = all` path "an emergency mitigation". Worth being more direct: a protocol upgrade is the *only* mitigation, which takes 1-2 weeks. The `max_number_of_shards = current + 1` cap is therefore the load-bearing safety mechanism — if a split goes wrong there's no faster recourse.

Possibly worth considering whether a runtime-config override path is feasible too, so operators have a faster knob in genuinely emergent situations. Currently there's no override; everything routes through the protocol upgrade cycle. Not a blocker for this NEP, but worth mentioning in the safety story.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+4. Descend into the middle child across all subtrees.
+5. At each position, evaluate two candidate splits:
+   - **Split A**: Use the current path as the boundary (middle memory goes to the right shard).
+   - **Split B**: Append "-0" to the current path, pushing the current account and its attached data (access keys, contract data) to the left shard and child accounts to the right. This handles "big middle accounts" with large contracts or many data entries.
+6. Track the best split (lowest `mem_diff = |left_memory - right_memory|`).
+7. Stop when the ACCOUNT subtree reaches a leaf or when further descent cannot improve the split.
+
+The resulting boundary account MUST satisfy NEAR's existing account ID validity rules (minimum 2 characters, valid UTF-8). If the natural descent terminates too early to produce a valid account ID, the algorithm force-descends into the first available child until the minimum path length is met.
+
+The algorithm works with both in-memory tries (memtries) and disk-based tries, and is deterministic and reproducible from state witnesses (recorded trie storage), which is critical for stateless validation.
```
shreyan-gupta: The algorithm prose covers the high-level shape, but a few details that are load-bearing for `InvalidChunkHeaderShardSplit` validation aren't fully specified here (force-descent on odd-nibble paths, `find_middle_child` fallback when no child crosses the threshold, left-excludes / right-includes boundary).

To avoid expanding the prose with pseudocode, maybe add one line at the end:

> "Protocol behavior is defined by the reference implementation at `core/store/src/trie/split.rs`."

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+At the last block of each epoch, the block producer:
+
+1. Collects `proposed_split` values from all chunk headers in the block.
+2. Calls `get_upcoming_shard_split()`, which:
+   - Verifies dynamic resharding is enabled (via `ShardLayoutConfig::Dynamic`).
+   - Checks the resharding cooldown: `epoch_height - last_resharding >= min_epochs_between_resharding`.
+   - Calls `pick_shard_to_split()` to select the winning shard: forced shards have priority; otherwise the shard with the highest `total_memory()` wins. In the (extremely unlikely in production, but possible in tests) case of a tie, the shard with the highest `ShardId` wins -- the selection key is the tuple `(total_memory, shard_id)`, ensuring determinism.
+3. Embeds the result as `shard_split: Option<(ShardId, AccountId)>` in `BlockHeaderInnerRestV6`.
+
+Only the last block of each epoch MAY carry a non-None `shard_split`.
```
shreyan-gupta, line 207: Two clarifications worth adding here:

1. Why both chunk and block headers carry split info. Chunk headers carry per-shard proposals (only the chunk producer for that shard has the trie state to compute one). The block header carries the picked winner — `EpochManager::finalize_epoch` consumes a single source of truth (`lib.rs:678`), and non-tracking validators / light clients need the decision without per-chunk data. A short paragraph here would close that gap.
2. When multiple shards are in `force_split_shards`, `chain/epoch-manager/src/lib.rs:1937` iterates by `Vec` insertion order — first match wins regardless of memory usage. One sentence so this isn't undefined behavior in the spec.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+### Deriving the New Shard Layout
+
+During `EpochManager::finalize_epoch()`, when finalizing epoch N:
+
+1. Read `block_info.shard_split()` from the last block of epoch N.
+2. Call `next_next_shard_layout()`, which has three phases:
```
shreyan-gupta: The "three phases" list is compact and the preconditions aren't quite clear from the prose. Could you expand them along the lines of:

1. **Static fallback** — when the next-next epoch still has `ShardLayoutConfig::Static`. Returns the static layout. Used for legacy / manual scheduling and tests.
2. **Transitional** — when the next-next epoch is the first under `ShardLayoutConfig::Dynamic`. Derives a fresh `ShardLayoutV3` from the previous layout's history via `derive_with_layout_history()`. Runs exactly once at the protocol upgrade that enables dynamic resharding.
3. **Dynamic** — both this and the next epoch are under `ShardLayoutConfig::Dynamic`. Applies this epoch's `shard_split` decision to derive the new layout.

Code at `chain/epoch-manager/src/lib.rs:651-706` if useful.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+The two-epoch delay (split decided at end of epoch N, takes effect in epoch N+2) gives the network one full epoch (N+1) to prepare: validators are reassigned, state migration begins, and memtries are pre-loaded.
+
+### Migration from Static to Dynamic Resharding
+
+When the network first enables dynamic resharding, the current shard layout (`ShardLayoutV2`) cannot directly become a `ShardLayoutV3` because V2 only stores single-generation parent-child relationships, not the full cumulative split history that V3 requires. The migration uses:
+
+1. `EpochManager::get_shard_layout_history()` reconstructs all historical layouts by iterating through protocol versions (newest to oldest).
+2. `ShardLayoutV3::derive_with_layout_history()` processes this history via `build_shard_split_map()`, which extracts parent-child relationships from `windows(2)` of the layout sequence.
+3. The resulting V3 layout has full ancestor tracking and is ready for further dynamic splits.
```
shreyan-gupta, line 229: Worth being explicit that this migration runs exactly once. After the transitional epoch, all subsequent splits derive from existing V3 layouts via `ShardLayoutV3::derive()`; `derive_with_layout_history()` isn't called again.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+### Validation
+
+Both `proposed_split` and `shard_split` MUST be validated to prevent forging:
+
+- **Chunk header validation**: During state witness validation, the `proposed_split` in the received chunk header is compared against the locally-computed `ChunkExtra.proposed_split()`. A mismatch produces `InvalidChunkHeaderShardSplit`.
+- **Block header validation**: During block processing, the `shard_split` in the block header is recomputed by calling `get_upcoming_shard_split()` with the block's chunk headers. A mismatch produces `InvalidBlockHeaderShardSplit`.
+
+Because the split algorithm is deterministic and operates on state that is provable via state witnesses, all honest validators MUST arrive at the same split proposal for a given shard state.
```
shreyan-gupta, line 240: The Validation section closes the forging surface but doesn't say what happens when two competing "last block of epoch N" candidates appear on a fork. We've seen this on forknet runs — wanted to flag it as worth documenting.

Each branch can carry a valid (self-consistent) `shard_split` value (possibly different shards, or one `None` and one `Some`). Each branch independently begins preparation in epoch N+1 — validators may pre-load based on their branch's view. Doomslug LFB voting canonicalizes one branch; the losing branch's layout decision is reorged out, and its pre-load drops at the next `retain_memtries` call. Correctness is preserved, but operators may see transient "phantom" pre-load activity.

Suggestion: add a short paragraph here noting that `shard_split` is provisional until finality, and that pre-loading on losing fork branches is expected behavior (not a bug).

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+### Memtrie Pre-loading
```
shreyan-gupta, line 246: Three things on this section:

1. "Typically a few minutes" understates the wall-clock. The background load is single-threaded by design (`core/store/src/trie/shard_tries.rs:535,565` set `parallelize: false` to avoid CPU contention with block production). For a 20 GB parent shard, expect more like 10–20 minutes. Could you replace "few minutes" with a more realistic range and note why it's single-threaded?
2. The pre-load bypasses operator memtrie config. `maybe_start_memtrie_preload_for_resharding` (`chain/chain/src/chain.rs:2186-2216`) doesn't check `load_memtries_for_shards` or `load_memtries_for_tracked_shards`. An operator who has explicitly disabled memtries for a shard still gets one loaded during the resharding preparation window. Worth one sentence so this isn't a surprise.
3. The synchronous restart-load fires only for shards the node currently tracks (per `cares_about_shard`). Probably worth adding the precondition to "loaded synchronously" so it's not ambiguous.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+To ensure the parent shard's memtrie is available when resharding executes, a background pre-loading mechanism is used:
+
+1. **Epoch N, last block**: `shard_split` is embedded in the block header.
+2. **Epoch N → N+1 boundary**: `finalize_epoch()` creates EpochInfo for epoch N+2 with the new layout. `maybe_start_memtrie_preload_for_resharding()` detects the layout change and spawns a background thread to load the parent shard's memtrie from flat storage.
+3. **During epoch N+1**: The background load completes (typically a few minutes). On the next block's `postprocess_ready_block()`, the loaded memtrie is received, delta catch-up is applied, and it is inserted into the active memtries map.
+
+To preserve the flat state deltas needed for catch-up, the parent shard's `FlatStorage` flat head is held (paused) during background loading. Multiple subsystems (state snapshots, memtrie loading, resharding) can independently hold the flat head; it only advances once all holds are released.
+
+If the node restarts during the preparation epoch, the pending resharding is detected at startup and the parent shard is loaded synchronously.
```
shreyan-gupta: Both NEP-508 and NEP-568 had dedicated state-sync sections — NEP-639 doesn't. The code handling is actually solid; worth surfacing for spec completeness. Something like:

> "**State sync across resharding boundaries.** A node that falls behind across a resharding boundary uses `EpochManagerAdapter::get_resharding_parent_shard_uid()` to detect a pending layout change and triggers a synchronous parent-memtrie load via the catchup path. Cross-shard receipts remain available across the boundary because the incoming-receipts fetch uses `ReceiptFilter::All`, and shard-membership filtering happens after the merkle path is verified."

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+| `BlockHeaderInnerRestV6` | **New version** | Adds `shard_split: Option<(ShardId, AccountId)>`, removes deprecated challenges fields |
+| `ShardChunkHeaderInnerV5` | **New version** | Adds `proposed_split: Option<TrieSplit>` |
+| `BlockInfoV4` | **New version** | Adds `shard_split`, removes deprecated `slashed` map |
+| `ChunkExtraV5` | **New version** | Adds `proposed_split: Option<TrieSplit>` |
+
+#### ShardLayoutV3
+
+`ShardLayoutV2` (used by [Resharding V3][NEP-568]) stores only a single-generation split map: it records which parent shard was split into which children for the most recent resharding event. This is sufficient for manually scheduled, one-off resharding but breaks down with dynamic resharding, where repeated automatic splits create multi-generational parent-child chains.
+
+`ShardLayoutV3` replaces V2 with a layout that stores the **full cumulative split history**:
```
shreyan-gupta: Small qualification: "full cumulative split history" is slightly overclaimed. `build_shard_split_map` walks layouts via `windows(2)` and only consumes V2+ entries — V1 predates the parent-child split map. In practice this is fine (V1 layouts had no split relationships to track), but readers might wonder what the first V3 layout actually contains.

Maybe reword to "cumulative split history from V2 onward" with a brief note that V1 pre-dates parent-child tracking.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+A non-deterministic split algorithm would cause validators to disagree on the shard layout, leading to a chain fork.
+
+### Validation of Split Proposals
+
+Both chunk producers and block producers could attempt to forge their respective resharding fields to manipulate which shard gets split or the boundary account:
+
+1. **Chunk header forgery**: A chunk producer could embed a false `proposed_split` in the chunk header (e.g., proposing a split for a shard that doesn't meet the threshold, or proposing a wrong boundary account). This is prevented by chunk validators, who recompute `proposed_split` from the state witness and reject mismatches (`InvalidChunkHeaderShardSplit`).
+2. **Block header forgery**: A block producer could embed a false `shard_split` in the block header (e.g., selecting a different shard than the one with the highest memory usage, or ignoring a valid proposal). This is prevented by block validators, who recompute `shard_split` from the chunk headers in the block and reject mismatches (`InvalidBlockHeaderShardSplit`).
+
+### DoS via Threshold Manipulation
```
shreyan-gupta, line 404: One thing worth adding to make the cost model concrete: storage staking *locks* (does not burn) NEAR — an attacker who inflates state to trigger a split can release the stake after deletion. So the effective cost is the opportunity cost of the locked stake over the inflation period, not the headline GB stake price. Doesn't need a design change, just worth being explicit about in the threat model.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+- **Pre-loading is the dominant cost**: Because dynamic resharding can split any shard at any epoch boundary, a validator whose assignment changes such that it must track a child of a soon-to-be-split shard needs to pre-load the parent memtrie before the split executes. During this preparation window (epoch N+1, between the split decision in epoch N's last block and execution at the N+1 → N+2 boundary), the validator holds **two memtries simultaneously**: its currently-tracked shard plus the parent shard being prepared.
+
+In the pessimistic case, validators MUST be provisioned with enough RAM to hold the two largest shards' memtries concurrently. This is the fundamental constraint dynamic resharding adds compared to [Resharding V3][NEP-568], where the parent shard to be loaded was known weeks in advance via the protocol upgrade schedule. Operators SHOULD treat "2× the size of the largest current shard's memtrie" as the minimum RAM headroom available beyond their currently-tracked shard.
```
shreyan-gupta, line 427: The "2× the largest current shard's memtrie" minimum is right at first principles. Two small things worth adding:

1. Worst-case framing for the RAM picture. In the typical case left/right are roughly balanced and the 2× headroom is conservative. The worst-case profile is set by the first split candidate (shard 5), where one large account skews the split to ~8 GB / ~32 GB. A validator that picks up the larger child briefly holds parent (~40 GB) + child (~32 GB) ≈ 72 GB during the pre-load overlap. Mentioning the worst case explicitly would help operators size defensively without overstating the typical load.
2. Multi-validator OOM as a liveness concern. `spawn_background_memtrie_loading_for_shard` panics after 3 retries with no back-off (`core/store/src/trie/shard_tries.rs:575,605,633`). For chunk producers picking up new child shards, correlated OOM could threaten shard-level liveness. Validators tracking existing shards already run memtries and aren't affected — the "2×" provisioning advice is really for producers whose assignment changes (and for RPCs that load multiple memtries).

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+Note that a transient case of holding two memtries also exists today whenever a validator's assignment changes between epochs (independently of resharding). Dynamic resharding makes this case more frequent and less predictable.
+
+## Alternatives
```
shreyan-gupta, line 431: There's another real alternative worth surfacing in this section: a two-protocol-upgrade rollout where the first upgrade hardcodes a new ShardLayout (with full split history baked in) and the second upgrade enables the dynamic behavior on top of that already-V3 layout. This would sidestep the V2→V3 in-place migration via `derive_with_layout_history()`.

What was the rationale for picking single-upgrade-with-migration over this? Fewer protocol cycles? Simpler operational sequence? Worth surfacing for the design record in case the migration path turns out to be problematic.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+- Eliminates the operational burden of manual shard layout analysis and deployment.
+- Reduces reaction time from weeks (manual protocol upgrade) to ~14.5 hours (two epochs on mainnet).
+- Fully compatible with stateless validation -- split proposals are verifiable through state witnesses.
+- Builds directly on the proven Resharding V3 execution machinery.
+
+### Neutral
+
+- The split algorithm performs a greedy descent that is equivalent to binary search on a monotonic cumulative distribution, so it finds the optimal boundary. However, a perfectly balanced split is not always achievable due to inherent constraints (indivisible accounts, minimum boundary length).
+- Thread pool sizes are based on `max_number_of_shards` rather than the current shard count, which is slightly wasteful but safe.
+- Mainnet and testnet shard layouts will diverge, since splits are driven by actual state size which differs between the two networks. This is expected and by design, but means tooling and tests cannot assume a fixed shard layout for either network.
```
shreyan-gupta: The "tooling and tests cannot assume a fixed shard layout" callout is right but the ecosystem implications could be more concrete. Maybe expand to:

> "Mainnet and testnet shard layouts will diverge over time (different state sizes → different split candidates). External tooling SHOULD always query `shard_layout` per-epoch via the `EXPERIMENTAL_protocol_config` RPC. Indexers consuming per-shard arrays (e.g. `IndexerShard`) keyed by array position will break when new shards appear — consumers should key by `shard_id` instead of by index."

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+### Negative
+
+- Additional computational cost during chunk application for shards near the epoch boundary (running the split algorithm).
+- Increased complexity in block/chunk header formats (new versioned structures).
+- Nodes must handle dynamically changing shard counts, which may complicate tooling and monitoring.
+
+### Backwards Compatibility
+
+Dynamic resharding is backwards compatible with existing protocol logic through the `ShardLayoutConfig` enum. When `ShardLayoutConfig::Static` is configured, the system behaves identically to Resharding V3. The transition to `ShardLayoutConfig::Dynamic` occurs as part of a protocol version upgrade that enables the `DynamicResharding` feature flag.
```
shreyan-gupta, line 480: Small naming-collision warning for external clients. `ProtocolConfigView` already has a `dynamic_resharding: bool` field (`core/chain-configs/src/genesis_config.rs:806`) — it's the legacy genesis flag, and every callsite sets it to `false`, unrelated to this NEP's `ShardLayoutConfig::Dynamic`. RPC consumers reading the existing bool will get a misleading signal. Worth one sentence here pointing tooling at `ShardLayoutConfig` (via `EpochInfo`'s `shard_layout_config`) instead.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+The existing `ShardLayoutV2` used by Resharding V3 is automatically migrated to `ShardLayoutV3` when dynamic resharding is first enabled, via `ShardLayoutV3::derive_with_layout_history()`. Subsequent dynamic splits derive new V3 layouts from existing V3 layouts using `ShardLayoutV3::derive()`.
+
+External tooling that queries shard layouts via the `EXPERIMENTAL_protocol_config` RPC endpoint will continue to work. However, tools that cache shard layouts MUST re-query at epoch boundaries, as the layout may now change between epochs at the same protocol version.
```
shreyan-gupta, line 484: The protocol-upgrade story is scattered across Backwards Compatibility, Migration, Limiting the Blast Radius, and Reference Implementation. Could you consolidate into a dedicated "Protocol Upgrade Path" subsection? Bits to cover:

1. `DynamicResharding` feature flag (currently PV 150 in nightly; final PV at stabilization).
2. The dependency on `StickyReshardingValidatorAssignment` (PV 153 in nightly), and the intended release order — sticky before / with / after DR's PV.
3. The epoch-config JSON commit that must accompany DR enablement (`ShardLayoutConfig::Dynamic` with initial parameter values).
4. The one-time V2→V3 migration at the transitional epoch.
5. The kill-switch path (protocol upgrade only).

Having it all in one place would make the rollout story easier to evaluate.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
```diff
@@ -0,0 +1,482 @@
...
+
+## Unresolved Issues
+
+- **SPICE integration**: Multiple integration points between the SPICE feature and resharding transitions remain incomplete (chunk executor, data distributor, chunk validation, chunk application).
```
shreyan-gupta, line 490: Unresolved Issues currently lists only SPICE. Two other items belong here:

1. Sharded RPC × dynamic resharding (also targeting 2.13). The remote routing path at `chain/jsonrpc/src/sharded_rpc.rs:307-346` does a plain `node.tracked_shards.contains(&shard_id)` membership check — no descendant resolution. The local path correctly uses `check_if_descendant_of_tracked_shard` (`chain/epoch-manager/src/shard_tracker.rs:480`), but the remote path explicitly notes "static config and not epoch-dependent" in code. When a shard splits, remote nodes configured with the old shard id fail routing. Worth acknowledging since both features land in the same release.
2. Cloud archival integration. Two TODOs in `cloud_archival_writer.rs`. One-line scope statement: cloud archival is out of scope for this NEP and tracked separately.

🤖 Comment from Claude Code on behalf of Shreyan

shreyan-gupta (commented):
Thanks @Wiezzel. Mechanics look sound — validation closes the forging surface, the algorithm is deterministic, and the V3 reuse story is clear. Recommending with changes; nothing below should block voting.

Comments fall into roughly three buckets:

1. Things missing from the NEP that are in master — sticky validator assignment most notably, plus state-sync handling at boundaries.
2. Spots where the text could be tightened — three-phase `next_next_shard_layout`, kill-switch latency, pre-load timing.
3. Operator-facing details worth surfacing — trie cache config drift on RPC nodes, memtrie preload behavior, mainnet/testnet layout divergence.

Most of these are spec-text changes, not design changes. Detailed line-anchored comments below.

🤖 Comment from Claude Code on behalf of Shreyan

neps/nep-0639.md
Trisfald: I think both sharded RPC and cloud archival integration with dynamic resharding fall within the scope of these projects.

neps/nep-0639.md
Wiezzel: The thing is that technically StickyReshardingValidatorAssignment is an independent protocol feature. It is related to dynamic resharding, but not strictly a part of it. Could work as well with the current resharding mechanism. That's why I didn't include it in the NEP originally. The release strategy is to stabilize them together. I can mention it here though, if you believe it's important.

neps/nep-0639.md
Wiezzel: This is the config that I've used for forknet tests and should be considered a candidate for the actual initial mainnet config:

```json
  "dynamic_resharding_config": {
    "memory_usage_threshold": 40000000000,
    "min_child_memory_usage": 10000000000,
    "max_number_of_shards": 10,
    "min_epochs_between_resharding": 2,
    "force_split_shards": [],
    "block_split_shards": []
  },
```

`min_epochs_between_resharding` doesn't actually matter, as the configured number of shards will be `10` i.e. one more shard than the current mainnet layout. I see `2` as a reasonable value for the future though.

neps/nep-0639.md
Wiezzel: I only included SPICE because that project has been in progress well before work on dynamic resharding started. As @Trisfald mentioned, sharded RPC and cloud archival integration are aware of dynamic resharding, and handling this is already in the scope of those projects, so I don't feel mentioning it here is necessary.

neps/nep-0639.md
Wiezzel: I am not quite sure about this, @Trisfald WDYT?

neps/nep-0639.md
Wiezzel: Added a 'Sticky Validator Assignment Across Resharding' subsection.

neps/nep-0639.md
Wiezzel: Added a 'Sticky Validator Assignment Across Resharding' subsection.

neps/nep-0639.md
Wiezzel: Fixed.

neps/nep-0639.md
Wiezzel: Fixed.

neps/nep-0639.md
Wiezzel: Added the reference-implementation pointer.

neps/nep-0639.md
Wiezzel: Expanded with explicit preconditions for each phase.

neps/nep-0639.md
Wiezzel: Added.

neps/nep-0639.md
Wiezzel: Added a 'State Sync Across Resharding Boundaries' subsection.

neps/nep-0639.md
Wiezzel: Reworded to 'cumulative split history from V2 onward' with a note about V1.

neps/nep-0639.md
Wiezzel: Clarified that storage staking locks (not burns) NEAR, so the effective cost is the opportunity cost of the locked stake.

neps/nep-0639.md
Wiezzel: Expanded with concrete tooling guidance (query per-epoch via RPC, key by shard_id not array index).

neps/nep-0639.md
Wiezzel: Added a warning about the legacy `dynamic_resharding: bool` field.

neps/nep-0639.md
Wiezzel: Good catch — this edge case has been raised before. I tried fixing it by resetting `proposed_split = None` on the child clone, but it broke some tests, and on reflection requiring `min_epochs_between_resharding > 0` is a reasonable constraint anyway. The plan is to add the `> 0` constraint in the code; once that's merged the NEP will be updated to reflect it.

neps/nep-0639.md
Wiezzel: @shreyan-gupta 
> A validator that picks up the larger child briefly holds parent (~40 GB) + child (~32 GB) ≈ 72 GB during the pre-load overlap

I don't think there's a possible situation where a node would at the same time hold independently parent and child memtrie. The child memtrie after all is created by freezing the parent. So I don't quite see where this 72 GB number comes from.

neps/nep-0639.md
Wiezzel: Added a note on the first-match-by-Vec-order behavior for `force_split_shards`.

neps/nep-0639.md
Wiezzel: Added a 'Behavior on Forks' subsection covering provisional `shard_split` and transient pre-load activity on losing branches.

neps/nep-0639.md
Wiezzel: Updated the pre-loading section: corrected the wall-clock to up to ~30 minutes, added the `parallelize: false` rationale, and clarified that the pre-load bypasses memtrie config but the restart path is still gated by tracked shards.

neps/nep-0639.md
shreyan-gupta: This was from the worst case perspective if ALL key values of the child change, in which case the math works out to be 40 GB for the parent (frozen and unchanged) + 32 GB from the churn of the child trie nodes...

But yeah, that's worst worst case and numbers would likely be way way lower. Can skip if you want, just added this for correctness.

neps/nep-0639.md
shreyan-gupta: Sounds good! Just wanted to make sure if we needed these in or not. I guess it depends on the timings of which feature is released first.

neps/nep-0639.md
shreyan-gupta: Yeah, high level, I just thought that it's better to have the "one time" protocol upgrade related changes in a specific sub section for mental grouping. But your call.

Trisfald (approved):
As an SME I recommend approving this NEP. All comments have been addressed by the author.

### Benefits

- Eliminates manual protocol-upgrade cycles for resharding: reaction time drops from weeks to hours.
- Largely reuses the Resharding V3 implementation for shard splitting: already ran successfully on mainnet.
- Safeguards: the max number of shards is bounded and resharding events are separated by a minimum number of epochs.
- Negligible performance impact: the split-search algorithm runs in sub-millisecond time per shard on mainnet state.

### Concerns

- Less predictable RAM provisioning for node operators: with static resharding, splits were known weeks in advance via the protocol upgrade schedule and operators could provision around them. Dynamic resharding can split any shard at any epoch boundary, so the ~2× largest-shard memtrie headroom must be kept free at all times.
  - In my opinion this is an acceptable trade off.

shreyan-gupta (approved):
As an SME I also recommend approving this NEP. Author has been thorough and responsive in addressing review comments.

### Benefits

- Removes operational toil from resharding: no more hand-picked boundary accounts, hardcoded layouts, or coordinated binary releases. Layout changes happen at epoch boundaries with no human in the loop.
- Builds on a tested foundation: the V3 execution machinery (memtrie splitting, flat state resharding, stateless validation) has already shipped to mainnet. Dynamic resharding is the scheduling layer on top.
- Sticky validator assignment keeps producer-to-shard assignments stable across splits, avoiding a network-wide state-sync stampede on every layout change. The companion protocol feature will be stabilized and released together with dynamic resharding.

### Concerns

- The split boundary granularity is AccountId — when one large account dominates a shard, no balanced split exists. The dry-run on mainnet shard 5 produces a ~8 GB / ~32 GB split for this reason. This is a property of the design, not a bug.
  - In my opinion this is an acceptable tradeoff — sub-account splitting would create much harder problems downstream.
- Asymmetric scaling: shards can split but cannot merge. Transient or speculative growth in a shard's state produces permanent shard-count overhead that the protocol cannot reclaim on its own.
  - Acceptable for now. Worth treating merge support as a planned follow-up. Shard merging is a technically hard problem.

gagdiez:
I see that both @shreyan-gupta and @Trisfald have given their approvals, so we are ready to move to voting.

I would like to schedule a meeting with the working group members next week, so we can present the NEP to the community, and the @near/wg-protocol can vote.

@Wiezzel @shreyan-gupta @Trisfald please share your availability here: https://app.rallly.co/invite/gcq05xH98kKi

In these meetings we usually start with an overview of the NEP (would be amazing if you could make a short introduction @Wiezzel), and for reviewers to give present their opinions on benefits vs concerns.

neps/nep-0639.md
Wiezzel: A single protocol upgrade is preferable to a double upgrade. Apart from the rollout delay in production environments, double upgrade would significantly slow down forknet testing. Also, it would require checking the code carefully if there is no assumption made that a new layout always triggers resharding (an assumption that would be broken if we introduce a no-op layout change for the mere sake of adding history). `derive_with_layout_history` is well-tested, so I don't see any potential problems here.

neps/nep-0639.md
Wiezzel: That's quite unlikely indeed. It is mentioned earlier that
> We recommend operators provision an additional **10–20%** of the largest shard's memtrie size as headroom for this effect.

Which seems to cover the realistic case.

wacban:
@gagdiez Can I also get an invite please?

neps/nep-0639.md
Wiezzel: https://github.com/near/nearcore/pull/15746

gagdiez:
@wacban of course, please just make sure to fill the form with your availability

neps/nep-0639.md
Wiezzel: Done. https://github.com/near/nearcore/pull/15746 has been merged, enforcing `min_epochs_between_resharding > 0` via the `NonZeroEpochHeight` type. NEP updated to reflect the new field type and the safety rationale.

thisisjoshford:
As @gagdiez is going on leave shortly, I'll be taking over hosting this discussion & vote. Based on everyone's availability [here](https://app.rallly.co/invite/gcq05xH98kKi), I've setup a call for Tuesday May 26th, 2026 @ 4:30PM UTC.

Please register to attend here: https://luma.com/gnjw1ls3

cc @shreyan-gupta @Trisfald @Wiezzel @shreyan-gupta @Trisfald @wacban

wacban:
@gagdiez @thisisjoshford Can you share the slides with @Wiezzel so that he can add his part to it please?

thisisjoshford:
> @gagdiez @thisisjoshford Can you share the slides with @Wiezzel so that he can add his part to it please?

Hey @wacban! I don't have any presentation slides (aside from just a basic intro to the meeting) and planning on handing of the presentation for this NEP to @Wiezzel as @gagdiez has done in the past. However, happy to run slides for him if that's what he prefers? If so, shoot them over to josh@near.dev :) Otherwise, just come prepared with your material and the floor will be yours.

birchmd (approved):
As a protocol working group member, I approve this NEP. It's exciting to see this step forward! Dynamic resharding has been on the Near roadmap since day 1 and I am happy we continue to make meaningful progress towards it.

One note I want to make about the sticky validator assignment: it only applies to chunk producers. This is important because regularly shuffling stateless validators is one of the security assumptions for the network (it is needed to make it difficult for an adversary to adaptively corrupt validators to sneak through an invalid state transition in one shard).

wacban:
@thisisjoshford Can we merge it now?

thisisjoshford:
> @thisisjoshford Can we merge it now?

Yes sir! NEP has been reviewed, discussed, and approved! :) 
See recording: https://www.youtube.com/watch?v=zJgyK7FK0Kg

thisisjoshford (approved):

thisisjoshford:
@Wiezzel could you resolve the broken link errors before merging?

gagdiez:
@thisisjoshford link errors come form other NEPs, so we should be good to merge this

gagdiez (approved):

PR #641, opened 2026-06-02
NEP-641: Offchain Authorizations for Smart Contracts

frol:
[Rendered NEP Markdown](https://github.com/near/NEPs/blob/nep-auth-resolve/neps/nep-0641.md)

It is open for discussion and we collaborate with DefuseLabs team on this.

---

For painting the picture of how this all fits together, you can go to https://trezu.app and use EVM wallets sign in flow (the implementation there is based on the older draft of the NEP, but I still find it helpful to paint the overall picture).

Sign in to Trezu requires "sign message" from the account owner, but EVM wallet is represented by a wallet contract on NEAR, so it cannot use NEAR's NEP-413 to sign message, as there are no access keys.

Trezu uses the signed message to issue JWT token to access the backend APIs (e.g. get the address book and history of confidential transactions).


neps/nep-0641.md
```diff
@@ -0,0 +1,763 @@
...
+<summary><strong>Opaque authorization blob</strong></summary>
+
+Each wallet contract defines its own internal format for the authorization blob: an EIP-712 envelope for an Ethereum-signing wallet, a WebAuthn assertion for a passkey wallet, a Borsh-encoded bundle of member authorizations for a multisig, and so on. This NEP intentionally does not standardize the blob format.
+
+Standardizing across all wallet types would require either lowest-common-denominator constraints (limiting innovation) or a fat schema (unmanageable). Companion NEPs may standardize specific blob formats for common schemes (EIP-712, WebAuthn, etc.) without modifying this NEP.
+
+</details>
+
+<details>
+<summary><strong>Two purposes: <code>PROVE_OWNERSHIP</code> and <code>APPROVE_OFFCHAIN_ACTION</code></strong></summary>
```
frol: I have already received a piece of feedback that the purpose might be unnecessary:

My original idea was to allow multisig members to sign in as DAO and for that one won't need to collect M-of-N signatures from other members, while for other actions you definitely want to require M-of-N consensus. For that, dApp should be able to signal to the wallet which purpose it is, and then the w_auth_resolve function to validate the provided authorization against the purpose.

Thus, the multisig contract could have defined two additional kinds of actions: "prove_ownership" and "approve_offchain_action", and then verify the policy requirements (1-of-N or M-of-N, etc) in the w_auth_resolve function.

However, multisig like SputnikDAO has quite a few kinds of actions (change policy, create proposal, approve proposal, etc), and it seems that it could be quite limiting to have only two kinds of actions. Still, the "function call" proposal is also quite generic, and does not allow bringing further granularity in the policy enforcement.

That said, for wallet contracts, where there is a single signer, PROVE_OWNERSHIP and APPROVE_OFFCHAIN_ACTION are redundant, and the purpose field only brings semantic difference that could be communicated to the user.

I still believe there is value in the purpose field even if it feels redundant in some cases and limiting on others.

There is a new emerging concept for SputnikDAO - automated members (steward contracts, that cast their vote on ping if the proposal matches their internal auto-approval policy). For those cases, w_auth_resolve("PROVE_OWNERSHIP") would always return INVALID, while for w_auth_resolve("APPROVE_OFFCHAIN_ACTION") it would depend on the action - if the authorization would have the action that is in line with the auto-approval policy, then the contract would return RESOLVED.

frol:
The use-cases that this NEP would enable are:

1. Sign in into a dApp with a proof of account ownership - e.g. Trezu allows to view the private address book (stored on Trezu backend) - PROVE_OWNERSHIP purpose
2. Approve confidential swap intents with passkey/metamask/... - sign off-chain message and submit it to the confidential API - APPROVE_OFFCHAIN_ACTION purpose

neps/nep-0641.md
```diff
@@ -0,0 +1,768 @@
...
+Today, dApps integrating with such accounts must hand-roll per-contract verification logic off-chain, duplicating the contract's authorization rules. This is fragile (contract upgrades break dApps), insecure (off-chain duplication invites divergence), and does not compose (a multisig of mixed wallet types cannot be verified by any single library).
+
+This NEP makes the **on-chain contract the single source of truth** for whether an off-chain authorization is valid for its account, accessible via a uniform view-function interface. Resolution is recursive, so heterogeneous compositions resolve correctly through generic caller-side logic.
+
+## Design Considerations
+
+The following considerations shaped the design. Each is documented so future revisions and implementers can see the trade-offs.
+
+<details>
+<summary><strong>Verification as a view function on the contract</strong></summary>
```
frol: The client-side SDK for verification could even have an implementation that would run the wasm contract locally on the client side, especially for well-known wallet contracts.

Here is a Node.js implementation for view-function calls: https://github.com/vgrichina/fast-near

frol:
This NEP is already implemented in the v0 of wallet contract for EIP-712 (Ethereum signMessage): https://github.com/near/intents/pull/256 and on the client side it is integrated into NEAR Connect: https://github.com/azbang/near-connect/pull/96, and it is live on https://trezu.org

frol:
One more consideration that I want to bring up - will `w_resolve_auth` be compatible with NEAR MPC? I need to model it before concluding this NEP.

frol:
This NEP is already implemented for Wallet Contracts and they are live in production on Trezu:
* https://github.com/near/intents/pull/319 (up-to-date implementation)
* https://github.com/near/intents/pull/256 (slightly outdated, so needs to be updated)

Thus, this NEP is now ready for review.

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+2. **Offchain action approval**: authorizing an MPC signing request, a confidential swap intent, or any other offchain action.
+
+Both reduce to one question — "does this account authorize this exact payload?" — which is what `w_resolve_auth` answers. What the payload *means* is the dApp's business, encoded in the payload itself (see [`JsonPayload`](#domain-separated-json-payload-recommended)).
+
+## Design Decisions
+
+<details>
+<summary><strong>Verification as a view function</strong></summary>
+
+The governing contract is the natural authority on what authorizes its account. A view function makes that rule authoritative (no offchain library to keep in sync), composable (one call verifies any wallet type), and free (no transaction, no gas). Cost: one RPC round-trip per node of the authorization graph — one call for the common single-account case. A verifier can even skip RPC entirely by executing the contract WASM locally against fetched state (see [`fast-near`](https://github.com/vgrichina/fast-near)).
```
mitinarseny, line 49: nit:
```suggestion
The governing contract is the natural authority on what authorizes its account. A view function makes that rule authoritative (no offchain library to keep in sync), composable (one call verifies any wallet type), and free (no transaction, no gas). Cost: one RPC round-trip per node of the authorization tree — one call for the common single-account case. A verifier can even skip RPC entirely by executing the contract WASM locally against fetched state (see [`fast-near`](https://github.com/vgrichina/fast-near)).
```

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+While the authorization blob is opaque, in case of Wallet contracts and Multisigs, where user signature is used to prove account ownership, adequate security measures MUST be taken to prevent replay attacks. Thus, this NEP requires the signer to include and sign the same five fields (`chain_id`, `signer_id`, `path`, `timestamp`, `payload`).
+
+Non-signature resolvers have nothing to sign and are exempt: their authority derives from chain state alone.
+
+</details>
+
+<details>
+<summary><strong>Caller-side recursion and the <code>path</code></strong></summary>
+
+A contract that cannot attest alone (multisig, extension wallet) returns its `payload` plus `pending` sub-authorizations, each with the sub-blob already **extracted** from its bundle — so the caller walks the graph without understanding any bundle format.
```
mitinarseny, line 83: nit:
```suggestion
A contract that cannot attest alone (e.g. multisig, extension wallet) returns its `payload` plus `pending` sub-authorizations, each with the sub-blob already **extracted** from its bundle (or state) — so the caller walks the graph without understanding any bundle format.
```

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+) -> AuthorizationResolution;
+```
+
+- `path` — traversal path from this contract's direct parent up to the top-level resolver; empty at top level. See [Path](#path).
+- `authorization` — opaque, resolver-defined blob. See [Authorization blob](#authorization-blob).
+
+Requirements:
+
+- MUST be a view function: no state mutation, callable via view-call RPC.
+- MUST be deterministic over its arguments and chain state.
```
mitinarseny, line 160: IIUC, there is no way a contract can return "non-deteministic" authorized payload if called with fixed arguments on a fixed block hash, because it has no other state to consider, no?

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+
+Example — `chain_id = "mainnet"`, `signer_id = "extension.near"`, `path = ["wallet.near", "v1.signer"]`, `timestamp = 2026-08-05T07:28:00.123456789Z`:
+
+```text
+recipient = "mainnet: extension.near -> wallet.near -> v1.signer @ 2026-08-05T07:28:00Z"
+```
+
+NEP-413 wallets render only `message` and `recipient` (`nonce` is an opaque hash), so this `recipient` puts the network, account, delegation path, and signing time in front of the user. Truncation to seconds is cosmetic; full precision stays bound through the `nonce`.
+
+Both `ed25519` and `secp256k1` are supported; `access_key` and `signature` curves MUST match. Existing NEP-413 wallets need no modification.
```
mitinarseny, line 342: Can you please indicate that any digital signature algorithm can be supported this way, including ML-DSA-65?

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+
+1. `msg.chain_id` equals the resolver's chain ID.
+2. `msg.signer_id` equals `account_id`.
+3. `msg.path` equals the current resolution [path](#path).
+4. The signature verifies per `via` against `access_key`.
+5. `msg.timestamp` is not later than the block's timestamp.
+6. `access_key` on `account_id` at the block:
+   - **key exists** → accept only with `FullAccess` permission (incl. gas-key full-access); reject function-call keys;
+   - **account exists, key absent** → reject — even for an implicit-derived ID, the owner may have removed the key deliberately;
+   - **account doesn't exist** → accept only if `account_id` is the implicit account derived from `access_key` (hex ed25519, or `0x` + Keccak for secp256k1) — it can be claimed under this key at any time.
```
mitinarseny, line 356: Technically, Universal Implicit Account IDs would be derived from UniversalStateInit, so we should check if there is a known StateInit having this `access_key` that derives to `account_id`.
```suggestion
   - **account doesn't exist or not initialized yet** → accept only if `account_id` is the implicit account derived from `access_key` (hex ed25519, or `0x` + Keccak for secp256k1) — it can be claimed under this key at any time.
```

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+
+### Pending sub-authorizations
+
+- Empty `pending` = leaf: the contract alone authorizes `payload`.
+- Non-empty `pending`: `payload` is authorized **iff** every entry, resolved recursively with the [extended path](#path), returns exactly its `expect`. Any failure or mismatch MUST fail the whole verification immediately.
+- The contract MUST extract each sub-blob from its own `authorization`; the caller passes blobs on opaquely.
+
+### Block consistency
+
+The resolver MUST resolve the whole graph against one block hash, SHOULD default to a recent *final* block, and SHOULD check each RPC response refers to the requested block. Historical resolution is possible (e.g. audits), but the timestamp rule keeps freshly-signed authorizations unresolvable against blocks preceding them.
```
mitinarseny, line 371: nit:
```suggestion
The resolver MUST resolve the authorization tree against one block hash, SHOULD default to a recent *final* block, and SHOULD check each RPC response refers to the requested block. Historical resolution is possible (e.g. audits), but the timestamp rule keeps freshly-signed authorizations unresolvable against blocks preceding them.
```

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+    c = try view_call(account, "w_resolve_auth",
+                      { "path": path, "authorization": auth },
+                      at_block = block)
+
+    if a succeeded: return a        # full-access key takes precedence
+    if c succeeded: return c
+    fail(a.error if a failed access-key verification else c.error)
+```
+
+Sibling pending entries MAY resolve concurrently. Finally, the dApp MUST check the returned payload equals the one it issued.
```
mitinarseny, line 429: > the dApp MUST check the returned payload equals the one it issued.

Sometimes, dApp / protocol do not "issue" any payload to sign - it can be constructed on client side, and offchain verifier might only need to deserialize and use it according to its own schema (e.g. `JsonPayload` with specific `domain`, `action`, and `payload` of specific format (can itself be JSON, too)).

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+
+    rect rgba(127, 127, 127, 0.15)
+    Note over FE,WC: Phase 3 — Verification
+    FE->>BE: POST /auth/verify<br/>{ account_id, authorization }
+    BE->>RPC: fetch latest final block
+    RPC-->>BE: block_hash
+    BE->>RPC: view_call("0s1234...", "w_resolve_auth",<br/>{ path: [], authorization },<br/>at_block = block_hash)
+    RPC->>WC: w_resolve_auth([], authorization)
+    WC->>WC: decode blob<br/>check chain_id, signer_id,<br/>path (= []), timestamp<br/>verify signature over msg
+    WC-->>RPC: { "payload": "<JsonPayload>" }
```
mitinarseny, line 472: This renders as `{ "payload": "" }` on GitHub for some reason

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+    RPC-->>BE: { payload, pending }
+
+    Note over BE: prepend parent: path = ["master.near"]<br/>recurse into extension.near
+    BE->>RPC: view_call(extension.near, w_resolve_auth,<br/>{ path: ["master.near"],<br/>authorization: inner_blob })
+    RPC->>E: w_resolve_auth(["master.near"], inner_blob)
+    E->>E: check chain_id, signer_id,<br/>signed path == arg path,<br/>timestamp not in future<br/>verify signature
+    E-->>RPC: { payload }
+    RPC-->>BE: { payload }
+
+    Note over BE: payload == expect ✓<br/>no more pending → done<br/>dApp checks payload == issued payload
```
mitinarseny, line 581: This note can look a bit misleading as it sounds like dApp checks the payload returned from the `extension.near`, not from `master.near`

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+- **Per-contract blob formats**: signing clients need format-specific encoders per wallet type. Unavoidable given wallet heterogeneity.
+- **Latency**: nested delegation costs one RPC round-trip per level (siblings parallelize; parent/child cannot). Depth rarely exceeds 2.
+- **Borsh + SHA3 in clients**: the canonical hash requires both in every signing client, including non-NEAR-native ones.
+- **Uninitialized deterministic accounts**: a NEP-616 account existing only as a `StateInit` derivation cannot answer view calls, so its contract authorizations cannot resolve yet (access-key authorizations for implicit accounts are exempt). Needs RPC support for view calls against pre-initialized state.
+
+## Future Possibilities
+
+- **Pre-init view calls**: RPC execution against a supplied `StateInit` (NEP-616), enabling resolution for not-yet-deployed deterministic accounts.
+- **More `AccessKeySchema` variants**, including post-quantum (e.g. ML-DSA).
+- **Local WASM execution** of `w_resolve_auth` in verifier SDKs (e.g. [`fast-near`](https://github.com/vgrichina/fast-near)) — no per-node RPC.
```
mitinarseny, line 654: Note that WASM bytecode currently deployed to the account is actually a part of its on-chain state and cannot be "cached" unconditionally.

mitinarseny (commented):
Awesome! This NEP formalizes a unified and generic way for smart-contracts to resolve off-chain authorizations, paving a way for truly off-chain interactions, where keeping governance on-chain.

neps/nep-0641.md
frol: @mitinarseny Yep, as of today, it cannot be non-deterministic, but to set the expectations explicitly, I believe this statement won't hurt.

neps/nep-0641.md
frol: Good catch. I fixed a few other problems there as well. Unfortunately, new lines are ignored by GitHub renderer :-(

neps/nep-0641.md
```diff
@@ -0,0 +1,658 @@
...
+    c = try view_call(account, "w_resolve_auth",
+                      { "path": path, "authorization": auth },
+                      at_block = block)
+
+    if a succeeded: return a        # full-access key takes precedence
+    if c succeeded: return c
+    fail(a.error if a failed access-key verification else c.error)
+```
+
+Sibling pending entries MAY resolve concurrently. Finally, the dApp MUST check the returned payload by either checking that it equals the one it issued or validating the payload (domain to match the dApp, action against user permissions, and nonce material preventing replays).
```
mitinarseny: nit: not every dApp/protocol has/needs a concept of "user permissions" or "nonces"
```suggestion
Sibling pending entries MAY resolve concurrently. Finally, the dApp MUST check the returned payload by either checking that it equals the one it issued or validating the payload (e.g. domain to match the dApp, action against user permissions, and nonce material preventing replays).
```

mitinarseny (approved):

frol:
As a Contract Standards working group member, I have a conflict of interest here, so I requested @mitinarseny to be the SME reviewer, and @graphite and @birchmd to act as working group approvers.

graphite (approved):
Thank you for the great NEP! between @frol's original work and @mitinarseny's review I literally have nothing to add here.

birchmd (approved):
I approve this NEP. Thanks for the great work @frol and @mitinarseny !

Defining an interface for off-chain services to authenticate users via their existing on-chain account abstractions makes it easier to develop such flows. And I like the design choices that are made in this proposal (e.g. opaque blobs, but with specific instances for common cases such as signatures).

The only downside from my perspective is that users will simply need to trust the off-chain service is implementing the standard correctly because there will not be any on-chain record of the authorization. But that is the nature of off-chain services, not really a flaw in the proposal itself.

PR #642, opened 2026-06-12
NEP-642: Account cost increase

jancionear:
This proposal makes creating a new account on NEAR roughly 10x more expensive than it is today. The cost of other operations remains unchanged.

This is needed to bring the cost of storage taken up by the account in line with the cost of storage paid for with storage deposits. Currently the storage taken up by new accounts is cheaper than storage purchased with storage deposits, which makes for an unfair balance.

The cost increase will be achieved by increasing the upfront gas purchase price 10x. For creating an account we will charge the full 10x gas price. For other operations we will charge a lower price and refund the difference, so that the final cost remains unchanged.

### DIFF
--- a/neps/nep-0642.md
+++ b/neps/nep-0642.md
@@ -1,0 +1,336 @@
+---
+NEP: 642
+Title: Account cost increase
+Authors: Jan Malinowski <jan.ciolek@nearone.org>
+Status: New
+DiscussionsTo: https://github.com/nearprotocol/neps/pull/642
+Type: Protocol
+Version: 1.0.0
+Created: 2026-06-08
+LastUpdated: 2026-06-08
+---
+
+<!-- NEP Header Preamble
+
+Each NEP must begin with an RFC 822 style header preamble. The headers must appear in the following order:
+
+NEP: The NEP title in no more than 4-5 words.
+
+Title: NEP title
+
+Author: List of author name(s) and optional contact info. Examples FirstName LastName <satoshi@fakenews.org>, FirstName LastName (@GitHubUserName)>
+
+Status: The NEP status -- New | Approved | Deprecated.
+
+DiscussionsTo (Optional): URL of current canonical discussion thread, e.g. GitHub Pull Request link.
+
+Type: The NEP type -- Protocol | Contract Standard | Wallet Standard | DevTools Standard.
+
+Requires (Optional): NEPs may have a Requires header, indicating the NEP numbers that this NEP depends on.
+
+Replaces (Optional): A newer NEP marked with a SupercededBy header must have a Replaces header containing the number of the NEP that it rendered obsolete.
+
+SupersededBy (Optional): NEPs may also have a SupersededBy header indicating that a NEP has been rendered obsolete by a later document; the value is the number of the NEP that replaces the current document.
+
+Version: The version number. A new NEP should start with 1.0.0, and future NEP Extensions must follow Semantic Versioning.
+
+Created: The Created header records the date that the NEP was assigned a number, should be in ISO 8601 yyyy-mm-dd format, e.g. 2022-12-31.
+
+LastUpdated: The Created header records the date that the NEP was assigned a number, should be in ISO 8601 yyyy-mm-dd format, e.g. 2022-12-31.
+
+See example above -->
+
+## Summary
+
+This proposal makes creating a new account on NEAR roughly 10x more expensive than it is today. The cost of other operations remains unchanged.
+
+This is needed to bring the cost of storage taken up by the account in line with the cost of storage paid for with storage deposits. Currently the storage taken up by new accounts is cheaper than storage purchased with storage deposits, which makes for an unfair balance.
+
+The cost increase will be achieved by increasing the upfront gas purchase price 10x. For creating an account we will charge the full 10x gas price. For other operations we will charge a lower price and refund the difference, so that the final cost remains unchanged.
+
+## Motivation
+
+Every account that exists on NEAR takes up permanent state that every node tracking the shard must store indefinitely. NEAR already accounts for the ongoing cost of state through [storage staking](https://docs.near.org/protocol/storage/storage-staking): an account must lock NEAR proportional to the bytes it occupies (currently `1 NEAR` per `100 KB`). A single account can take up to 770 bytes of storage without paying any storage deposits (zero balance account), which should cost `0.0077 NEAR` at the price used in storage deposits.
+
+The fee charged for creating that account, however, has been disconnected from this cost. The `create_account` action fee is ~7.7 Tgas, which at the network minimum gas price (`0.0001 NEAR/TGas`) costs `0.00077 NEAR`. Because the network almost always runs at the minimum gas price, account creation has effectively been an order of magnitude cheaper than the normal storage cost. This pricing imbalance is not ideal, storing data should cost the same, no matter where the bytes come from.
+
+## Specification
+
+The general idea is that the gas will be purchased at a price that is 10x higher than the current minimum gas price. For operations that don't create an account the runtime will charge the normal gas price and refund the difference, so the final cost remains unchanged. For operations that create an account the runtime will charge more to cover the proper cost of creating the account.
+
+All behavior described here is gated behind a new protocol feature and only takes effect from the protocol version in which it is activated.
+
+### New runtime parameters
+
+The proposal adds two runtime configuration parameters and changes how gas attached to a receipt is priced and refunded. 
+
+| Parameter | Value | Meaning |
+| :-- | :-- | :-- |
+| `min_gas_purchase_price` | `0.001 NEAR/TGas` | Minimum price at which the gas attached to a receipt is *purchased*. This is 10x the network minimum gas price of `0.0001 NEAR/TGas`. |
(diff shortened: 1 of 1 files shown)


walnut-the-cat:
Thank you @jancionear for submitting this NEP.

As a moderator, I reviewed this NEP and it meets the proposed template guidelines. I am moving this NEP to the REVIEW stage and suggest @jakmeier and @mitinarseny as SME reviewers to complete a technical review (see expectations). @near/wg-protocol working group members, please let us know if you have better suggestions. 

<details>

</details>

jakmeier (commented):
# SME report by @jakmeier

I recommend approving this proposal.

## Summary of benefits

- This change allows setting the cost for long-term storage related to account creation to a NEAR denominated value. Without the proposal, such a cost change would disrupt existing on-chain applications.
- The new cost of creating a new account is a minimum of 0.007 NEAR, regardless of how the account is created. (Burning the cost for ZBA creation vs. reserving the fee on the account for storage staking.) This aligns the paid cost with the fair cost as considered by the protocol team.
- Potential future changes to the cost of storage will be easier to adopt.


## Summary of concerns and blockers

| Concern | Status |
| ------------ | --------- | 
| The proposal **undoes the gas pricing model simplification** achieved by removing pessimistic gas pricing. With this proposal, accounts will need **higher minimal balances** and access keys need **higher allowances** to keep functioning as they do today. | Dismissed: Discussions with dApp developers showed that this is not a big concern. They are able to increase the balances and allowances where necessary. |
| The proposal **undoes  the reduction in refund receipts** achieved by removing pessimistic gas pricing. | Dismissed: The overhead of additional receipts is comparatively low and future possibilities to address this issue are listed in the proposal. |
| Increasing account creation gas parameters fees can lead to more expensive transfers to implicit accounts. Today this always includes the ZBA gas fee even when the target account already exists. | Resolved: The implementation resolves this with two tricks. (1) Shift the bulk of the gas fee from SEND to EXEC and (2) only apply the extra cost when the account didn't exist at the start of the receipt execution. |

Note that we discussed these concerns outside what is documented here on GitHub. I raised them early in the design phase of the proposal and the team worked together to resolve the issues ahead of time. My thanks go to everyone listed as contributor, with special thanks to @jancionear and @Trisfald who put in the most work to resolve all known issues.

neps/nep-0642.md
```diff
@@ -0,0 +1,354 @@
...
+
+All behavior described here is gated behind a new protocol feature and only takes effect from the protocol version in which it is activated.
+
+### New runtime parameters
+
+The proposal adds two runtime configuration parameters and changes how gas attached to a receipt is priced and refunded. 
+
+| Parameter | Value | Meaning |
+| :-- | :-- | :-- |
+| `min_gas_purchase_price` | `0.001 NEAR/TGas` | Minimum price at which the gas attached to a receipt is *purchased*. This is 10x the network minimum gas price of `0.0001 NEAR/TGas`. |
```
mitinarseny, line 69: Do we really need an additional config parameter here or can it be derived from already existing `storage_amount_per_byte` somehow?

neps/nep-0642.md
```diff
@@ -0,0 +1,354 @@
...
+burned_cost         = gas_burn_price * create_account_gas           # already burned during execution
+amount_to_charge    = max(0, desired_cost - burned_cost)            # top-up taken from the surplus
+```
+
+`amount_to_charge` is then subtracted from the refund and burned. The total cost of creating an account is therefore `max(desired_cost, burned_cost)`:
+
+* At low (typical) gas prices, `burned_cost` is small, the top-up brings the total up to the fixed `0.007 NEAR`.
+* At high gas prices, the naturally-burned gas already exceeds `0.007 NEAR`, no top-up is taken, and the user simply pays the (higher) gas cost.
+
+This applies regardless of *how* the account is created - an explicit `CreateAccount` action, implicit account creation by transferring to a not-yet-existing implicit account, a `DeterministicStateInit` action, or an account created by a contract through a cross-contract call. Because the receipt that creates the account is not known to be account-creating at the time its gas is purchased, all receipts must purchase gas at `min_gas_purchase_price`; the charge is only taken from the surplus of receipts that actually create an account.
```
mitinarseny, line 123: Maybe it's worth to mention here that despite Implicit AccountIds are created on receiving a first native transfer, but there are following exceptions:
* refunds of gas or total attached deposit to a failed receipt
* non-existing `beneficiary_id` from `DeleteAccount` action

Also, can there be a case when a not-yet-existing implicit account is created on execution of the first [`DelegateAction`](https://github.com/near/NEPs/blob/master/neps/nep-0366.md#delegateaction) signed by account's "implicit access key"?

mitinarseny (approved):
Hey @jancionear, thank you for working on this NEP!

Taking into account all the nuances around gas & storage management, it seems that the proposed solution removes the cost imbalance in an efficient way without breaking backwards compatibility and other significant tradeoffs.

Left couple of minor comments but, in general, looks good to me 👍🏻

neps/nep-0642.md
jancionear: Using `storage_amount_per_byte` would be nice, but then the charge would be 0.0077 NEAR (770 bytes), while we have only 7.2 TGas to burn, which gives us 0.0072 NEAR. It was easier to define an explicit charge for each created account. Orr we could raise gas price even higher, but I wanted to avoid doing that because it negatively affects users.

neps/nep-0642.md
jancionear: Good point, I'll add a section about them.

> Also, can there be a case when a not-yet-existing implicit account is created on execution of the first [DelegateAction](https://github.com/near/NEPs/blob/master/neps/nep-0366.md#delegateaction) signed by account's "implicit access key"?

I think this can't happen, a delegate action fails if the transaction author doesn't exist (see `check_account_existence` in `apply_action`)

thisisjoshford:
Looks like we're ready to schedule a meeting to present this NEP to the community and have @near/wg-protocol vote.

@jancionear @walnut-the-cat @mitinarseny @jakmeier - [please share your availability here](https://app.rallly.co/invite/IXHzgiKQwe60)

birchmd:
As a protocol working member I lean towards approving this proposal. As the NEP notes, it is important that storage costs are aligned, regardless of how the bytes which need to be stored are created. It is unfortunate that we are effectively seeing the return of pessimistic gas pricing (undoing the gains of NEP-536), but I agree with the authors of the proposal that maintaining backwards compatibility (thus ruling out the possibility of increasing the cost of account creation directly) is more important. Hopefully future work on how users pay for storage can make progress again.

thisisjoshford:
> Looks like we're ready to schedule a meeting to present this NEP to the community and have @near/wg-protocol vote.
> 
> @jancionear @walnut-the-cat @mitinarseny @jakmeier - [please share your availability here](https://app.rallly.co/invite/IXHzgiKQwe60)

Friendly bump on this @mitinarseny :)

mfornet:
As PWG member I also lean toward approving this PR. I agree that we should change equally everywhere in the protocol for storage, and the change for end users (even if it is 10x) should no have a big impact with regard to price. I guess this will affect more applications onboarding millions of users, but still considering the expected value a user will bring to the platform absorbs the cost of the price increase.

thisisjoshford:
Based on everyone's availability [here](https://app.rallly.co/invite/IXHzgiKQwe60), I've setup a call for Friday July 3rd, 2026 @ 4:00PM UTC.

@jancionear  please prepare a short overview for this NEP to get us started and then we will follow up with SME's review. Please register to attend here: https://luma.com/xatywncd

cc @jancionear @walnut-the-cat @mitinarseny @jakmeier @birchmd @mfornet @gagdiez

thisisjoshford:
@jancionear @mitinarseny @jakmeier can you [update your availability here](https://app.rallly.co/invite/IXHzgiKQwe60). Unfortunately, one of the SME's had a schedule change for this Friday and is no longer able to attend. Pushing to next week dependent on everyone's availability.

thisisjoshford:
> @jancionear @mitinarseny @jakmeier can you [update your availability here](https://app.rallly.co/invite/IXHzgiKQwe60). Unfortunately, one of the SME's had a schedule change for this Friday and is no longer able to attend. Pushing to next week dependent on everyone's availability.

@mitinarseny please add your availability so we can get this meeting rescheduled.

gagdiez:
As the moderator, I would like to thank @jancionear for submitting this NEP, and for the group members for your review.

Based on the discussion it seems like this proposal is approved we are therefore scheduling a group meeting next week. Everyone is welcome to join.

Meeting Info
📅 Date: Wednesday July 8, 1pm UTC
✏️ [Register](https://luma.com/wvtpjqbm)

cc @jancionear @mitinarseny @jakmeier

gagdiez:
Thank you to everyone who attended the Contracts Standard meeting on Wednesday! The working group members reviewed the NEP and reached the following consensus:

Status: Approved

https://github.com/near/NEPs/pull/642#issuecomment-4781033600
https://github.com/near/NEPs/pull/642#issuecomment-4798193268

Meeting Recording: https://youtu.be/zdDVwdG3yzw

@jancionear thanks for authoring this NEP!

Next Steps:

If anyone is interested in revisiting this NEP, please submit a new one.

gagdiez (approved):

PR #649, opened 2026-07-30
NEP-649: Deterministic ML-DSA-65 Key Derivation from BIP-39 Seeds (SLIP-0010)

vsavchyn-dev:
# ML-DSA-65 key derivation for wallets

NEP text: https://github.com/vsavchyn-dev/NEPs/blob/feat/mldsa-derivation/neps/nep-0649.md
Reference implementation: https://github.com/vsavchyn-dev/NEPs/blob/feat/mldsa-derivation/neps/nep-0649.md#reference-implementation


neps/nep-0649.md
```diff
@@ -0,0 +1,967 @@
...
+
+With the introduction of [NEP-645], users got access to PQ transaction signatures on NEAR Protocol, allowing them to protect their accounts against a future quantum computer threat by adding an ML-DSA-65 access key. Meanwhile, self-custodial HD wallets, which serve as the main point of entry for users on NEAR, do not have a specification to follow on how to derive those keys deterministically.
+
+[SLIP-10] specification does not yet cover ML-DSA and has an open PR ([slips#1968]) that proposes adding it.
+
+This NEP specifies the standard on how wallets should deterministically derive ML-DSA-65 key pairs from a BIP-39 mnemonic using [SLIP-10] hardened-only derivation, aligned with the construction proposed in [slips#1968].
+
+## Motivation
+
+NEAR Protocol is one of the early adopters of post-quantum keys, and users need a way to deterministically derive their ML-DSA-65 access keys, similar to how it is done with Ed25519 keys specified in [SLIP-10]. This has several applications, for instance enabling recovery of the key from a [BIP-39] mnemonic alone, and cross-wallet portability.
```
r-near: We should explain a bit why hardened-only is forced (ML-DSA rounds its public key so there's no parent/child homomorphism and xpub can never exist)

r-near: Also we should cite the paper: https://eprint.iacr.org/2026/380 since that explains it

neps/nep-0649.md
```diff
@@ -0,0 +1,967 @@
...
...71f3ae5c32413a641589b8a1004063fb3d866ef45610b486b4cd3cc68a7ee2a92ff7d16cdb497ffee7fd0beed3fde1edfdb7cfa7aa859a2f8d5177eb1c9ff80db41f71bfdab80bb77c6484a6dbc2b030199461d85caa3e9ded4462990720a2268331e17742db2b2190f03f6883a398c5d52d84aad912957fb733e262bfda90ca32d48d5b05c6f44f66a38791ecf34f6f6eb76635873ed7836e1c8441b781b7bd069095b0bf588e70a1aa55e9b2614efd33c4cec6c5ae95c27ef082ca5faaafa65e80727c8f4cffa6efc1413d7878e3140bbaa876985c0cafb0da6af9a04f63666b5f69dde021ca549c52853f03d4c0fbd0c6a6026e8eaee337c11f5026143264fbe5d7e9de25ca912adabf25f0072a3ea58dcab6860b145d6591b1a852b695247077be39144e2dc7a86f00784565204e241f09670517b8f30169810ac859eedf2988f831a9a5332361162e1f1d45e9592067c30bfc6597df6bb3caf9da4a5d4553f3dd2ddc99526d5027da114967b0c1807b126220498d4226e8002b247f5efbded7ed69f07c48f80e048a46411f34bdb2c650421dd0f2738ea8b7340650158d7fdadf36832a3446d50ec887fd6a7daaeedc3437f48f8af1d4e6db16385f873cb9d25b919608e19c4e27b280b00aafede6c47838c73a192e44ff26a73e093f12a18f8158c7c16dd403d8e4811e899435c644c429f7b63a9df6ad83d93875768264642f20f083cc2be3085546b25fd6d8f2551c15ac8b85175131535a8df9234fd398b26feffa65b49438f56d77f697f0958aacfb35e1c4a8e6e03c53601d2ed3080b722c94159f8b626a90aca7cc8d51c06e7b4dcaffec03
```
r-near, line 892: We should also add "MUST NOT reuse the ed25519 node secret as ξ" somewhere. This proves a different (also true) thing. But we wanna make sure no one accidentally pipes `I_L` straight into ML-DSA keygen

r-near, line 681: We should probably specify these encodings in the prose as well, not just the code.

r-near, line 854: For the vectors, perhaps making it more machine readable? Like moving this to an `assets/NEP-649/vectors.json` file?

neps/nep-0649.md
```diff
@@ -0,0 +1,967 @@
...
+### Derivation Path
+
+This NEP adapts [BIP-44] paths. The coin type MUST be `397'`, which was assigned to NEAR in [SLIP-44].
+
+Since ML-DSA derivation is hardened-only (see [Child Key Derivation](#child-key-derivation)), every level of the path MUST be hardened. This deviates from [BIP-44], which defines the `change` and `address_index` levels as non-hardened.
+
+- Wallets MUST support `m/44'/397'/index'`, where index selects the key.
+- Wallets SHOULD support a full [BIP-44] path in form `m/44'/397'/0'/0'/index'`.
+- Wallets MAY support deeper [BIP-44]-style paths, with every level hardened and condition that the wallet stores the exact path, so the key can be re-derived.
+- When the user does not specify a path, wallets SHOULD derive at `m/44'/397'/0'` path. Wallets using the full form SHOULD default to `m/44'/397'/0'/0'/1'` path that matches NEAR Ledger convention.
```
r-near: The default path is ambiguous here. 2 different paths with the same mnemonic and different ML-DSA key. Key recovery needs to be standardized across wallets. 

Honestly this may be pragmatic since we already have that path split for ed25519, but we should call it out explicitly and require the key recovery to scan both shapes. Right now it looks like the scanning requirement is across schemes (not path shapes)

neps/nep-0649.md
```diff
@@ -0,0 +1,967 @@
...
+
+The function ser<sub>32</sub>(i) is as defined in [BIP-32]: serialize a 32-bit unsigned integer i as a 4-byte sequence, Big-endian.
+
+1. Check whether i ≥ 2<sup>31</sup> (whether the child is a hardened key).
+   - If so (hardened child): let I = HMAC-SHA512(Key = c<sub>par</sub>, Data = 0x00 || k<sub>par</sub> || ser<sub>32</sub>(i)). (Note: The 0x00 pads the node secret to make it 33 bytes long.)
+   - If not (normal child): return failure.
+2. Split I into two 32-byte sequences, I<sub>L</sub> and I<sub>R</sub>.
+3. The returned child node secret k<sub>i</sub> is I<sub>L</sub>. The returned chain code c<sub>i</sub> is I<sub>R</sub>.
+
+The HMAC-SHA512 function is specified in [RFC 4231](https://datatracker.ietf.org/doc/html/rfc4231).
```
birchmd: The HMAC-SHA512 function is also used in the `Master Key Generation` section above. The citation for the function should appear after its first usage, instead of here.

neps/nep-0649.md
```diff
@@ -0,0 +1,967 @@
...
+
+To ensure recoverability of the user's account with ML-DSA-65 generated keys:
+- Wallets MUST accept any valid [BIP-39] mnemonic (12-24 words) for recovery.
+- The default path to derive ML-DSA-65 key from MUST be the same as Ed25519 recovery path.
+
+Also, wallets MAY generate a new [BIP-39] mnemonic for ML-DSA-65 keys, at the cost of a second backup and loss of single-mnemonic recovery. See [Mnemonic as a security bottleneck](#mnemonic-as-a-security-bottleneck) on why the new mnemonic MUST be from 18 to 24 words.
+
+Account discovery is out of scope (see [Scope](#scope)). However, wallets that discover accounts via the existing Ed25519 flow are encouraged to check if the same seed phrase produces Ed25519 and ML-DSA-65 keys. Note that wallets might need to compute ML-DSA-65 hash per [NEP-645] for discovering the account. Additionally, it is worth checking if the key has a `FullAccess` permission to allow for a full usage of account. 
+
+### Master Key Generation
```
birchmd, line 75: Nit: can `Generation` and `Derivation` be used interchangeably?

neps/nep-0649.md
```diff
@@ -0,0 +1,967 @@
...
+
+### Recoverability
+
+Many wallets enable users to recover their accounts with deriving Ed25519 key at the default `m/44'/397'/0'` path from [BIP-39] mnemonic.
+
+To ensure recoverability of the user's account with ML-DSA-65 generated keys:
+- Wallets MUST accept any valid [BIP-39] mnemonic (12-24 words) for recovery.
+- The default path to derive ML-DSA-65 key from MUST be the same as Ed25519 recovery path.
+
+Also, wallets MAY generate a new [BIP-39] mnemonic for ML-DSA-65 keys, at the cost of a second backup and loss of single-mnemonic recovery. See [Mnemonic as a security bottleneck](#mnemonic-as-a-security-bottleneck) on why the new mnemonic MUST be from 18 to 24 words.
```
birchmd: I initially thought there was a discrepancy between the minimum length of 18 here and 12 above, but then I realized the difference is in the accepted minimum for a user-provided mnemonic vs a new mnemonic generated by the wallet. I think this could be made a little more clear (it is clearer in the security section below, but should also be clear here).

birchmd (commented):
Nice work on this NEP @vsavchyn-dev ! I only have a few minor comments.

neps/nep-0649.md
vsavchyn-dev: Moved citation a bit higher in https://github.com/near/NEPs/pull/649/changes/e8664ae75f9c95e6006cd5aa891a72f24a48ebc6#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR80

neps/nep-0649.md
vsavchyn-dev: I merged this whole paragraph into a list and reworded it a bit - it should be more clear now: https://github.com/near/NEPs/pull/649/changes/e8664ae75f9c95e6006cd5aa891a72f24a48ebc6#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR66-R70

neps/nep-0649.md
vsavchyn-dev: I guess calling it out in both sections is sufficient: https://github.com/near/NEPs/pull/649/changes/e8664ae75f9c95e6006cd5aa891a72f24a48ebc6#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR55-R67

At the same time I made it mandatory to support full BIP-44 path - at the end of the day, we shouldn't restrict users to only deriving from `m/44'/397'/index'` or `m/44'/397'/0'/0'/index'`, we should support both.

Thinking about it, while I did write down "full [BIP-44] path in form `m/44'/397'/0'/0'/index'`", wouldn't it make sense to mention full BIP-44 path (with `m/44'/397'` ofc)? Right now it reads for me as "we permit `m/44'/397'/index'` and `m/44'/397'/0'/0'/index'` only, and you may use something similar to BIP-44 path, no `m/44'/397'/account'/change'/index'`. Wdyt?

neps/nep-0649.md
vsavchyn-dev: Mentioned encodings in https://github.com/near/NEPs/pull/649/changes/e8664ae75f9c95e6006cd5aa891a72f24a48ebc6#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR110

vsavchyn-dev: Wrote it down in `Key Pair Derivation` section: https://github.com/near/NEPs/pull/649/changes/e8664ae75f9c95e6006cd5aa891a72f24a48ebc6#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR104

r-near: Sorry, I was unclear, I meant the NEP-645 string formats rather than hex vs base58.

Something like this in Key Pair Generation or the vectors intro:

> - public key string: `ml-dsa-65:` + base58(1952-byte pk)
> - on-trie handle: `ml-dsa-65-hash:` + base58(SHA3-256(`"near:ml-dsa-65-pubkey-hash:v1"` || pk))

neps/nep-0649.md
r-near: Yeah agreed. Let's generalize support to `m/44'/397'/account'/change'/index'` but keep the default pinned at `0'/0'/1'`.

And the recovery MUST scoped to just the two defaults

r-near: Maybe also clarify that `change'` has no BIP-44 meaning here (it's just Ledger/near-cli compat)

neps/nep-0649.md
vsavchyn-dev: I don't think that paper explains specifically why hardened-only derivation needed (aka it had few lines about that), however, I wrote explanation down as people might wonder why hardened-only is forced:  https://github.com/near/NEPs/pull/649/commits/66b1ea2d0ce1248cf5e722de53f8617cbb947605

neps/nep-0649.md
vsavchyn-dev: In this section, I tried to follow the SLIP-10 paper, which stated `Master Key Generation`. I think by definition `Generation` and `Derivation` can be used interchangeably: `HMAC-SHA512` generates a hash, but at the same time, the hash is derived from the domain-separator seed and mnemonic. However, in the context of this NEP, you are right; it is confusing to write `Generation` anywhere except parts with mention of `ML-DSA.KeyGen`, as we derive the seed to generate ML-DSA key.

Following this, I just wrote "Master key derivation MUST..." instead of "Master key generation MUST..." (https://github.com/near/NEPs/pull/649/commits/70cf84fb9eb9cb227db66af4dfc73179dadd7c16), everything else in the paper already states things properly.

neps/nep-0649.md
vsavchyn-dev: Will this mention of the format be sufficient? [3458558](https://github.com/near/NEPs/pull/649/commits/345855824d5387e623d18340f3fb5a946d850974)

Or did you mean something else?

neps/nep-0649.md
vsavchyn-dev: Fixed in [246b69e](https://github.com/near/NEPs/pull/649/commits/246b69ee0beaab5fd796dc6991a20a134b157fe7)

I also changed wrote `account'` instead of `index'` for the short path to be more BIP-44-like.

neps/nep-0649.md
vsavchyn-dev: Added them in [30d8c0d](https://github.com/near/NEPs/pull/649/commits/30d8c0d9c2b77545040a4fcd6fd3d8b783ee3f6d)

vsavchyn-dev:
@near/nep-moderators I believe this nep is ready for the review - `markdown-link-check` is unrelated to this NEP.

Additionally, @darioush, I am curious if this NEP have any significant implications for `nearcore`? I want to ensure that the [ML-DSA key recovery logic](https://github.com/vsavchyn-dev/NEPs/blob/feat/mldsa-derivation/neps/nep-0649.md#recoverability) and [ML-DSA key pair generation + storage of public key](https://github.com/vsavchyn-dev/NEPs/blob/feat/mldsa-derivation/neps/nep-0649.md#key-pair-generation-fips-204) I've proposed is perfectly in sync with the protocol's implementation of post-quantum signatures. For instance, maybe I need to mention that per NEP-645, storing ML-DSA key only is available with enabled protocol feature?

gagdiez:
thanks @vsavchyn-dev, this NEP is ready for review. @walnut-the-cat could you help by selecting two SME to give their review?

Maybe Simon and Darioush, who reviewed the original ML-DSA-65 NEP.

SimonRastikian:
Looking into it

neps/assets/nep-0649/test-vectors.json
SimonRastikian, line 1: I don't know about this but will check the .md file

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+
+With the introduction of [NEP-645], users got access to PQ transaction signatures on NEAR Protocol, allowing them to protect their accounts against a future quantum computer threat by adding an ML-DSA-65 access key. Meanwhile, self-custodial HD wallets, which serve as the main point of entry for users on NEAR, do not have a specification to follow on how to derive those keys deterministically.
+
+[SLIP-10] specification does not yet cover ML-DSA and has an open PR ([slips#1968]) that proposes adding it.
+
+This NEP specifies the standard on how wallets should deterministically derive ML-DSA-65 key pairs from a BIP-39 mnemonic using SLIP-10 hardened-only derivation, aligned with the construction proposed in [slips#1968].
+
+## Motivation
+
+NEAR Protocol is one of the early adopters of post-quantum keys, and users need a way to deterministically derive their ML-DSA-65 access keys, similar to how it is done with Ed25519 keys specified in SLIP-10. This has several applications, for instance enabling recovery of the key from a [BIP-39] mnemonic alone, and cross-wallet portability.
```
SimonRastikian: "NEAR Protocol is one of the early adopters of post-quantum keys" seems a bit pretentious and not really meant for a NEP

SimonRastikian:
```suggestion
NEAR Protocol is one of the early adopters of post-quantum keys, and users need a way to deterministically derive their ML-DSA-65 access keys, similar to how it is done with Ed25519 keys specified in SLIP-10. This has several applications, for instance enabling key recovery from a [BIP-39] mnemonic alone, and cross-wallet portability.
```

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+
+**In scope (this NEP):**
+
+- ML-DSA-65 deterministic key derivation using 32-bytes output from SLIP-10 function with BIP-39 seed for Algorithm 6 `ML-DSA.KeyGen_internal(𝜉)` from FIPS 204.
+
+**Out of scope:**
+
+- Derivation for other ML-DSA key variants and a way to generate them (while possible, NEAR only enables ML-DSA-65).
+- Derivation for other key types available for users on NEAR (Ed25519 and Secp256k1).
+- Definition for Non-hardened ML-DSA keys (see [Child Key Derivation](#child-key-derivation)).
```
SimonRastikian:
```suggestion
- Definition for non-hardened ML-DSA keys (see [Child Key Derivation](#child-key-derivation)).
```

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+
+### Master Key Generation
+
+Master key derivation MUST follow the [SLIP-10] process described below.
+
+The use of the node secret as key generation seed is specified in [Key Pair Generation](#key-pair-generation-fips-204).
+
+Let S be a seed byte sequence of 128 to 512 bits. This is the same as the seed byte sequence used in BIP-32. The value of S SHOULD be the binary seed obtained from a [BIP-39] mnemonic and optional passphrase.
+
+Let Curve be the ASCII string used as the key for HMAC-SHA512 ([RFC 4231](https://datatracker.ietf.org/doc/html/rfc4231)), identifying the signature scheme. This NEP defines derivation only for `Curve = "ML-DSA-65 seed"`.
```
SimonRastikian, line 83: I strongly dislike what is happening here.
Curve is only meant for elliptic curve schemes like Ed25519 not for post-quantum (lattice-based) scheme ML-DSA

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+
+1. Calculate I = HMAC-SHA512(Key = Curve, Data = S)
```
SimonRastikian, line 85: Please fix the curve thingy here

SimonRastikian: If it was up to me I would have used HKDF (to double check if SLIP does that already)instead of HMAC but I think this should be fine here. The advantage would be that you run HKDF-Extract(salt = Curve, IKM = S). Then I_R <- HKDF.expand(info="master chain code")  and I_L <- HKDF.expand(info="master node secret").

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+2. Split I into two 32-byte sequences, I<sub>L</sub> and I<sub>R</sub>.
+3. Use I<sub>L</sub> as master node secret, and I<sub>R</sub> as master chain code.
+
+For ML-DSA, the last step always succeeds (note that there is no step 4 compared to [SLIP-10]), as any 32-byte sequence is a valid seed for the ML-DSA key generation.
```
SimonRastikian, line 89: Nicely flagged  "no step 4"

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+Master key derivation MUST follow the [SLIP-10] process described below.
+
+The use of the node secret as key generation seed is specified in [Key Pair Generation](#key-pair-generation-fips-204).
+
+Let S be a seed byte sequence of 128 to 512 bits. This is the same as the seed byte sequence used in BIP-32. The value of S SHOULD be the binary seed obtained from a [BIP-39] mnemonic and optional passphrase.
+
+Let Curve be the ASCII string used as the key for HMAC-SHA512 ([RFC 4231](https://datatracker.ietf.org/doc/html/rfc4231)), identifying the signature scheme. This NEP defines derivation only for `Curve = "ML-DSA-65 seed"`.
+
+1. Calculate I = HMAC-SHA512(Key = Curve, Data = S)
+2. Split I into two 32-byte sequences, I<sub>L</sub> and I<sub>R</sub>.
```
SimonRastikian, line 86: If you still want to use HMAC-SHA512 I would personally prefer to domain separate both I_L and I_R when generating them. I do not think however the current method has real security implications

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
...502573720120251532870081162155872730338036774463410712115468581716630816618831633164470502545542008646162730814204462474281377752855276172176020631244445161885338623802547227386665711008310772758024712716757655774126018170241126162103377215211331805605761685827305156304830578088321144703514861775123611233651118187210787226250262186032126063284766357531001673867036550455404332228751437304612605671282238057087842733053247837473888610762752416284415272071480734815557505767220886681023706765610455657864637733242555721276357001333445237562322428712374510022782676665637735584440843773148582560510414471708777518758550302525065601683845656056785314525875477201487228307873684018176868636024370302827856835176167347550133354285036167671432285601046423346547825872670203882772675014251354874813114766025756861661313428058284764623311228053348057846248528722705644012464440666872136217484548363587544650035368405814782088685580057118870852738714847863745800372552172756221375247710340225163677881276573027514251120456825583857473855562744570378846380626337335578616725648133837071240214455330882005314778162270102253307154461644866017428468358662311667754530710786385387301711783010864034872380830001836
```
SimonRastikian, line 960: What is this supposed to represent?

SimonRastikian (approved):

neps/nep-0649.md
vsavchyn-dev: These are test vectors people can compare things to. First two vectors generated from same seed and path as SLIP-10 test vectors [here](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#test-vectors), and last two are just compliant with what is written in this nep (aka 24-word BIP-39 and hardened BIP-44 path).

Each level has both hex-encoded (what is used in SLIP-10 format) and base58-encoded (what we use in Near) private and public keys.

neps/assets/nep-0649/test-vectors.json
vsavchyn-dev: `test-vectors.json` contains same vectors as in "Test Vectors" section, but just in computer-readable format.

neps/nep-0649.md
vsavchyn-dev: This NEP tries to be as compatible with SLIP-10 as possible, so the name `Curve` is inherited for consistency. Also, it is purely a domain-separation string and I believe it carries no elliptic curve meaning.

Still, I am not opposed to idea of changing `Curve` to `Key` (similar to how it's written in [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#master-key-generation:~:text=Calculate%20I%20%3D%20HMAC%2DSHA512(Key%20%3D%20%22Bitcoin%20seed%22%2C%20Data%20%3D%20S)))

neps/nep-0649.md
vsavchyn-dev: done in [6c131dc](https://github.com/near/NEPs/pull/649/commits/6c131dc828c53b09accdfcf9a07b75cfa95e35be)

neps/nep-0649.md
vsavchyn-dev: done in [6c131dc](https://github.com/near/NEPs/pull/649/commits/6c131dc828c53b09accdfcf9a07b75cfa95e35be)

neps/nep-0649.md
vsavchyn-dev: Something like:

```
Let Key be the ASCII string used as the key for HMAC...
This NEP defines derivation only for `Key = "ML-DSA-65 seed"`.

1. Calculate I = HMAC-SHA512(Key = "ML-DSA-65 seed", Data = S)
```

Wdyt?

neps/nep-0649.md
vsavchyn-dev: So, no SLIP uses HKDF. It's plain `HMAC-SHA512` throughout Master Key Gen and CKD.

That said, I believe 1st step of Master Key Gen and CKD looks like `HKDF.Extract`. The only difference is 2nd step of Master Key Gen ("Split I into two 32-byte sequences") - SLIP-10 reads I_L and I_R directly from the 64-byte output instead of deriving the with two `HKDF.Expand(info = "...")` calls.

Unfortunately, `HKDF.Expand` produces different bytes, breaking compatibility with SLIP-10 and the [slips#1968] test vectors, so I don't think we should use `HKDF` here.

neps/nep-0649.md
vsavchyn-dev: Similar to my point of not using HKDF - this nep needs to be compatible with SLIP-10, so domain separation of I_L and I_R would just break this compatibility imo

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+NEP: 649
+Title: Deterministic ML-DSA-65 Key Derivation from BIP-39 Seeds (SLIP-10)
+Authors: Vladyslav Savchyn (@vsavchyn-dev)
+Status: Draft
+DiscussionsTo: https://github.com/near/neps/pull/649
+Type: Wallet Standard
+Requires: 645
+Version: 1.0.0
+Created: 2026-07-30
+LastUpdated: 2026-07-30
```
r-near, line 11: Maybe update the date here?

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
+---
+NEP: 649
+Title: Deterministic ML-DSA-65 Key Derivation from BIP-39 Seeds (SLIP-10)
+Authors: Vladyslav Savchyn (@vsavchyn-dev)
+Status: Draft
+DiscussionsTo: https://github.com/near/neps/pull/649
```
r-near, line 6: Should it be caps? 
```suggestion
DiscussionsTo: https://github.com/near/NEPs/pull/649
```

neps/nep-0649.md
r-near: +1 to `Key`, it's what BIP-32 uses and it drops the elliptic curve connotation. 

Maybe keep a clause saying it's the same string SLIP-0010 calls `Curve`? So the https://github.com/satoshilabs/slips/pull/1968 cross-reference stays obvious

Line [904](https://github.com/near/NEPs/pull/649/changes#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR904) & [920](https://github.com/near/NEPs/pull/649/changes#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR920) say the same thing so maybe rename it there too

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
...201202515328700811621558727303380367744634107121154685817166308166188316331644705025455420086461627308142044624742813777528552761721760206312444451618853386238025472273866657110083107727580247127167576557741260181702411261621033772152113318056057616858273051563048305780883211447035148617751236112336511181872107872262502621860321260632847663575310016738670365504554043322287514373046126056712822380570878427330532478374738886107627524162844152720714807348155575057672208866810237067656104556578646377332425557212763570013334452375623224287123745100227826766656377355844408437731485825605104144717087775187585503025250656016838456560567853145258754772014872283078736840181768686360243703028278568351761673475501333542850361676714322856010464233465478258726702038827726750142513548748131147660257568616613134280582847646233112280533480578462485287227056440124644406668721362174845483635875446500353684058147820886855800571188708527387148478637458003725521727562213752477103402251636778812765730275142511204568255838574738555627445703788463806263373355786167256481338370712402144553308820053147781622701022533071544616448660174284683586623116677545307107863853873017117830108640348723808300018360082756
```
r-near, line 656: The JSON has also has NEAR mnemonic vectors. 

Also the JSON doesn't have the `ml-dsa-65-hash` handle, which is what a wallet compares against `view_access_key_list`. The tables above have it, so `print_node` could probably just emit it 🤷

r-near, line 920: Should this be both trees at both default paths now (`m/44'/397'/0'` and `m/44'/397'/0'/0'/1'`)?

r-near, line 926: Same as above, it's four derivations now rather than two trees instead of one

r-near, line 890: FYI SDKs already match this. https://github.com/r-near/near-kit-rs/pull/278 and https://github.com/r-near/near-kit/pull/234 default ML-DSA phrases to 24 words and refuse under 18.

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+
+### Key Pair Generation (FIPS 204)
+
+The ML-DSA-65 key pair is generated with Algorithm 6 `ML-DSA.KeyGen_internal(𝜉)` from [FIPS 204]. A 32-byte node secret I<sub>L</sub> from the [Child Key Derivation](#child-key-derivation) section MUST be used as a seed `𝜉` for `ML-DSA.KeyGen_internal(𝜉)`. At the same time, the 32-byte Ed25519 node secret from [SLIP-10] MUST NOT be reused as a seed `𝜉` for the ML-DSA-65 key generation. The resulting public key is a standard ML-DSA-65 public key and MAY be added to the user's account with the `AddKey` action as specified in [NEP-645].
+
+On NEAR, the resulting key appears in two string formats:
+
+- Public key string: `<key_type>:<base58-data>`, where `key_type` is `ml-dsa-65` and `base58-data` is the base58 encoding of the 1952-byte public key. This follows NEAR's convention for public key strings.
+- On-trie handle: `<key_handle_type>:<base58-data>`, where `key_handle_type` is `ml-dsa-65-hash`, and `base58-data` is the base58 encoding of the 32-byte `SHA3-256(b"near:ml-dsa-65-pubkey-hash:v1" || public_key)` hash, as specified in [NEP-645]. This is the form returned by `view_access_key_list`, since the chain stores only the hash.
```
r-near, line 112: Should we add a secret-key string here too? Maybe `ml-dsa-65:<base58(ξ)>` as the form wallets SHOULD export, with importers telling forms apart by decoded length (32 = seed, 4032 = expanded, anything else rejected)?

r-near: And if the 4032-byte form stays importable, maybe a MUST to validate it on import?

neps/nep-0649.md
r-near, line 900: FYI the two reviewers on [#1968](https://github.com/satoshilabs/slips/pull/1968) (@andrewkozlik, @onvej-sl) are co-authoring their own draft now, [SLIP-0027](https://github.com/satoshilabs/slips/pull/2017).

It derives ML-DSA seeds from SLIP-0021 plus a counter-mode Expand, so it gives different keys for the same mnemonic 👀 

Maybe mention SLIP-0027 here, and add something like: NEP-649 outputs are fixed by its vectors whatever [#1968](https://github.com/satoshilabs/slips/pull/1968) / [SLIP-0027](https://github.com/satoshilabs/slips/pull/2017) / Ledger's BIP end up doing, and any future NEAR ML-DSA derivation uses a different key string.

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+
+Usage of the 32-byte seed as the default key format is explicitly permitted by NIST in [Post Quantum FAQ](https://csrc.nist.gov/Projects/post-quantum-cryptography/faqs#Rdc7).
```
r-near, line 114: I think the FAQ covers the seed as a key format (storing/importing a seed that came out of KeyGen), not deriving ξ from a mnemonic. FIPS 204 §3.6.1 still wants ξ fresh from an approved RBG.  

Maybe say these are standard ML-DSA-65 keys but the derivation itself isn't FIPS 204 key generation?

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+
+Account discovery is out of scope (see [Scope](#scope)). However, wallets that discover accounts via the existing Ed25519 flow are encouraged to check if the same seed phrase and default path specified in this section produce Ed25519 and ML-DSA-65 keys. Note that wallets might need to compute the ML-DSA-65 hash per [NEP-645] for discovering the account. Additionally, it is worth checking whether the key has `FullAccess` permission to enable full use of the account. 
+
+### Master Key Generation
+
+Master key derivation MUST follow the [SLIP-10] process described below.
+
+The use of the node secret as key generation seed is specified in [Key Pair Generation](#key-pair-generation-fips-204).
+
+Let S be a seed byte sequence of 128 to 512 bits. This is the same as the seed byte sequence used in BIP-32. The value of S SHOULD be the binary seed obtained from a [BIP-39] mnemonic and optional passphrase.
```
r-near, line 81: Should this be a MUST when a BIP-39 mnemonic is used? With SHOULD, a wallet could feed in the raw entropy or its own KDF output and still be compliant, right? And cross-wallet recovery would break?

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+
+1. Check whether i ≥ 2<sup>31</sup> (whether the child is a hardened key).
+   - If so (hardened child): let I = HMAC-SHA512(Key = c<sub>par</sub>, Data = 0x00 || k<sub>par</sub> || ser<sub>32</sub>(i)). (Note: The 0x00 pads the node secret to make it 33 bytes long.)
+   - If not (normal child): return failure.
+2. Split I into two 32-byte sequences, I<sub>L</sub> and I<sub>R</sub>.
+3. The returned child node secret k<sub>i</sub> is I<sub>L</sub>. The returned chain code c<sub>i</sub> is I<sub>R</sub>.
+
+### Key Pair Generation (FIPS 204)
+
+The ML-DSA-65 key pair is generated with Algorithm 6 `ML-DSA.KeyGen_internal(𝜉)` from [FIPS 204]. A 32-byte node secret I<sub>L</sub> from the [Child Key Derivation](#child-key-derivation) section MUST be used as a seed `𝜉` for `ML-DSA.KeyGen_internal(𝜉)`. At the same time, the 32-byte Ed25519 node secret from [SLIP-10] MUST NOT be reused as a seed `𝜉` for the ML-DSA-65 key generation. The resulting public key is a standard ML-DSA-65 public key and MAY be added to the user's account with the `AddKey` action as specified in [NEP-645].
```
r-near, line 107: Could we widen this to any classical key, not just the Ed25519 node secret? [Cosmos shipped](https://github.com/cosmos/cosmos-sdk/pull/26436/) exactly that with the secp256k1 BIP-32 key as ξ, and NEAR has secp256k1 keys too. Maybe "ξ MUST NOT be a private key or node secret from any other key tree"?

neps/nep-0649.md
r-near: Sorry, missed this! Yep, that's what I meant 👍 (one more ask on [L111](https://github.com/near/NEPs/pull/649/changes#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR111))

neps/nep-0649.md
r-near: Also while touching `print_node`: could it emit the `ml-dsa-65:` pk string and the seed-form secret key (`ml-dsa-65:<base58(node secret)`>), and label the 4032-byte sk as the expanded form?

neps/nep-0649.md
r-near: Oh and [L34](https://github.com/near/NEPs/pull/649/changes#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR34) ("curve string") and the [L675](https://github.com/near/NEPs/pull/649/changes#diff-e514832a368228580b962a3a87f14be3812d43f9682fba381e6de8e4cc3f991cR675) comment ("seed modifier for HMAC-512") too.

neps/nep-0649.md
r-near, line 914: Small thing: QIP-0002 uses "Dilithium seed" as the key and defaults to ML-DSA-87, so same shape but different keys. Maybe "structurally similar to" so nobody expects compatibility?

neps/nep-0649.md
```diff
@@ -0,0 +1,975 @@
...
+- Definition for non-hardened ML-DSA keys (see [Child Key Derivation](#child-key-derivation)).
+- Post-derivation usage of the ML-DSA-65 keys by wallets (account discovery, key rotation).
+
+### Derivation Path
+
+This NEP adapts [BIP-44] paths. The coin type MUST be `397'`, which was assigned to NEAR in [SLIP-44]. Every level of the path MUST be hardened; this deviates from BIP-44, which defines the `change` and `index` levels as non-hardened.
+
+- Wallets MUST support `m/44'/397'/account'`, where `account'` selects the key; the default path in this form is `m/44'/397'/0'`.
+- Wallets MUST support a full BIP-44 path in the form `m/44'/397'/account'/change'/index'`; the default path in this form is `m/44'/397'/0'/0'/1'`.
+- Wallets MAY support deeper BIP-44-style paths, with every level hardened, and require that the wallet store the exact path so the key can be re-derived.
```
r-near, line 55: I think the MAY also covers "require" here, so recording the path reads as optional.

Maybe "Wallets MAY support deeper paths; a wallet that does MUST record the exact path"? 

Other wallets only scan the two defaults, so it's worth showing to the user too

neps/nep-0649.md
SimonRastikian: If you  decide on changing it, even better than key would be something like Scheme or SchemeType maybe? In any case key is also strictly better than curve

neps/nep-0649.md
SimonRastikian: It's fine, HKDF uses in its underlying structure two instances of HMAC so I am fine keeping HMAC. It's just HKDF is more appropriate to derive keys (HKDF = Hash Key Derivation Function)

neps/nep-0649.md
SimonRastikian: What are the consequences of this NEP not being compatible with SLIP-10? Would it generate practical issues?

neps/nep-0649.md
vsavchyn-dev: The thing is, I thought there might be some other BIPs or SLIPs that can be passed. For instance SLIP-39 can be used here as long as it provides 192 bits of entropy, no?

But, MUST here would fit imo

neps/nep-0649.md
vsavchyn-dev: While secret-key is used for signing, I think it's better to avoid adding it here.

I would argue that this "Resulting key appears on near..." should be replaced by "Resulting key MUST be formatted according to NEP-645" or something. This NEP should mostly care about deriving, while saying that "you must store keys according to nep645"

neps/nep-0649.md
vsavchyn-dev: Regarding the main comment - I think it would be better to just leave one encoding (I would prefer base58 in this case to make it easy for wallets), and don't include ml-dsa-65 hash to at least make it similar to slip-10 test vectors.

Regarding second comment - It's bad idea imo, we clearly have node secret that is seed for keygen and a sk + pk for a path. Not having it is more clear and even easier to compare against.

Additionally, I think it would be better to delete all test vectors in markdown and just link to json file to reduce the size of this markdown....

vsavchyn-dev: I think "aligned" means "similar to, not strictly the same", so it should be probably fine...?

neps/nep-0649.md
vsavchyn-dev: I opted in for `Key` as it is generalized a bit more: [12109b6](https://github.com/near/NEPs/pull/649/commits/12109b6b45b291faccfa3dbf1a45db5305bdb8b6)

neps/nep-0649.md
vsavchyn-dev: Done [12109b6](https://github.com/near/NEPs/pull/649/commits/12109b6b45b291faccfa3dbf1a45db5305bdb8b6)

neps/nep-0649.md
vsavchyn-dev: Done [12109b6](https://github.com/near/NEPs/pull/649/commits/12109b6b45b291faccfa3dbf1a45db5305bdb8b6)

neps/nep-0649.md
vsavchyn-dev: Fixed in [12109b6](https://github.com/near/NEPs/pull/649/commits/12109b6b45b291faccfa3dbf1a45db5305bdb8b6)

neps/nep-0649.md
vsavchyn-dev: Done in [12109b6](https://github.com/near/NEPs/pull/649/commits/12109b6b45b291faccfa3dbf1a45db5305bdb8b6)

neps/nep-0649.md
vsavchyn-dev: Fixed in [12109b6](https://github.com/near/NEPs/pull/649/commits/12109b6b45b291faccfa3dbf1a45db5305bdb8b6)

neps/nep-0649.md
vsavchyn-dev: Fixed in [cecbad3](https://github.com/near/NEPs/pull/649/commits/cecbad372435edd17607a0a45401ab2cf5839fc8)

PR #652, opened 2026-08-19
NEP-652: Early Chunk Producer Reassignment

stedfn:
Protocol NEP proposing early chunk producer reassignment: removing a persistently failing chunk producer from its shard's chunk production rotation within an epoch, instead of waiting for the existing end-of-epoch kickout, which cannot change the validator set until two epoch boundaries later.

Today a producer that goes offline mid-epoch keeps receiving its share of that shard's chunk slots until the epoch ends. Every one of those slots is a missed chunk, so a shard can spend most of an epoch producing at a fraction of its nominal rate. The existing `chunk_producer_kickout_threshold` is a rotation control, not a liveness control: it decides who keeps a seat two epochs out, not what the shard does for the rest of this one.

The proposal blacklists a producer per shard once it has missed at least 100 chunks in the epoch and its production ratio is below 80%, and reassigns its slots to the remaining eligible producers by stake-weighted sampling over the survivors. A keep-one safety valve guarantees a shard is never left without an eligible producer. A 1000-block start-of-epoch grace window absorbs correlated restarts, primarily protocol upgrades.

Determinism is the core safety argument: the blacklist basis is the anchor's last-final block, which is header-derived, and the resolved producer is persisted and read back verbatim, with reads failing closed rather than falling back.

An excluded producer receives no new slots on that shard, so once its earlier assignments have been counted its production statistics there stop changing unless it returns. Those are the statistics the end-of-epoch kickout and the reward calculation read, so exclusion also settles how the boundary check reads the epoch. That consequence is intended, and the document states the conditions on it: the stake exemption, the guard that keeps one validator when every validator would otherwise be kicked out, and validator selection for the next-but-one epoch succeeding.

Reference implementation is merged in nearcore behind `ProtocolFeature::EarlyKickout` and stabilized at protocol version 87; the PRs are linked in the Reference Implementation section. Mainnet currently runs 86, so the mechanism is not yet active there.

### DIFF
--- a/neps/nep-0652.md
+++ b/neps/nep-0652.md
@@ -1,9 +1,9 @@
 ---
-NEP: 0
+NEP: 652
 Title: Early Chunk Producer Reassignment
 Authors: Stefan Neamtu <stefan.neamtu@nearone.org>
 Status: Draft
-DiscussionsTo: https://github.com/near/NEPs/pull/0000
+DiscussionsTo: https://github.com/near/NEPs/pull/652
 Type: Protocol
 Version: 1.0.0
 Created: 2026-08-17
--- a/neps/nep-0000.md
+++ b/neps/nep-0000.md
@@ -1,0 +1,848 @@
+---
+NEP: 0
+Title: Early Chunk Producer Reassignment
+Authors: Stefan Neamtu <stefan.neamtu@nearone.org>
+Status: Draft
+DiscussionsTo: https://github.com/near/NEPs/pull/0000
+Type: Protocol
+Version: 1.0.0
+Created: 2026-08-17
+LastUpdated: 2026-08-17
+---
+
+## Summary
+
+This proposal does not directly remove a validator from the validator set. It deterministically
+excludes a persistently failing chunk producer from future chunk-slot assignment on its shard for
+the remainder of the epoch, instead of letting its slots go on being missed until the epoch
+boundary. (The implementation names the feature `ProtocolFeature::EarlyKickout`; this document
+uses *reassignment* for the mechanism and reserves *kickout* for the existing end-of-epoch rules.)
+
+Today a chunk producer that goes offline mid-epoch keeps its share of that shard's chunk slots
+until the epoch ends, because the existing kickout only reacts at epoch finalization. For a fully
+unavailable producer every such slot is a missed chunk, so a shard can spend most of an epoch
+producing at a fraction of its nominal rate.
+
+This proposal derives a per-shard *exclusion set* — a *blacklist* — from the in-epoch chunk
+statistics the protocol already aggregates: a producer is excluded on a shard once it has missed
+at least 100 chunks there in the current epoch *and* its inclusion ratio is below 80%. Its future
+slots go to the remaining eligible producers by the existing stake-weighted rule. A keep-one
+safety valve guarantees a shard always retains one eligible producer, and a 1000-block
+start-of-epoch grace window absorbs coordinated restarts such as protocol upgrades. The mechanism
+is deterministic: the decision basis is a finality-derived block, and the resolved producer for
+each chunk is persisted and read back verbatim.
+
+The principal trade-offs are one new storage column and a hot-path read, higher load on the
+remaining producers, and exclusion that is sticky for the rest of the epoch regardless of cause —
+freezing the producer's chunk statistics and thereby locking in the end-of-epoch kickout outcome.
+
+## Motivation
+
+Chunk production availability is currently protected by a single, epoch-granular control:
+`chunk_producer_kickout_threshold`. At the end of an epoch, a producer whose
+`produced_chunks / expected_chunks` ratio is below the threshold is kicked out of the validator
+set for the following epoch. [NEP-509](https://github.com/near/NEPs/blob/master/neps/nep-0509.md)
+("Stateless validation Stage 0") describes this kickout family and extends it to chunk validators.
+
+That control is a *rotation* mechanism, not a *liveness* mechanism. It answers "should this node
+keep its seat next epoch?". It does not answer "what should the shard do for the rest of this
+epoch?". The gap this leaves is concrete:
+
+* A chunk producer that crashes early in an epoch keeps being sampled for its shard's chunk slots
+  for the remainder of the epoch. On mainnet-scale epochs that is on the order of tens of
+  thousands of block heights.
+* For a fully unavailable producer, every one of those slots is a missed chunk on that shard. The
+  shard still advances (block production is unaffected), but its transactions and receipts are not
+  processed at those heights, so the shard's effective throughput drops in proportion to the
+  failing producer's share of the settlement.
+* The impact is *per shard*. A producer assigned to one shard degrades that shard specifically,
+  which makes the user-visible effect uneven across the network.
+* The failure does not have to be adversarial. Hardware failure, a bad deploy, a network partition
+  affecting one operator, or an unusually long upgrade restart all produce the same shape.
+
+With four producers `A B C D` rotating on a shard and `C` failing mid-epoch, the shard looks like
+this today (`X` marks a missed chunk):
+
+```
+A B C D A B C D A B C D ...
+    ^ C fails
+A B X D A B X D A B X D ...     until the epoch boundary
+```
+
+and like this with in-epoch reassignment, once the failure has been detected:
+
+```
+A B X D A B X D A B D A B D ...
+                ^ C's slots reassigned to the survivors
+```
+
+Reassigning those slots mid-epoch converts a long tail of guaranteed misses into chunks produced
+by healthy producers. The protocol already computes, per epoch and per shard, the
+`(produced, expected)` counters this decision needs; what is missing is a rule that consumes them
+before the epoch boundary and a deterministic way for every node to agree on the resulting
+assignment.
+
+NEP-509 also documents a limitation that motivates the shape of this proposal rather than its
+existence: chunk validator quality is measured by `included_chunks / expected_chunks` rather than
+by endorsements, because the endorsement mask is not part of the block header. This proposal
+deliberately stays within the signals that *are* header-derived and already aggregated, so it
+introduces no new consensus inputs. Closing NEP-509's endorsement-blind gap is listed under
+Future possibilities.
+
+## Specification
+
+### Concepts
+
+**Chunk production statistics.** For each epoch, the protocol aggregates per shard and per
+validator a pair of counters, `produced` and `expected`, over the blocks of that epoch. These are
+the same counters that feed end-of-epoch kickouts. `missed` is defined as
+`expected - produced` (saturating at zero), and the *inclusion ratio* is `produced / expected`.
+Note that, as NEP-509 records, `produced` in fact counts chunks *included in blocks*: the block
+producer makes the final inclusion decision, so a chunk that was produced but not delivered, not
+validated in time, or simply not included counts as missed all the same. This proposal inherits
+that property of the signal; its consequences are discussed under Security Implications.
+
+**Chunk producer settlement.** The per-shard sequence of chunk producers that validator selection
+assigned to a shard for the epoch; the canonical sampler draws each height's producer from it.
+The *distinct producers of a shard* are the set of validator ids appearing in that shard's
+settlement. The statistics map also carries entries for validators that only endorsed chunks on
+the shard and never produced any; those entries MUST NOT be treated as producers.
+
+**Block info.** The per-block record of consensus metadata — height, epoch id, the last-final
+hash, and so on — that every node persists for every block it processes.
+
+**Last-final block.** The most recent block that is final under Doomslug as of a given block,
+recorded in that block's info. It is a function of the block's header, so every node derives the
+same last-final block for the same block.
+
+**Anchor.** For a chunk at height `H` built on parent block `P`, the anchor `A` is `P`'s parent,
+i.e. the chunk's *grandparent*. `A.height + 2` is the height used for sampling. `A` is the key
+under which the resolved producer is persisted and looked up. The grandparent, rather than the
+parent, is the anchor because chunks are produced and distributed concurrently with the
+propagation of their parent block: a node must be able to resolve a chunk's producer — to verify
+its signature — before it has fully processed `P`, and the grandparent is the newest block it is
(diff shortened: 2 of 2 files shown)


walnut-the-cat:
Thank you @stedfn for submitting this NEP.

As a moderator, I reviewed this NEP and it meets the proposed template guidelines. I am moving this NEP to the REVIEW stage and suggest @jancionear and @birchmd as SME reviewers to complete a technical review (see expectations). @near/wg-protocol working group members, please let us know if you have better suggestions. 

<details>

</details>

neps/nep-0652.md
```diff
@@ -0,0 +1,962 @@
...
+---
+
+## Summary
+
+This proposal does not directly remove a validator from the validator set. It deterministically
+excludes a persistently failing chunk producer from future chunk-slot assignment on its shard for
+the remainder of the epoch, instead of letting its slots go on being missed until the existing
+end-of-epoch kickout can remove it, two epoch boundaries later. (The implementation names the
+feature `ProtocolFeature::EarlyKickout`; this document uses *reassignment* for the mechanism and
+reserves *kickout* for the existing end-of-epoch rules.)
```
jancionear: > This proposal does not directly remove a validator from the validator set.

It would be nice to start the summary with a general description of what the proposed feature does. This sentence is pretty confusing as a start.

neps/nep-0652.md
```diff
@@ -0,0 +1,962 @@
...
+**Block info.** The per-block record of consensus metadata — height, epoch id, the last-final
+hash, and so on — that every node persists for every block it processes.
+
+**Last-final block.** The block named by a block header's `last_final_block` field: the most
+recent block with full BFT finality as of that block. The separate `last_ds_final_block` field
+records the weaker Doomslug finality and is not used here. Block info stores the former. Nodes
+reading the same header therefore derive the same last-final block.
+
+**Anchor.** For a chunk at height `H` built on parent block `P`, the anchor `A` is `P`'s parent,
+i.e. the chunk's *grandparent*. `A.height + 2` is the height used for sampling. `A` is the key
```
jancionear: It's interesting that the height is always `A.height + 2`, even when `A` might be a few heights behind. I think it should work okay in general.

neps/nep-0652.md
```diff
@@ -0,0 +1,962 @@
...
+The valve guarantees exactly one thing: the eligible set handed to the sampler is never empty. It
+does not guarantee shard liveness, and it does not preserve the original assignment distribution.
+If every producer on a shard is a candidate, the retained producer receives *all* subsequent chunk
+slots on that shard and may be unable to sustain that load. More generally, excluding producers
+that hold a fraction `x` of the shard's eligible stake multiplies every survivor's expected
+assignment rate by `1 / (1 - x)`; sustained partial failures can therefore concentrate load, and
+in the extreme push borderline survivors over the exclusion thresholds themselves. This is an
+explicit worst-case trade-off: the protocol prefers one deterministic eligible producer over
+either an undefined empty-set assignment or re-including producers whose chunks are currently not
+landing at all. Softer variants are discussed under Alternatives.
```
jancionear, line 198: For the safety valve I feel like it would be a bit better to include all validators instead of allowing only one.

A likely scenario I can imagine is that there is some general issue affecting all nodes in the network. Maybe a protocol bug, maybe a performance problem. All nodes go below the kickout threshold, but the safety valve allows only one, even though all of them were affected by the same thing. When all nodes go below the threshold, I don't see any real reason to not allow all of them to produce chunks, there's a higher likelihood that one of them will be functional than when picking only one.

Picking only one is actually a bit dangerous, what if it is dead? That would mean a long stream of missing chunks, which could cause problems for the network.

neps/nep-0652.md
```diff
@@ -0,0 +1,962 @@
...
+also be deterministic, since aggregation follows the anchor's own ancestry; the last-final block
+is preferred because it keeps the aggregation walk off the not-yet-final suffix during a finality
+stall.
+
+Two further requirements follow:
+
+* The epoch start height used by the grace check MUST be derived by walking block info from the
+  basis, not read from a per-epoch shared record. Sibling blocks at an epoch boundary overwrite
+  such a shared record, which would make the grace decision depend on the order in which a node
+  happened to process forks.
```
jancionear: Could you explain a bit more? I couldn't understand what the issue is exactly from the description.

neps/nep-0652.md
```diff
@@ -0,0 +1,962 @@
...
+| Inclusion ratio threshold | 80% | Matches `chunk_producer_kickout_threshold` on mainnet, where the end-of-epoch chunk-production and block-production thresholds are 80% and the chunk-validator-only threshold is 70%. Exclusion and the chunk-production kickout therefore use the same ratio on mainnet; other chains can configure a different boundary threshold. See Interaction with end-of-epoch kickout and rewards. |
+| `EARLY_KICKOUT_EPOCH_GRACE_BLOCKS` | 1000 | Absorbs correlated start-of-epoch downtime, primarily protocol-upgrade restarts, which land on epoch boundaries and affect the whole validator set at once. Without it a slow network-wide restart could blacklist many producers simultaneously in the first blocks of an epoch. |
+
+The two thresholds are conjunctive, so both a large absolute failure count and a low ratio are
+required; see Detection latency for what that means in time. The values were chosen after
+soliciting feedback from validators, who accepted them
+<!-- TODO(author): link the validator feedback venue/thread if public -->. In the reference
```
jancionear: There's a TODO here

neps/nep-0652.md
```diff
@@ -0,0 +1,962 @@
...
+  canonical sampler's picks. On this path a missing row, or a stored account that is not in the
+  epoch's validator set, falls back to canonical sampling at `A.height + 2`. If the parent's or the
+  anchor's block info is missing, as for the block right after an epoch-sync install, aggregation
+  samples canonically at the parent's height plus one instead. These fallbacks feed the statistics
+  the boundary kickout reads, so each must return the producer consensus resolved. Rows are written
+  and garbage collected together with the anchor's block info, so a same-epoch anchor that has
+  block info also has its rows, and the first fallback is reachable only through a damaged
+  database. The second fallback serves chunks whose anchor lies in the previous epoch, which the
+  consensus reader also resolves canonically. A dedicated test checks that every anchor with a
+  non-empty blacklist has its rows.
```
jancionear: WTF?

jancionear: This section was very hard to parse for me. I had to get GPT to explain it.

AFAIU the gist of it is that epoch sync provides the last few blocks of the previous epoch, but we can't find the chunk producers for those blocks, so the statistics are wrong, but it's fine because they're not used anywhere.

This sounds alright, I see that the epoch sync proof contains `next_next_epoch_info` which means it doesn't need the statistics for the previous epoch to calculate next next epoch info.

neps/nep-0652.md
```diff
@@ -0,0 +1,962 @@
...
+producer was excluded. Exclusion is scoped to that epoch: the blacklist is derived from `T`'s
+aggregated statistics, and those statistics reset when `T+1` begins, so no exclusion is carried
+forward. The producer is assigned slots again in `T+1` if `T+1`'s settlement — fixed when `T-1` was
+finalized, before this failure — still places it on the shard. If it is still failing there, `T+1`
+can exclude it again, but that is not guaranteed: the grace window, the 100-miss evidence floor and
+the keep-one valve all apply afresh, and an epoch too short to accumulate 100 misses never excludes
+anyone. Meanwhile, if `T`'s final statistics fail the boundary check, the producer loses its seat
+from the start of `T+2`, subject to the conditions above. An alternative that decouples assignment
+from these statistics is discussed under Alternatives, and a probe mechanism that lets a recovered
+producer return mid-epoch under Future possibilities.
```
jancionear: This was very hard to parse.

> The timeline across the boundary is worth spelling out, writing `T` for the epoch in which the
producer was excluded.

This sentence is super confusing

jancionear (commented):
Overall the NEP looks good to me.

Chunk producers going down but not being kicked out for a long time is a real problem that is worth fixing. We have observed cases where a single validator goes down, which causes a lot of missing chunks, which in turn increases the block times as the block producer waits for the chunk that will never arrive, overall degrading performance of the network. This can be avoided with a solution like this one.

The approach looks sound to me - using the state at anchor/basis blocks is deterministic and works well even in tricky fork cases. The epoch match check gracefully handles epoch transitions.

The only problem I see is with the safety valve - it feels unnecessarily risky for me, but it's not strictly a blocker. (https://github.com/near/NEPs/pull/652#discussion_r4027086052)

Overall I think this is a good change that will improve performance of the network, worth implementing.

One general complaint I have is that the NEP gives a very Claude-generated feel. Claude's writing is very unpleasant to read through, many times I had to reread the same sentence a couple of times to understand what it's supposed to mean. For future NEPs I would propose to use GPT, which has a much more pleasant writing style, or writing things manually.

neps/nep-0652.md
stedfn: Done. The summary now starts with the in epoch reassignment behavior

neps/nep-0652.md
stedfn: as discussed on [zulip](https://near.zulipchat.com/#narrow/channel/615213-releases.2F2.2E14/topic/Early.20kickout/near/624887068), if the sole producer goes down, its metrics will look worse and worse until it gets replaced by another 'least bad' producer.

neps/nep-0652.md
stedfn: rephrased in [447ed57](https://github.com/near/NEPs/pull/652/commits/447ed57ce4e68503ae6908c251397ecd5f15a9d9), let me know if it is better

neps/nep-0652.md
stedfn: deleted, thanks

neps/nep-0652.md
stedfn: made it more clear in [447ed57](https://github.com/near/NEPs/pull/652/commits/447ed57ce4e68503ae6908c251397ecd5f15a9d9)

neps/nep-0652.md
stedfn: made it more clear in [447ed57](https://github.com/near/NEPs/pull/652/commits/447ed57ce4e68503ae6908c251397ecd5f15a9d9), let me know if it is better

neps/nep-0652.md
```diff
@@ -0,0 +1,981 @@
...
+slots go to the remaining eligible producers by the existing stake-weighted rule. A keep-one
+safety valve guarantees a shard always retains one eligible producer, and a 1000-block
+start-of-epoch grace window absorbs coordinated restarts such as protocol upgrades. The mechanism
+is deterministic: the decision basis is a finality-derived block, and the resolved producer for
+each chunk is persisted and read back verbatim.
+
+The principal trade-offs are one new storage column and a hot-path read, higher load on the
+remaining producers, and exclusion that can last for the rest of the epoch regardless of cause.
+An excluded producer gets no new assignments on that shard, so once its earlier assignments have
+been counted its chunk statistics there stop changing unless it returns. Those statistics are
```
birchmd: > chunk statistics there stop changing unless it returns

How will we know the producer has returned if it does not receive any more slots for the remainder of the epoch?

neps/nep-0652.md
```diff
@@ -0,0 +1,981 @@
...
+  statistics from the header bitmap, and the boundary kickout uses them for chunk-validator-only
+  nodes. Applying exclusion to chunk validators would also require changing their assignments.
+  A missing chunk credits no endorsements to anyone, so a future rule would have to decide how
+  to attribute that case before excluding a validator for failing to endorse.
+* **Threshold tuning from mainnet data.** Once the feature is active, observed reassignment rates
+  and safety-valve firings give direct evidence for revisiting the parameters.
+* **Extending to block producers.** Block production has an analogous failure mode. It is out of
+  scope here because block producer assignment interacts with Doomslug consensus, which this
+  proposal deliberately does not touch.
+* **Faster recovery.** A blacklisted producer gets no new assignments on the affected shard, so
```
birchmd, line 813: Along the same lines as this: instead of a hard boundary after which the validator is blacklisted, there could be a sort of exponential backoff. Effectively the validator stake would be reduced by some function of the `(produced, expected)` statistics to have them be sampled less frequently. In the extreme case where the validator consistently misses then their stake would effectively drop to zero and they would functionally be blacklisted.

birchmd (commented):
As a SME reviewer and protocol working group member, I recommend approving this NEP. Responding to unavailable chunk producers faster than one or two epochs keeps the network responsive even in the face of degraded validator performance. At the same time, I do not think this proposal meaningfully changes the safety properties of the protocol.

As I was reading this proposal I was thinking about whether this has an impact on an attack known as "adaptive corruption" in which a malicious entity initially does not have enough stake to disrupt the protocol safety guarantees but gains enough over the course of an epoch by selectively targeting certain participants in the protocol. This attack is especially relevant in a sharded protocol because the malicious entity can have a globally small share of the stake, but still enough concentrated in a single shard to include an invalid state transition for example. This proposal involves giving extra weight to the remaining producers when one becomes blacklisted, which could be part of an adaptive corruption attack. However, I do not think that is a risk here because of stateless validation. It is the stateless validators that must endorse each chunk, so even if all the producers in a shard are malicious, they still cannot force through an invalid state transition because honest stateless validators will not allow it. And this proposal has no impact on the way stateless validators are assigned.

I left one minor comment on the future possibility of faster recovery, but I think this proposal can be accepted as-is.

Out of curiosity, I am also wondering how often a chunk producer has gone completely offline for multiple epochs in practice. Was there an inciting incident that led to this proposal?

jancionear (approved):
As a SME reviewer I recommend approving this NEP.

neps/nep-0652.md
jancionear: Makes sense, thanks!

gagdiez:
I see that both @birchmd and @jancionear have given their approvals, so we are ready to move to voting.

I would like to schedule a meeting with the working group members next week, so we can present the NEP to the community, and the @near/wg-protocol can vote.

@birchmd @jancionear @stedfn please share your availability here: https://rallly.co/invite/TMWB0orAasPf

In these meetings we usually start with an overview of the NEP (would be amazing if you could make a short introduction @stedfn nothing fancy, 2 or 3 slides would suffice), and for reviewers to give present their opinions on benefits vs concerns.

PR #654, opened 2026-09-02
NEP-641: non-normative clarifications from a second implementation

jatinsahijwani:
Non-normative clarifications to NEP-641, per NEP-001's provision for
updating Final NEPs. Opened at @frol's request in the forum thread so we can
discuss the specific wording.

These came out of writing a second implementation of the spec in TypeScript.
Each one is a detail an implementer has to get exactly right that is
currently discoverable only by reading the Rust reference, and each is now pinned by a committed test vector in near/intents#350 or near/intents#351, so none of this is an assertion about behaviour.

Seven added lines and one changed character, in three places:

**Timestamp representations**, after the existing clock-skew paragraph. No
fractional part is emitted when the nanosecond component is zero, otherwise
nanoseconds are padded to nine digits and trailing zeros are stripped. An
implementation that always emits nine digits passes every test the reference
ships and still disagrees with it for any millisecond-precision timestamp,
which is everything a JavaScript client produces. The same paragraph notes
that the `u64` Borsh form makes pre-epoch instants unrepresentable, which the
spec currently gives no range for.

**`path` in Borsh**, after the canonical hash. `skip_serializing_if` governs
JSON only, so an empty path still contributes its zero count to the hash
preimage. The type listing reproduces the attribute without saying which
encoding it governs, so a reader working from the prose has no way to tell.

**The two curve rules**, as sub-bullets on step 4 of the verification
procedure, mirroring the formatting step 6 already uses. ed25519
verification is cofactorless after an explicit small-order rejection, and
the secp256k1 recovery byte is range-checked even though verification never
recovers a key.

Item 4 is the one I would most want in the spec. "All native NEAR Protocol
key types SHOULD be supported; `access_key` and `signature` curves MUST
match" is currently the entire specification of signature verification, and
it leaves the equation choice to the implementer. The two answers disagree
on real inputs, and the cofactored form that most high-level libraries
expose by default is the permissive direction, so an implementation can
accept authorizations deployed contracts reject.

I have deliberately used lowercase "must match" rather than a normative
MUST, to keep this within NEP-001's allowance for non-normative
clarifications on a Final NEP. If you would rather that one be a real
requirement, uppercasing that single word is the whole change.

`Version`, `Status` and `LastUpdated` are untouched. Happy to bump
`LastUpdated` if that is the convention here.

Background and full citations: `docs/SPEC-NOTES.md` in
https://github.com/jatinsahijwani/nep641-ts
Forum thread:
https://gov.near.org/t/building-the-typescript-resolver-for-nep-641-currently-listed-as-tbd/42479


neps/nep-0641.md
```diff
@@ -254,12 +254,16 @@ The verifying contract MUST panic if:
 
 Clients SHOULD set `timestamp` ~60 seconds before signing time to absorb clock skew and block-time lag.
 
+In the RFC-3339 form, no fractional part is emitted when the nanosecond component is zero; otherwise nanoseconds are padded to nine digits and trailing zeros are stripped, keeping at least one digit (`2026-08-05T07:28:00Z`, `2026-08-05T07:28:00.1Z`, `2026-08-05T07:28:00.123456789Z`). The Borsh form is `u64` nanoseconds, so instants before the UNIX epoch cannot be represented and fail to serialize.
```
frol: I think it would be better to put a shorter version of this comment inline to the struct definition.

neps/nep-0641.md
```diff
@@ -254,12 +254,16 @@ The verifying contract MUST panic if:
...
+
 Canonical hash:
 
 ```text
 SHA3-256(b"NEAR_NEP641_OFFCHAIN_MESSAGE/V1" || borsh(msg))
 ```
 
+The `skip_serializing_if` on `path` governs JSON only: an empty `path` is absent from the JSON form but still contributes its four-byte little-endian count of zero to `borsh(msg)`, and therefore to the hash.
```
frol: Indeed, for non-Rust developers it could be ambiguous. Let's refine the struct definition with more comments regarding JSON & Borsh serialization.

neps/nep-0641.md
```diff
@@ -348,7 +352,9 @@ The resolver MUST verify an `AccessKeyAuthorization` for `account_id` at the pin
 1. `msg.chain_id` equals the resolver's chain ID.
 2. `msg.signer_id` equals `account_id`.
 3. `msg.path` equals the current resolution [path](#path).
-4. The signature verifies per `via` against `access_key`.
+4. The signature verifies per `via` against `access_key`:
+   - **ed25519** → reject small-order public keys, then verify with the **cofactorless** equation, comparing compressed `R` bytes; implementations must match this, since the cofactored equation most high-level libraries expose by default accepts signatures the reference rejects.
+   - **secp256k1** → the recovery byte is range-checked and rejected unless it is 0 to 3, even though verification never recovers a public key.
```
frol: Ideally, we should reference [the protocol reference documentation](https://nomicon.io/) or at least NEP-413 for signature verification instead of re-inventing and inlining the specifics as this will diverge over time.

frol (changes requested):

neps/nep-0641.md
jatinsahijwani: Done. Folded into the `timestamp` field comment and the standalone paragraph is gone.

neps/nep-0641.md
jatinsahijwani: Done, and I used the `(RFC-3339 string in JSON, u64 nanoseconds in Borsh)` shape the `timestamp` comment already had, since that pattern was already in this struct rather than something I would be introducing.

neps/nep-0641.md
jatinsahijwani: I went looking for something to cite and could not find it, so I have kept these inline but re-anchored them. Reporting what I checked in case it is useful beyond this PR.

Neither nomicon nor NEP-413 specifies any of the three points. Searching everything indexed on nomicon.io returns no occurrences of cofactor, small order, torsion, `verify_strict`, `ecrecover` or `recovery_id`. The transaction-signature spec at `RuntimeSpec/Transactions.md` is a single sentence about whether the transaction is signed by the access key corresponding to `public_key`. NEP-413's only adjacent line specifies the preimage, that signatures are over a SHA-256 hash rather than the raw message, not how verification is evaluated.

The case that convinced me a reference would not work is the small-order rejection. That rule is NEP-641's own, layered above protocol-level verification: `crates/crypto/src/ed25519.rs:17-22` calls `is_weak()` before delegating, while nearcore's `ed25519_verify` (`runtime/near-vm-runner/src/wasmtime_runner/logic.rs:1911,1965`) uses dalek's non-strict `verify` and accepts those keys. A pointer at nomicon or nearcore for that line would document the opposite of what NEP-641 requires.

What I have done instead: kept the rules inline, dropped any claim about what other libraries do, and noted that the small-order rejection sits above protocol verification with the reference implementation as the source of truth. Happy to cut them entirely if you would still rather they were not here.

This is not new ground in nearcore. near/nearcore#2700 raised the same inconsistency in 2020, and its own framing was that "Merely pointing to external standards is not sufficient, because they may not prescribe a single correct behavior, and having such a behavior is necessary for consensus safety." It is closed, but the reasoning applies directly here.

Separately, and not for this PR: nomicon's Math API page documents four host functions while nearcore registers `ecrecover`, `ed25519_verify`, `p256_verify`, `ml_dsa_verify`, the `sha3_*` family, and the `alt_bn128_*` and `bls12381_*` families. `ed25519_verify` and `ecrecover` are not documented at all. Happy to open that against the docs separately if it would be welcome.

jatinsahijwani:
Pushed `d049c8b`, addressing all three. Second commit rather than an amend so the delta is reviewable on its own.

While revising I also found two things wrong in my own original text and fixed them.

The claim that the cofactored equation is "what most high-level libraries expose by default" was an overclaim, and backwards. Taming the Many EdDSAs (eprint 2020/1244) tested OpenSSL, dalek, libsodium, Zebra, donna and Go among others, and found every tested library except Zebra implements a cofactorless variant. RFC 8032 itself prescribes the cofactored equation, which is part of why this is easy to get wrong. My finding was specifically about `@noble/curves` in TypeScript and I generalised it further than the evidence supports. The requirement is now stated plainly, with no claim about anyone else's defaults.

One caveat on that citation: the paper tested ed25519-dalek 1.0.0-pre.4 and nearcore pins 2.2.0, so I checked `raw_verify` in 2.2.0 directly rather than relying on the paper for the current version.

"Even though verification never recovers a public key" was wrong for the on-chain path, which trial-recovers because `env::ecrecover` requires a recovery id. The range check happens at parse time, so that is what it now says.

One more of the same class that I have left out: the reference rejects high-`s` secp256k1 signatures as malleable, and that is not documented anywhere either. Happy to add it here or leave it for a follow-up, whichever you prefer.

jatinsahijwani:
@frol no rush, flagging in case this dropped off the list.

I pushed d049c8b on 3 September addressing all three comments. The timestamp and path notes are now field comments on the struct rather than standalone paragraphs, and I corrected two claims of my own that were wrong, both noted in the comment above.

The one thing still open is whether you want the ed25519 rule to be normative. Lowercase "must match" keeps it inside NEP-001's non-normative allowance for a Final NEP, but uppercasing that single word is the whole change if you would rather it were a requirement. Happy either way, just did not want to make that call on your spec myself.

gagdiez:
hey @jatinsahijwani, thanks for the NEP, will nudge Frol next week so he reviews it, thanks!

PR #655, opened 2026-09-15
NEP-655: Universal Accounts

Wiezzel:
## Summary

This draft NEP introduces **Universal Accounts**: one unified, post-quantum-safe scheme for deriving both implicit (user-created) and deterministic (contract-created) account ids, which have until now needed separate mechanisms. A universal account id (`0u...`) is the SHA3-256 hash of a `StateInit` naming the account's code, storage and access keys; which fields are populated decides whether the account is a key-only EOA, a contract, or both. It documents the feature as stabilized at protocol version 87 (neard 2.14):

- The `0u` address format: 54 characters, Crockford base32, no checksum.
- `UniversalStateInit`, a versioned flat struct carried on the wire as opaque bytes, with the account id derived from exactly those bytes and content canonicity left to the producer.
- The new `UniversalStateInit` action, its validation rules, and the receipt-total bounds on committed access keys and storage entries.
- The uninitialized account state, and the two ways an account is created and initialized: funded and initialized in one batch by a relayer or contract, or self-signed by the account's own future owner against a key the id commits to.
- Two host functions that let a contract built against this scheme derive and create accounts of a `StateInit` version that did not exist when it was compiled.
- The rejected designs, including the address checksum that was built and then removed, enforced `StateInit` canonicity, the per-kind `StateInit` enum, and a dedicated "claim" transaction.

The reference implementation lives in nearcore behind `ProtocolFeature::UniversalAccounts`; the PRs that make it up are listed in the Reference Implementation section.


Wiezzel:
cc @wacban @mitinarseny

walnut-the-cat:
Thank you @Wiezzel for submitting this NEP.

As a moderator, I reviewed this NEP and it meets the proposed template guidelines. I am moving this NEP to the REVIEW stage and suggest @mitinarseny and @matejpavlovic as SME reviewers to complete a technical review (see expectations). @near/wg-protocol working group members, please let us know if you have better suggestions. 

<details>

</details>

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+Universal accounts are a new, disjoint scheme for *new* accounts only. The three existing derivations (NEAR-implicit, ETH-implicit, and [NEP-616]'s deterministic `0s` accounts) are untouched and keep working; there is no migration.
+
+A universal account can be created two ways: a relayer or contract can fund and initialize it in one batch (`Transfer` + `StateInit` + optional follow-up call), or the account's own future owner can sign a transaction directly against the not-yet-existing id, self-authorizing initialization with a key the id itself commits to. Two new host functions let a contract built against this scheme derive and create accounts of a `StateInit` version that did not exist when the contract was compiled, closing the same forward-compatibility gap [NEP-616]'s equivalent mechanism left open.
```
matejpavlovic, line 19:
```suggestion
A universal account can be created in two ways: a relayer or contract can fund and initialize it in one batch (`Transfer` + `StateInit` + optional follow-up call), or the account's own future owner can sign a transaction directly against the not-yet-existing id, self-authorizing initialization with a key the id itself commits to. Two new host functions let a contract built against this scheme derive and create accounts of a `StateInit` version that did not exist when the contract was compiled, closing the same forward-compatibility gap [NEP-616]'s equivalent mechanism left open.
```

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+```
+0u | 52 symbols encoding SHA3-256(RawStateInit bytes)
+```
+
+The address carries **no checksum**. An earlier design added one (see [Alternatives](#alternatives)); it was removed before stabilization because it could not be safely enforced.
+
+The only validity rule on the string itself is canonicity of that final symbol: with 260 bits of symbol capacity carrying a 256-bit hash, only 2 of the 32 possible final symbols (`0` and `g`) have their trailing 4 bits zero. A string with any other final symbol, or that otherwise does not match this shape, is not a universal account id; it is classified as an ordinary named account instead. This canonicity rule is enforceable (the prefix and length are the protocol's own choice) and is unrelated to the `StateInit`-content canonicity question discussed below.
+
+Classifying a string as a universal account id (`AccountType::UniversalAccount`) is implemented in the `near-account-id` crate, alongside the other account-type predicates, not in nearcore itself.
```
matejpavlovic, line 51: While this is useful information, it seems more like implementation rather than part of the specification. But no strong feelings about it, just resolve and ignore this comment if you want to keep it here.

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+    /// Contract code, by reference to a global contract. Absent for a key-only account.
+    pub code: Option<GlobalContractIdentifier>,
+    /// Storage entries to populate on initialization.
+    pub data: BTreeMap<Vec<u8>, Vec<u8>>,
+    /// Access keys to install, all full-access (see Keys, below).
+    pub access_keys: BTreeSet<PublicKeyHandle>,
+}
+```
+
+`GlobalContractIdentifier` is the type [NEP-591] introduced for referencing global contract code by hash or by account id. `PublicKeyHandle` is the on-trie key representation introduced alongside [NEP-645]: for a classical key it is the key itself; for an ML-DSA-65 key it is a 32-byte, domain-separated SHA3-256 hash of the 1952-byte public key, so committing a post-quantum key to a `StateInit` costs no more than a classical one.
```
matejpavlovic, line 72: Does this mean that a universal state init can only be used with a global contract? Or is it possible to somehow use it with arbitrary code?

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+4. The total number of `access_keys` committed across every `UniversalStateInit` action in the same receipt MUST NOT exceed `max_universal_state_init_keys` (1024).
+5. The total number of storage entries across every `DeterministicStateInit` and `UniversalStateInit` action in the same receipt MUST NOT exceed `max_state_init_entries` (1500, from protocol version 87 onward; effectively unbounded before, when only `DeterministicStateInit` existed). This limit is shared with the older, unrelated `0s` action; see [Security Implications](#security-implications) for why.
+
+Rules 4 and 5 are both receipt totals, not per-action limits, and both recurse into any `Delegate`/`DelegateV2` inner actions. Every action in a receipt shares one receiver, and a state init has to derive to that receiver, so a receipt can carry many byte-identical copies of the same payload and pays for each one; a per-action bound would therefore bound nothing. The two counts are taken from a single decode of each action so that they cannot drift apart again.
```
matejpavlovic, line 117: I don't quite understand this. If it's just my lack of background, feel free to ignore and resolve this comment. What would the byte-identical payloads be? Many copies of the same action? I looked at "Bounding state-init size", but I still don't really get it.

neps/nep-0655.md
mitinarseny: Technically, all `0u` accounts exist from day zero, but they were not initialized yet.
```suggestion
A universal account can be created two ways: a relayer or contract can fund and initialize it in one batch (`Transfer` + `StateInit` + optional follow-up call), or the account's own future owner can sign a transaction directly against the not-yet-initialized id, self-authorizing initialization with a key the id itself commits to. Two new host functions let a contract built against this scheme derive and create accounts of a `StateInit` version that did not exist when the contract was compiled, closing the same forward-compatibility gap [NEP-616]'s equivalent mechanism left open.
```

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+If the target account does not yet exist, or exists but is uninitialized, applying the action installs `code` (via the same global-contract-reference mechanism [NEP-591] and [NEP-616] already use), writes every `data` entry, and adds every `access_keys` entry as a full-access key, all seeded with the same nonce (see [Keys](#keys)). If the account is already initialized, none of that repeats: the action only settles `deposit` against the account's storage-staking requirement, refunding any excess through the same refund-routing mechanism [NEP-616] introduced, which this action shares. This makes a repeated `UniversalStateInit` to an already-initialized account a safe no-op rather than an error, which matters for the initialization flows below.
```
matejpavlovic, line 119: Is there a case where the account is initialized, but needs the storage staking deposit needs to be settled?

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+### Account lifecycle
+
+A universal account can be in one of three states: nonexistent, uninitialized, or initialized. This third possibility (funded but not yet initialized) is what distinguishes universal accounts from `0s` deterministic accounts, which never need it: a `0s` account's code is mandatory, so "no code yet" already means "not yet initialized," with no separate state required. A universal account's code is optional, so that trick does not work; a key-only universal account is fully initialized with no code at all, which is indistinguishable from "not yet initialized" unless the protocol tracks the distinction directly.
```
matejpavlovic, line 123:
```suggestion
A universal account can be in one of three states: nonexistent, uninitialized, or initialized. The "uninitialized" possibility means the account is funded but not yet initialized. This is what distinguishes universal accounts from `0s` deterministic accounts, which never need it: a `0s` account's code is mandatory, so "no code yet" already means "not yet initialized," with no separate state required. A universal account's code is optional, so that trick does not work; a key-only universal account is fully initialized with no code at all, which is indistinguishable from "not yet initialized" unless the protocol tracks the distinction directly.
```

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+The account's own actor id is never mutated as a side effect of creating it via transfer. An earlier version of the implicit-account-creation code did set it, which was harmless only because the single-action rule below made the assignment unreachable; removing that assignment, rather than relying on it staying unreachable, is what makes the batched creation flow below safe (see [Security Implications](#security-implications)).
```
matejpavlovic, line 144: I don't know what an actor id is in this context, but that might not be an issue for other readers. Feel free to ignore, just pointing it out in case you want to reach a broader set of readers.

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+Because there is no access key yet, there is also no ordinary nonce to check the transaction against. The account's `bootstrap_nonce`, seeded at creation from the block height, stands in for it: the self-signed transaction's nonce MUST equal `bootstrap_nonce + 1`, checked in forced strict-nonce mode regardless of what the transaction itself requests. Exactly one nonce is admissible at a time, which is what closes a replay path a looser mode would leave open, since a failed init does not flip the account's state and the same signed bytes would otherwise stay valid indefinitely.
```
matejpavlovic, line 154: What would a replay attack look like here, if the initialization is idempotent?

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+Base and per-byte are anchored to the same account-creation and storage economics as the existing `DeterministicStateInit` action, not to a measured-compute estimate: an internal gas-estimator pass was run against them and its output was judged too inconsistent to anchor a fee to. Charging the per-byte term over the entire raw payload, rather than only the bytes making up `data` entries as `DeterministicStateInit`'s is, means that at equal size the universal fee can only be greater than or equal to the deterministic one.
+
+Base, per-byte, and per-key each split into a send portion, paid by whoever creates the action (a transaction's signer at conversion, or a contract appending the promise action), and an execution portion, paid when the account is actually installed. Per-entry has no send portion at all; it is charged purely at execution. For the promise host function specifically, the send portions of base and per-byte are charged unconditionally, and the send portion of per-key only once the payload has been decoded (see [Host functions](#host-functions)).
```
matejpavlovic, line 204: Is the execution portion also paid by the creator of the action? Or is the execution somehow charged to the balance of the account being initialized, even if the initialization was triggered by another contract?

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+- [#16408](https://github.com/near/nearcore/pull/16408) capped the number of state-init storage entries.
+- [#16415](https://github.com/near/nearcore/pull/16415) moved the access-key cap from the action to the receipt total.
+
+End-to-end coverage lives in `test-loop-tests/src/tests/universal_account_id.rs`: key-only and contract accounts, repeated initialization, initialization after a plain transfer, self-signed initialization, and the three relayer-driven batch shapes, including a negative test that a relayer funding an account cannot smuggle in an `AddKey`.
+
+## Security Implications
+
+### A checksum-free address is never invalid, only nonexistent
+
+Without a checksum, a mistyped `0u` address is still a well-formed account id. It simply names an account that almost certainly does not exist, since with a 256-bit hash, mistyping into an account that does exist is infeasible. A transfer to it therefore fails the ordinary way, as "account does not exist," and the funds are refunded.
```
matejpavlovic, line 236: Earlier on, the NEP says that

"A transfer (or any other receipt that would create an account) to a nonexistent universal account id creates it in the Uninitialized state..."

It looks like if someone sends funds to a mistyped address, those funds will with very high probability be lost forever. I think this is fine though. It's the same with implicit accounts, after all (if I understand it correctly).

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+2. The bytes MUST decode as a `UniversalStateInit` (failure is `MalformedUniversalStateInit`, at both validation and action-error level, so it can be produced whether the check runs ahead of execution or the malformed bytes reach the runtime handler directly).
+3. Each storage key and value MUST NOT exceed the ordinary storage-write length limits (`max_length_storage_key` = 2048 bytes, `max_length_storage_value` = 4 MiB); there are no separate, universal-account-specific length limits.
+4. The total number of `access_keys` committed across every `UniversalStateInit` action in the same receipt MUST NOT exceed `max_universal_state_init_keys` (1024).
+5. The total number of storage entries across every `DeterministicStateInit` and `UniversalStateInit` action in the same receipt MUST NOT exceed `max_state_init_entries` (1500, from protocol version 87 onward; effectively unbounded before, when only `DeterministicStateInit` existed). This limit is shared with the older, unrelated `0s` action; see [Security Implications](#security-implications) for why.
```
mitinarseny, line 115: Can we please clarify what should happen with duplicate entries in `access_keys` and `data`, if any?

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+A transfer (or any other receipt that would create an account) to a nonexistent universal account id creates it in the `Uninitialized` state, holding only balance and storage usage, plus a `bootstrap_nonce` seeded from the creating block's height. Only a `UniversalStateInit` action can move an account out of this state, by calling `initialize()`, which clears the placeholder fields and installs the real `code`/`data`/`access_keys`. Every other action that would need existing state to act on (`DeployContract`, `AddKey`, `DeleteAccount`, `FunctionCall`, `Stake`, `Delegate`, and so on) is rejected against an uninitialized account with `ActionErrorKind::AccountNotInitialized`, a distinct error from "account does not exist": the account is there, it simply has nothing installed yet. Once initialized, a universal account is deleted the same way as any other account, including one referencing a global contract; nothing about `DeleteAccount` is universal-account-specific.
```
mitinarseny, line 142: > Every other action that would need existing state to act on (`DeployContract`, `AddKey`, `DeleteAccount`, `FunctionCall`, `Stake`, `Delegate`, and so on) is rejected against an uninitialized account with `ActionErrorKind::AccountNotInitialized`, a distinct error from "account does not exist": the account is there, it simply has nothing installed yet.

1. Do we really need a separate error variant for that? Shouldn't these actions fail at the validation stage, since they are only allowed if sender-is-receiver and no receipt can originate from an uninitialized account?
2. What would happen if a universal account creates a promise with `[DeleteAccount, AddKey]` actions to itself?

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+**Self-signed.** The account's own future owner can also initialize it directly, with no relayer and no pre-existing access key, by signing a transaction that is addressed to the id its own state init derives to, using one of the keys that state init names in `access_keys`. The verifier recognizes this shape (`Transaction::is_state_init_bootstrap()`: signer and receiver match, a top-level action's state init derives to that id, and the signing key is among the committed keys) and authorizes the whole transaction accordingly, rather than looking up an access key that does not exist yet. Because the transaction is genuinely self-authorized, it MAY carry further top-level actions after the `UniversalStateInit`, including ones, like `AddKey` or `DeleteAccount`, that the batched flow above cannot authorize.
```
mitinarseny, line 152: > The verifier recognizes this shape (`Transaction::is_state_init_bootstrap()`: signer and receiver match, ...

Does it mean that, in case of access keys, the owner of the private key cannot start using already pre-funded but not-yet-initialized account right away, but needs to send a separate initialization transaction first, since it's required that sender is receiver? I.e. the owner cannot do both `StateInit` and `Transfer` to another account in the same transaction.

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
...ainst repeating a specific mistake already made once, in [NEP-616]'s deterministic accounts: mutating a receipt's actor id as a side effect of account creation. Exempting universal accounts from the single-action rule on implicit-account-creating transfers (to permit the batched creation flow) would have reopened exactly that vector if an unrelated piece of legacy code had not also been fixed: `action_implicit_account_creation_transfer` used to unconditionally set the receipt's actor id to the newly created account, a no-op under the old single-action rule (there was never a subsequent action to authorize), but a live privilege-escalation bug the moment a subsequent action becomes possible. A relayer batching `[Transfer, UniversalStateInit, AddKey]` or `[Transfer, UniversalStateInit, DeleteAccount]` could otherwise have that trailing action authorized against the account it had just funded, installing a key the account's id never committed to, or draining it outright. The fix removes the actor-id assignment entirely, for every implicit account kind; it is provably unobserved for the older kinds because their creation gate still requires the transfer to be the receipt's only action.
```
mitinarseny, line 254: > it is provably unobserved for the older kinds because their creation gate still requires the transfer to be the receipt's only action.

Hmm, what about `[Transfer, Transfer]` to a not-yet-existing implicit account id?

mitinarseny (commented):
@Wiezzel Thank you for submitting this NEP and @walnut-the-cat for the honor to be a SME for it!

I like how this proposal introduces a new universal implicit account id type that can be used for both access-key(s)- and contract-owned accounts, while being post-quantum safe and forward-compatible with future variants, new access key types and External Contract Calls. 

Leaving here a couple of comments and looking forward to have it usable on mainnet soon! 

PS: I hope this would be the last time we need to introduce yet another account id scheme 😅🤞🏻

neps/nep-0655.md
```diff
@@ -0,0 +1,414 @@
...
+
+A third cost is not gas-metered at all. While selecting transactions, a chunk producer decodes and hashes a candidate payload several times before any balance check can reject it. That work is bounded by the chunk producer's own transaction-preparation time limit, which is local policy rather than a protocol rule, so it protects only a producer running current software.
+
+## Alternatives
+
+### A checksum on the address
+
+The address format originally included a 6-symbol checksum (a BCH polymod over the Crockford alphabet, in the style of Bech32m), bringing the total to 60 characters. It was fully implemented, known-answer tested, and agreed at a design-review meeting before being reversed, for the funds-safety reason given in [Security Implications](#security-implications). A sub-debate over which checksum algorithm to use (an off-the-shelf CRC versus the custom Bech32m-derived one) turned out to be moot once the concept of a protocol-level checksum was dropped entirely. Whether a checksum belongs at the wallet/UI layer instead, similar to TON's user-facing address format, was left an open question, not decided either way.
```
matejpavlovic, line 268: Not quite sure how the connection between the checksum and the "funds-safety reason" are connected (the safety of the funds is questioned in a previous comment). I think the checksum (along with appropriate tooling) would prevent certain mistakes that could lead to lost funds.

A more compelling argument against is a different one in my opinion: It would introduce non-negligible complexity to the protocol, which would need to verify the validity of user accounts (using the checksum). It would need to distinguish between addresses provided externally which are potentially invalid and addresses that are already checked. This is not unsurmountable, but still inconvenient and error-prone. I'd argue that the core protocol should be kept as simple as possible and sanitizing user input should be kept outside of it. A special user-facing address format (TON-style), like already mentioned in this paragraph, seems like a much better idea to me.

matejpavlovic (commented):
Nice, well written!

I added a few comments, mostly minor ones. The biggest one is about transferring funds to mistyped addresses and everything that relates to it.

There were some technicalities I didn't fully understand (e.g. around actor IDs), but it didn't seem to be about fundamentally concerning things.

neps/nep-0655.md
Wiezzel: Yes, the "payloads" simply means duplicates of the same action.

neps/nep-0655.md
Trisfald: If I understand it correctly, yes it's true.

For example, after withdrawing NEAR from an exchange into a new 0u account, the user needs two transaction to send money to someone else:
- tx1: UniversalStateInit
- tx2: Transfer

Maybe it can be improved by TXv2 adding the optional state init? Or maybe we can consider ECC the optimized flow?

neps/nep-0655.md
Wiezzel: They are deduplicated on decode, but I assumed this is just a borsh deserialization detail, not part of the Universal Accounts specs per se.

mitinarseny: I see, then does it make sense to specify explicitly that:
1. For any duplicate key the last (or first?) seen value is taken?
2. These per-receipt limits apply to the total number of *distinct* or *all listed* entries?

neps/nep-0655.md
mitinarseny: > a universal state init can only be used with a global contract?

IIUC, yes. If we want to extend it to "inline" contract code later, then we can either introduce `UniversalStateInit::V2` variant or, alternatively, there can be a new `ContractCode` enum as an extension of `GlobalContractId` to be used in `UniversalStateInit::code` field.

neps/nep-0655.md
Wiezzel: > Do we really need a separate error variant for that? Shouldn't these actions fail at the validation stage, since they are only allowed if sender-is-receiver and no receipt can originate from an uninitialized account?

It can't be a pure pre-execution check, because `UniversalStateInit` can flip the account into initialized state partway through the same receipt (that's exactly how self-signed `[UniversalStateInit, AddKey]` is meant to work). And `FunctionCall` specifically isn't actor-gated at all, so it needs this check directly.

> What would happen if a universal account creates a promise with `[DeleteAccount, AddKey]` actions to itself?

That would fail due to a general rule unrelated to universal accounts: `DeleteAccount` must be the receipt's final action.

mitinarseny: > It can't be a pure pre-execution check

Sorry, I meant inside [`check_actor_permissions()`](https://github.com/near/nearcore/blob/ad26727eafe8d8fc019578da870ee819c3222f1b/runtime/runtime/src/actions.rs#L770-L777) which is [called](https://github.com/near/nearcore/blob/ad26727eafe8d8fc019578da870ee819c3222f1b/runtime/runtime/src/lib.rs#L606) before actually "applying" each action.

neps/nep-0655.md
Wiezzel: > For any duplicate key the last (or first?) seen value is taken?

`access_keys` are a set, so it doesn't matter. For `data` I honestly don't know if it's the first, the last, or something else. And again I'm not sure if this should be part of this spec. It's not even part of the borsh spec, but an implementation detail of a particular serializer library.

> These per-receipt limits apply to the total number of distinct or all listed entries?

They are applied post-deserialization, so they bound distinct entries.

neps/nep-0655.md
Wiezzel: > Hmm, what about `[Transfer, Transfer]` to a not-yet-existing implicit account id?

That's forbidden as well. The transfer must be the *only single action*, not the *only type of action*. I will reword it a bit to make it clearer.

neps/nep-0655.md
Wiezzel: Yes, that's true. I will add this to the future possibilities section.

neps/nep-0655.md
Wiezzel: Will add to the future possibilities section.

neps/nep-0655.md
Wiezzel: I will rephrase add an example to make it clear.

neps/nep-0655.md
Wiezzel: I believe the account's storage usage can grow after initialization through ordinary contract execution, so a later universal state init carrying a deposit could be used to top up storage stake.

neps/nep-0655.md
Wiezzel: I will add a short explanation.

neps/nep-0655.md
Wiezzel: That's a really good question, and the answer is not obvious, so I'll put it in the doc. A **failed** self-signed attempt still burns the conversion fee from the account's own balance at verification, regardless of whether the action succeeds or not. Without this guard, if anyone submitted a failing state init, it could be replayed by a malicious actor indefinitely, burning through the account's balance.

neps/nep-0655.md
Wiezzel: Yes, gas is paid the standard way – by the creator of the action. Will add a sentence to clarify.

neps/nep-0655.md
Wiezzel: Yes, this is correct.

neps/nep-0655.md
Wiezzel: I agree that a client-side checksum could be good. What was rejected was a protocol-level checksum that caused some previously-assumed-valid account IDs to become rejected. And that was a problem, because a contract predating this feature wouldn't have a way to verify the checksum and could potentially send funds to such address. These funds would be irrecoverably lost.

neps/nep-0655.md
Wiezzel: `check_actor_permissions` and `check_account_existence` are semantically different – the first one answers "is the caller allowed to act upon this account" and the second one "is there anything to act upon". And the answer might differ. In particular `FunctionCall` and `Delegate` don't require actor-equals-account, but they do require account-exists-and-is-initialized.

I hope this answers your question.

Wiezzel:
@mitinarseny @matejpavlovic I have replied to and/or addressed in the doc all comments. Please take another look.

neps/nep-0655.md
matejpavlovic: But doesn't the contract execution that happens after initialization already have to pay for the storage deposit (because it would otherwise fail)? In other words, a contract can never have a "debt" on storage deposit, right? And if the account already has enough balance to cover its storage stake, then I don't understand the "the action only settles `deposit` against the account's storage-staking requirement".

Wiezzel: Yes, now I see that I was mistaken here. You are correct that a contract can never have a storage deposit debt. So for an already-initialized account this is effectively a full refund. I will fix that paragraph to make it clear.

